PR #3, opened 2016-01-27
fixed LDAP operation on empty value

skoef:
this is in addition to commit f2b9fc797d59730c531afc1339130b89351cf322, since that only checks the value being empty on all but the first field


matt-jordan:
This is a mirror of the Asterisk project, not the actual repo. No PRs will be merged through Github.

If you'd like to propose this patch to the project, please push the patch to the Asterisk repo on Gerrit:

https://gerrit.asterisk.org/#/q/status:open

Instructions for doing so can be found on the Asterisk wiki:

https://wiki.asterisk.org/wiki/display/AST/Gerrit+Usage

PR #7, opened 2017-06-06
add -y to install aptitude

Revent:


jcolp:
This github project is a mirror of the Asterisk project's Gerrit. As a result we don't accept pull requests on here. If you'd like to contribute you can find instructions on the wiki[1].

[1] https://wiki.asterisk.org/wiki/display/AST/Patch+Contribution+Process

PR #10, opened 2018-03-18
Removing 10 seconds delay after mp3play

ricardomaraschini:
This patch changes the way asterisk polls output from mpg123, instead
of waiting for 10 seconds(when playing an http url) it now uses a
timeout of one second and iterates 10 times using this same timeout.

The main difference is that for every timeout asterisk receives it now
checks if mpg123 is still running before poll again.

### DIFF
--- a/apps/app_mp3.c
+++ b/apps/app_mp3.c
@@ -35,6 +35,7 @@
 #include "asterisk.h"
 
 #include <sys/time.h>
+#include <sys/types.h>
 #include <signal.h>
 
 #include "asterisk/lock.h"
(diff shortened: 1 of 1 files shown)


jcolp:
Thanks for the PR.

Unfortunately, we don't accept PRs on Github. The Asterisk repo here is just a mirror.

If you'd like to submit the change back to the project, we'd appreciate it. Please push the change up to [Gerrit](https://gerrit.asterisk.org/). Instructions for doing so can be found on the Asterisk [wiki](https://wiki.asterisk.org/wiki/display/AST/Patch+Contribution+Process).

Thanks!

PR #33, opened 2023-04-08
stasis.c : Fix issue with stasis_topic_pool_delete_topic function

Yashyk456:
Fix issue in stasis_topic_pool_delete_topic function

The stasis_topic_pool_delete_topic function in stasis.c includes a call to the stasis_topic_name function which can return NULL. Previously, the function was being compared directly with strncmp, which caused a segmentation fault in some cases.

To fix this, I added a null check in the if statement to ensure that stasis_topic_name is not NULL before performing strncmp. This prevents the segmentation fault and ensures that the function works as intended.

This commit resolves the issue and improves the stability of the stasis_topic_pool_delete_topic function.

### DIFF
--- a/main/stasis.c
+++ b/main/stasis.c
@@ -1872,7 +1872,7 @@ void stasis_topic_pool_delete_topic(struct stasis_topic_pool *pool, const char *
 	int pool_topic_name_len = strlen(pool_topic_name);
 	const char *search_topic_name;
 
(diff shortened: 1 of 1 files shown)


seanbright:
The Asterisk project is not currently on Github.This is merely a mirror. The contribution process is documented on our wiki at https://wiki.asterisk.org/wiki/display/AST/Patch+Contribution+Process

Issue #35, opened 2023-04-28
[New Feature]: chan_dahdi: Allow disabling pulse or tone dialing


InterLinked1:
### Feature Description

Currently, pulse and tone dialing are enabled on all FXS lines and there is no way to control this at all.

Adds a "dialmode" option to DAHDI that can be set to four different options:

1) both (default)
2) pulse
3) dtmf or tone
4) none

This allows for specifying, on a per-line basis, which types of signaling methods are allowed on FXS lines. Additionally, the dialmode can be updated during a call.

Imported from: https://issues.asterisk.org/jira/browse/ASTERISK-29992

PR #36, opened 2023-04-28
chan_dahdi: Add dialmode option for FXS lines.

InterLinked1:
Currently, both pulse and tone dialing are always enabled on all FXS lines, with no way of disabling one or the other.

In some circumstances, it is desirable or necessary to disable one of these, and this behavior can be problematic.

A new "dialmode" option is added which allows setting the methods to support on a per channel basis for FXS (FXO signalled lines). The four options are "both", "pulse", "dtmf"/"tone", and "none".

Additionally, integration with the CHANNEL function is added so that this setting can be updated for a channel during a call.

Resolves: #35
ASTERISK-29992
Imported from: https://gerrit.asterisk.org/c/asterisk/+/18304

UserNote: A "dialmode" option has been added which allows specifying, on a per-channel basis, what methods of subscriber dialing (pulse and/or tone) are permitted.

Additionally, this can be changed on a channel
at any point during a call using the CHANNEL
function.


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: master

channels/chan_dahdi.c
```diff
@@ -7009,6 +7021,30 @@ static int dahdi_func_read(struct ast_channel *chan, const char *function, char
 		}
 		ast_mutex_unlock(&p->lock);
 #endif	/* defined(HAVE_PRI) */
+	} else if (!strcasecmp(data, "dialmode")) {
+		struct analog_pvt *analog_p = p->sig_pvt;
+		if (analog_p) {
```
jcolp, line 7026: This is insufficient. sig_pvt is a void*, so if sig_pvt is set to any private structure then analog_p will be set - though the value will likely be garbage if it's not an analog_pt

channels/chan_dahdi.c
```diff
@@ -7114,6 +7150,27 @@ static int dahdi_func_write(struct ast_channel *chan, const char *function, char
 			ast_log(LOG_WARNING, "Unsupported value '%s' provided for '%s' item.\n", value, data);
 			res = -1;
 		}
+	} else if (!strcasecmp(data, "dialmode")) {
+		struct analog_pvt *analog_p = p->sig_pvt;
+		if (!analog_p) {
```
jcolp, line 7155: Same applies here

jcolp (approved):

jcolp (approved):

Issue #39, opened 2023-04-29
[Bug]: Remove .gitreview from repository.


seanbright:
### Severity

Trivial

### Versions

GIT

### Components/Modules

Core

### Operating Environment

N/A

### Frequency of Occurrence

None

### Issue Description

Remove the, now obsolete, `.gitreview` file from the root of the project.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
I wonder if alongside this we should prune any Jenkins stuff that isn't relevant any longer, too.

seanbright:
Make sense. I’ll take a look.

PR #40, opened 2023-04-29
core: Cleanup gerrit and JIRA references.

seanbright:
Resolves: #39


seanbright:
cherry-pick-to: 20
cherry-pick-to: master

jcolp (approved):

gtjoseph (changes requested):
The .gitreview file is used by build_tools/make_version and the build scripts to determine the base branch so we still need it.  We could remove the gerrit references though...
```
[github]
defaultbranch=master
basebranch=master
```
Or we could find another way to determine base branch.

jcolp:
We haven't had .gitreview since the beginning of Gerrit, it was added afterwards, so how was it done before?

gtjoseph:
> We haven't had .gitreview since the beginning of Gerrit, it was added afterwards, so how was it done before?

The build/install/test processes were much simpler back then and didn't need to know the base branch.  One way we use it today is to determine which builds of the external codecs and DPMA to download (or NOT download in case of master).

jcolp:
According to the git log it did know the base branch back then, @matt-jordan hardcoded it in instead of parsing it out of .gitreview at the time. Our use of things got more "fancy" afterwards.

gtjoseph:
Well, however we determine it, we still need to know and the scripts need to be changed accordingly.

seanbright:
New proposal is that we put it in configure.ac and pull it from there if needed.

jcolp (approved):
I'm down with this approach.

seanbright:
> You need to run bootstrap.sh to update ./configure.
> Actually creating new variable for this and exporting it to makeopts.in and autoconfig.in, etc. would make things easier in the future.

I will run bootstrap, this was just to propose the approach. 

I’m pretty sure there is already a `PACKAGE_VERSION` variable that ends up in `autoconfig.h` but I will confirm.

jcolp (approved):

gtjoseph:
Feel free to do the other branches.

seanbright:
> Feel free to do the other branches.

[done](https://github.com/asterisk/asterisk/pull/57) and [done](https://github.com/asterisk/asterisk/pull/58).

Issue #45, opened 2023-05-01
[bug]: Non-bundled PJSIP check for evsub pending NOTIFY check is insufficient/ineffective


jcolp:
### Severity

Minor

### Versions

18, 20, master

### Components/Modules

configure

### Operating Environment

All

### Frequency of Occurrence

Constant

### Issue Description

The autoconf logic at:

https://github.com/asterisk/asterisk/blob/1b95f6ee3fd032884d9ad06311e2ddac7875afef/configure.ac#L2514

Assumes that the source code for PJSIP will be present when it is not guaranteed. This can lead to improper behavior.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #47, opened 2023-05-01
configure: Fix pending NOTIFY check for non-bundled pjsip.

seanbright:
Resolves: #45


seanbright:
Assuming this gets approved I will submit separate PRs for 18 and 20. Cherry-picking changes to generated files never feels correct.

mbradeen (approved):
I could not find any external indications of this change other than version so in this case I think that's all we have.

jcolp (approved):

mbradeen (approved):

jcolp (approved):

gtjoseph:
Go ahead and do the cherry-pick PRs.

Issue #55, opened 2023-05-02
[bug]: res_sorcery_memory_cache: Memory leak when calling sorcery_memory_cache_open


nphantom:
### Severity

Minor

### Versions

20, 18

### Components/Modules

Resources/res_sorcery_memory_cache

### Operating Environment

CentOS 8.5

### Frequency of Occurrence

Constant

### Issue Description

The pointer returned by ast_strdup had never been freed.
Imported from: https://issues.asterisk.org/jira/browse/ASTERISK-30429

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #56, opened 2023-05-02
res_sorcery_memory_cache.c: Fix memory leak.

nphantom:
The pointer returned by ast_strdup had never been freed.

Resolves: https://github.com/asterisk/asterisk/issues/55
[ASTERISK-30429](https://issues.asterisk.org/jira/browse/ASTERISK-30429)

### DIFF
--- a/res/res_sorcery_memory_cache.c
+++ b/res/res_sorcery_memory_cache.c
@@ -1507,7 +1507,7 @@ static int age_cmp(void *a, void *b)
  */
 static void *sorcery_memory_cache_open(const char *data)
 {
-	char *options = ast_strdup(data), *option;
+	char *options = ast_strdupa(data), *option;
 	RAII_VAR(struct sorcery_memory_cache *, cache, NULL, ao2_cleanup);
 
 	cache = ao2_alloc_options(sizeof(*cache), sorcery_memory_cache_destructor, AO2_ALLOC_OPT_LOCK_NOLOCK);


res/res_sorcery_memory_cache.c
```diff
@@ -1507,8 +1507,9 @@ static int age_cmp(void *a, void *b)
  */
 static void *sorcery_memory_cache_open(const char *data)
 {
-	char *options = ast_strdup(data), *option;
```
jcolp, line 1510: I don't see why this can't just use strdupa, other modules do that.

jcolp (changes requested):

res/res_sorcery_memory_cache.c
nphantom: Initially, I intended to switch to strdupa. However, since it was unclear whether the original use of strdup was due to a spelling mistake or to account for potential long input data, I decided to add resource deallocation based on that. Of course, using strdupa would have been a simpler and more efficient option.

jcolp: Other implementations do it, and this isn't a loop where it will continually allocate, so it's safe.

jcolp (approved):

gtjoseph (changes requested):
Can you squash the 2 commits then do a `git push --force` so there's only 1 commit in the pull request.

nphantom:
> Can you squash the 2 commits then do a `git push --force` so there's only 1 commit in the pull request.

I have completed your request, now there's only one commit in the pull request.

gtjoseph (approved):

gtjoseph:
If you want this PR cherry-picked, please add the appropriate comment as described in
https://wiki.asterisk.org/wiki/display/AST/Code+Contribution

nphantom:
cherry-pick-to: 20
cherry-pick-to: 18

Issue #46, opened 2023-05-01
[bug]: Stir/Shaken: Wrong CID used when looking up certificates


lukeescude:
### Severity

Trivial

### Versions

16, 18, 20

### Components/Modules

STIR/SHAKEN

### Operating Environment

CentOS 7 x64

### Frequency of Occurrence

Constant

### Issue Description

When signing outbound calls, Asterisk is using the wrong callerID to look up certificates.

There is a patch available in progress, I just wanted to get it into the Github issue system: 

https://gerrit.asterisk.org/c/asterisk/+/16174

Dial plan example:
exten => 507,1,Set(CALLERID(num)=442081254195)

507 is the extension, 442081254195 is the Caller ID number (network number) which should be utilized for cert check

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
@seanbright @bkford I'm lost in the history here.  The gerrit review comments make my head ache.  Does that review need to be recreated here as a PR?

seanbright:
Probably. The review on gerrit is -2 because the tests are failing and needed to be reworked.

seanbright:
I suggest reading the comments on [the abandoned 16 review](https://gerrit.asterisk.org/c/asterisk/+/16153).

lukeescude:
Just to confirm, this fix has not made it into Asterisk yet, correct? I'm surprised this isn't bothering more people, since stir/shaken is pretty much a requirement now.

jcolp:
This issue remains open, no fix has been merged. It will automatically close upon merging in a fix.

PR #62, opened 2023-05-04
res_pjsip_stir_shaken: Use correct Caller ID to lookup certificate.

seanbright:
Determine the Caller ID to use for certificate lookup based on the same logic we use when setting up the INVITE session. This allows Caller ID set via the CALLERID() function to be used for STIR/SHAKEN certificate lookup.

Resolves: #46


res/res_pjsip_stir_shaken.c
```diff
@@ -510,17 +513,27 @@ static void stir_shaken_outgoing_request(struct ast_sip_session *session, pjsip_
...
+	ast_party_id_copy(&connected_id, &effective_id);
+	ast_channel_unlock(session->channel);
+
+	/* XXX: This if statement is logically the same as the first if statement in
+	   res_pjsip_caller_id.c:add_id_headers. They should be combined. */
+	if (!connected_id.number.valid
+		|| (!session->endpoint->id.trust_outbound
+			&& (ast_party_id_presentation(&connected_id) & AST_PRES_RESTRICTION) != AST_PRES_ALLOWED)) {
+		ast_party_id_free(&connected_id);
 		return;
```
seanbright, line 528: It would be better to do what this comment says than to kick it down the road.

jcolp: Can you clarify why "combined" means within this context? Having the STIR/SHAKEN logic be in callerid, for example?

jcolp (approved):

InterLinked1 (approved):

gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20

bkford (approved):

gtjoseph:
Given all the test failures, what's the path forward for this PR?

trinicomcom:
Hi guys, not sure if I am helping here but been trying to get asterisk to send S/S identity header for a week now.
I found another issue with the stir shaken identity header that is created by asterisk to send out is that the To Caller ID is also the Orig Caller ID in the identity header. I am pretty sure all we need is to put the $callerid(num) from extensions.conf for outbound calls in the origin in the header creation and just send that for all outbound calls and not try to match a cert to a caller ID number since all we are doing is sending out a master cert for the owner of all the numbers in the switch to prove the outbound call is legit and the owner is certified.  I hope this makes sense. I can help test any releases if you want I would just need some direction on how to compile it on my lab server.

jcolp:
@seanbright Is this now good to review again?

seanbright:
It’s ready to be reviewed. Was I the hold up? There are failing tests that need to be fixed but I do not personally plan on fixing them.

jcolp:
No, I just didn't know what the actual state of things were currently.

seanbright:
> Given all the test failures, what's the path forward for this PR?

You can do one of the follow:

1. Fix the tests
2. Disable the tests

This simple patch has been sitting for **2 years** (originally published to gerrit 7/21) waiting to be committed.

Issue #66, opened 2023-05-05
[improvement]: Fix preserve reason of pause when Asterisk is restared for realtime queues


pnlarsson:
### Improvement Description

When the Asterisk is restared it does not preserve paused reason for members of realtime queues. This was fixed for non-realtime queues in https://gerrit.asterisk.org/c/asterisk/+/2131/

Patch exists - pr coming soon

PR #67, opened 2023-05-05
apps/app_queue: Preserve reason for realtime queues

pnlarsson:
When Asterisk is restarted it does not preserve paused reason for
members of realtime queues. This was fixed for non-realtime queues in
ASTERISK_25732

Resolves: #66

UpgradeNote: Add a new column to the queue_member table:
reason_paused VARCHAR(80) so the reason can be preserved.

UserNote: Make paused reason in realtime queues persist an
Asterisk restart. This was fixed for non-realtime
queues in ASTERISK_25732.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -1606,6 +1606,9 @@ static int log_membername_as_agent;
 /*! \brief name of the ringinuse field in the realtime database */
 static char *realtime_ringinuse_field;
 
+/*! \brief does realtime backend support reason_paused */
+static int realtime_reason_paused;
+
 enum queue_result {
 	QUEUE_UNKNOWN = 0,
 	QUEUE_TIMEOUT = 1,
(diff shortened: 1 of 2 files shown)


pnlarsson:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp:
You've put an upgrade note stating that reason_paused needs to be added, but the code makes it appear as though it is optional. Is it optional? If so then the upgrade note is not accurate.

jcolp (changes requested):
Adding request changes to get attention on my previous comment.

jcolp:
The UpgradeNote needs to be clarified then to explain that. Right now it reads as if this is a breaking schema change.

pnlarsson:
> The UpgradeNote needs to be clarified then to explain that. Right now it reads as if this is a breaking schema change.

Changed the UpgradeNote a bit

jcolp:
The review hasn't been updated with the change.

pnlarsson:
> The review hasn't been updated with the change.

Strange I changed the top comment - is that not what is used for commit message?

jcolp (approved):

gtjoseph (approved):

Issue #65, opened 2023-05-05
[bug]: heap overflow by default at startup


asarubbo:
### Severity

Trivial

### Versions

18.17.0

### Components/Modules

core?

### Operating Environment

Gentoo

### Frequency of Occurrence

Constant

### Issue Description

By compiling asterisk with asan (but it can be noticed as well via valgrind) I see an off-by-one that happens by default.

### Relevant log output

```shell
~ $ /usr/sbin/asterisk -C /etc/asterisk/asterisk.conf -f -U asterisk
Unable to access the running directory (Permission denied).  Changing to '/' for compatibility.
PBX UUID: 80ab15e7-983a-4e45-87da-dfdf67426b1d
[May  5 11:46:06] NOTICE[2506348]: loader.c:2415 load_modules: 314 modules will be loaded.
[May  5 11:46:06] NOTICE[2506348]: cdr.c:4568 cdr_toggle_runtime_options: CDR simple logging enabled.
[May  5 11:46:06] WARNING[2506348]: config.c:2041 process_text_line: parse error: No category context for line 310 of /etc/asterisk/geolocation.conf
[May  5 11:46:06] ERROR[2506348]: res_sorcery_config.c:334 sorcery_config_internal_load: Contents of config file 'geolocation.conf' are invalid and cannot be parsed
[May  5 11:46:06] WARNING[2506348]: config.c:2041 process_text_line: parse error: No category context for line 310 of /etc/asterisk/geolocation.conf
[May  5 11:46:06] ERROR[2506348]: res_sorcery_config.c:334 sorcery_config_internal_load: Contents of config file 'geolocation.conf' are invalid and cannot be parsed
=================================================================
==2506348==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6030002c0458 at pc 0x7f741af1f286 bp 0x7fff5e6db130 sp 0x7fff5e6da8a8
READ of size 1 at 0x6030002c0458 thread T0
    #0 0x7f741af1f285  (/usr/lib/gcc/x86_64-pc-linux-gnu/12/libasan.so.8+0x74285)
    #1 0x7f741af212c5 in __vsnprintf_chk (/usr/lib/gcc/x86_64-pc-linux-gnu/12/libasan.so.8+0x762c5)
    #2 0x6e14fd in vsnprintf /usr/include/bits/stdio2.h:68
[...]

PR #69, opened 2023-05-09
xml.c: Process XML Inclusions recursively.

seanbright:
If processing an XInclude results in new `<xi:include>` elements, we need to run XInclude processing again. This continues until no replacement occurs or an error is encountered.

There is a separate issue with dynamic strings (ast_str) that will be addressed separately.

Resolves: #65

### DIFF
--- a/main/xml.c
+++ b/main/xml.c
@@ -68,6 +68,29 @@ int ast_xml_finish(void)
 	return 0;
 }
 
+/*!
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (approved):

PR #74, opened 2023-05-10
chan_pjsip: Allow topology/session refreshes in early media state

### LINKED ISSUE #73: [new-feature]: pjsip: Allow topology/session refreshes in early media state
### Feature Description

When an SDP negotiation has finished during an initial INVITE transaction and the SDP was sent using a reliable 1xx response, session modifications are allowed in the early media state by sending UPDATEs ([RFC 6337](https://datatracker.ietf.org/doc/html/rfc6337) contains many details, especially [section 3.2](https://datatracker.ietf.org/doc/html/rfc6337#section-3.2)).

Asterisk does not allow this yet, but there are use cases in which session modifications in the early media state are necessary.

maximilianfridrich:
With this change, session modifications in the early media state are
possible if the SDP was sent reliably and confirmed by a PRACK. For
details, see RFC 6337, escpecially section 3.2.

Resolves: #73

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -167,6 +167,14 @@ static struct ast_sip_session_supplement chan_pjsip_ack_supplement = {
 	.incoming_request = chan_pjsip_incoming_ack,
 };
 
+static int chan_pjsip_incoming_prack(struct ast_sip_session *session, struct pjsip_rx_data *rdata);
+
+static struct ast_sip_session_supplement chan_pjsip_prack_supplement = {
+	.method = "PRACK",
+	.priority = AST_SIP_SUPPLEMENT_PRIORITY_CHANNEL,
+	.incoming_request = chan_pjsip_incoming_prack,
+};
+
 /*! \brief Function called by RTP engine to get local audio RTP peer */
 static enum ast_rtp_glue_result chan_pjsip_get_rtp_peer(struct ast_channel *chan, struct ast_rtp_instance **instance)
 {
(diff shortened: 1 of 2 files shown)


maximilianfridrich:
cherry-pick-to: 20
cherry-pick-to: 18

channels/chan_pjsip.c
```diff
@@ -1564,13 +1572,18 @@ static int send_topology_change_refresh(void *data)
 {
 	struct topology_change_refresh_data *refresh_data = data;
 	struct ast_sip_session *session = refresh_data->session;
+	enum ast_channel_state state = ast_channel_state(session->channel);
+	enum ast_sip_session_refresh_method method = AST_SIP_SESSION_REFRESH_METHOD_INVITE;
 	int ret;
 	SCOPE_ENTER(3, "%s: %s\n", ast_sip_session_get_name(session),
 		ast_str_tmp(256, ast_stream_topology_to_str(refresh_data->media_state->topology, &STR_TMP)));
 
+	if (session->early_confirmed && (state == AST_STATE_RINGING || state == AST_STATE_RING)) {
```
jcolp, line 1581: Add a comment specifying RFC 6337, as well as why ringing and ring.

jcolp (changes requested):

channels/chan_pjsip.c
maximilianfridrich: Done. Also, now the `early_confirmed` flag is also set whenever we receive a reliable 1xx response with SDP, which is how it should be.

jcolp (approved):

gtjoseph (approved):

PR #75, opened 2023-05-10
configure: fix test code to match gethostbyname_r prototype.

jkroonza:
This enables the test to work with CC=clang.

Without this the test for 6 args would fail with:

utils.c:99:12: error: static declaration of 'gethostbyname_r' follows non-static declaration static int gethostbyname_r (const char *name, struct hostent *ret, char *buf,
           ^
/usr/include/netdb.h:177:12: note: previous declaration is here extern int gethostbyname_r (const char *__restrict __name,
           ^

Fixing the expected return type to int sorts this out.

### DIFF
--- a/configure
+++ b/configure
@@ -17537,7 +17537,7 @@ $as_echo_n "checking for gethostbyname_r with 6 arguments... " >&6; }
 int
 main ()
 {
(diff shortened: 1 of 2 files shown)


seanbright (approved):

seanbright:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (approved):

gtjoseph (approved):

Issue #87, opened 2023-05-16
[bug]: app_followme: Setting enable_callee_prompt=no breaks timeout


alex2grad:
### Severity

Minor

### Versions

18.17.1, 20.2.0

### Components/Modules

app_followme

### Operating Environment

Linux x86_64

### Frequency of Occurrence

Constant

### Issue Description

Setting enable_callee_prompt=no, makes FollowMe application dial only first 'number' entry forever, ignoring the configured timeout.

https://issues.asterisk.org/jira/browse/ASTERISK-30326

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #88, opened 2023-05-16
app_followme: fix issue with enable_callee_prompt=no

alex2grad:
If the FollowMe option 'enable_callee_prompt' is set to 'no' then asterisk
incorretly sets a winner channel to the channel from any control frame was read.

This fix sets the winner channle only to the answred channel.

Resolves: #87

ASTERISK-30326


InterLinked1 (changes requested):
GitHub doesn't seem to allow commenting on commit messages like Gerrit does, so this is not inline:

asterisk -> Asterisk
incorretly -> incorrectly
from -> from which
channle -> channel
answred -> answered

alex2grad:
> GitHub doesn't seem to allow commenting on commit messages like Gerrit does, so this is not inline:
> 
> asterisk -> Asterisk incorretly -> incorrectly from -> from which channle -> channel answred -> answered

done

InterLinked1:
> > GitHub doesn't seem to allow commenting on commit messages like Gerrit does, so this is not inline:
> > asterisk -> Asterisk incorretly -> incorrectly from -> from which channle -> channel answred -> answered
> 
> done

It isn't done; it was the commit message, not the PR message, to which I was referring.

alex2grad:
> > > GitHub doesn't seem to allow commenting on commit messages like Gerrit does, so this is not inline:
> > > asterisk -> Asterisk incorretly -> incorrectly from -> from which channle -> channel answred -> answered
> > 
> > 
> > done
> 
> It isn't done; it was the commit message, not the PR message, to which I was referring.

done

jcolp (approved):

Issue #94, opened 2023-05-18
[new-feature]: sig_analog: Add full Caller ID support for incoming calls


InterLinked1:
### Feature Description

A previous patch from last year added internal support for full Caller ID to the Asterisk core. This now updates chan_dahdi to take advantage of this support.

Imported from: ASTERISK-30331

PR #95, opened 2023-05-18
sig_analog: Add fuller Caller ID support.

InterLinked1:
A previous change, ASTERISK_29991, made it possible to send additional Caller ID parameters that were
not previously supported.

This change adds support for analog DAHDI channels to now be able to receive these parameters for
on-hook Caller ID, in order to enhance the usability of CPE that support these parameters.

Resolves: #94
ASTERISK-30331

UserNote: Additional Caller ID properties are now supported on incoming calls to FXS stations, namely the
redirecting reason and call qualifier.

Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19601

### DIFF
--- a/channels/sig_analog.c
+++ b/channels/sig_analog.c
@@ -1073,6 +1073,8 @@ int analog_call(struct analog_pvt *p, struct ast_channel *ast, const char *rdest
 			}
 
 		}
+
+		/* Name and Number */
 		n = ast_channel_connected(ast)->id.name.valid ? ast_channel_connected(ast)->id.name.str : NULL;
 		l = ast_channel_connected(ast)->id.number.valid ? ast_channel_connected(ast)->id.number.str : NULL;
 		if (l) {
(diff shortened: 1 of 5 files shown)


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 18

channels/sig_analog.c
```diff
@@ -1087,12 +1089,29 @@ int analog_call(struct analog_pvt *p, struct ast_channel *ast, const char *rdest
...
 			p->caller.id.name.presentation = ast_channel_connected(ast)->id.name.presentation;
 			p->caller.id.number.presentation = ast_channel_connected(ast)->id.number.presentation;
+
+			/* Redirecting Reason */
+			p->redirecting_reason = ast_channel_redirecting(ast)->from.number.valid ? ast_channel_redirecting(ast)->reason.code : -1;
+
+			/* Call Qualifier */
+			ast_channel_lock(ast);
+			/* XXX In the future, we may want to make this a CALLERID or CHANNEL property and fetch it from there. */
+			qual_var = pbx_builtin_getvar_helper(ast, "CALL_QUALIFIER");
```
jcolp, line 1108: This is not documented anywhere.

channels/sig_analog.c
```diff
@@ -1087,12 +1089,29 @@ int analog_call(struct analog_pvt *p, struct ast_channel *ast, const char *rdest
...
+			if (!ast_strlen_zero(qual_var) && !strcmp(qual_var, "1")) {
```
jcolp, line 1109: This can just use ast_true

jcolp (changes requested):

channels/sig_analog.c
InterLinked1: What would be the best way to document it? It's not part of an application, so there's not really an appropriate home I can think of for it in the xmldocs. It isn't a CHANNEL function variable, it's not even chan_dahdi specific, per se.

It is Caller ID related, however. Would it be appropriate to document it in `func_callerid`, even though it's not currently set there? Might also be useful if it's eventually migrated to be part of the caller ID structure.

jcolp: It is currently only applicable in this case, so it is chan_dahdi specific therefore honestly probably multiple places. The DAHDI configuration file and func_callerid with explicit statement that it only works on DAHDI.

jcolp (approved):

gtjoseph (approved):

Issue #96, opened 2023-05-18
[bug]: make install-logrotate causes logrotate to fail on service restart


InterLinked1:
### Severity

Minor

### Versions

20.1.0

### Components/Modules

asterisk.logrotate

### Operating Environment

Debian 11

### Frequency of Occurrence

Constant

### Issue Description

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-30442

Commit 55c53de022db2237723d57eb87305d709e4c808c changed the first line of contrib/scripts/asterisk.logrotate to read

``__LOGDIR__/debug.log __LOGDIR__/console __LOGDIR__/full.log __LOGDIR__/messages.log __LOGDIR__/*log {``

Installing asterisk and running the service creates the file /var/log/asterisk/messages.log.
Running make install-logrotate installs /etc/logrorate.d/asterisk from the above template. Then running systemctl restart logrotate results in a failed service for logrotate. Manually calling logrotate from the command line reveals the error is

```
error: asterisk:1 duplicate log entry for /var/log/asterisk/messages.log
error: found error in file asterisk, skipping
```

The "*log" in the first line of the logrotate config is expanding to messages.log, which now means that file is listed twice, and the configuration is bad. The obvious fix is to remove all explicit .log files and just allow the file globbing to take care of that.

``__LOGDIR__/console __LOGDIR__/*.log {``

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #97, opened 2023-05-18
logrotate: Fix duplicate log entries.

InterLinked1:
The Asterisk logrotate script contains explicit
references to files with the .log extension,
which are also included when *log is expanded.
This causes issues with newer versions of logrotate. This fixes this by ensuring that a log file cannot be referenced multiple times after expansion occurs.

Resolves: #96
ASTERISK-30442
Reported by: EN Barnett
Tested by: EN Barnett

Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19929


InterLinked1:
cherry-pick-to: 20

jcolp (approved):

Issue #98, opened 2023-05-18
[new-feature]: callerid: Allow timezone to be specified at runtime


InterLinked1:
### Feature Description

The time zone used for Caller ID generation is currently hardcoded to the system time zone, and cannot be specified by modules using the Caller ID function. This makes it possible to optionally specify a tz format timezone.

PR #99, opened 2023-05-18
callerid: Allow specifying timezone for date/time.

InterLinked1:
The Caller ID generation routine currently is hardcoded to always use the system time zone. This makes it possible to optionally specify any TZ-format time zone.

Resolves: #98
ASTERISK-30330

Imported from: https://gerrit.asterisk.org/c/asterisk/+/19600


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 18

Issue #100, opened 2023-05-18
[bug]: sig_analog: hidecallerid setting is broken


InterLinked1:
### Severity

Minor

### Versions

20.0.1

### Components/Modules

sig_analog

### Operating Environment

Debian

### Frequency of Occurrence

Constant

### Issue Description

The hidecallerid option in chan_dahdi does not currently work properly. This fixes it.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #101, opened 2023-05-18
chan_dahdi: Fix broken hidecallerid setting.

InterLinked1:
The hidecallerid setting in chan_dahdi.conf currently is broken for a couple reasons.

First, the actual code in sig_analog to "allow" or "block" Caller ID depending on this setting improperly used ast_set_callerid instead of updating the presentation. This issue was mostly fixed in ASTERISK_29991, and that fix is carried forward to this code as well.

Secondly, the hidecallerid setting is set on the DAHDI pvt but not carried forward to the analog pvt properly. This is because the chan_dahdi config loading code improperly set permhidecallerid to permhidecallerid from the config file, even though hidecallerid is what is actually set from the config file. (This is done correctly for call waiting, a few lines above.) This is fixed to read the proper value.

Thirdly, in sig_analog, hidecallerid is set to permhidecallerid only on hangup. This can lead to potential security vulnerabilities as an allowed Caller ID from an initial call can "leak" into subsequent calls if no hangup occurs between them. This is fixed by setting hidecallerid to permcallerid when calls begin, rather than when they end. This also means we don't need to also set hidecallerid in chan_dahdi.c when copying from the config, as we would have to otherwise.

Fourthly, sig_analog currently only allows dialing *67 or *82 if that would actually toggle the presentation. A comment is added clarifying that this behavior is okay.

Finally, a couple log messages are updated to be more accurate.

Resolves: #100
ASTERISK-30349 #close
Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19707

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -13178,7 +13178,8 @@ static struct dahdi_pvt *mkintf(int channel, const struct dahdi_chan_conf *conf,
 				analog_p->canpark = conf->chan.canpark;
 				analog_p->dahditrcallerid = conf->chan.dahditrcallerid;
 				analog_p->immediate = conf->chan.immediate;
-				analog_p->permhidecallerid = conf->chan.permhidecallerid;
+				analog_p->permhidecallerid = conf->chan.hidecallerid; /* hidecallerid is the config setting, not permhidecallerid (~permcallwaiting above) */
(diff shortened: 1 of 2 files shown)


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 18

gtjoseph:
The unit test failure is my fault.  I'll re-run when it's fixed.

gtjoseph (approved):

Issue #104, opened 2023-05-18
[improvement]: Add AMI action to get a list of connected channels


bkford:
### Improvement Description

Adding an AMI action that will allow you to specify a channel and get all channels that are connected to it, including local channels and what they are connected to

PR #105, opened 2023-05-18
AMI: Add CoreShowChannelMap action.

bkford:
Adds a new AMI action (CoreShowChannelMap) that takes in a channel name
and provides a list of all channels that are connected to that channel,
following local channel connections as well.

Resolves: #104

UserNote: New AMI action CoreShowChannelMap has been added.

### DIFF
--- a/main/manager.c
+++ b/main/manager.c
@@ -68,6 +68,7 @@
 #include "asterisk/module.h"
 #include "asterisk/config.h"
 #include "asterisk/callerid.h"
+#include "asterisk/core_local.h"
 #include "asterisk/lock.h"
 #include "asterisk/cli.h"
 #include "asterisk/app.h"
@@ -1040,6 +1041,33 @@
 			</list-elements>
 			<xi:include xpointer="xpointer(/docs/managerEvent[@name='CoreShowChannelsComplete'])" />
 		</responses>
+	</manager>
+	<managerEvent language="en_US" name="CoreShowChannelMapComplete">
+		<managerEventInstance class="EVENT_FLAG_CALL">
+			<synopsis>Raised at the end of the CoreShowChannelMap list produced by the CoreShowChannelMap command.</synopsis>
+			<syntax>
+				<parameter name="EventList">
+					<para>Conveys the status of the command response list</para>
+				</parameter>
+				<parameter name="ListItems">
+					<para>The total number of list items produced</para>
+				</parameter>
+			</syntax>
+		</managerEventInstance>
+	</managerEvent>
+	<manager name="CoreShowChannelMap" language="en_US">
+		<synopsis>
+			List all channels connected to the specified channel.
+		</synopsis>
+		<syntax>
+			<parameter name="Channel">
+				<para>The channel to get the mapping for. Requires a channel name.</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>List all channels currently connected to the specified channel. This can be any channel, including
+			Local channels, and Local channels will be followed through to their other half.</para>
+		</description>
 	</manager>
 	<manager name="LoggerRotate" language="en_US">
 		<synopsis>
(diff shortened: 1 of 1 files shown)


main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+{
+	char *str;
+	int size;
+
+	/* If we've already added this channel, don't add it again */
+	if (AST_VECTOR_GET_CMP(v, s, !strcmp)) {
+                return -1;
+        }
+
+	size = strlen(s) + 1;
```
InterLinked1, line 6922: I'm probably missing something here, but what's the reason for using malloc + ast_copy_string rather than ast_strdup?

Also, if the allocation succeeds, there is no need to bounds check. strcpy would be safe.

main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+		}
+
+		/* If this is a local channel that we haven't seen yet, let's go ahead and find out what else is connected to it */
+		if (!add_channel_res && ast_begins_with(current_channel_snapshot->base->name, "Local")) {
+			struct ast_channel_snapshot *other_local_snapshot;
+			struct ast_bridge_snapshot *other_bridge_snapshot;
+			char *other_local;
+			int size = strlen(current_channel_snapshot->base->name);
+
+			other_local = ast_malloc(size + 1);
```
InterLinked1, line 6980: Same question here regarding strdup...

main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+		current_channel_snapshot = ast_channel_snapshot_get_latest(current_channel_uid);
+		if (!current_channel_snapshot) {
+			ast_log(LOG_ERROR, "Unable to get snapshot of channel\n");
+			res = 1;
+			break;
+		}
+
+		add_channel_res = coreshowchannelmap_add_to_map(channel_map, current_channel_snapshot->base->name);
+		if (add_channel_res > 0) {
+			ast_log(LOG_ERROR, "Unable to add channel to vector\n");
```
InterLinked1, line 6967: coreshowchannelmap_add_to_map already logs an error if res > 0, so the additional error log here is redundant

main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+			other_local_snapshot = ast_channel_snapshot_get_latest_by_name(other_local);
+			if (!other_local_snapshot) {
+				ast_log(LOG_ERROR, "Unable to get other local channel\n");
+				ao2_ref(current_channel_snapshot, -1);
+				ast_free(other_local);
+				return 1;
+			}
+
+			if (coreshowchannelmap_add_to_map(channel_map, other_local_snapshot->base->name)) {
+				ast_log(LOG_ERROR, "Unable to add channel to vector\n");
```
InterLinked1, line 7001: Is this log also redundant? If res > 0, something is already logged, and if res == -1, it was already in the vector so this is misleading.

main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+				ast_free(other_local);
+				ao2_cleanup(other_local_snapshot);
+				break;
+			}
+
+			other_bridge_snapshot = ast_bridge_get_snapshot_by_uniqueid(other_local_snapshot->bridge->id);
+			if (other_bridge_snapshot) {
+				res = coreshowchannelmap_add_connected_channels(channel_map, other_local_snapshot, other_bridge_snapshot);
+			}
+
```
InterLinked1, line 7013: Unnecessary extra line here

InterLinked1 (changes requested):

main/manager.c
```diff
@@ -1041,6 +1042,32 @@
...
+			</syntax>
+		</managerEventInstance>
+	</managerEvent>
+	<manager name="CoreShowChannelMap" language="en_US">
+		<synopsis>
+			List all channels connected to the specified channel.
+		</synopsis>
+		<syntax>
+			<parameter name="Channel">
+				<para>The channel to get the mapping for</para>
```
jcolp, line 1064: Specify that this accepts a channel name

main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+
+	channel_snapshot = ast_channel_snapshot_get_latest_by_name(channel_name);
+	if (!channel_snapshot) {
+		astman_send_error(s, m, "Could not get channel snapshot\n");
+		return 0;
+	}
+
+	bridge_snapshot = ast_bridge_get_snapshot_by_uniqueid(channel_snapshot->bridge->id);
+	if (!bridge_snapshot) {
+		astman_send_error(s, m, "Channel is not in a bridge\n");
```
jcolp, line 7061: I think it is better from a user perspective to simply behave as if the channel is not connected to anything, and not return an error.

main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+	RAII_VAR(struct ast_channel_snapshot *, channel_snapshot, NULL, ao2_cleanup);
+
+	if (!ast_strlen_zero(actionid)) {
+		snprintf(id_text, sizeof(id_text), "ActionID: %s\r\n", actionid);
+	} else {
+		id_text[0] = '\0';
+	}
+
+	if (ast_strlen_zero(channel_name)) {
+		ast_log(LOG_ERROR, "CoreShowChannelMap requires a channel.\n");
```
jcolp, line 7049: This should be an AMI error

main/manager.c
```diff
@@ -6873,6 +6900,203 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+		return 0;
+	}
+
+	bridge_snapshot = ast_bridge_get_snapshot_by_uniqueid(channel_snapshot->bridge->id);
+	if (!bridge_snapshot) {
+		astman_send_error(s, m, "Channel is not in a bridge\n");
+		return 0;
+	}
+
+	channel_map = ao2_alloc_options(sizeof(struct ast_vector_string), channel_map_vector_destructor, AO2_ALLOC_OPT_LOCK_NOLOCK);
```
jcolp, line 7065: I think instead of using an ao2_alloced vector string it would be much easier to allocate an ao2 string container:

`channel_map = ast_str_container_alloc_options(AO2_ALLOC_OPT_LOCK_NOLOCK | AO2_CONTAINER_ALLOC_OPT_DUPS_OBJ_REJECT, 1);`

This will also reject duplicates.

and then:

`ast_str_container_add(channel_map, name);`

main/manager.c
jcolp: I would say you don't even need to do that. Channel names are a fixed maximum size. Just use a defined buffer of that maximum size and copy the existing name in. No having to deal with memory alloc/free in that case.

jcolp (changes requested):

main/manager.c
seanbright: > Channel names are a fixed maximum size

I don't think that is true any longer. They are stringfields on both `struct ast_channel` and `struct ast_channel_snapshot_base` and nothing appears to enforce a maximum length.

jcolp: We still have AST_CHANNEL_NAME for the maximum size, and it's used in various places so that's... fun.

bkford:
cherry-pick-to: 20
cherry-pick-to: 18

main/manager.c
```diff
@@ -6873,6 +6901,175 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+		struct ast_channel_snapshot *current_channel_snapshot;
+		int add_channel_res;
+
+		/* Don't add the original channel to the list - it's either already in there,
+		 * or it's the channel we want the map for */
+		if (!strcmp(current_channel_uid, channel_snapshot->base->uniqueid)) {
+			continue;
+		}
+
+		current_channel_snapshot = ast_channel_snapshot_get_latest(current_channel_uid);
```
jcolp, line 6940: I think in all cases where the channel snapshot can't be found it should be treated as graceful - because it is entirely possible that the underlying channel has gone away between when we got its uniqueid and when we asked for the snapshot.

That includes the Local case down below.

main/manager.c
```diff
@@ -6873,6 +6901,175 @@ static int action_coreshowchannels(struct mansession *s, const struct message *m
...
+
+		add_channel_res = coreshowchannelmap_add_to_map(channel_map, current_channel_snapshot->base->name);
+		if (add_channel_res) {
+			res = 1;
+			ao2_ref(current_channel_snapshot, -1);
+			break;
+		}
+
+		/* If this is a local channel that we haven't seen yet, let's go ahead and find out what else is connected to it */
+		if (!add_channel_res && ast_begins_with(current_channel_snapshot->base->name, "Local")) {
```
jcolp, line 6955: add_channel_res can never be non-zero here

jcolp (changes requested):

main/manager.c
jcolp: If add_channel_res is non-zero then at line 6948 you break out of the loop so line 6955 wouldn't execute. It's only if add_channel_res is zero that it could ever occur.

jcolp (changes requested):

Issue #68, opened 2023-05-08
[bug]: Long running channel iterator (in ChanSpy) blocks CEL HANGUP events of unrelated channels


wdoekes:
### Severity

Major

### Versions

master,20,19,18,16

### Components/Modules

chan_spy, channels

### Operating Environment

Right now, testing with branch 18, but this should be on all versions.

### Frequency of Occurrence

Frequent

### Issue Description

_The `ChanSpy()` channel iterator holds references to channels for the duration of the listening/spying. These referenced channels will get destroyed first when the channel iteration resumes (when done listening to that one channel). This is a problem for the CEL HANGUP event, because all channels still in the iterator will not get any HANGUP reporting until this resumption: anyone relying on CEL HANGUP time for call duration statistics will get wrong values._

A.k.a. CEL HANGUP times can be wrong (late) and appear "batched".

----

**Components**

- ChanSpy -- iterates over all channels using (e.g. ast_channel_iterator_by_name_new) -- keeps the iterator until the spying is done:
  - https://github.com/asterisk/asterisk/blob/1a7866b172b3ef2853d3ab4fb2e2229d29a50082/apps/app_chanspy.c#L956-L964
  - https://github.com/asterisk/asterisk/blob/1a7866b172b3ef2853d3ab4fb2e2229d29a50082/apps/app_chanspy.c#L994-L999
  - https://github.com/asterisk/asterisk/blob/1a7866b172b3ef2853d3ab4fb2e2229d29a50082/apps/app_chanspy.c#L1146
  - https://github.com/asterisk/asterisk/blob/1a7866b172b3ef2853d3ab4fb2e2229d29a50082/apps/app_chanspy.c#L1184

- CEL hangup events -- are generated because of a publish by the ast_channel destructor:
  - https://github.com/asterisk/asterisk/blob/1a7866b172b3ef2853d3ab4fb2e2229d29a50082/main/channel.c#L2206-L2213
  - https://github.com/asterisk/asterisk/blob/1a7866b172b3ef2853d3ab4fb2e2229d29a50082/main/stasis_channels.c#L948-L950
  - https://github.com/asterisk/asterisk/blob/1a7866b172b3ef2853d3ab4fb2e2229d29a50082/main/cel.c#L910-L921

To get accurate HANGUP timings from CEL, channel destruction should never be delayed for more than a short while.

But because some components (ChanSpy) can hold references to channels that are hung up, destruction can be delayed for an indetermined amount of time.

[...]

PR #114, opened 2023-05-24
app_chanspy: Refactor arguments to allow splitting large functions

wdoekes:
Because of upcoming planned changes/refactoring to app_chanspy, it is convenient to pass some of the many arguments as a struct. This changeset adds a channel_spy_context struct to pass around.

Related: #68

### DIFF
--- a/apps/app_chanspy.c
+++ b/apps/app_chanspy.c
@@ -504,7 +504,7 @@ static struct ast_generator spygen = {
 	.generate = spy_generate,
 };
 
-static int start_spying(struct ast_autochan *autochan, const char *spychan_name, struct ast_audiohook *audiohook, struct ast_flags *flags)
+static int start_spying(struct ast_autochan *autochan, const char *spychan_name, struct ast_audiohook *audiohook, const struct ast_flags *flags)
 {
 	int res;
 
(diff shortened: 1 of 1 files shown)


wdoekes:
cherry-pick-to: 20
cherry-pick-to: 19
cherry-pick-to: 18
cherry-pick-to: 16

wdoekes:
- As discussed on IRC, I want the commits applied without squashing (they all compile and run fine if merged in sequence);
- [AsteriskGateTestMatrix (18, pjs2)](https://github.com/asterisk/asterisk/actions/runs/5071396702/jobs/9108016214?pr=114#logs) failed, but that looks unrelated.
- I did not add any `UserNote:` stuff. does not feel needed.
- As for the fix itself (the last commit): I'm open to alternative solutions. I mention memory hoarding. This could be fixed by removing items from the vector that are not in the iterator. But then we'd want to switch to a linkedlist/hashmap instead.

wdoekes:
@jcolp: Do you have a quick example of how to use those snapshots? Some bigger refactoring might be worth it.

We've been running this on prod now, but we did run into 2 similar deadlocks. I cannot prove that this patch caused it, but cannot disprove it either.

The two deadlocks occurred with two threads in:
`got_optimized_out -> try_swap_optimize_out -> bridge_do_move -> bridge_complete_join -> bridge_channel_complete_join -> simple_bridge_join -> ast_channel_request_stream_topology_change -> simple_bridge_join -> ast_channel_request_stream_topology_change -> unreal_colp_stream_topology_request_change -> ast_unreal_lock_all`
According to the core dumps we collected, they were stuck at `ast_channel_lock_both(p->chan, p->owner)`. Both threads were supposedly already owner of one of the locks. Either the core dump misinformed, or that shouldn't happen. (Especially the thread that was holding on to p->owner but not p->chan, because it causes a locking inversion.)

If this changeset is to blame, I would put my money on some ast_autochan behaviour (swapping (locked?) channels?), but I don't really see how I changed anything in this respect... especially since there was only one ChanSpy invocation in one of these Asterisk runs, and it was 13 hours prior to the deadlock.

jcolp:
I'm not sure what example would be applicable for this exactly, but "core show channels" uses the channel snapshot cache and snapshots. It doesn't pull from the channels container.

Issue #120, opened 2023-05-25
[bug]: chan_dahdi: Fix broken presentation for FXO caller ID


InterLinked1:
### Severity

Major

### Versions

20.0.0

### Components/Modules

chan_dahdi

### Operating Environment

Debian

### Frequency of Occurrence

Constant

### Issue Description

Caller ID presentation is always set to "available" on incoming calls to FXO ports, even though it's not. This fixes this by taking the appropriately set presentation flag and setting the presentation on the channel.

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-30333

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #121, opened 2023-05-25
chan_dahdi: Fix Caller ID presentation for FXO ports.

InterLinked1:
Currently, the presentation for incoming channels is always available, because it is never actually set, meaning the channel presentation can be nonsensical. If the presentation from the incoming Caller ID spill is private or unavailable, we now update the channel presentation to reflect this.

Resolves: #120
ASTERISK-30333
ASTERISK-21741

Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19603


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (approved):

PR #128, opened 2023-05-28
res_pjsip_device_features: Add forwarding and DND synchronization.

### LINKED ISSUE #127: [new-feature]: res_pjsip_device_features: Add PJSIP support for synchronized forwarding and DND
### Feature Description

This is the first of a couple new features made possible by #81. This adds support for synchronized call forwarding and do not disturb features to PJSIP, which many IP phones (e.g. Polycom SoundPoint, VVX) support, providing for a significantly better user experience. This is implemented according to the official Broadworks specifications and designed to be drop-in compatible for Broadworks.

This also helps bridge feature parity as this has been achievable for a long time using the Call Manager patches for chan_sip.

InterLinked1:
This adds support for synchronization of device features, specifically call forwarding, ring control, and Do Not Disturb, according to the Broadworks specifications for device feature key synchronization. This allows Asterisk to support IP phones that support device feature key sync the same way that a Broadworks server would.

The implementation of this feature is designed to be flexible enough to allow administrators to configure their own process for handling feature toggle and set requests. To allow this, AMI events and actions are present to interface with an administrator's business logic.

Resolves: #127

UpgradeNote: Device feature key sync for call forwarding and Do Not Disturb has been added to PJSIP.


res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
...
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
+ * at the top of the source tree.
+ */
+
+/*** MODULEINFO
+	<depend>pjproject</depend>
+	<depend>res_pjsip</depend>
+	<depend>res_pjsip_pubsub</depend>
+	<support_level>core</support_level>
```
jcolp, line 23: This will not be core supported.

InterLinked1: Ah, thanks for the catch. I knew that of course, but missed that when I copy/pasted.

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+#include "asterisk/res_pjsip_pubsub.h"
+#include "asterisk/res_pjsip_body_generator_types.h"
+#include "asterisk/module.h"
+#include "asterisk/logger.h"
+#include "asterisk/astobj2.h"
+#include "asterisk/sorcery.h"
+#include "asterisk/xml.h"
+#include "asterisk/taskprocessor.h"
+#include "asterisk/app.h"
+#include "asterisk/astdb.h"
```
alex2grad, line 50: Why use old AstDB instead of Sorcery?
I think all new modules especially PJSIP modules should use Sorcery.
In case of Sorcery the user can choose where to store synchronized data: astdb, memory, realtime DB, memory_cache.
All PJSIP objects are stored using Sorcery, so these PJSIP device features should also be stored using Sorcery.

alex2grad (changes requested):

res/res_pjsip_device_features.c
InterLinked1: AstDB is mainly used as a cache between restarts, the same way PJSIP uses it for subscriptions, for example. It's really the admin's responsibility to store them in the desired place anyways as part of the approval process, which is outside of the module.

alex2grad: The PJSIP subscription_persistence uses Sorcery and astdb storage by default.
ast_sorcery_apply_default(sorcery, "subscription_persistence", "astdb", "subscription_persistence");
But the user can choose another storage.
As I know nor PJSIP core neither PJSIP modules use astdb directly.

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+	unsigned int have_dnd:1;
+	unsigned int have_callforwardalways:1;
+	unsigned int have_callforwardbusy:1;
+	unsigned int have_callforwardnoanswer:1;
+};
+
+/* Whether we should auto-approve device feature changes.
+ * If disabled, system will use the AMI event to update feature data and write PJSIP_DEVICE_FEATURES to trigger a NOTIFY.
+ * If enabled, we will "auto approve": internal cache will automatically update per the request and a NOTIFY will be sent. */
+static int auto_approve = 0;
```
alex2grad, line 126: It's constant. I could not find the code which can change this.
I think this should be PJSIP Endpoint parameter, it means some endpoints are allowed to change it automatically, some endpoints not.

alex2grad (changes requested):

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+	ast_sip_subscription_notify(task_data->feature_state_sub->sip_sub, &data, task_data->terminate);
+	return 0;
+}
+
+#define NUM_FEATURES 4
+static int send_update(struct feature_state_subscription *feature_state_sub)
+{
+	int i;
+
+	/* XXX This is kind of a kludge. In order to send updates for more than one feature,
```
alex2grad, line 578: Why do this "kind of a kludge" and send many NOTIFYs instead of just one with multipart?
Please look at pjsip_multipart_* functions.
https://www.pjsip.org/pjsip/docs/html/group__PJSIP__MULTIPART.htm

alex2grad (changes requested):

res/res_pjsip_device_features.c
alex2grad: According to https://www.cisco.com/c/dam/en/us/td/docs/voice_ip_comm/broadworks/FD/AS/DeviceFeatureKeySynchronizationFD.pdf
2.2 Call Flows
2.2.1 Initial SUBSCRIBE and NOTIFY
"When a supported phone powers up, it sends a SUBSCRIBE request to the Application
Server to create a subscription for the as-feature-event package and to get the **full** feature
status. "

"The Application Server sends a NOTIFY request to the phone
with a body that contains the **complete** feature status, and the phone sends a 200
response. The body is a **MIME multipart body**, containing an XML document for **each**
feature assigned to the subscriber".

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2022, Naveen Albert
```
jcolp, line 4: 2023

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+		</synopsis>
+		<syntax>
+			<parameter name="endpoint" required="true">
+				<para>Name of PJSIP endpoint, without the technology.</para>
+			</parameter>
+			<parameter name="feature" required="true">
+				<para>Name of the feature.</para>
+				<para>Must be one of:</para>
+				<enumlist>
+					<enum name="donotdisturb"/>
```
jcolp, line 68: I think the words should be separated by _ ala do_not_disturb

res/res_pjsip_device_features.c
jcolp: This absolutely needs to be configurable per-endpoint.

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+static int send_ami(const char *device_id, const char *endpoint, const char *number, const char *feature, int status, int ring_count, const char *data)
+{
+	/*** DOCUMENTATION
+		<managerEvent language="en_US" name="DeviceFeatureSync">
+			<managerEventInstance class="EVENT_FLAG_CALL">
+				<synopsis>Raised when a SIP device tries to synchronize a server-side feature</synopsis>
+					<syntax>
+						<parameter name="DeviceID">
+							<para>The unique device ID sent by the endpoint.</para>
+							<para>The same device ID can be used in NOTIFY requests, but this is not used within Broadworks.</para>
```
jcolp, line 166: What does this comment regarding NOTIFY have to do with this event exactly?

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+						</parameter>
+						<parameter name="Endpoint">
+							<para>The endpoint (device name).</para>
+						</parameter>
+						<parameter name="CallerIDNum">
```
jcolp, line 171: How is the caller ID number determined?

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+						</parameter>
+						<parameter name="Data">
+							<para>For call forwarding enablement, the forward-to number.</para>
+						</parameter>
+					</syntax>
+			</managerEventInstance>
+		</managerEvent>
+	***/
+
+	if (status != 0 && status != 1) {
```
jcolp, line 197: This needs to be better commented about... why these are invalid

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+static int parse_incoming_xml(struct feature_state_subscription *feature_state_sub, char *xmlbody, int xmllen, const char *endpoint_name, const char *number)
+{
+	int res = 0;
+	const char *nodename;
+	const char *tmpstr;
+	struct ast_xml_node *root, *tmpnode = NULL;
+	struct ast_xml_doc *xmldoc = ast_xml_read_memory(xmlbody, xmllen);
+
+	if (!xmldoc) {
+		ast_log(LOG_WARNING, "Failed to parse as XML: %.*s\n", xmllen, xmlbody);
```
jcolp, line 237: This needs to be more detailed for a user, such as including the endpoint name and what actually happened to cause this to occur

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+		tmpnode = ast_xml_find_child_element(root, "doNotDisturbOn", NULL, NULL);
+		tmpstr = ast_xml_get_text(tmpnode);
+		dnd_disp = ast_true(tmpstr) ? 1 : 0; /* ast_true returns -1 on success */
+		ast_xml_free_text(tmpstr);
+
+		if (!res) {
+			res = send_ami(feature_state_sub->sync_data.deviceid, endpoint_name, number, "DoNotDisturb", dnd_disp, 0, NULL);
+		}
+		if (!res) {
+			/* Note: We do NOT set sync_data.dnd = dnd_disp directly here! That's not our job!!!
```
jcolp, line 299: There needs to be a large comment at the top of this module detailing specifically how it works, the interactions, and the responsibilities of all entities involved.

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+	struct feature_state_subscription *feature_state_sub;
+	struct subscription_item *subitem;
+	char tmp[16];
+	const char *endpoint_name = ast_sorcery_object_get_id(endpoint);
+
+	subitem = ast_calloc(1, sizeof(*subitem));
+	if (!subitem) {
+		return NULL;
+	}
+	subitem->endpoint = ast_strdup(endpoint_name);
```
jcolp, line 464: This doesn't need to be a separate allocation, it can be allocated as part of subitem

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+	if (!subitem->endpoint) {
+		ast_free(subitem);
+		return NULL;
+	}
+
+	feature_state_sub = ao2_alloc(sizeof(*feature_state_sub), feature_state_subscription_destructor);
```
jcolp, line 470: Does this need to be allocated with a lock?

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+	if (ast_sip_subscription_is_terminated(feature_state_sub->sip_sub)) {
+		ast_log(LOG_WARNING, "Subscription for %s is already terminated\n", args.endpoint);
+		ao2_ref(feature_state_sub, -1);
+		return -1;
+	}
+
+	/* Get the new status, cache in AstDB, and then send a NOTIFY. */
+	if (!strcasecmp(args.feature, "donotdisturb")) {
+		int new_dnd_status = ast_true(args2.status) ? 1 : 0; /* Use 1, not -1 */
+		DEVICE_FEATURE_UPDATE(args.endpoint, "donotdisturb", update_needed_dnd, dnd, new_dnd_status, new_dnd_status ? "1" : "0");
```
jcolp, line 752: There is no protection or serialization around these. Is it possible for updates to occur at the same time?

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+	}
+
+	ast_debug(2, "%d subscription(s) added for %s\n", added_subscriptions, ast_sorcery_object_get_id(endpoint));
+	send_update(feature_state_sub); /* Send it all of its current settings. */
+
+	ao2_cleanup(feature_state_sub);
+	return 0;
+}
+
+static void sync_data_destructor(void *obj)
```
jcolp, line 948: This is unused and doesn't need to exist

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+{
+	//struct ast_sip_device_feature_sync_data *sync_data = obj;
+	return;
+}
+
+static struct ast_sip_device_feature_sync_data *sync_data_alloc(struct ast_sip_subscription *sip_sub, struct feature_state_subscription *feature_state_sub)
+{
+	struct ast_sip_device_feature_sync_data *sync_data;
+	sync_data = ao2_alloc(sizeof(*sync_data), sync_data_destructor);
```
jcolp, line 957: Does this need a lock? What is this used for?

res/res_pjsip_device_features.c
```diff
@@ -0,0 +1,1005 @@
...
+
+static void *get_notify_data(struct ast_sip_subscription *sub)
+{
+	struct feature_state_subscription *feature_state_sub;
+
+	feature_state_sub = get_feature_state_sub(sub);
+	if (!feature_state_sub) {
+		return NULL;
+	}
+	return sync_data_alloc(sub, feature_state_sub);
```
jcolp, line 969: Why is feature_state_sub passed in, if it is unused?

res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2022, Naveen Albert
```
jcolp, line 4: 2023

res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
...
+	char *buf;
+	int len;
+	int active;
+	struct ast_xml_doc *doc;
+	struct ast_xml_node *root, *tmpnode;
+	const char *fwd_str = forward_type_str(fwdtype);
+
+	ast_assert(!ast_strlen_zero(fwd_str));
+
+	ast_debug(1, "ForwardingEvent update required (%s)\n", fwd_str);
```
jcolp, line 89: This should include more information

res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
...
+	doc = ast_xml_new();
+	if (!doc) {
+		ast_log(LOG_ERROR, "Could not create new XML document\n");
```
jcolp, line 92: This is meaningless to a user

res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
...
+	 */
+
+	tmpnode = ast_xml_new_node("device");
+	if (!tmpnode) {
+		goto cleanup;
+	}
+	/* The Broadworks spec says device is mandatory, but the actual value is used by neither server nor client. So if we don't have one, make something up. */
+	ast_xml_set_text(tmpnode, S_OR(sync_data->deviceid, "123456"));
+	if (!sync_data->deviceid[0]) {
+		ast_debug(2, "This was the first NOTIFY with body data for this endpoint\n");
```
jcolp, line 136: This needs more information (such as endpoint)

res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
...
+		}
+		ast_xml_set_text(tmpnode, forward_to);
+		ast_xml_add_child(root, tmpnode);
+	} /* else, forwarding is not active. No additional info needed. */
+
+	/* Finalize. */
+	ast_xml_doc_dump_memory(doc, &buf, &len);
+	ast_xml_close(doc);
+	if (len <= 0) {
+		ast_log(LOG_WARNING, "XML document has length %d?\n", len);
```
jcolp, line 169: This is meaningless to a user

res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
...
+	}
+	if (buf) {
+		ast_str_append(bodytext, 0, "%s", buf);
+		ast_xml_free_text(buf);
+	}
+	return 0;
+cleanup:
+	ast_xml_close(doc);
+	ast_log(LOG_ERROR, "Could not create new XML root node\n");
```
jcolp, line 178: This is meaningless to a user

res/res_pjsip_device_features_body_generator.c
```diff
@@ -0,0 +1,325 @@
...
+	int len;
+	struct ast_str **bodytext = body;
+	struct ast_xml_doc *doc;
+	struct ast_xml_node *root, *tmpnode;
+	struct ast_sip_device_feature_sync_data *sync_data = data;
+
+	int updates_made = 0;
+
+	/* This callback is called for *all* NOTIFYs, so we should only add XML to the body if actually necessary. */
+	ast_debug(2, "Generating body content for %s/%s\n", FEATURE_TYPE, FEATURE_SUBTYPE);
```
jcolp, line 194: This needs more information such as the endpoint, and so on for other log messages, and meaningless messages for users too

jcolp (changes requested):

jcolp:
I also agree with Alexei's comments

include/asterisk/res_pjsip_body_generator_types.h
```diff
@@ -74,4 +74,30 @@ struct ast_sip_message_accumulator {
...
 };
 
+/*!
+ * \brief Synchronization data used for as-feature-event XML bodies
+ *
+ * This is used for x-as-feature-event+xml bodies.
+ */
+struct ast_sip_device_feature_sync_data {
+	/* Device ID */
+	char deviceid[64];
```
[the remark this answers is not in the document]
InterLinked1: It would hold the device name (channel name before the -).
Where is the #define for the max device name length? I'd been looking for that for a long time and could never find one. I know you and Sean were discussing the one for channel name length, but that would be longer than necessary here.

res/res_pjsip_device_features.c
alex2grad: The endpoint property. For example

ast_sorcery_object_field_register(sip_sorcery, "endpoint", "allow_device_features_sync", "no", OPT_BOOL_T, 1, FLDSET(struct ast_sip_endpoint, allow_device_features_sync));

include/asterisk/res_pjsip_body_generator_types.h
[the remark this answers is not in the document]
alex2grad: This "struct ast_sip_device_feature_sync_data" should be the sorcery object. For example
struct ast_sip_device_feature_sync_data {
  /*! Sorcery object details, the id is the ENDPOINT name */
  SORCERY_OBJECT(details);
  AST_DECLARE_STRING_FIELDS(
    AST_STRING_FIELD(fwd_exten_always);
    AST_STRING_FIELD(fwd_exten_busy);
    AST_STRING_FIELD(fwd_exten_noanswer);
  );
  unsigned int dnd;
  unsigned int ring_count;
};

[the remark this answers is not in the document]
alex2grad: Sorry, forget deviceid
AST_STRING_FIELD(deviceid);

alex2grad:
Could you please explain "administrator's business logic."?

As I understand with this module the SIP phones can synchronize with the Application Server the status of the features: DND, CFA, CFB, and CFNA, i.e. if a user changes the status of one of these features via web portal the Application Server notifies the phone about the status change. Conversely, if the user changes the feature status via a button or a menu on the phone, the phone notifies the Application Server of the status change.

I see the business logic as:
If the SIP endpoint is not allowed to use the Device Feature Key Synchronization feature then the Asterisk rejects all such requests from SIP endpoint.
If the Asterisk receives the status change from SIP endpoint, the Asterisk saves the status change and notifies the Application Server using AMI event.
If the Asterisk receives the status change from the Application Server via AMI command, the Asterisk saves the status change and notifies the SIP endpoint via SIP NOTIFY.

InterLinked1:
> Could you please explain "administrator's business logic."?
> 
> As I understand with this module the SIP phones can synchronize with the Application Server the status of the features: DND, CFA, CFB, and CFNA, i.e. if a user changes the status of one of these features via web portal the Application Server notifies the phone about the status change. Conversely, if the user changes the feature status via a button or a menu on the phone, the phone notifies the Application Server of the status change.
> 
> I see the business logic as: If the SIP endpoint is not allowed to use the Device Feature Key Synchronization feature then the Asterisk rejects all such requests from SIP endpoint. If the Asterisk receives the status change from SIP endpoint, the Asterisk saves the status change and notifies the Application Server using AMI event. If the Asterisk receives the status change from the Application Server via AMI command, the Asterisk saves the status change and notifies the SIP endpoint via SIP NOTIFY.

Business logic here meaning that Asterisk itself may neither be capable of approving requests itself nor be the source of the truth for this information.

The TL;DR is that this module does not fully control any of these features. For example, I may have an IP phone and an analog DAHDI line that share the same feature state, and there may be dialing codes and other mechanisms to control the features. How they do that is totally implementation dependent.

The PJSIP module here is only to handle the SIP side of synchronized feature state. It is not *really* intended to handle the feature implementations themselves. In theory it could (that's basically what auto approve is), but that is simplistic and overly limiting, and so control is explicitly configurable by the administrator if desired.

As an example, I'll describe my implementation. The PJSIP module here is only part of how the system works. I can activate Do Not Disturb by dialing *77, or toggling the key on a Polycom phone, or a Recent Change message or other out of band mechanism that updates the switch's internal database (which is outside of Asterisk). All of these processes need to agree upon a common way of communicating. This is not dictated (and should not be dictated) by the PJSIP module, In my case, my feature data is permanently stored in the switch database and certain small things in AstDB (like whether a feature is active or not), but the actual live status lives in memory in a proprietary Asterisk module (it's literally a single bit whether DND is active for a particular "line"... line being a logical construct in my system that Asterisk itself doesn't know about). I need this to be updated whenever any of the above mechanisms update the feature status. Otherwise, the PJSIP module can "think" the feature is toggled, and it internally might think it is, but it's wrong because this module doesn't actually know or control the feature status directly. When a call goes to a line, it will check the Do Not Disturb bit in memory, it doesn't care or know about the PJSIP module at all. That's business logic because everyone might implement this differently.

Similar things apply to all the other settings, too. And in the reverse direction, if I dial *87 on a phone, PJSIP needs to send a NOTIFY to the phone that the feature is turned off. Of course, it won't know my business logic, so I'll need to tell it explicitly about that. That's why there's an interface to expose activation/deactivation requests and set the status for the device: how it interacts with the rest of the system will vary greatly.

alex2grad:
Also big question about multi subscriptions/registrations.
For example if the SIP endpoint has 2 registrations and both of which support the feature synchronization then the Asterisk should send NOTIFY to both subscriptions on the status change.

InterLinked1:
> Also big question about multi subscriptions/registrations. For example if the SIP endpoint has 2 registrations and both of which support the feature synchronization then the Asterisk should send NOTIFY to both subscriptions on the status change.

Multiple contacts on the same AOR? That should already work by default I believe.

And even if they are separate AORs, it's easy to achieve in business logic. You can just look up all the devices that should share a common DND state and then send them all updates. But again, that's business logic, not built in.

alex2grad:
> > Could you please explain "administrator's business logic."?
> > As I understand with this module the SIP phones can synchronize with the Application Server the status of the features: DND, CFA, CFB, and CFNA, i.e. if a user changes the status of one of these features via web portal the Application Server notifies the phone about the status change. Conversely, if the user changes the feature status via a button or a menu on the phone, the phone notifies the Application Server of the status change.
> > I see the business logic as: If the SIP endpoint is not allowed to use the Device Feature Key Synchronization feature then the Asterisk rejects all such requests from SIP endpoint. If the Asterisk receives the status change from SIP endpoint, the Asterisk saves the status change and notifies the Application Server using AMI event. If the Asterisk receives the status change from the Application Server via AMI command, the Asterisk saves the status change and notifies the SIP endpoint via SIP NOTIFY.
> 
> Business logic here meaning that Asterisk itself may neither be capable of approving requests itself nor be the source of the truth for this information.

Please read "2.1.3 Assigned or Unassigned Features".
Asterisk should know what features are assigned or unassigned using the Sorcery object for this SIP endpoint which can contain this information. Where the data is stored ("the source of the truth for this information.") and how it's maintained is up to admin ("Business logic"). So the Asterisk is capable of approving requests.

> 
> The TL;DR is that this module does not fully control any of these features. For example, I may have an IP phone and an analog DAHDI line that share the same feature state, and there may be dialing codes and other mechanisms to control the features. How they do that is totally implementation dependent.

The Sorcery data can be modified by any other mechanisms independent from the Asterisk.

> 
> The PJSIP module here is only to handle the SIP side of synchronized feature state. It is not _really_ intended to handle the feature implementations themselves. In theory it could (that's basically what auto approve is), but that is simplistic and overly limiting, and so control is explicitly configurable by the administrator if desired.

The PJSIP module shares this data/information with the Application Server, so both the Asterisk and the Application Server can modified this data and notify about changes.

> As an example, I'll describe my implementation. The PJSIP module here is only part of how the system works. I can activate Do Not Disturb by dialing *77, or toggling the key on a Polycom phone, or a Recent Change message or other out of band mechanism that updates the switch's internal database (which is outside of Asterisk). All of these processes need to agree upon a common way of communicating. This is not dictated (and should not be dictated) by the PJSIP module, In my case, my feature data is permanently stored in the switch database and certain small things in AstDB (like whether a feature is active or not), but the actual live status lives in memory in a proprietary Asterisk module (it's literally a single bit whether DND is active for a particular "line"... line being a logical construct in my system that Asterisk itself doesn't know about). I need this to be updated whenever any of the above mechanisms update the feature status. Otherwise, the PJSIP module can "think" the feature is toggled, and it internally might think it is, but it's wrong because this module doesn't actually know or control the feature status directly. When a call goes to a line, it will check the Do Not Disturb bit in memory, it doesn't care or know about the PJSIP module at all. That's business logic because everyone might implement this differently.
> 

Following you example.
"activate Do Not Disturb by dialing *77" - the Application Server modifies DND state in Sorcery data source and sends AMI request to the Asterisk to notify about this change.
"toggling the key on a Polycom phone" - the Asterisk  modifies DND state in Sorcery data source and sends AMI event which is received and processed by the Application Server to do actions outside of the Asterisk.
"the switch's internal database" - the data (DB table/view) of the information about these features can be shared with the Asterisk using Sorcery. In this case the Application Server can rely on data modifications (for example DB triggers) instead of AMI events.

InterLinked1:
> > > Could you please explain "administrator's business logic."?
> > > As I understand with this module the SIP phones can synchronize with the Application Server the status of the features: DND, CFA, CFB, and CFNA, i.e. if a user changes the status of one of these features via web portal the Application Server notifies the phone about the status change. Conversely, if the user changes the feature status via a button or a menu on the phone, the phone notifies the Application Server of the status change.
> > > I see the business logic as: If the SIP endpoint is not allowed to use the Device Feature Key Synchronization feature then the Asterisk rejects all such requests from SIP endpoint. If the Asterisk receives the status change from SIP endpoint, the Asterisk saves the status change and notifies the Application Server using AMI event. If the Asterisk receives the status change from the Application Server via AMI command, the Asterisk saves the status change and notifies the SIP endpoint via SIP NOTIFY.
> > 
> > 
> > Business logic here meaning that Asterisk itself may neither be capable of approving requests itself nor be the source of the truth for this information.
> 
> Please read "2.1.3 Assigned or Unassigned Features". Asterisk should know what features are assigned or unassigned using the Sorcery object for this SIP endpoint which can contain this information. Where the data is stored ("the source of the truth for this information.") and how it's maintained is up to admin ("Business logic"). So the Asterisk is capable of approving requests.

Not necessarily. Yes if it's purely binary (on/off), but there may be other factors that influence it which change during runtime.

In the simplest auto approve case though, configurable per endpoint (not currently, but it will be), that's how it would work though.

> > The TL;DR is that this module does not fully control any of these features. For example, I may have an IP phone and an analog DAHDI line that share the same feature state, and there may be dialing codes and other mechanisms to control the features. How they do that is totally implementation dependent.
> 
> The Sorcery data can be modified by any other mechanisms independent from the Asterisk.

*If* there's a sorcery backend that supports it. I'm not using anything that would be exposed in a sorcery backend currently, so that doesn't help me.

The manual way already requires something that needs to be set regardless either way (right now, it's a dialplan function).

> > The PJSIP module here is only to handle the SIP side of synchronized feature state. It is not _really_ intended to handle the feature implementations themselves. In theory it could (that's basically what auto approve is), but that is simplistic and overly limiting, and so control is explicitly configurable by the administrator if desired.
> 
> The PJSIP module shares this data/information with the Application Server, so both the Asterisk and the Application Server can modified this data and notify about changes.

The key point here is that I may not want Asterisk to be allowed to change the state. Changes should succeed only if I tell them to.

This is certainly a more complicated use case, and for the auto approve case is irrelevant. But I think in the case that admins opt to not auto approve, they probably have complicated requirements anyways and it's best to be more flexible than less.

> > As an example, I'll describe my implementation. The PJSIP module here is only part of how the system works. I can activate Do Not Disturb by dialing *77, or toggling the key on a Polycom phone, or a Recent Change message or other out of band mechanism that updates the switch's internal database (which is outside of Asterisk). All of these processes need to agree upon a common way of communicating. This is not dictated (and should not be dictated) by the PJSIP module, In my case, my feature data is permanently stored in the switch database and certain small things in AstDB (like whether a feature is active or not), but the actual live status lives in memory in a proprietary Asterisk module (it's literally a single bit whether DND is active for a particular "line"... line being a logical construct in my system that Asterisk itself doesn't know about). I need this to be updated whenever any of the above mechanisms update the feature status. Otherwise, the PJSIP module can "think" the feature is toggled, and it internally might think it is, but it's wrong because this module doesn't actually know or control the feature status directly. When a call goes to a line, it will check the Do Not Disturb bit in memory, it doesn't care or know about the PJSIP module at all. That's business logic because everyone might implement this differently.
> 
> Following you example. "activate Do Not Disturb by dialing *77" - the Application Server modifies DND state in Sorcery data source and sends AMI request to the Asterisk to notify about this change.

What I do currently is change the bit in memory and then call the dialplan function to tell Asterisk of the new state, which then triggers the NOTIFY to go out.

> "toggling the key on a Polycom phone" - the Asterisk modifies DND state in Sorcery data source and sends AMI event which is received and processed by the Application Server to do actions outside of the Asterisk.

No, because maybe I deny changing the setting. In that case, the change should not have any effect and Sorcery data should not change, because it would be wrong.

>"the switch's internal database" - the data (DB table/view) of the information about these features can be shared with the Asterisk using Sorcery. In this case the Application Server can rely on data modifications (for example DB triggers) instead of AMI events.

Not in my case, the database has a long term store of feature assignments but it's not used by Asterisk directly at all. In my case, the information accessed in real time is generally in memory. Unless a Sorcery frontend /driver were built for that, which I have no intention of doing, Sorcery can't work with that at all, to my knowledge (though not having used Sorcery before, I'm not 100% positive).

I *think* we're mostly on the same page here... in the auto-approve mode, Asterisk is the application server (kind of... it doesn't actually implement the feature, but the feature status could be readily checked somehow). When disabled though, all the module should do is notify the application server about something a device has requested, and communicate to a device a feature setting that the application server has told it is the current setting. All the application logic would sit outside of Asterisk itself, so people can implement it however they like.

alex2grad:
There is Sorcery In-Memory storage module.
https://wiki.asterisk.org/wiki/display/AST/Sorcery

InterLinked1:
> There is Sorcery In-Memory storage module. https://wiki.asterisk.org/wiki/display/AST/Sorcery

Good to know, but not really relevant here. That's just a backend that stores settings in memory, it's not how *I* store settings in memory.

Fundamentally there is no getting around in a custom environment (no auto approve), the application server will need to tell Asterisk/PJSIP of the new state to send it the current module. PJSIP cannot be the source of truth, because it isn't the center of the feature, it's just hooking into it. I might run a system without PJSIP at all, and it makes no sense to be storing my feature config in a way that is dictated by PJSIP, Sorcery or otherwise. I am not changing my feature implementation or where that data is stored, full stop. The reasonable thing to do is add hooks for getting events and notifying PJSIP, and nothing else needs to be changed.

Going back to the original point here, I can look into changing the cache from AstDB to Sorcery, since, at least with auto approve enabled, that might matter somewhat. Aside from that, Sorcery doesn't provide any major benefits I can see here, since it isn't the source of truth, in which case the best option is simply the one that is convenient for the programmer, since end users aren't interacting with the cache directly.

asteriskteam:

asteriskteam:

InterLinked1:
@asteriskteam Could someone also reopen this one please? Thank you!

res/res_pjsip_device_features.c
InterLinked1: I don't think there is any current way with res_pjsip_pubsub to do this, though let me know if I'm wrong about this. This only happens when a device comes online, so it's not really a big deal, but could be a small improvement later on along with pubsub modifications.

jcolp:
To be up front, don't expect any of your updates across your larger PRs to be reviewed or decided upon anytime soon.

res/res_pjsip_device_features.c
InterLinked1: Like the documentation says, it's simply the caller ID of the PJSIP endpoint

InterLinked1:
> To be up front, don't expect any of your updates across your larger PRs to be reviewed or decided upon anytime soon.

I didn't expect otherwise, this is just some long overdue work I'm addressing at the moment and updating here. I know you'll get to it eventually...

InterLinked1:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

InterLinked1:
> * [ ]  A change was detected to the pjsip sample configuration files in ./config/samples but no Alembic change was detected. If this PR includes changes that contain new configuration parameters or a change to existing configuration parameters for pjsip, chances are that a change to the realtime database schema is also required. (sample-configs)

I'll hold off on the Alembic change until this is good to go.

Issue #133, opened 2023-05-31
[bug]: unlock channel after moh state access


mbradeen:
### Severity

Trivial

### Versions

18, 20, master

### Components/Modules

res_musiconhold

### Operating Environment

ubuntu

### Frequency of Occurrence

None

### Issue Description

CommUnity UCaaS submission - moh_files_generator releases channel lock while still accessing channel moh state.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #134, opened 2023-05-31
res_musiconhold: avoid moh state access on unlocked chan

mbradeen:
Move channel unlock to after moh state access to avoid
potential unlocked access to state.

Resolves: #133

### DIFF
--- a/res/res_musiconhold.c
+++ b/res/res_musiconhold.c
@@ -458,24 +458,22 @@ static void moh_files_write_format_change(struct ast_channel *chan, void *data)
 
 static int moh_files_generator(struct ast_channel *chan, void *data, int len, int samples)
 {
-	struct moh_files_state *state = ast_channel_music_state(chan);
+	struct moh_files_state *state;
 	struct ast_frame *f = NULL;
-	int res = 0;
+	int res = 0, sample_queue = 0;
 
+	ast_channel_lock(chan);
+	state = ast_channel_music_state(chan);
 	state->sample_queue += samples;
+	/* save the sample queue value for un-locked access */
+	sample_queue = state->sample_queue;
+	ast_channel_unlock(chan);
 
-	while (state->sample_queue > 0) {
+	while (sample_queue > 0) {
 		ast_channel_lock(chan);
 		f = moh_files_readframe(chan);
-
-		/* We need to be sure that we unlock
-		 * the channel prior to calling
-		 * ast_write. Otherwise, the recursive locking
-		 * that occurs can cause deadlocks when using
-		 * indirect channels, like local channels
-		 */
-		ast_channel_unlock(chan);
 		if (!f) {
+			ast_channel_unlock(chan);
 			return -1;
 		}
 
@@ -489,6 +487,19 @@ static int moh_files_generator(struct ast_channel *chan, void *data, int len, in
 		if (ast_format_cmp(f->subclass.format, state->mohwfmt) == AST_FORMAT_CMP_NOT_EQUAL) {
 			ao2_replace(state->mohwfmt, f->subclass.format);
 		}
+
+		/* We need to be sure that we unlock
+		 * the channel prior to calling
+		 * ast_write, but after our references to state
+		 * as it refers to chan->music_state. Update
+		 * sample_queue for our loop
+		 * Otherwise, the recursive locking that occurs
+		 * can cause deadlocks when using indirect
+		 * channels, like local channels
+		 */
+		sample_queue = state->sample_queue;
+		ast_channel_unlock(chan);
+
 		res = ast_write(chan, f);
 		ast_frfree(f);
 		if (res < 0) {


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (dismissed):

seanbright (changes requested):
Can you describe - ideally in both a code comment and the commit message - what this is accomplishing? It's not clear to me how the channel lock and the what `state` are related.

mbradeen:
`ast_channel_music_state` returns `chan->music_state`.  One of our partners ran into issues on their live systems.

seanbright (changes requested):
I _think_ you are saying that the channel lock is being used to protect `state` so you have to hold the channel lock while reading/writing to it.

Unfortunately there are numerous places where unprotected access to `state` occurs throughout the file (including a couple times near the top of the same function you are updating).

The [documentation in `channel.h`](https://github.com/asterisk/asterisk/blob/e0ed0db41f32a47b49a655c737b93f66b292cb04/include/asterisk/channel.h#L4103) indicates that the channel must be locked before `ast_channel_music_state(...)` is called which is not occurring in many places throughout this file.

A potential resolution is to give `struct moh_files_state` its own mutex instead of relying on the channel lock. You would still need to take the channel lock when calling `ast_channel_music_state(...)` and `ast_channel_music_state_set(...)` but any manipulation of the state itself could be down under its own lock.

seanbright:
(as an aside it would still be great to understand what this is actually intended to fix)

jcolp:
> I _think_ you are saying that the channel lock is being used to protect `state` so you have to hold the channel lock while reading/writing to it.
> 
> Unfortunately there are numerous places where unprotected access to `state` occurs throughout the file (including a couple times near the top of the same function you are updating).
> 
> The [documentation in `channel.h`](https://github.com/asterisk/asterisk/blob/e0ed0db41f32a47b49a655c737b93f66b292cb04/include/asterisk/channel.h#L4103) indicates that the channel must be locked before `ast_channel_music_state(...)` is called which is not occurring in many places throughout this file.
> 
> A potential resolution is to give `struct moh_files_state` its own mutex instead of relying on the channel lock. You would still need to take the channel lock when calling `ast_channel_music_state(...)` and `ast_channel_music_state_set(...)` but any manipulation of the state itself could be down under its own lock.

Channel is locked when all generator callbacks are invoked, except for generate (which is on purpose according to a comment in the generator API code). That's why there is some locking in the generate callback. The other cases of starting/stopping MOH also have locking since they're outside of generator land, and appear to do it correctly.

seanbright:
OK so it is just the two cases of missing locks at the beginning of `moh_files_generator(...)` that need to be addressed.

mbradeen:
waiting on feedback from original reporter.

jcolp (approved):

mbradeen:
Example output:
`=======================================================================
=== GIT-master-4099f965e1M
=== Currently Held Locks at Time: 468334:09:29.088484 =================
=======================================================================
===
=== <pending> <lock#> (<file>): <lock type> <line num> <function> <lock name> <lock addr>
=== <locked at>, <failed at>, <held for> (attempts, times locked)
===
=== Thread ID: 0x7f8bff3ca700 LWP:30685 (lock_thread          started at [   83] test_named_lock.c named_lock_test())
=== ---> Lock #0 (test_named_lock.c): MUTEX 49 lock_thread lock 0x7f8c00004ef0
===      468334:09:29.088294, 00:00:00.000000, 00:00:00.000251 (1, 1)
	asterisk __ast_pthread_mutex_lock()
	asterisk __ao2_lock()
	[0x7f8c8dc3a449] test_named_lock.so test_named_lock.c:50 lock_thread()
	asterisk <unknown>()
	[0x7f8cd707b609] libpthread.so.0 pthread_create.c:478 start_thread()
	[0x7f8cd6cb4133] libc.so.6 :0 clone()
=== -------------------------------------------------------------------
===
=== Thread ID: 0x7f8bff34d700 LWP:30686 (lock_thread          started at [   84] test_named_lock.c named_lock_test())
=== ---> Lock #0 (test_named_lock.c): MUTEX 49 lock_thread lock 0x7f8c0c005070
===      468334:09:29.089931, 00:00:00.000000, 00:00:00.026615 (1, 1)
	asterisk __ast_pthread_mutex_lock()
	asterisk __ao2_lock()
	[0x7f8c8dc3a449] test_named_lock.so test_named_lock.c:50 lock_thread()
	asterisk <unknown>()
	[0x7f8cd707b609] libpthread.so.0 pthread_create.c:478 start_thread()
	[0x7f8cd6cb4133] libc.so.6 :0 clone()
=== -------------------------------------------------------------------
===
=======================================================================
`

seanbright:
So `moh_files_generator(...)` is the only function that reads/writes to the `sample_queue` member of `struct moh_files_state` so do we actually need it as part of the struct or can we just use a local variable in that function? That would simplify the locking a bit if that is the case.

jcolp:
> So `moh_files_generator(...)` is the only function that reads/writes to the `sample_queue` member of `struct moh_files_state` so do we actually need it as part of the struct or can we just use a local variable in that function? That would simplify the locking a bit if that is the case.

I believe it exists on state such that if the returned frame has a greater number of samples than is currently needed, it tries its best to not produce more on the next invocation. For example the generator is asked for 160 samples but gets back 320 samples, the next invocation then becomes a noop if 160 is requested again.

I'm not sure this happens in practice (though it may depending on underlying codecs).

seanbright:
> > So `moh_files_generator(...)` is the only function that reads/writes to the `sample_queue` member of `struct moh_files_state` so do we actually need it as part of the struct or can we just use a local variable in that function? That would simplify the locking a bit if that is the case.
> 
> I believe it exists on state such that if the returned frame has a greater number of samples than is currently needed, it tries its best to not produce more on the next invocation. For example the generator is asked for 160 samples but gets back 320 samples, the next invocation then becomes a noop if 160 is requested again.
> 
> I'm not sure this happens in practice (though it may depending on underlying codecs).

The only nominal way out of `moh_files_generate(...)` is with `state->sample_queue` being `0` (or less than `0`). If we leave for any other reason (`return -1`) the generator is stopped, so there wouldn't be a subsequent invocation.

jcolp:
> > > So `moh_files_generator(...)` is the only function that reads/writes to the `sample_queue` member of `struct moh_files_state` so do we actually need it as part of the struct or can we just use a local variable in that function? That would simplify the locking a bit if that is the case.
> > 
> > 
> > I believe it exists on state such that if the returned frame has a greater number of samples than is currently needed, it tries its best to not produce more on the next invocation. For example the generator is asked for 160 samples but gets back 320 samples, the next invocation then becomes a noop if 160 is requested again.
> > I'm not sure this happens in practice (though it may depending on underlying codecs).
> 
> The only nominal way out of `moh_files_generator(...)` is with `state->sample_queue` being `0` (or less than `0`). If we leave for any other reason (`return -1`) the generator is stopped, so there wouldn't be a subsequent invocation.

Right, in the case I mentioned state->sample_queue would go negative.

jcolp:
I've given this some further thought and I think we should keep sample_queue on state to cover the scenario I mentioned, I'm not comfortable making it local and potentially changing that behavior for some people and causing unintendend consequences.

seanbright (approved):

gtjoseph (approved):

Issue #116, opened 2023-05-24
[bug]: SIP Reason: "Call completed elsewhere" no longer propagating


AstersikRoss:
### Severity

Minor

### Versions

18, 20, 21, master

### Components/Modules

res_pjsip_rfc3326

### Operating Environment

Cent OS

### Frequency of Occurrence

None

### Issue Description

Since Gerrit commit 19570 the 'Call completed elsewhere' reason is not passed on to endpoints.

The setup is as follows:

    Endpoint -> Asterisk 1 (via Call Group) -> Asterisk 2 -> Endpoint

To test make a call from one Endpoint on Asterisk 1 which contains a group with two endpoints on Asterisk 2. Answer the call on one of the endpoints in the group. Asterisk 1 sends headers:

    CSeq: 29930 CANCEL
    Reason: SIP;cause=200;text="Call completed elsewhere"
    Reason: Q.850;cause=26
    Max-Forwards: 70
    User-Agent: Asterisk
    Content-Length: 0

Asterisk 2 does not send on the 'Reason' headers.

Asterisk 1 does not contain the patch and passes on the Headers correctly.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

AstersikRoss:
> static void rfc3326_use_reason_header(struct ast_sip_session *session, struct pjsip_rx_data *rdata)

This method appears to remove the cause text from the header and doesn't account for the string 'Call completed elsewhere' being present indicating that the cause code within Asterisk should be set to AST_CAUSE_ANSWERED_ELSEWHERE

seanbright:
@AstersikRoss can you please test the patch at https://github.com/asterisk/asterisk/pull/161 and let us know if it resolves this?

PR #151, opened 2023-06-06
res_pjsip_rfc3326: Allow propogation of 'Answered Eleswhere'

AstersikRoss:
res_pjsip_rfc3326: Allow propagation of 'Answered Elsewhere' through Asterisk 

Resolves: #116


AstersikRoss:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp:
The contributor license agreement has not been signed. Until that is done this can not be reviewed.

res/res_pjsip_rfc3326.c
```diff
@@ -52,24 +52,36 @@ static void rfc3326_use_reason_header(struct ast_sip_session *session, struct pj
...
+			} else if (sscanf(cause, "cause=%30d", &code) != 1) {
 				continue;
 			}
 		} else {
 			continue;
 		}
-		if (cause_q850) {
-			ast_channel_hangupcause_set(session->channel, code & 0x7f);
+		
+		if(cause_sip && code == AST_CAUSE_ANSWERED_ELSEWHERE) {
```
jcolp, line 76: Space between if and (

res/res_pjsip_rfc3326.c
```diff
@@ -52,24 +52,36 @@ static void rfc3326_use_reason_header(struct ast_sip_session *session, struct pj
...
+				if(!strncasecmp(text, "text=\"Call completed elsewhere\"", 31)) {
+					code = AST_CAUSE_ANSWERED_ELSEWHERE;
+				}
+	
 				*text = '\0';
-			}
+			} 
 
-			if (sscanf(cause, "cause=%30d", &code) != 1) {
+			if(code) {
```
jcolp, line 67: Space between if and (

res/res_pjsip_rfc3326.c
```diff
@@ -52,24 +52,36 @@ static void rfc3326_use_reason_header(struct ast_sip_session *session, struct pj
 
 		cause_q850 = !strncasecmp(cause, "Q.850", 5);
 		cause_sip = !strncasecmp(cause, "SIP", 3);
+
 		if ((cause_q850 || cause_sip) && (cause = strstr(cause, "cause="))) {
-			/* If text is present get rid of it */
+			/* If text is present, process and get rid of it */
 			if ((text = strchr(cause, ';'))) {
+
+				if(!strncasecmp(text, "text=\"Call completed elsewhere\"", 31)) {
```
jcolp, line 60: Space between if and (, and why is this a strncasecmp of a specific length (add comment why)?

jcolp (changes requested):
This needs to be a single commit. The commit message needs to be updated according to [our standard](https://wiki.asterisk.org/wiki/display/AST/Commit+Messages). A UserNote needs to be added to the commit message.

res/res_pjsip_rfc3326.c
InterLinked1: strncasecmp accepts 3 arguments (the last one is the length), you can't just remove the argument.

What Josh is asking is where the number 31 came from. We don't like magic numbers just sprinkled in the code. Here, it's the length of the string so I guess to me it's obvious, but it may not be for everyone. A comment indicating what 31 means would help.

I'm not sure if gcc is smart enough to optimize away a call like strlen("text=\"Call completed elsewhere\"") in lieu of the length. I have a special macro for this that I use, that avoids having to hardcode numbers at all.

jcolp: The Q.850 and SIP checks do it so they don't have to copy and parse the string. In the case of answered elsewhere that shouldn't occur because the rest of the string should strictly be the string being checked. That's why I asked, to see if you had a case where the remaining part of the string was not:

`text=”Call completed elsewhere”`

As that could indicate other problems.

jcolp:
This still appears to be two commits as well.

res/res_pjsip_rfc3326.c
jcolp: What does that mean that it could be something different in reference to this change?

jcolp: If you are referring to a DIFFERENT text string for some other thing, then yes, of course it could be. I was strictly referring to the Call completed elsewhere case.

At this point just leave the strncasecmp.

seanbright:
I fixed all of the formatting issues (and a bug I noticed) and pushed it up.

res/res_pjsip_rfc3326.c
```diff
@@ -53,12 +53,17 @@ static void rfc3326_use_reason_header(struct ast_sip_session *session, struct pj
...
 			if ((text = strchr(cause, ';'))) {
+				if (!strncasecmp(text, "text=\"Call completed elsewhere\"", 31)) {
+					code = AST_CAUSE_ANSWERED_ELSEWHERE;
+				}
 				*text = '\0';
 			}
 
-			if (sscanf(cause, "cause=%30d", &code) != 1) {
+			if (code) {
+				continue;
```
seanbright, line 65: This seems wrong. If `code` is set here I assume you want to fall through to the `ast_channel_hangupcause_set(...)` branches below.

res/res_pjsip_rfc3326.c
```diff
@@ -53,12 +53,17 @@ static void rfc3326_use_reason_header(struct ast_sip_session *session, struct pj
 		cause_q850 = !strncasecmp(cause, "Q.850", 5);
 		cause_sip = !strncasecmp(cause, "SIP", 3);
 		if ((cause_q850 || cause_sip) && (cause = strstr(cause, "cause="))) {
-			/* If text is present get rid of it */
+			/* If text is present, process and get rid of it */
 			if ((text = strchr(cause, ';'))) {
+				if (!strncasecmp(text, "text=\"Call completed elsewhere\"", 31)) {
```
seanbright, line 58: This seems fragile. I've seen examples in the wild where this is `text="completed elsewhere"`. Maybe we just want to check for `elsewhere` and cross our fingers?

seanbright:
Aside -  you could create a simple macro for the "starts with" code that has magic numbers:

```
#define STARTS_WITH(str, literal) \
    !strcasecmp(str, literal, sizeof(literal) - 1)
```

and then:

```
if (STARTS_WITH(text, "text=\"Call completed elsewhere\"")) {
```

This avoids hard-coding the length of the "needle" assuming it will always be a string literal.

res/res_pjsip_rfc3326.c
seanbright: I didn't fix it. As the code is currently written it will not fall through.

jcolp:
Please try the fix in #161 as it is a better approach and returns the behavior that existed previously in a more complete fashion.

Issue #145, opened 2023-06-05
[bug]: ABI issue with pjproject and pjsip_inv_session


bkford:
### Severity

Major

### Versions

18.17.0, certified/18.9, 20.3.0

### Components/Modules

res_pjsip

### Operating Environment

Ubuntu

### Frequency of Occurrence

Constant

### Issue Description

pjproject added a field to the pjsip_inv_session struct, which caused direct access of things past it to crash Asterisk if the module was compiled against a different version of pjproject and expected it to be somewhere it isn't anymore. Adding an API call to Asterisk to fetch the information relevant to the crash.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #153, opened 2023-06-06
res_pjsip_session: Added new function calls to avoid ABI issues.

bkford:
Added two new functions (ast_sip_session_get_dialog and
ast_sip_session_get_pjsip_inv_state) that retrieve the dialog and the
pjsip_inv_state respectively from the pjsip_inv_session on the
ast_sip_session struct. This is due to pjproject adding a new field to
the pjsip_inv_session struct that caused crashes when trying to access
fields that were no longer where they were expected to be if a module
was compiled against a different version of pjproject.

Resolves: #145

### DIFF
--- a/include/asterisk/res_pjsip_session.h
+++ b/include/asterisk/res_pjsip_session.h
@@ -799,6 +799,26 @@ void ast_sip_session_send_request_with_cb(struct ast_sip_session *session, pjsip
 struct ast_sip_session *ast_sip_dialog_get_session(pjsip_dialog *dlg);
 
 /*!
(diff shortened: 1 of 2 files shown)


bkford:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (changes requested):
See other review

jcolp (changes requested):
See other review. Which I thought was master, but isn't.

jcolp (approved):

gtjoseph (approved):

PR #154, opened 2023-06-06
res_pjsip_session: Added new function calls to avoid ABI issues.

bkford:
Added two new functions (ast_sip_session_get_dialog and
ast_sip_session_get_pjsip_inv_state) that retrieve the dialog and the
pjsip_inv_state respectively from the pjsip_inv_session on the
ast_sip_session struct. This is due to pjproject adding a new field to
the pjsip_inv_session struct that caused crashes when trying to access
fields that were no longer where they were expected to be if a module
was compiled against a different version of pjproject.

Resolves: #145

### DIFF
--- a/include/asterisk/res_pjsip_session.h
+++ b/include/asterisk/res_pjsip_session.h
@@ -795,6 +795,26 @@ void ast_sip_session_send_request_with_cb(struct ast_sip_session *session, pjsip
 struct ast_sip_session *ast_sip_dialog_get_session(pjsip_dialog *dlg);
 
 /*!
+ * \brief Retrieves a dialog from a session
+ *
+ * \param session The session to retrieve the dialog from
+ *
+ * \retval non-NULL if dialog exists
+ * \retval NULL if no dialog
+ */
+pjsip_dialog *ast_sip_session_get_dialog(const struct ast_sip_session *session);
+
+/*!
+ * \brief Retrieves the pjsip_inv_state from a session
+ *
+ * \param session The session to retrieve the state from
+ *
+ * \retval state if inv_session exists
+ * \retval PJSIP_INV_STATE_NULL if inv_session is NULL
+ */
+pjsip_inv_state ast_sip_session_get_pjsip_inv_state(const struct ast_sip_session *session);
+
+/*!
  * \brief Resumes processing of a deferred incoming re-invite
  *
  * \param session The session which has a pending incoming re-invite
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -3641,6 +3641,28 @@ struct ast_sip_session *ast_sip_dialog_get_session(pjsip_dialog *dlg)
 	return session;
 }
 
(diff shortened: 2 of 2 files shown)


include/asterisk/res_pjsip_session.h
```diff
@@ -794,6 +794,26 @@ void ast_sip_session_send_request_with_cb(struct ast_sip_session *session, pjsip
...
+
+/*!
+ * \brief Retrieves the pjsip_inv_state from a session
+ *
+ * \param session The session to retrieve the state from
+ *
+ * \retval state if inv_session exists
+ * \retval PJSIP_INV_STATE_NULL if inv_session is NULL
+ */
+pjsip_inv_state ast_sip_session_get_pjsip_inv_state(struct ast_sip_session *session);
```
jcolp, line 815: const struct

res/res_pjsip_session.c
```diff
@@ -3641,6 +3641,28 @@ struct ast_sip_session *ast_sip_dialog_get_session(pjsip_dialog *dlg)
...
+	pjsip_inv_session *inv_session = session->inv_session;
+
+	if (!inv_session) {
+		return NULL;
+	}
+
+	return inv_session->dlg;
+}
+
+pjsip_inv_state ast_sip_session_get_pjsip_inv_state(struct ast_sip_session *session)
```
jcolp, line 3655: constify

jcolp (changes requested):

include/asterisk/res_pjsip_session.h
```diff
@@ -794,6 +794,26 @@ void ast_sip_session_send_request_with_cb(struct ast_sip_session *session, pjsip
...
+
+/*!
+ * \brief Retrieves the pjsip_inv_state from a session
+ *
+ * \param session The session to retrieve the state from
+ *
+ * \retval state if inv_session exists
+ * \retval PJSIP_INV_STATE_NULL if inv_session is NULL
+ */
+const enum pjsip_inv_state ast_sip_session_get_pjsip_inv_state(struct ast_sip_session *session);
```
jcolp, line 815: I meant the ast_sip_session passed in should be const, since it's just returning an enum

include/asterisk/res_pjsip_session.h
```diff
@@ -794,6 +794,26 @@ void ast_sip_session_send_request_with_cb(struct ast_sip_session *session, pjsip
...
 
+/*!
+ * \brief Retrieves a dialog from a session
+ *
+ * \param session The session to retrieve the dialog from
+ *
+ * \retval non-NULL if dialog exists
+ * \retval NULL if no dialog
+ */
+const struct pjsip_dialog *ast_sip_session_get_dialog(struct ast_sip_session *session);
```
jcolp, line 805: This can't return const since the underlying structure will likely end up getting used in non-const ways

jcolp (changes requested):

jcolp (approved):

gtjoseph (approved):

Issue #155, opened 2023-06-09
[bug]: GCC 13 is catching a few new trivial issues


gtjoseph:
### Severity

Major

### Versions

18,20,master

### Components/Modules

All

### Operating Environment

Fedora 38, GCC 13.1

### Frequency of Occurrence

Constant

### Issue Description

* gcc 13 is now catching when a function is declared as returning
  an enum but defined as returning an int or vice versa.

* gcc 13 is also now (incorrectly) complaining of dangling pointers
  when assigning a pointer to a local char array to a char *.

### Relevant log output

```shell
test_utils.c: In function ‘quoted_escape_test’:
test_utils.c:150:22: error: storing the address of local variable ‘small’ in ‘tests[1].buf’ [-Werror=dangling-pointer=]
  150 |         tests[1].buf = small;

stasis_message.c:56:5: error: conflicting types for ‘stasis_message_type_create’ due to enum/integer mismatch; have ‘int(const char *, struct stasis_message_vtable *, struct stasis_message_type **)’ [-Werror=enum-int-mismatch]
   56 | int stasis_message_type_create(const char *name,
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~

loader.c:1824:5: error: conflicting types for ‘ast_load_resource’ due to enum/integer mismatch; have ‘int(const char *)’ [-Werror=enum-int-mismatch]
 1824 | int ast_load_resource(const char *resource_name)
      |     ^~~~~~~~~~~~~~~~~

app.c:188:25: error: conflicting types for ‘ast_app_getdata’ due to enum/integer mismatch; have ‘enum ast_getdata_result(struct ast_channel *, const char *, char *, int,  int)’ [-Werror=enum-int-mismatch]
  188 | enum ast_getdata_result ast_app_getdata(struct ast_channel *c, const char *prompt, char *s, int maxlen, int timeout)
      |                         ^~~~~~~~~~~~~~~
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #157, opened 2023-06-09
build: Fix a few gcc 13 issues

gtjoseph:
* gcc 13 is now catching when a function is declared as returning
  an enum but defined as returning an int or vice versa.  Fixed
  a few in app.h, loader.c, stasis_message.c.

* gcc 13 is also now (incorrectly) complaining of dangling pointers
  when assigning a pointer to a local char array to a char *. Had
  to change that to an ast_alloca.

Resolves: #155

### DIFF
--- a/include/asterisk/app.h
+++ b/include/asterisk/app.h
@@ -137,7 +137,7 @@ int ast_ivr_menu_run(struct ast_channel *c, struct ast_ivr_menu *menu, void *cbd
  *  is pressed during playback, it will immediately break out of the message and continue
  *  execution of your code.
  */
-int ast_app_getdata(struct ast_channel *c, const char *prompt, char *s, int maxlen, int timeout);
+enum ast_getdata_result ast_app_getdata(struct ast_channel *c, const char *prompt, char *s, int maxlen, int timeout);
 
 /*! \brief Plays a stream and gets DTMF data from a channel
  * \param c Which channel one is interacting with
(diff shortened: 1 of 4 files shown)


InterLinked1 (changes requested):
Can't comment inline on commit messages, but occurrrances is misspelled.

retrofun:
This change causes a _use without previous declaration_ error for me with asterisk-20.4.0-rc1:
```
/usr/include/asterisk/app.h:142:6: error: use of enum 'ast_getdata_result' without previous declaration
 enum ast_getdata_result ast_app_getdata(struct ast_channel *c, const char *prompt, char *s, int maxlen, int timeout);
      ^~~~~~~~~~~~~~~~~~
/usr/include/asterisk/app.h:159:6: error: use of enum 'ast_getdata_result' without previous declaration
 enum ast_getdata_result ast_app_getdata_terminator(struct ast_channel *c, const char *prompt, char *s, int maxlen, int timeout, char *terminator);
      ^~~~~~~~~~~~~~~~~~
```

_ast_getdata_result_ is defined later in app.h, [line 1198](https://github.com/asterisk/asterisk/blob/6c8b23a68870a107a0ebe667b3b6337d1f0f6efe/include/asterisk/app.h#L1198).

jcolp:
@retrofun Please open a new issue, and include the compiler version you are using.

Issue #158, opened 2023-06-09
[bug]: test_stasis_endpoints.c: Unit test channel_messages is unstable


gtjoseph:
### Severity

Major

### Versions

18,20,master

### Components/Modules

test_stasis_endpoints

### Operating Environment

Rocky 8 CI environment

### Frequency of Occurrence

Occasional

### Issue Description

The channel_messages test occasionally fails.  It looks like it's expecting stasis to deliver messages in a specific order which isn't guaranteed.

### Relevant log output

```shell
[test_stasis_endpoints.c:channel_messages:264]: Condition failed: ast_channel_snapshot_type() == type
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #159, opened 2023-06-09
test_stasis_endpoints.c: Make channel_messages more stable

gtjoseph:
The channel_messages test was assuming that stasis would return
messages in a specific order.  This is an incorrect assumption as
message ordering was never guaranteed.  This was causing the test
to fail occasionally.  We now test all the messages for the
required message types instead of testing one by one.

Resolves: #158


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20

jcolp (approved):

PR #160, opened 2023-06-09
build: Fix a few gcc 13 issues

gtjoseph:
* gcc 13 is now catching when a function is declared as returning
  an enum but defined as returning an int or vice versa.  Fixed
  a few in app.h, loader.c, stasis_message.c.

* gcc 13 is also now (incorrectly) complaining of dangling pointers
  when assigning a pointer to a local char array to a char *. Had
  to change that to an ast_alloca.

Resolves: #155


gtjoseph:
No cherry-picks required.
This is certified/18.9 specific.

jcolp (approved):

Issue #64, opened 2023-05-04
[bug]: app_voicemail_imap wrong behavior when losing IMAP connection


OlafTitz:
### Severity

Minor

### Versions

18.14.0

### Components/Modules

app_voicemail_imap

### Operating Environment

Ubuntu 22.04

### Frequency of Occurrence

Frequent

### Issue Description

Seen by user: When the IMAP server is down, a wrong MWI is delivered. My desk phone shows 2 waiting messages (when there were none previously).

Apparent cause: the function static int messagecount(const char *mailbox_id, const char *folder) just returns the result from calling __messagecount(context, mailbox, folder). But that can be negative in case of error. Then the clients get a NOTIFY with "-1" messages waiting. The function has_voicemail(const char *mailbox, const char *folder) similarly checks for the result of __messagecount being != 0 instead of >0.

This happens when the IMAP connection is lost and can't be reestablished (e.g. IMAP server shut down), but judging from the code it can happen in other error situations too. 

### Relevant log output

```shell
When the IMAP server shuts down, asterisk regularly logs:

[May  4 20:42:51] ERROR[1642]: app_voicemail_imap.c:3287 mm_log: IMAP Error: Can't connect to 10.9.4.2,993: Connection refused
[May  4 20:42:51] ERROR[1642]: app_voicemail_imap.c:2497 __messagecount: Houston we have a problem - IMAP mailstream is NULL

The clients get this bogus NOTIFY:

NOTIFY sip:30@***;transport=tcp;x-reg=*** SIP/2.0
Via: SIP/2.0/TCP ***;rport;branch=***;alias
From: <sip:30@***>;tag=***
To: <sip:30@***;x-reg=***>
Contact: <sip:30@***;transport=TCP>
Call-ID: ***
CSeq: *** NOTIFY
Subscription-State: terminated
Event: message-summary
Allow-Events: message-summary, presence, dialog, refer
Max-Forwards: 70
User-Agent: Asterisk PBX 18.14.0~dfsg+~cs6.12.40431414-1
Content-Type: application/simple-message-summary
Content-Length:   104

Messages-Waiting: yes
Message-Account: sip:*90@***;transport=TCP
Voice-Message: -1/0 (0/0)
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

OlafTitz:
This patch fixes the problem.
[...]

PR #163, opened 2023-06-16
app_voicemail_imap: Fix message count when IMAP server is unavailable

OlafTitz:
Some callers of __messagecount did not correctly handle error return, instead returning a -1 message count.
This caused a notification with "Messages-Waiting: yes" and "Voice-Message: -1/0 (0/0)" if the IMAP server was unavailable.

Fixes: #64


OlafTitz:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (approved):

gtjoseph (approved):

Issue #135, opened 2023-05-31
[bug]: func_curl.c error adding httpheaders


scgm11:
### Severity

Minor

### Versions

all

### Components/Modules

func_curl

### Operating Environment

Ubuntu 22

### Frequency of Occurrence

Constant

### Issue Description

using func_curl from dialplan and adding a header that already exists duplicates it and leave the request in a bad formar

Example

`exten=> .XXXXXXXX,1,Set(CURLOPT(httpheader)=Authorization: basic TEST)
exten=> .XXXXXXXX,2,Set(CURLOPT(httpheader)=Content-Type: application/json)
exten=> .XXXXXXXX,3,Set(CALLERID(num)=${CURL(https://${SERVERDOMAIN}/test)}))
exten=> .XXXXXXXX,4,Set(CURLOPT(httpheader)=Authorization: basic TEST)
exten=> .XXXXXXXX,5,Set(CURLOPT(httpheader)=Content-Type: application/json)
exten=> .XXXXXXXX,3,Set(CALLERID(num)=${CURL(https://${SERVERDOMAIN}/test)}))
`

one solution could be use 

`AST_LIST_HEAD_INIT(list);
`
in acf_curl_helper after the unlock if this is ok for you I will make a patch here

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
What does the RFC state for multiple headers of the same name? If not allowed, then I believe the code should be updated to handle that situation. If allowed then the ability should be added to have a user overwrite what already exists instead of having multiple headers.

As for your fix, I believe that wouldn't work and would eliminate all configured options.

scgm11:
ok I saw that now, but seems to be an issue withe the Asterisk list then, shouldn't it be init somewhere?
as seems gets duplicated each time you call curl. As it is now if you have to call 2 WS with different options you cant

jcolp:
The problem is isolated to httpheaders because of the code:

https://github.com/asterisk/asterisk/blob/master/funcs/func_curl.c#L440

Thus why I stated before:

What does the RFC state for multiple headers of the same name? If not allowed, then I believe the code should be updated to handle that situation. If allowed then the ability should be added to have a user overwrite what already exists instead of having multiple headers.

Just wiping everything clean after CURL() completes is not an option because that goes against the documented and existing behavior, so it needs to be determined what the RFC allows for httpheaders in order to determine the best way to fix this.

[...]

PR #165, opened 2023-06-19
func_curl: add new headers or replace them if already exists

scgm11:
Resolves: #135

### DIFF
--- a/funcs/func_curl.c
+++ b/funcs/func_curl.c
@@ -332,6 +332,37 @@ static int parse_curlopt_key(const char *name, CURLoption *key, enum optiontype 
 	return 0;
 }
 
+static void replace_or_add_header(struct curl_slist **headers, const char *header) {
+    struct curl_slist *current = *headers;
+    struct curl_slist *prev = NULL;
+
+    while (current) {
+        const char *colon = strchr(current->data, ':');
+        if (colon) {
+            size_t len = colon - current->data;
+            if (strncmp(current->data, header, len) == 0) {
+                struct curl_slist *new_header = curl_slist_append(NULL, header);
+                if (new_header) {
+                    if (prev) {
+                        prev->next = new_header;
+                    } else {
+                        *headers = new_header;
+                    }
+                    new_header->next = current->next;
+                    return;
+                } else {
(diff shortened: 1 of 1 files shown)


scgm11:
cherry-pick-to: 20
cherry-pick-to: 18

funcs/func_curl.c
```diff
@@ -332,6 +332,37 @@ static int parse_curlopt_key(const char *name, CURLoption *key, enum optiontype
...
+                } else {
+                	ast_log(LOG_ERROR, "Unable to allocate memory for new header: %s\n", header);
+                    return;
+                }
+            }
+        }
+        prev = current;
+        current = current->next;
+    }
+
```
scgm11, line 362: I think I may be missing a curl_slist_free_all(current); here ?

funcs/func_curl.c
```diff
@@ -698,7 +729,7 @@ static int acf_curl_helper(struct ast_channel *chan, struct curl_args *args)
 		if (cur->key == CURLOPT_SPECIAL_HASHCOMPAT) {
 			hashcompat = (long) cur->value;
 		} else if (cur->key == CURLOPT_HTTPHEADER) {
-			headers = curl_slist_append(headers, (char*) cur->value);
+			replace_or_add_header(&headers, (char*) cur->value);
```
jcolp, line 732: I do not believe it should be the responsibility of the acf_curl_helper function to do this replacement. The list of headers on the datastore should only ever contain a single header of the given name. The global headers list should not add a header to the request if the same header is present in the datastore list.

jcolp (changes requested):

funcs/func_curl.c
scgm11: so you want the change in acf_curlopt_write ath the end where 	AST_LIST_INSERT_TAIL(list, new, list); ?

jcolp: That is where I would expect such functionality to be written, but there would also have to be logic in acf_curl_helper for the global / channel datastore situation.

scgm11: ok sorry cant understand what you want, you can close this if you feel its not good enough as I don't have the knowledge to do it as you want.

jcolp:
Closed per comments, as well as comment on issue regarding implementation.

PR #178, opened 2023-06-27
func_curl.c: Add CURLResetDefaults() application.

seanbright:
This application clears all customizations made by the CURLOPT() function and resets everything to its default state.

Related to #135

ASTERISK-30088 #close

UserNote: A new application - CURLResetDefaults() - has been added which resets all customizations made by the CURLOPT() dialplan function.

### DIFF
--- a/funcs/func_curl.c
+++ b/funcs/func_curl.c
@@ -200,8 +200,17 @@
 		<see-also>
 			<ref type="function">CURL</ref>
 			<ref type="function">HASH</ref>
+			<ref type="application">CURLResetDefaults</ref>
 		</see-also>
 	</function>
+	<application name="CURLResetDefaults" language="en_US">
+		<synopsis>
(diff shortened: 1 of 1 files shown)


seanbright:
So this is my WIP curl reset code that I mentioned in #135. There was something about this approach that I didn't like which prevented me from uploading before but I am having a hard time remembering what that was. If someone wants to run with this, please do. If not I will get back to it when I have some time.

seanbright:
So looking over it again I believe the issue is that the curl instance is thread local and not related to a channel. Maybe a better first step would be to make curl instances be part of a channel datastore?

jcolp:
I think this also doesn't reset all things to their default for CURLOPT, since state is held on the datastore and applied when the CURL operation is invoked. This + clearing the datastore would be needed I think.

seanbright:
You are correct. This is a turd.

Issue #183, opened 2023-06-30
[deprecation]: Deprecate users.conf


InterLinked1:
### Improvement Description

This deprecates `users.conf`, which is largely irrelevant today and has been criticized since the beginning as something that never should have been added.

There isn't a "deprecation" category so I'm marking this as an "improvement", since it will simplify the codebase a fair bit from a developer perspective and also simplify for users since they won't get cryptic warnings about something they shouldn't be doing anyways.

PR #184, opened 2023-06-30
users.conf: Deprecate users.conf configuration.

InterLinked1:
This deprecates the users.conf config file, which
is no longer as widely supported but still integrated with a number of different modules.

Because there is no real mechanism for marking a
configuration file as "deprecated", and users.conf is not just used in a single place, this now emits a warning to the user when the PBX loads to notify about the deprecation.

This configuration mechanism has been widely criticized and discouraged since its inception, and is no longer relevant to the configuration that most users are doing today. Removing it will allow for some simplification and cleanup in the codebase.

Resolves: #183

UpgradeNote: The users.conf config is now deprecated and will be removed in a future version of Asterisk.

### DIFF
--- a/pbx/pbx_config.c
+++ b/pbx/pbx_config.c
@@ -1974,6 +1974,27 @@ static void append_interface(char *iface, int maxlen, char *add)
 	}
 }
 
+static void startup_event_cb(void *data, struct stasis_subscription *sub, struct stasis_message *message)
+{
+	struct ast_json_payload *payload;
+	const char *type;
+
+	if (stasis_message_type(message) != ast_manager_get_generic_type()) {
+		return;
+	}
+
+	payload = stasis_message_data(message);
+	type = ast_json_string_get(ast_json_object_get(payload->json, "type"));
+
+	if (strcmp(type, "FullyBooted")) {
+		return;
+	}
+
+	ast_log(LOG_WARNING, "users.conf is deprecated and will be removed in a future version of Asterisk\n");
+
+	stasis_unsubscribe(sub);
+}
+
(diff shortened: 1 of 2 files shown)


InterLinked1:
No cherry picks required

seanbright:
While not technically a module deprecation this will need to be discussed on the asterisk-dev list.

pbx/pbx_config.c
```diff
@@ -1994,6 +1994,9 @@ static void pbx_load_users(void)
 	if (!cfg)
 		return;
 
+	/*! \todo Remove users.conf support in Asterisk 23 */
+	ast_log(LOG_WARNING, "users.conf is deprecated and will be removed in a future version of Asterisk\n");
```
jcolp, line 1998: I think this should use the same approach in other places and subscribe to know that Asterisk is fully booted, at which point the users.conf warning can be output so it appears after all of the other logging that can occur.

jcolp (changes requested):

pbx/pbx_config.c
InterLinked1: If we do this in other places as well, should it use a common subscription so that a single warning is emitted, or all separate? I'm just not sure how'd you do that without symbols being visible outside of individual files.

My other thought would be instead of every place where the config loads, would it be fine to put it in ast_config_load?

e.g.

`if (!strcmp(cfg_filename, "users.conf")) warning()`

That would cover allow a single warning to cover anything possibly using users.conf. On the downside, it would add a slight amount of overhead to every config load.

gtjoseph (approved):

PR #185, opened 2023-06-30
res_rtp_asterisk: Move ast_rtp_rtcp_report_alloc using `rtp->themssrc_valid` into the scope of the rtp_instance lock.

nphantom:
res_rtp_asterisk: Move ast_rtp_rtcp_report_alloc using `rtp->themssrc_valid` into the scope of the rtp_instance lock.

From the gdb information, it was found that when calling __ast_free, the size of the
allocated space pointed to by the pointer matches the size created when rtp->themssrc_valid
is equal to 0. However, in reality, when reading the value of rtp->themssrc_valid in gdb,
it is found to be 1.

Within ast_rtcp_write(), the call to ast_rtp_rtcp_report_alloc() uses rtp->themssrc_valid,
which is outside the protection of the rtp_instance lock. However,
ast_rtcp_generate_report(), which is called by ast_rtcp_generate_compound_prefix(), uses
rtp->themssrc_valid within the protection of the rtp_instance lock.

This can lead to the possibility that the value of rtp->themssrc_valid used in the call to
ast_rtp_rtcp_report_alloc() may be different from the value of rtp->themssrc_valid used
within ast_rtcp_generate_report().

Resolves: asterisk#63

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -4901,9 +4901,7 @@ static int ast_rtcp_write(const void *data)
 	struct ast_sockaddr remote_address = { { 0, } };
 	unsigned char *rtcpheader;
 	unsigned char bdata[AST_UUID_STR_LEN + 128] = ""; /* More than enough */
-	RAII_VAR(struct ast_rtp_rtcp_report *, rtcp_report,
-			ast_rtp_rtcp_report_alloc(rtp->themssrc_valid ? 1 : 0),
-			ao2_cleanup);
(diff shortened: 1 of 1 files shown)


jcolp:
Have you signed the CLA? If not, that needs to be done before this can be reviewed.

jcolp (approved):

nphantom:
cherry-pick-to: 20
cherry-pick-to: 18

gtjoseph (approved):

Issue #194, opened 2023-07-06
[bug]: Segfault/double-free in bundled pjproject using TLS transport


gtjoseph:
### Severity

Major

### Versions

16,19

### Components/Modules

res_pjsip

### Operating Environment

Unknown

### Frequency of Occurrence

Occasional

### Issue Description

Original issue: [ASTERISK-30471](https://issues.asterisk.org/jira/browse/ASTERISK-30471)
Original reporter: Florentin Mayer(f.mayer)

We recently found out that the 16.30 LTS version of Asterisk crashes reproduceably when using TLS as transport, as described in the following.

The pjproject version 2.12 included in the Asterisk’s “16” branch still contains the bug https://github.com/pjsip/pjproject/issues/3069 which leads to a crash of Asterisk if e.g. a path to a certificate/CA is wrong or the certificate does not meet the openssl requirements (e.g. “key too small”). It is fairly easy to reproduce, configure a TLS transport with invalid server certificate (or CA) and connect a telephone with TLS. Then either restart asterisk (double free) or connect a TLS endpoint again (segfault).
The first case initially occurred with our systems on the current debian build (1:16.28.0~dfsg-0+deb11u1), and is described in the pjproject bug report.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #195, opened 2023-07-06
bundled_pjproject: Backport 2 SSL patches from upstream

gtjoseph:
* Fix double free of ossock->ossl_ctx in case of errors
https://github.com/pjsip/pjproject/commit/863629bc65d6

* free SSL context and reset context pointer when setting the cipher
  list fails
https://github.com/pjsip/pjproject/commit/0fb32cd4c0b2

Resolves: #194


gtjoseph:
cherry-pick-to: 19

jcolp (approved):

Issue #200, opened 2023-07-10
[bug]: Regression: In app.h an enum is used before its declaration.


gtjoseph:
### Severity

Major

### Versions

18.19.0-rc1,20.4.0-rc1

### Components/Modules

app

### Operating Environment

any

### Frequency of Occurrence

Constant

### Issue Description

In app.h, ast_app_getdata and ast_app_getdata_terminator return the enum ast_getdata_result but that enum isn't declared until about 1000 lines latter in the file.

Originally reported by user @retrofun

### Relevant log output

```shell
/usr/include/asterisk/app.h:142:6: error: use of enum 'ast_getdata_result' without previous declaration
 enum ast_getdata_result ast_app_getdata(struct ast_channel *c, const char *prompt, char *s, int maxlen, int timeout);
      ^~~~~~~~~~~~~~~~~~
/usr/include/asterisk/app.h:159:6: error: use of enum 'ast_getdata_result' without previous declaration
 enum ast_getdata_result ast_app_getdata_terminator(struct ast_channel *c, const char *prompt, char *s, int maxlen, int timeout, char *terminator);
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #201, opened 2023-07-10
app.h: Move declaration of ast_getdata_result before its first use

gtjoseph:
The ast_app_getdata() and ast_app_getdata_terminator() declarations
in app.h were changed recently to return enum ast_getdata_result
(which is how they were defined in app.c).  The existing
declaration of ast_getdata_result in app.h was about 1000 lines
after those functions however so under certain circumstances,
a "use before declaration" error was thrown by the compiler.
The declaration of the enum was therefore moved to before those
functions.

Resolves: #200


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20

jcolp (approved):

Issue #205, opened 2023-07-10
[new-feature]: sig_analog: Allow flash to time out to silent hold


InterLinked1:
### Feature Description

Imported from https://issues.asterisk.org/jira/browse/ASTERISK-30004

Currently, if using simple switch (immediate=no and threewaycalling=yes), if you flash to get a recall dial tone, the dial tone will continue forever to allow using it as a "hold" feature.

To replicate the way some real CO switches handle this, this adds the ability to time out to a "silent hold". Rather than playing dial tone for potentially a while, this allows the dial tone to time out normally, but to silence as opposed to any kind of "permanent signal" condition. This allows the user to then have the call on hold be silent, if this is their actual intention, which is potentially much less annoying.

InterLinked1:
@jcolp sig_analog is core supported, not extended supported, is this the right label?

jcolp:
I've changed it to core.

PR #206, opened 2023-07-10
sig_analog: Allow three-way flash to time out to silence.

InterLinked1:
sig_analog allows users to flash and use the three-way dial tone as a primitive hold function, simply by never timing it out.

Some systems allow this dial tone to time out to silence, so the user is not annoyed by a persistent dial tone. This option allows the dial tone to time out normally to silence.

Resolves: #205
ASTERISK-30004 #close
Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19716

UserNote: The threewaysilenthold option now allows the three-way dial tone to time out to silence, rather than continuing forever.

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -12898,6 +12898,7 @@ static struct dahdi_pvt *mkintf(int channel, const struct dahdi_chan_conf *conf,
 		tmp->usedistinctiveringdetection = usedistinctiveringdetection;
 		tmp->callwaitingcallerid = conf->chan.callwaitingcallerid;
 		tmp->threewaycalling = conf->chan.threewaycalling;
+		tmp->threewaysilenthold = conf->chan.threewaysilenthold;
 		tmp->adsi = conf->chan.adsi;
 		tmp->use_smdi = conf->chan.use_smdi;
 		tmp->permhidecallerid = conf->chan.hidecallerid;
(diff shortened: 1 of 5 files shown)


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (approved):

gtjoseph (changes requested):
Per https://docs.asterisk.org/latest/Development/Policies-and-Procedures/Commit-Messages/  Please do NOT intersperse unsupported commit message trailers among the supported ones.  The "ASTERISK-30004..." line MUST come before the standard Resolves, Usernote and UpgradeNote trailers.

InterLinked1:
> Per https://docs.asterisk.org/latest/Development/Policies-and-Procedures/Commit-Messages/ Please do NOT intersperse unsupported commit message trailers among the supported ones. The "ASTERISK-30004..." line MUST come before the standard Resolves, Usernote and UpgradeNote trailers.

https://docs.asterisk.org/latest/Development/Policies-and-Procedures/Commit-Messages/ is a 404, but should be fixed now!

InterLinked1:
@gtjoseph By the way, I think the same thing has happened here: for some reason, GitHub isn't discussing the stale feedback for an addressed issue, so I'm assuming this probably isn't showing up in the "needs attention queue" or however it works (since it says "Changes requested" which is wrong). But it only seems to happen sometimes, not all the time, trying to see what that might be. Could be my fault since I acknowledged it, maybe I shouldn't be doing that.

gtjoseph:
Ok. I'll re-review in a bit.

gtjoseph (approved):

PR #216, opened 2023-07-18
func_cut: Add example to documentation.

phoneben:
This adds an example to the XML documentation clarifying usage of the 
CUT function to address a common misusage.

### DIFF
--- a/funcs/func_cut.c
+++ b/funcs/func_cut.c
@@ -75,6 +75,10 @@
 		</syntax>
 		<description>
 			<para>Cut out information from a string (<replaceable>varname</replaceable>), based upon a named delimiter.</para>
+					<example title="The 'varname' parameter can only accept a variable name, not a variable expression">
+				exten => s,1,Set(foo=${CUT(bar,,2)}); This is correct syntax
+				exten => s,2,Set(foo=${CUT(${bar},,2)}); This is invalid syntax (unless bar contains the name of another variable)
+ 					</example>
 		</description>
 	</function>
  ***/


phoneben:
cherry-pick-to: 20
cherry-pick-to: 18

InterLinked1 (changes requested):
Your commit message still doesn't follow the Asterisk commit message guidelines.

A good title would be `func_cut: Add clarifying example usage.` or something along those lines, with a brief description below. Don't include any code in the commit message, you don't need to repeat your change. And yes, you're updating the wiki indirectly, but you're really updating the documentation (xmldocs).

Make sure to use `git commit -a --amend` to amend existing commit and not make a new one.

Change itself looks fine now, though.

funcs/func_cut.c
```diff
@@ -75,6 +75,11 @@
 		</syntax>
 		<description>
 			<para>Cut out information from a string (<replaceable>varname</replaceable>), based upon a named delimiter.</para>
+					<example title="The 'varname' parameter could only accept a variable name, not a variable expression">
+					<para>Beleow are examples illustrating the correct and invalid syntax.</para>
```
InterLinked1, line 79: Below is misspelled, and <para> tags do not belong inside an example, this should go before the example tag.

InterLinked1 (changes requested):
The commit message isn't any better still.

funcs/func_cut.c
phoneben: fixed thank you

InterLinked1 (changes requested):
It's just getting worse each time. Do not include your comments on the commit messages in your commit message. That is not the place.

If you don't want to keep doing this all day, here's one you could use:

```
func_cut: Add example to documentation.

This adds an example to the XML documentation clarifying usage
of the CUT function to address a common misusage.

```

I don't see a related issue but if there was one, that should also be mentioned per the guidelines.

phoneben:
I did not file an issue. Just wanted to update the docs so the next person comming around should find it .

thank you for your help 

the commit note you are referring is just the first correct ?

InterLinked1:
> I did not file an issue. Just wanted to update the docs so the next person comming around should find it .

Generally the project likes it if issues are created; I don't know if they'll insist on that for a small improvement like this; I guess someone else can chime in if it's an issue (no pun intended...)

> thank you for your help
> 
> the commit note you are referring is just the first correct ?

Not sure what you mean by "the first" - there should only be one commit, and that's what I'm referring to.

In other words, amend your commit and replace the commit message with something along the lines of the example I gave, or anything else that meets the commit message guidelines: https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/

phoneben:
hope i got this correct for now !

InterLinked1 (changes requested):
Almost... the first line is part of the message so don't do "Update func_cut.c". Just start with the "func_cut: " part immediately.

The f in `func_cut` should not be capitalized.
The last sentence should be line wrapped at ~72 characters.

phoneben:
did i get it correct ?

InterLinked1 (changes requested):
Just one last issue I see: there's an extra line in the middle of the commit message body:

```
This adds an example to the XML documentation clarifying usage of the

CUT function to address a common misusage
```
Should be:

```
This adds an example to the XML documentation, clarifying usage of the
CUT function to address a common misusage
```

Remove that extra line and it should be good.

phoneben:
done ty

funcs/func_cut.c
```diff
@@ -75,6 +75,11 @@
 		</syntax>
 		<description>
 			<para>Cut out information from a string (<replaceable>varname</replaceable>), based upon a named delimiter.</para>
+						<para>Below are examples illustrating the correct and invalid syntax.</para>
+					<example title="The 'varname' parameter could only accept a variable name, not a variable expression">
```
InterLinked1, line 79: Commit message looks good now.

One other thing I'm seeing here now: `could` implies past tense, change this to `can`.
Also, there is only one example, so I personally don't really think the "Below are examples illustrating the correct and invalid syntax" is needed... the comments in the example already show which is why, so this doesn't add anything.

InterLinked1 (changes requested):

jcolp (approved):

phoneben:
is there anything i am missing ?

PR #217, opened 2023-07-19
extensions.conf.sample: Remove reference to missing context.

seanbright:
c3ff4648 removed the `[iaxtel700]` context but neglected to remove references to it.

This commit addresses that and also removes iaxtel and freeworlddialup references from other config files.


seanbright:
cherry-pick-to: 18

gtjoseph (approved):

PR #220, opened 2023-07-28
Update codec_builtin.c maximum_ms changed to multiple of 20 for alaw and ulaw

eduardomazolini:
The telephone carrier Vivo (Grupo Telefônica no Brasil), the largest in Brazil, in its SIP service, fails to complete calls to Apple phones (iPhone) when the maxptime is not a multiple of 20. Rightly or wrongly, Vivo and Apple are giants and do not care to solve this problem. This is how I managed to solve the problem on my PBX. The only codec accepted by them is alaw but in some countries ulaw is used. I changed only what was necessary. Version 16 is used in FreePBX 15.
cherry-pick-to: 16
cherry-pick-to: 18


jcolp:
Changes can not be merged into release branches like 16.3, and 16 no longer receives changes. I am closing this pull request out.

Issue #224, opened 2023-08-04
[new-feature]: chan_dahdi: Allow automatic reorigination on hangup


InterLinked1:
### Feature Description

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-30357

When a DAHDI user is hung up on, right now he just hears reorder tone forever. In reality, this is not reflective of the behavior of most telephone switches and is not what users are used to in the "real world".

Many switches do an immediate reoriginate, so the user can make another call without having to manually depress and release the switchhook.

This adds an option to reoriginate (provide dial tone again) if all calls towards a user have cleared, which can be much more convenient. Lines configured for kewlstart will still do a loop disconnect as usual.

PR #225, opened 2023-08-04
chan_dahdi: Allow autoreoriginating after hangup.

InterLinked1:
Currently, if an FXS channel is still off hook when all calls on the line have hung up, the user is provided reorder tone until going back on hook again.

In addition to not reflecting what most commercial switches actually do, it's very common for switches to automatically reoriginate for the user so that dial tone is provided without the user having to depress and release the hookswitch manually. This can increase convenience for users.

This behavior is now supported for kewlstart FXS channels. It's supported only for kewlstart (FXOKS) mainly because the behavior doesn't make any sense for ground start channels, and loop start signalling doesn't provide the necessary DAHDI event that makes this easy to implement. Likely almost everyone is using FXOKS over FXOLS anyways since FXOLS is pretty useless these days.

Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19714

ASTERISK-30357 #close

Resolves: #224

UserNote: The autoreoriginate setting now allows for kewlstart FXS channels to automatically reoriginate and provide dial tone to the user again after all calls on the line have cleared. This saves users from having to manually hang up and pick up the receiver again before making another call.

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -6565,6 +6565,36 @@ hangup_out:
 	ast_free(p->cidspill);
 	p->cidspill = NULL;
 
+	if (p->reoriginate && p->sig == SIG_FXOKS && dahdi_analog_lib_handles(p->sig, p->radio, 0)) {
+		/* Automatic reorigination: if all calls towards a user have hung up,
+		 * give dial tone again, so user doesn't need to cycle the hook state manually. */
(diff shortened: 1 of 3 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20

jcolp (approved):

Issue #226, opened 2023-08-04
[improvement]: 


MikeNaso:
### Improvement Description

During an incoming call from my provider, when the call is answer on my end asterisk is send a 200 OK (after the 180 ringing), however, my provider discard the answer, since the contact doesn't include the contact_user which is my phone number.
Including the contact_user is not supposed to be against the RFC
Here what I did to included it, 

```
--- ../1/asterisk-16.30.1/res/res_pjsip.c	2023-07-07 15:00:32.000000000 -0400
+++ res/res_pjsip.c	2023-08-04 16:21:02.930673143 -0400
@@ -962,9 +962,11 @@
 
 	contact.ptr = pj_pool_alloc(rdata->tp_info.pool, PJSIP_MAX_URL_SIZE);
 	contact.slen = pj_ansi_snprintf(contact.ptr, PJSIP_MAX_URL_SIZE,
-			"<%s:%s%.*s%s:%d%s%s>",
+			"<%s:%s%s%s%.*s%s:%d%s%s>",
 			uas_use_sips_contact(rdata) ? "sips" : "sip",
 			(type & PJSIP_TRANSPORT_IPV6) ? "[" : "",
+			(!ast_strlen_zero(endpoint->contact_user))?endpoint->contact_user:"",
+			(!ast_strlen_zero(endpoint->contact_user))?"@":"",
 			(int)transport->local_name.host.slen,
 			transport->local_name.host.ptr,
 			(type & PJSIP_TRANSPORT_IPV6) ? "]" : "",
```

jcolp:
Inline patches are not accepted as they are not properly licensed. I have edited your issue to remove it.

MikeNaso:
Can you point me the right guide to follow? I found this https://wiki.asterisk.org/wiki/display/AST/Patch+Contribution+Process but the first link don't work

seanbright:
https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution

PR #232, opened 2023-08-08
res_pjsip.c: Add the contact_user to the Contact header

MikeNaso:
If the contact_user in declared in the endpoint, the contact_user will be inserted in the message for the incoming calls which it was not the case before. The contact_user is already used in the outgoing calls. 

Resolves: #226

### DIFF
--- a/res/res_pjsip.c
+++ b/res/res_pjsip.c
@@ -1000,9 +1000,11 @@ static pjsip_dialog *create_dialog_uas(const struct ast_sip_endpoint *endpoint,
 
 	contact.ptr = pj_pool_alloc(rdata->tp_info.pool, PJSIP_MAX_URL_SIZE);
 	contact.slen = pj_ansi_snprintf(contact.ptr, PJSIP_MAX_URL_SIZE,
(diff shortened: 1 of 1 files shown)


MikeNaso:
cherry-pick-to: 16
cherry-pick-to: 18
cherry-pick-to: 20

MikeNaso:
> The commit message is not clear or explicit enough. It should state that this extends the contact_user setting for incoming calls while it already worked on outgoing calls.

Done

MikeNaso:
Now the test failed changing the commit description, very strange

jcolp (changes requested):
The first line is still not detailed enough. I think the following would be descriptive enough:

    res_pjsip.c: Set contact_user on incoming call local Contact header
    
    If the contact_user is configured on the endpoint it will now be
    set on the local Contact header URI for incoming calls. The
    contact_user has already been set on the local Contact header
    URI for outgoing calls.

MikeNaso:
Done

jcolp (approved):

PR #236, opened 2023-08-09
manager: Tolerate stasis messages with no channel snapshot.

jcolp:
In some cases I have yet to determine some stasis messages may be created without a channel snapshot. This change adds some tolerance to this scenario, preventing a crash from occurring.


jcolp:
cherry-pick-to: 18
cherry-pick-to: 20

gtjoseph (changes requested):
Maybe throw some parens around the bitwise &?  Even though the precedence is correct, I had to go remind myself of it.

gtjoseph (approved):

Issue #240, opened 2023-08-09
[new-feature]: sig_analog: Add Called Subscriber Held capability


InterLinked1:
### Feature Description

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-30372

Most analog switches allow for incoming calls to hold up local subscriber lines as a convenience mechanism. This adds the capability to do this to sig_analog.

PR #241, opened 2023-08-09
sig_analog: Add Called Subscriber Held capability.

InterLinked1:
This adds support for Called Subscriber Held for FXS lines, which allows users to go on hook when receiving a call and resume the call later from another phone on the same line, without disconnecting the call. This is a convenience mechanism that most real PSTN telephone switches support.

ASTERISK-30372 #close

Resolves: #240

UserNote: Called Subscriber Held is now supported for analog FXS channels, using the calledsubscriberheld option. This allows a station  user to go on hook when receiving an incoming call and resume from another phone on the same line by going on hook, without disconnecting the call.


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

Issue #244, opened 2023-08-09
[bug]: config.c: Template inheritance is incorrect for ast_variable_retrieve


InterLinked1:
### Severity

Major

### Versions

20.0.1 (really all versions, ever)

### Components/Modules

config.c

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-30370

If values are specified in a template and then overridden, behavior is currently wrong when explicitly retrieving a variable, as reading a config line by line will result in the overriden value being used (correct), but explicitly asking for it will return the earliest set value, which is that provided by the template. 

As a result, modules can pick up the wrong values when loading config items in this manner, and users can get very confused. 

This fixes this by adjusting ast_variable_retrieve to return the last match instead of the first (the equivalent of -1 to the AST_CONFIG function), so that the most recently specified/overridden value is returned and behavior is consistent.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #245, opened 2023-08-09
config.c: Make ast_variable_retrieve return last match.

InterLinked1:
ast_variable_retrieve currently returns the first match for a variable, as opposed to the last one. This is problematic because modules that load config settings by explicitly calling ast_variable_retrieve on a variable name (as opposed to iterating through all the directives as specified) will end up taking the first specified value, such as the default value from the template rather than the actual effective value in an individual config section, leading to the wrong config.

This fixes this by making ast_variable_retrieve return the last match, or the most recently overridden one, as the effective setting. This is similar to what the -1 index in the AST_CONFIG function does.

There is another function, ast_variable_find_last_in_list, that does something similar. However, it's a slightly different API, and it sees virtually no usage in Asterisk. ast_variable_retrieve is what most things use so this is currently the relevant point of breakage.

In practice, this is unlikely to cause any breakage, since there would be no logical reason to use an inherited value rather than an explicitly overridden value when loading a config.

ASTERISK-30370 #close

Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19744

Resolves: #244

UpgradeNote: Config variables retrieved explicitly by name now return the most recently overriding value as opposed to the base value (e.g. from a template). This is equivalent to retrieving a config setting using the -1 index to the AST_CONFIG function. The major implication of this is that modules processing configs by explicitly retrieving variables by name will now get the effective value of a variable as overridden in a config rather than the first-set value (from a template), which is consistent with how other modules load config settings.

### DIFF
--- a/main/config.c
+++ b/main/config.c
@@ -784,11 +784,19 @@ const char *ast_config_option(struct ast_config *cfg, const char *cat, const cha
 const char *ast_variable_retrieve(struct ast_config *config, const char *category, const char *variable)
 {
 	struct ast_variable *v;
+	const char *match = NULL;
+
+	/* We can't return as soon as we find a match, because if a config section overrides
+	 * something specified in a template, then the actual effective value is the last
+	 * one encountered, not the first one.
+	 * (This is like using the -1 index for the AST_CONFIG function.)
+	 * Also see ast_variable_find_last_in_list
+	 */
 
 	if (category) {
 		for (v = ast_variable_browse(config, category); v; v = v->next) {
 			if (!strcasecmp(variable, v->name)) {
-				return v->value;
+				match = v->value;
 			}
 		}
 	} else {
@@ -797,13 +805,13 @@ const char *ast_variable_retrieve(struct ast_config *config, const char *categor
 		for (cat = config->root; cat; cat = cat->next) {
 			for (v = cat->root; v; v = v->next) {
 				if (!strcasecmp(variable, v->name)) {
-					return v->value;
+					match = v->value;
 				}
 			}
 		}
 	}
 
-	return NULL;
+	return match;
 }
 
 const char *ast_variable_retrieve_filtered(struct ast_config *config,


InterLinked1:
cherry-pick-to: 21

jcolp (commented):
I'm not comfortable putting this into master at this time. This is a fundamental change in behavior which should go into a standard release first to gather feedback from users. While it is reasonable to believe that it shouldn't impact people, in practice this doesn't always happen.

InterLinked1:
> I'm not comfortable putting this into master at this time. This is a fundamental change in behavior which should go into a standard release first to gather feedback from users. While it is reasonable to believe that it shouldn't impact people, in practice this doesn't always happen.

Makes sense, but I thought breaking changes were always typically master only anyways? If it's going into a standard release already, why wouldn't it also go into master, since that's always the furthest ahead?

jcolp:
Once 21 was branched breaking changes went off the table unless absolutely necessary for some reason. The cut off point is when it is branched, not when the first release candidate occurs.

InterLinked1:
> Once 21 was branched breaking changes went off the table unless absolutely necessary for some reason. The cut off point is when it is branched, not when the first release candidate occurs. The master branch will next be an LTS.

Okay, so just to confirm, like some other things at the moment, this will need to wait for ~a year before it can be merged, i.e. it cannot go into either 21 or master at this time?

jcolp:
Correct.

InterLinked1:
> Correct.

Got it.

Could I suggest a new tag for issues/PRs, if it's not too much trouble? Something like breaking-change or what not, just so we can easily identify and keep track of things that will need to wait until such a time as breaking changes can be introduced again, e.g. this one and #258

InterLinked1:
> Unless it was done in an optional manner that defaults to maintaining existing behavior.

I'd rather not add unneeded complexity to work around a bug just to get it merged sooner. I'll wait until next year so a proper fix can go in, that's fine.

jcolp:
Label created and added.

Allan-N (approved):
Thank you for this fix!

gtjoseph:
@InterLinked1 This is eligible to go into master now since 23 will be a standard release.  You'll need to rebase it on current master and change the cherry-pick line to "none".

gtjoseph (dismissed):
Rebase and change cherry-pick comment to "none"

jcolp (approved):

Issue #246, opened 2023-08-09
[bug]: res_pjsip_logger: Reload disables logging


InterLinked1:
### Severity

Minor

### Versions

All

### Components/Modules

res_pjsip_logger

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-29912

When `res_pjsip` is reloaded, any logging enabled in `res_pjsip_logger` via the CLI is disabled. This is a behavior difference from basically every other module, including `chan_sip`, and has caused confusion and annoyance for many users.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #247, opened 2023-08-09
res_pjsip_logger: Preserve logging state on reloads.

InterLinked1:
Currently, reloading res_pjsip will cause logging
to be disabled. This is because logging can also
be controlled via the debug option in pjsip.conf
and this defaults to "no".

To improve this, logging is no longer disabled on
reloads if logging had not been previously
enabled using the debug option from the config.
This ensures that logging enabled from the CLI
will persist through a reload.

ASTERISK-29912 #close
Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19575

Resolves: #246

UserNote: Issuing "pjsip reload" will no longer disable logging if it was previously enabled from the CLI.

### DIFF
--- a/res/res_pjsip_logger.c
+++ b/res/res_pjsip_logger.c
@@ -598,6 +598,19 @@ static char *pjsip_set_logger_pcap(int fd, const char *arg)
 	return CLI_SUCCESS;
 }
 
+enum pjsip_logger_mask {
+	AST_PJSIP_LOGGER_UNSET = 0,
+	AST_PJSIP_LOGGER_NONE = (1 << 0),
+	AST_PJSIP_LOGGER_HOST = (1 << 1),
+	AST_PJSIP_LOGGER_METHOD = (1 << 2),
+	AST_PJSIP_LOGGER_VERBOSE = (1 << 3),
+	AST_PJSIP_LOGGER_PCAP = (1 << 4),
+	AST_PJSIP_LOGGER_ALL = (1 << 5),
+};
+
+static enum pjsip_logger_mask logger_cli_settings = AST_PJSIP_LOGGER_UNSET;
+static enum pjsip_logger_mask logger_config_settings = AST_PJSIP_LOGGER_UNSET;
+
 static char *pjsip_set_logger(struct ast_cli_entry *e, int cmd, struct ast_cli_args *a)
 {
 	static const char * const method_choices[] = {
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

res/res_pjsip_logger.c
```diff
@@ -665,15 +668,27 @@ static void check_debug(void)
...
 
 	if (ast_true(debug)) {
+		if (logger_enabled_via_config) {
+			return;
+		}
 		pjsip_enable_logger_all(-1);
 		return;
 	}
 
+	logger_enabled_via_config = 1;
```
jcolp, line 691: This doesn't behave as described. The logger_enabled_via_config is only set when host based logging is enabled from the configuration file. If you merely enable it globally for all traffic, it's never set.

res/res_pjsip_logger.c
```diff
@@ -665,15 +668,27 @@ static void check_debug(void)
...
 
 	if (ast_false(debug)) {
-		pjsip_disable_logger(-1);
+		/* If the logger was enabled via the CLI instead of
+		 * through the config file, then we shouldn't
+		 * disable it on a reload.
+		 * Only disable logging if the logger was previously
+		 * enabled via the config file. */
+		if (logger_enabled_via_config) {
+			pjsip_disable_logger(-1);
```
jcolp, line 677: This would not handle the case if you do CLI based logging and config based logging simultaneously and change the config based logging. It would turn off logging.

jcolp (changes requested):

asteriskteam:

res/res_pjsip_logger.c
```diff
@@ -664,16 +684,35 @@ static void check_debug(void)
...
+		}
+		logger_config_settings = AST_PJSIP_LOGGER_NONE;
 		return;
 	}
 
+	/* Got directive to enable debug */
 	if (ast_true(debug)) {
-		pjsip_enable_logger_all(-1);
+		/* If logger already enabled via config, then nothing has changed. */
+		if (!(logger_config_settings & AST_PJSIP_LOGGER_ALL)) {
```
mbradeen, line 705: The cli setting isn't considered here, and the config is set to NONE in the ast_false case either way so I'm not sure what this check or the conf setting enum is for.
If a cli no should over-ride the config yes, then I think this needs to mirror the ast_false case.  Otherwise I don't think this check or the config enum is needed.

mbradeen (dismissed):

gtjoseph (changes requested):
There was a coredump during the unit tests and there were more than usual failures in the testsuite.  Can you rebase this change on current master and re-push?

gtjoseph (approved):

jcolp (approved):

mbradeen (approved):

Issue #248, opened 2023-08-09
[bug]: core_local: Local channels cannot have slashes in the destination


InterLinked1:
### Severity

Major

### Versions

18.9.0

### Components/Modules

core_local

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-30013

Currently, a slash cannot be used in dialing any local channels. The code for core_local uses strchr to parse options on the first '/' it finds, which butchers the result if this is done. 

For example: Local/PJSIP/mydevice and Local/foo/bar will both currently fail. 

However, including slashes in extensions, generally, is perfectly fine and works in most other contexts in Asterisk. This fixes this oversight with the Local channel driver so that extensions containing the slash can be specified.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #249, opened 2023-08-09
core_local: Fix local channel parsing with slashes.

InterLinked1:
Currently, trying to call a Local channel with a slash in the extension will fail due to the parsing of characters after such a slash as being dial modifiers. Additionally, core_local is inconsistent and incomplete with
its parsing of Local dial strings in that sometimes it uses the first slash and at other times it uses the last.

For instance, something like DAHDI/5 or PJSIP/device is a perfectly usable extension in the dialplan, but Local channels in particular prevent these from being called.

This creates inconsistent behavior for users, since using a slash in an extension is perfectly acceptable, and using a Goto to accomplish this works fine, but if specified through a Local channel, the parsing prevents this.

This fixes this by explicitly parsing options from the last slash in the extension, rather than the first one, which doesn't cause an issue for extensions with slashes.

ASTERISK-30013 #close
Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/18369

Resolves: #248

### DIFF
--- a/main/core_local.c
+++ b/main/core_local.c
@@ -315,12 +315,6 @@ static int local_devicestate(const char *data)
 	struct local_pvt *lp;
 	struct ao2_iterator it;
 
-	/* Strip options if they exist */
-	opts = strchr(exten, '/');
-	if (opts) {
-		*opts = '\0';
-	}
-
 	context = strchr(exten, '@');
 	if (!context) {
 		ast_log(LOG_WARNING,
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #230, opened 2023-08-08
[bug]: PJSIP_RESPONSE_HEADERS function documentation is misleading


alipsitz-sf:
### Severity

Trivial

### Versions

18.16.0

### Components/Modules

pjsip, Dialplan

### Operating Environment

Debian bullseye/sid, FreePBX

### Frequency of Occurrence

None

### Issue Description

The [documentation for PJSIP_HEADERS](https://docs.asterisk.org/Asterisk_18_Documentation/API_Documentation/Dialplan_Functions/PJSIP_HEADERS/) describes a parameter called `prefix` and uses the wording "if specified", which heavily implies that the parameter is optional.

However, [in the actual implementation](https://github.com/asterisk/asterisk/blob/master/res/res_pjsip_header_funcs.c#L761-L764) a zero-length `prefix` is not allowed.

Admittedly, it would be more convenient for my use-case if the function worked as described in the documentation (in other words, the parameter is optional and when not provided the function returns all headers). I understand that borders on a feature request, so it would be nice to have the documentation say "required" instead of "if specified".

### Relevant log output

```shell
res_pjsip_header_funcs.c: This function requires a pattern.
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #250, opened 2023-08-09
res_pjsip_header_funcs: Make prefix argument optional.

InterLinked1:
The documentation for PJSIP_HEADERS claims that
prefix is optional, but in the code it is actually not. However, there is no inherent reason for this, as users may want to retrieve all header names, not just those beginning with a certain prefix.

This makes the prefix optional for this function,
simply fetching all header names if not specified. As a result, the documentation is now correct.

Resolves: #230

UserNote: The prefix argument to PJSIP_HEADERS is now optional. If not specified, all header names will be returned.


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #179, opened 2023-06-28
[bug]: Queue strategy “Linear” with Asterisk 20 on Realtime


celsoannes:
### Severity

Minor

### Versions

Asterisk 20.1.0

### Components/Modules

app_queue

### Operating Environment

```
Distributor ID: Debian
Description:    Debian GNU/Linux 11 (bullseye)
Release:        11
Codename:       bullseye
```
```
psql (PostgreSQL) 13.11 (Debian 13.11-0+deb11u1)
```

### Frequency of Occurrence

Constant

### Issue Description

When using realtime to create a linear queue using PostgreSQL, the order in which extensions should be called is not respected.

I am defining the order in which extensions should be called based on the **`uniqueid`** column of the **`queue_members`** table. However, regardless of the order I specify or the order in which I perform the **`INSERT`** operations, Asterisk always sorts based on the **`interface`** column:

```
SELECT * FROM public.queue_members ORDER BY uniqueid ;
 queue_name | interface  | membername | state_interface  | penalty | paused | uniqueid | wrapuptime | ringinuse
------------+------------+------------+------------------+---------+--------+----------+------------+-----------
 demo       | PJSIP/4001 |            | hit:4001@default |         |      0 |        1 |            |
 demo       | PJSIP/2000 |            | hit:2000@default |         |      0 |        2 |            |
 demo       | PJSIP/4000 |            | hit:4000@default |         |      0 |        3 |            |
(3 rows)
```
The order defined in the database was:

1. 4001
2. 2000
3. 4000

Asterisk always load and ring queue based on the **`interface`** numeric order:

```
CLI> queue show
demo has 0 calls (max unlimited) in 'linear' strategy (0s holdtime, 0s talktime), W:0, C:0, A:0, SL:0.0%, SL2:0.0% within 0s
   Members:
      PJSIP/2000 (ringinuse disabled) (realtime) (Invalid) has taken no calls yet (login was 1191 secs ago)
      PJSIP/4000 (ringinuse disabled) (realtime) (Invalid) has taken no calls yet (login was 1191 secs ago)
      PJSIP/4001 (ringinuse disabled) (realtime) (Invalid) has taken no calls yet (login was 1191 secs ago)
   No Callers
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

seanbright:
[...]

PR #252, opened 2023-08-10
extconfig: Allow explicit DB result set ordering to be disabled.

seanbright:
Added a new boolean configuration flag - `order_multi_row_results_by_initial_column` - to both res_pgsql.conf and res_config_odbc.conf that allows the administrator to disable the explicit `ORDER BY` that was previously being added to all generated SQL statements that returned multiple rows.

Fixes: #179


seanbright:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #253, opened 2023-08-10
app_gosub patch appear to have broken predial handlers that utilize macros that call gosubs


creslin287:
### Severity

Minor

### Versions

18.19.0, probably other branches as well

### Components/Modules

app_stack

### Operating Environment

CentOS 7.x

### Frequency of Occurrence

Frequent

### Issue Description

We've got a number of macro predial handlers that we're still using (boo!), and some of them call gosub within them (as part of the effort to migrate away from macros).

There's some sort of problem we're running into caused by patch 617dad4cba1513dddce87b8e95a61415fb587cf1.  It seems to have broke them.  More specifically, when the Return() app is called within the gosub, it can't seem to find the return patch back to the calling macro (again, gosub called from a predial macro), and outputs the error, and app_stack:return_exec() returns -1 to app_macro, which then bubbles up to app_dial() causing the call to hangup.

Didn't want to dig too far into the patch to figure out the fix, so I'm just going to push up a revert commit.

### Relevant log output

```shell
[2023-08-09 09:39:29] DEBUG[396][C-00000006]: pbx_variables.c:798 pbx_substitute_variables_helper_full_location: Function FC_GET_TENANT_AREA_CODE_CONTROL(Snow1) result is '0'
[2023-08-09 09:39:29] DEBUG[396][C-00000006]: pbx_variables.c:874 pbx_substitute_variables_helper_full_location: Expression result is '1'
[2023-08-09 09:39:29] DEBUG[396][C-00000006]: pbx.c:2939 pbx_extension_helper: Launching 'ExecIf'
    -- Executing [s@area-code-recording-settings:5] ExecIf("Local/2566948922@dial-local-callgroup-external-00000004;1", "1?Return()") in new stack
[2023-08-09 09:39:29] ERROR[396][C-00000006]: app_stack.c:423 return_exec: Return without Gosub: stack is empty
[2023-08-09 09:39:29] DEBUG[396][C-00000006]: app_macro.c:476 _macro_exec: Spawn extension (area-code-recording-settings,s,5) exited non-zero on 'Local/2566948922@dial-local-callgroup-external-00000004;1' in macro 'inbound-group-recording'
  == Spawn extension (area-code-recording-settings, s, 5) exited non-zero on 'Local/2566948922@dial-local-callgroup-external-00000004;1' in macro 'inbound-group-recording'
[...]

PR #254, opened 2023-08-10
Revert "app_stack: Print proper exit location for PBXless channels."

creslin287:
This reverts commit 617dad4cba1513dddce87b8e95a61415fb587cf1.

apps/app_stack.c: Revert buggy gosub patch

This seems to break the case when a predial macro calls a gosub.
When the gosub calls return, the Return function outputs:

app_stack.c:423 return_exec: Return without Gosub: stack is empty

This returns -1 to the calling macro, which returns to app_dial
and causes the call to hangup instead of proceeding with the macro
that invoked the gosub.

Resolves: #253

### DIFF
--- a/apps/app_stack.c
+++ b/apps/app_stack.c
@@ -375,26 +375,6 @@ static int pop_exec(struct ast_channel *chan, const char *data)
 	return res;
 }
 
-static int frames_left(struct ast_channel *chan)
-{
-	struct ast_datastore *stack_store;
-	struct gosub_stack_list *oldlist;
-	int exists;
-
-	ast_channel_lock(chan);
-	stack_store = ast_channel_datastore_find(chan, &stack_info, NULL);
-	if (!stack_store) {
-		ast_channel_unlock(chan);
-		return -1;
-	}
-	oldlist = stack_store->data;
-	AST_LIST_LOCK(oldlist);
-	exists = oldlist->first ? 1 : 0;
-	AST_LIST_UNLOCK(oldlist);
-	ast_channel_unlock(chan);
-	return exists;
-}
-
 static int return_exec(struct ast_channel *chan, const char *data)
 {
 	struct ast_datastore *stack_store;
@@ -402,7 +382,6 @@ static int return_exec(struct ast_channel *chan, const char *data)
 	struct gosub_stack_list *oldlist;
 	const char *retval = data;
 	int res = 0;
-	int lastframe;
 
 	ast_channel_lock(chan);
 	if (!(stack_store = ast_channel_datastore_find(chan, &stack_info, NULL))) {
@@ -414,7 +393,6 @@ static int return_exec(struct ast_channel *chan, const char *data)
 	oldlist = stack_store->data;
 	AST_LIST_LOCK(oldlist);
 	oldframe = AST_LIST_REMOVE_HEAD(oldlist, entries);
-	lastframe = oldlist->first ? 0 : 1;
 	AST_LIST_UNLOCK(oldlist);
 
 	if (!oldframe) {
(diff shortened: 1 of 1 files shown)


creslin287:
cherry-pick-to: 20
cherry-pick-to: 18

seanbright (approved):

creslin287:
> Given that you didn't cherry pick to 21, is this not applicable to master since it only affects macros? If this is only happening with macros I don't think this is applicable to master.

Hey!  I can cherry pick to 21 as well. To be honest, I didn't dig deep enough into the patch to figure out what was going on enough to feel that there wouldn't be any other unintended behavioral consequences that I hadn't caught, perhaps in gosub usage. It just seemed safer to do a consistent across all applied branches (and keeps the branches more consistent).

I'd forgotten about 21 though, I can add the cherry pick there as well.

I guess taking a step back though, @jcolp, @gtjoseph, @bkford, and others - what are your thoughts?  Should I leave the patch in master and 21 and revert only for 20 and 18 or should I just revert across the board?

InterLinked1:
> > Given that you didn't cherry pick to 21, is this not applicable to master since it only affects macros? If this is only happening with macros I don't think this is applicable to master.
> 
> Hey! I can cherry pick to 21 as well. To be honest, I didn't dig deep enough into the patch to figure out what was going on enough to feel that there wouldn't be any other unintended behavioral consequences that I hadn't caught, perhaps in gosub usage. It just seemed safer to do a consistent across all applied branches (and keeps the branches more consistent).
> 
> I'd forgotten about 21 though, I can add the cherry pick there as well.
> 
> I guess taking a step back though, @jcolp, @gtjoseph, @bkford, and others - what are your thoughts? Should I leave the patch in master and 21 and revert only for 20 and 18 or should I just revert across the board?

You said it only happens with macros. Macros don't exist in 21 or master anymore, that's why I asked, it might not be capable of causing any issues there, and then the fix needs to be readded again if there isn't an issue.

I don't immediately see what the issue here would be with macros, but I never used them myself so I'm not familiar with that code, just that they can cause problems elsewhere so wondering if that's what happened here.

Do you have a POC snippet handy that causes this, that I could use to do some further testing?

InterLinked1:
@creslin287 I'm able to reproduce your issue and understand the root cause. Inside a macro, the channel has no PBX, so it's triggering the modification logic.

A simple fix is to not do that if the channel is executing a macro, because macros cause some different behavior. A simple diff:

```
- if (ast_channel_pbx(chan) || !lastframe) {
+ if (ast_channel_pbx(chan) || !lastframe || ast_channel_macrocontext(chan)) {
```

Still, this change does not apply to 21 or master, only 18 and 20, as this symbol isn't even available there.

gtjoseph (changes requested):
So if this doesn't apply to 21 or master then I think the PR needs to be resubmitted against the 20 branch with a cherry-pick to 18, yes?

InterLinked1:
> So if this doesn't apply to 21 or master then I think the PR needs to be resubmitted against the 20 branch with a cherry-pick to 18, yes?

I would think so, yes.
I can submit a patch as proposed above which fixes the issue here without undoing the previous bug fix.

seanbright:
We should merge this PR. Any subsequent changes to this will need changes to the testsuite to confirm we are not introducing bugs like this.

jcolp:
I agree. A way to ensure there is test coverage under 21/master should also occur.

jcolp (approved):

creslin287:
Hey, what is left as far as this PR goes?  I saw that George requested changes.  Is the consensus that I should apply the revert to only the 18 and 20 branches and @InterLinked1 will submit a follow up patch, or is it revert across the branches and @InterLinked1 will attempt a version 2 at some point with tests?

InterLinked1:
> Hey, what is left as far as this PR goes? I saw that George requested changes. Is the consensus that I should apply the revert to only the 18 and 20 branches and @InterLinked1 will submit a follow up patch, or is it revert across the branches and @InterLinked1 will attempt a version 2 at some point with tests?

I think the plan is to let your revert go through as is for all branches at this moment, and I will follow-up later to re-fix the original bug (so I think you're good). But I think George has to change his review to reflect that... GitHub PRs seem to get "stuck" otherwise and won't show up so might need @gtjoseph 's attention.

gtjoseph (approved):

gtjoseph:
Yeah, I got distracted, sorry.  Cherry-pick testing in progress, then I'll merge.

Issue #255, opened 2023-08-11
[bug]: pjsip_endpt_register_module: Assertion "Too many modules registered"


gtjoseph:
### Severity

Major

### Versions

18,20,21,master

### Components/Modules

pjproject_bundled

### Operating Environment

Rocky8

### Frequency of Occurrence

Constant

### Issue Description

PJSIP_MAX_MODULE controls how many modules can be registered for callbacks on a pjproject endpoint.  The default is 32 with about 8 taken up by pjproject itself leaving us with 24 available slots.  With recent commits, we've exhausted those slots which results in the error message.  If compiled in development mode, Asterisk will abort on this condition but if compiled in production mode, Asterisk will continue but the modules that failed to register for callbacks probably won't operate correctly.

We just need to increase PJSIP_MAX_MODULE.

### Relevant log output

```shell
asterisk: ../src/pjsip/sip_endpoint.c:183: pjsip_endpt_register_module: Assertion `!"Too many modules registered!"' failed.
[1]    2675363 IOT instruction  sudo asterisk -fcg
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #256, opened 2023-08-11
pjproject_bundled: Increase PJSIP_MAX_MODULE to 38

gtjoseph:
The default is 32 with 8 being used by pjproject itself.  Recent
commits have put us over the limit resulting in assertions in
pjproject.  Since this value is used in invites, dialogs,
transports and subscriptions as well as the global pjproject
endpoint, we don't want to increase it too much.

Resolves: #255


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

mbradeen (approved):

Issue #265, opened 2023-08-21
[bug]: app_macro isn't locking around channel datastore access


creslin287:
### Severity

Minor

### Versions

18

### Components/Modules

app_macro

### Operating Environment

CentOS 7.x

### Frequency of Occurrence

Occasional

### Issue Description

app_macro isn't properly locking around channel datastore access, causing channel datastore list corruption in some cases which can cause a crash.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

InterLinked1:
@gtjoseph Shouldn't this issue have been closed by the merge of #266?
I wonder if merges to non-master branches are not closing issues when they should be.

gtjoseph:
> @gtjoseph Shouldn't this issue have been closed by the merge of #266?
> I wonder if merges to non-master branches are not closing issues when they should be.

Yeah it's a GitHub thing. They only automatically link issues to PRs on the master branch. It's one of the reasons for us changing to submitting to master and cherry picking down instead of submitting to the lowest branch and submitting up.  I'll add an automation to fix this eventually.

PR #266, opened 2023-08-21
app_macro: Fix locking around datastore access

creslin287:
app_macro sometimes would crash due to datastore list corruption on the
channel because of lack of locking around find and create process for
the macro datastore. This patch locks the channel lock prior to protect
against this problem.

Resolves: #265


creslin287:
cherry-pick-to: 18

jcolp (approved):

PR #270, opened 2023-08-24
app_audiosocket: Fixed timeout with -1 to avoid busy loop.

nphantom:
Resolves: asterisk#234

### DIFF
--- a/apps/app_audiosocket.c
+++ b/apps/app_audiosocket.c
@@ -180,7 +180,7 @@ static int audiosocket_run(struct ast_channel *chan, const char *id, int svc)
 	chanName = ast_channel_name(chan);
 
 	while (1) {
-
+		ms = -1;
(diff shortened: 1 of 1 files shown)


nphantom:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

apps/app_audiosocket.c
```diff
@@ -180,7 +180,7 @@ static int audiosocket_run(struct ast_channel *chan, const char *id, int svc)
 	chanName = ast_channel_name(chan);
 
 	while (1) {
-
+		ms = 10;
```
jcolp, line 183: Can this not just be -1? Then the function will block until there is activity on either the channel or passed in file descriptor.

nphantom: > Can this not just be -1? Then the function will block until there is activity on either the channel or passed in file descriptor.

You're right, -1 is a better choice.

jcolp (approved):

Issue #271, opened 2023-08-24
[new-feature]: sig_analog: Add Call Waiting Deluxe support.


InterLinked1:
### Feature Description

Imported from JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-30373

This adds support for Call Waiting Deluxe to sig_analog, i.e. the advanced disposition options to enhance Call Waiting.

PR #272, opened 2023-08-24
sig_analog: Add Call Waiting Deluxe support.

InterLinked1:
Adds support for Call Waiting Deluxe options to enhance the current call waiting feature.

As part of this change, a mechanism is also added that allows a channel driver to queue an audio file for Dial() to play, which is necessary for the announcement function.

ASTERISK-30373 #close

Imported from Gerrit: https://gerrit.asterisk.org/c/asterisk/+/19748

Resolves: #271

### DIFF
--- a/apps/app_dial.c
+++ b/apps/app_dial.c
@@ -1686,6 +1686,15 @@ static struct ast_channel *wait_for_answer(struct ast_channel *in,
 				case AST_CONTROL_PVT_CAUSE_CODE:
 					ast_indicate_data(in, AST_CONTROL_PVT_CAUSE_CODE, f->data.ptr, f->datalen);
 					break;
+				case AST_CONTROL_PLAYBACK_BEGIN:
+					if (!f->data.ptr) {
+						ast_log(LOG_WARNING, "Got playback begin directive without filename on %s\n", ast_channel_name(c));
+					} else {
+						const char *filename = f->data.ptr;
+						ast_verb(3, "Playing audio file %s on %s\n", filename, ast_channel_name(in));
+						ast_streamfile(in, filename, ast_channel_language(in));
+					}
+					break;
 				case -1:
 					if (single && !caller_entertained) {
 						ast_verb(3, "%s stopped sounds\n", ast_channel_name(c));
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -9191,6 +9191,26 @@ static int dahdi_write(struct ast_channel *ast, struct ast_frame *frame)
 		return -1;
 	}
 
+	if (p->sig == SIG_FXOLS || p->sig == SIG_FXOKS || p->sig == SIG_FXOGS) {
+		struct analog_pvt *analog_p = p->sig_pvt;
+		if (analog_p->callwaitdeluxepending) {
+			unsigned int mssinceflash = ast_tvdiff_ms(ast_tvnow(), analog_p->flashtime);
+			if (mssinceflash >= 1000) {
+				/* Timer expired: the user hasn't yet selected an option. Take the default action and get on with it. */
+				/* Note: If in the future Advanced Call Waiting Deluxe (*76) is supported, then as part of the
+				 * dialing code, we'll need to automatically invoke the preselected behavior about 2-3 seconds after
+				 * the call waiting begins (this allows for the SAS, CAS, and CWCID spill to be sent first).
+				 */
+				analog_p->callwaitdeluxepending = 0;
+				analog_callwaiting_deluxe(analog_p, 0);
+			}
+			ast_mutex_unlock(&p->lock);
+			/* The user shouldn't hear anything after hook flashing, until a decision is made, by the user or when the timer expires. */
+			ast_debug(5, "Dropping frame since Call Waiting Deluxe pending on %s\n", ast_channel_name(ast));
+			return 0;
+		}
+	}
+
 	if (p->dialing) {
 		ast_mutex_unlock(&p->lock);
 		ast_debug(5, "Dropping frame since I'm still dialing on %s...\n",
@@ -12947,6 +12967,7 @@ static struct dahdi_pvt *mkintf(int channel, const struct dahdi_chan_conf *conf,
 
 		tmp->usedistinctiveringdetection = usedistinctiveringdetection;
 		tmp->callwaitingcallerid = conf->chan.callwaitingcallerid;
+		tmp->callwaitingdeluxe = conf->chan.callwaitingdeluxe; /* Not used in DAHDI pvt, only analog pvt */
 		tmp->threewaycalling = conf->chan.threewaycalling;
 		tmp->threewaysilenthold = conf->chan.threewaysilenthold;
 		tmp->calledsubscriberheld = conf->chan.calledsubscriberheld; /* Not used in chan_dahdi.c, just analog pvt, but must exist on the DAHDI pvt anyways */
@@ -13285,6 +13306,7 @@ static struct dahdi_pvt *mkintf(int channel, const struct dahdi_chan_conf *conf,
 					break;
 				}
 				analog_p->callwaitingcallerid = conf->chan.callwaitingcallerid;
+				analog_p->callwaitingdeluxe = conf->chan.callwaitingdeluxe;
 				analog_p->ringt = conf->chan.ringt;
 				analog_p->ringt_base = ringt_base;
 				analog_p->onhooktime = time(NULL);
@@ -18413,6 +18435,8 @@ static int process_dahdi(struct dahdi_chan_conf *confp, const char *cat, struct 
 			confp->chan.callwaiting = ast_true(v->value);
 		} else if (!strcasecmp(v->name, "callwaitingcallerid")) {
 			confp->chan.callwaitingcallerid = ast_true(v->value);
+		} else if (!strcasecmp(v->name, "callwaitingdeluxe")) {
+			confp->chan.callwaitingdeluxe = ast_true(v->value);
 		} else if (!strcasecmp(v->name, "context")) {
 			ast_copy_string(confp->chan.context, v->value, sizeof(confp->chan.context));
 		} else if (!strcasecmp(v->name, "language")) {
--- a/channels/chan_iax2.c
+++ b/channels/chan_iax2.c
@@ -1447,6 +1447,7 @@ static int iax2_is_control_frame_allowed(int subtype)
 	case AST_CONTROL_STREAM_RESTART:
 	case AST_CONTROL_STREAM_REVERSE:
 	case AST_CONTROL_STREAM_FORWARD:
+	case AST_CONTROL_PLAYBACK_BEGIN: /* Only supported by app_dial currently */
 		/* None of these playback stream control frames should go across the link. */
 	case AST_CONTROL_RECORD_CANCEL:
 	case AST_CONTROL_RECORD_STOP:
--- a/channels/sig_analog.c
+++ b/channels/sig_analog.c
@@ -1590,6 +1590,161 @@ static int analog_handles_digit(struct ast_frame *f)
 	}
 }
 
+enum callwaiting_deluxe_option {
+	CWD_CONFERENCE = '3',
+	CWD_HOLD = '6',
+	CWD_DROP = '7',
+	CWD_ANNOUNCEMENT = '8',
+	CWD_FORWARD = '9',
+};
+
+static const char *callwaiting_deluxe_optname(int option)
+{
+	switch (option) {
+	case CWD_CONFERENCE:
+		return "CONFERENCE";
+	case CWD_HOLD:
+		return "HOLD";
+	case CWD_DROP:
+		return "DROP";
+	case CWD_ANNOUNCEMENT:
+		return "ANNOUNCEMENT";
+	case CWD_FORWARD:
+		return "FORWARD";
+	default:
+		return "DEFAULT";
+	}
+}
+
+int analog_callwaiting_deluxe(struct analog_pvt *p, int option)
+{
+	const char *announce_var;
+	char announcement[PATH_MAX];
+
+	ast_debug(1, "Handling Call Waiting on channel %d with option %c: treatment %s\n", p->channel, option, callwaiting_deluxe_optname(option));
+
+	if (!p->subs[ANALOG_SUB_CALLWAIT].owner) {
+		/* This can happen if the caller hook flashes and the call waiting hangs up before the CWD timer expires (1 second) */
+		ast_debug(1, "Call waiting call disappeared before it could be handled?\n");
+		return -1;
+	}
+
+	analog_lock_sub_owner(p, ANALOG_SUB_CALLWAIT);
+	if (!p->subs[ANALOG_SUB_CALLWAIT].owner) {
+		ast_log(LOG_WARNING, "Whoa, the call-waiting call disappeared.\n");
+		return -1;
+	}
+
+	/* Note that when p->callwaitdeluxepending, dahdi_write will drop incoming frames to the channel,
+	 * since the user shouldn't hear anything after flashing until either a DTMF has been received
+	 * or it's been a second and the decision is made automatically. */
+
+	switch (option) {
+	case CWD_CONFERENCE:
+		/* We should never have a call waiting if we have a 3-way anyways, but check just in case,
+		 * there better be no existing SUB_THREEWAY since we're going to make one (and then swap the call wait to it) */
+		if (p->subs[ANALOG_SUB_THREEWAY].owner) {
+			ast_channel_unlock(p->subs[ANALOG_SUB_CALLWAIT].owner);
+			ast_log(LOG_ERROR, "Already have a 3-way call on channel %d, can't conference!\n", p->channel);
+			return -1;
+		}
+
+		/* To conference the incoming call, swap it from SUB_CALLWAIT to SUB_THREEWAY,
+		 * and then the existing 3-way logic will ensure that flashing again will drop the call waiting */
+		analog_alloc_sub(p, ANALOG_SUB_THREEWAY);
+		analog_swap_subs(p, ANALOG_SUB_THREEWAY, ANALOG_SUB_CALLWAIT);
+		analog_unalloc_sub(p, ANALOG_SUB_CALLWAIT);
+
+		ast_verb(3, "Building conference call with %s and %s\n", ast_channel_name(p->subs[ANALOG_SUB_THREEWAY].owner), ast_channel_name(p->subs[ANALOG_SUB_REAL].owner));
+		analog_set_inthreeway(p, ANALOG_SUB_THREEWAY, 1);
+		analog_set_inthreeway(p, ANALOG_SUB_REAL, 1);
+
+		if (ast_channel_state(p->subs[ANALOG_SUB_THREEWAY].owner) == AST_STATE_RINGING) {
+			ast_setstate(p->subs[ANALOG_SUB_THREEWAY].owner, AST_STATE_UP);
+			ast_queue_control(p->subs[ANALOG_SUB_THREEWAY].owner, AST_CONTROL_ANSWER);
+			/* Stop the ringing on the call wait channel (yeah, apparently this is how it's done) */
+			ast_queue_hold(p->subs[ANALOG_SUB_THREEWAY].owner, p->mohsuggest);
+			ast_queue_unhold(p->subs[ANALOG_SUB_THREEWAY].owner);
+		}
+		analog_stop_callwait(p);
+
+		ast_channel_unlock(p->subs[ANALOG_SUB_THREEWAY].owner); /* Unlock what was originally SUB_CALLWAIT */
+		break;
+	case CWD_HOLD: /* The CI-7112 Visual Director sends "HOLD" for "Play hold message" rather than "ANNOUNCEMENT". For default behavior, nothing is actually sent. */
+	case CWD_ANNOUNCEMENT:
+		/* We can't just call ast_streamfile here, this thread isn't responsible for media on the call waiting channel.
+		 * Indicate to the dialing channel in app_dial that it needs to play media.
+		 *
+		 * This is a lot easier than other ways of trying to send early media to the channel
+		 * (such as every call from the core to dahdi_read, sending the channel one frame of the audio file, etc.)
+		 */
+
+		/* There's not a particularly good stock audio prompt to use here. The Pat Fleet library has some better
+		 * ones but we want one that is also in the default Allison Smith library. "One moment please" works okay.
+		 * Check if a variable containing the prompt to use was specified on the call waiting channel, and
+		 * fall back to a reasonable default if not. */
+
+		/* The SUB_CALLWAIT channel is already locked here, no need to lock and unlock to get the variable. */
+		announce_var = pbx_builtin_getvar_helper(p->subs[ANALOG_SUB_CALLWAIT].owner, "CALLWAITDELUXEANNOUNCEMENT");
+		ast_copy_string(announcement, S_OR(announce_var, "one-moment-please"), sizeof(announcement));
+		ast_debug(2, "Call Waiting Deluxe announcement for %s: %s\n", ast_channel_name(p->subs[ANALOG_SUB_CALLWAIT].owner), announcement);
+		ast_channel_unlock(p->subs[ANALOG_SUB_CALLWAIT].owner);
+		/* Tell app_dial what file to play. */
+		ast_queue_control_data(p->subs[ANALOG_SUB_CALLWAIT].owner, AST_CONTROL_PLAYBACK_BEGIN, announcement, strlen(announcement) + 1);
+		/* Unlike all the other options, the call waiting is still active with this option,
+		 * so we don't call analog_stop_callwait(p)
+		 * The call waiting will continue to be here, and at some later point the user can flash again and choose a finalizing option
+		 * (or even queue the announcement again... and again... and again...)
+		 */
+		break;
+	case CWD_FORWARD:
+		/* Go away, call waiting, call again some other day... */
+		analog_stop_callwait(p);
+		/* Can't use p->call_forward exten because that's for *72 forwarding, and sig_analog doesn't
+		 * have a Busy/Don't Answer call forwarding exten internally, so let the dialplan deal with it.
+		 * by sending the call to the 'f' extension.
+		 */
+		ast_channel_call_forward_set(p->subs[ANALOG_SUB_CALLWAIT].owner, "f");
+		ast_channel_unlock(p->subs[ANALOG_SUB_CALLWAIT].owner);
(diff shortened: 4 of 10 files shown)


InterLinked1:
@asteriskteam Would someone mind giving this review in particular a look soon, if they have time? Don't want to nag, but because of the monolithic architecture of chan_dahdi, it's very difficult to do concurrent development for chan_dahdi so things have to be serialized. Thanks!

include/asterisk/channel.h
```diff
@@ -4076,6 +4077,7 @@ const char *ast_channel_linkedid(const struct ast_channel *chan);
 const char *ast_channel_parkinglot(const struct ast_channel *chan);
 const char *ast_channel_hangupsource(const struct ast_channel *chan);
 const char *ast_channel_dialcontext(const struct ast_channel *chan);
+const char *ast_channel_dial_announcement(const struct ast_channel *chan);
```
jcolp, line 4080: I am not a fan of using the channel API or adding a stringfield for the sole purpose of communicating to app_dial that an announcement  is to be played on an interface when dial. I don't have an alternative of how to do it, but an alternative should be explored.

jcolp (changes requested):

jcolp:
> @asteriskteam Would someone mind giving this review in particular a look soon, if they have time? Don't want to nag, but because of the monolithic architecture of chan_dahdi, it's very difficult to do concurrent development for chan_dahdi so things have to be serialized. Thanks!

The reality is that community members have no interest in reviewing these kind of changes, and personally I put such changes at the bottom of my todo list. You can continue to do development of chan_dahdi for your needs/interests/purposes, but the review times here are likely to continue to be a long time.

include/asterisk/channel.h
InterLinked1: It was the cleanest way I could think to do it at the time, but I agree it's slightly awkward.

I have two thoughts about this:
- A variable is already set on the channel containing the filename to play (since this is the most flexible, least core intrusive way to do it...). The API is simply to indicate that the file should be played. Is there an existing control frame that might be appropriate to use here? Alternately, would adding a new control frame altogether be more appropriate than the current implementation?
- If the answer to both of the above is 'no', would setting another variable on the channel to indicate that a file should be played be acceptable to you? I didn't like this approach because I suspect it will perform worse, given that every time the channel is polled, we'll have to traverse all the channel variables to check if this variable exists, making this linear time rather than a constant time operation. I would prefer an approach that doesn't have a performance impact.
- Finally, if the first bullet is not acceptable but the second is, how about using a control frame to signal to app_dial that "something interesting" has happened, and only then checking if the variable exists? I think there are things like SRCCHANGE that seemed to be used for things like this.

jcolp: There is no existing control frame specifically for this. There are other playback stream control frames, provided that it is done in a generic fashion and documented a control frame could be added alongside these to start playback.

include/asterisk/frame.h
```diff
@@ -312,6 +312,7 @@ enum ast_control_frame_type {
 	AST_CONTROL_STREAM_TOPOLOGY_REQUEST_CHANGE = 35,    /*!< Channel indication that a stream topology change has been requested */
 	AST_CONTROL_STREAM_TOPOLOGY_CHANGED = 36,           /*!< Channel indication that a stream topology change has occurred */
 	AST_CONTROL_STREAM_TOPOLOGY_SOURCE_CHANGED = 37,    /*!< Channel indication that one of the source streams has changed its source */
+	AST_CONTROL_PLAYBACK_BEGIN = 38,                    /*!< Channel indication that a dialing interface should begin playback of an audio file */
```
jcolp, line 315: This should be part of AST_CONTROL_STREAM further below.

jcolp (changes requested):

apps/app_dial.c
```diff
@@ -1686,6 +1687,19 @@ static struct ast_channel *wait_for_answer(struct ast_channel *in,
 				case AST_CONTROL_PVT_CAUSE_CODE:
 					ast_indicate_data(in, AST_CONTROL_PVT_CAUSE_CODE, f->data.ptr, f->datalen);
 					break;
+				case AST_CONTROL_PLAYBACK_BEGIN:
+					filename = pbx_builtin_getvar_helper(c, "DIALING_PLAYBACK_ANNOUNCEMENT");
```
jcolp, line 1691: The filename should be a data payload on the frame, and not just put as a dialplan variable.

channels/chan_iax2.c
```diff
@@ -1447,6 +1447,7 @@ static int iax2_is_control_frame_allowed(int subtype)
 	case AST_CONTROL_STREAM_RESTART:
 	case AST_CONTROL_STREAM_REVERSE:
 	case AST_CONTROL_STREAM_FORWARD:
+	case AST_CONTROL_PLAYBACK_BEGIN:
```
jcolp, line 1450: It should be documented that AST_CONTROL_PLAYBACK_BEGIN is only supported by app_dial.

jcolp (changes requested):
Additionally, unless someone else gives this a review I will not be accepting the change. You'll have to convince someone else to review this as well.

asteriskteam:

asteriskteam:

asteriskteam:

asteriskteam:

InterLinked1:
@jcolp Would you mind reopening this? Thanks!

channels/chan_dahdi.c
```diff
@@ -9191,6 +9191,26 @@ static int dahdi_write(struct ast_channel *ast, struct ast_frame *frame)
 		return -1;
 	}
 
+	if (p->sig == SIG_FXOLS || p->sig == SIG_FXOKS || p->sig == SIG_FXOGS) {
+		struct analog_pvt *analog_p = p->sig_pvt;
+		if (analog_p->callwaitdeluxepending) {
+			unsigned int mssinceflash = ast_tvdiff_ms(ast_tvnow(), analog_p->flashtime);
```
Allan-N, line 9197: I want to question anything that's driven by timing. Does having a 1 second timer suggest that if I'm really quick to do something then I can get a different behavior? What happens if Asterisk (or the system it's hosted on) is faster/slower then expected, could that result in a change in behavior? What if I just can't flash+DTMF within this short window, do I miss out on being able to do something important? One should strive for very predictable results.

channels/sig_analog.c
```diff
@@ -1590,6 +1590,161 @@ static int analog_handles_digit(struct ast_frame *f)
 	}
 }
 
+enum callwaiting_deluxe_option {
+	CWD_CONFERENCE = '3',
+	CWD_HOLD = '6',
+	CWD_DROP = '7',
+	CWD_ANNOUNCEMENT = '8',
+	CWD_FORWARD = '9',
+};
```
Allan-N, line 1599: Why 3, 6, 7, 8, and 9 ?

channels/sig_analog.c
```diff
@@ -1590,6 +1590,161 @@ static int analog_handles_digit(struct ast_frame *f)
...
+
+int analog_callwaiting_deluxe(struct analog_pvt *p, int option)
+{
+	const char *announce_var;
+	char announcement[PATH_MAX];
+
+	ast_debug(1, "Handling Call Waiting on channel %d with option %c: treatment %s\n", p->channel, option, callwaiting_deluxe_optname(option));
+
+	if (!p->subs[ANALOG_SUB_CALLWAIT].owner) {
+		/* This can happen if the caller hook flashes and the call waiting hangs up before the CWD timer expires (1 second) */
```
Allan-N, line 1627: Do we really want to be sensitive to timing?

channels/sig_analog.c
```diff
@@ -1590,6 +1590,161 @@ static int analog_handles_digit(struct ast_frame *f)
...
+		ast_debug(1, "Call waiting call disappeared before it could be handled?\n");
+		return -1;
+	}
+
+	analog_lock_sub_owner(p, ANALOG_SUB_CALLWAIT);
+	if (!p->subs[ANALOG_SUB_CALLWAIT].owner) {
+		ast_log(LOG_WARNING, "Whoa, the call-waiting call disappeared.\n");
```
Allan-N, line 1634: Whoa?

channels/sig_analog.c
```diff
@@ -1590,6 +1590,161 @@ static int analog_handles_digit(struct ast_frame *f)
...
+		return -1;
+	}
+
+	/* Note that when p->callwaitdeluxepending, dahdi_write will drop incoming frames to the channel,
+	 * since the user shouldn't hear anything after flashing until either a DTMF has been received
+	 * or it's been a second and the decision is made automatically. */
```
Allan-N, line 1640: Do we really want to be sensitive to timing?

channels/sig_analog.c
```diff
@@ -1590,6 +1590,161 @@ static int analog_handles_digit(struct ast_frame *f)
...
+		 * Check if a variable containing the prompt to use was specified on the call waiting channel, and
+		 * fall back to a reasonable default if not. */
+
+		/* The SUB_CALLWAIT channel is already locked here, no need to lock and unlock to get the variable. */
+		announce_var = pbx_builtin_getvar_helper(p->subs[ANALOG_SUB_CALLWAIT].owner, "CALLWAITDELUXEANNOUNCEMENT");
+		ast_copy_string(announcement, S_OR(announce_var, "one-moment-please"), sizeof(announcement));
+		ast_debug(2, "Call Waiting Deluxe announcement for %s: %s\n", ast_channel_name(p->subs[ANALOG_SUB_CALLWAIT].owner), announcement);
+		ast_channel_unlock(p->subs[ANALOG_SUB_CALLWAIT].owner);
+		/* Tell app_dial what file to play. */
+		ast_queue_control_data(p->subs[ANALOG_SUB_CALLWAIT].owner, AST_CONTROL_PLAYBACK_BEGIN, announcement, strlen(announcement) + 1);
```
Allan-N, line 1693: It's OK (or essential) to call ast_queue_control_data() after ast_channel_unlock() ?

channels/sig_analog.c
```diff
@@ -1590,6 +1590,161 @@ static int analog_handles_digit(struct ast_frame *f)
...
+			ast_verb(3, "Holding original call and swapping to call waiting on %s\n", ast_channel_name(p->subs[ANALOG_SUB_REAL].owner));
+		}
+
+		/* Stop ringing on the incoming call */
+		ast_queue_hold(p->subs[ANALOG_SUB_REAL].owner, p->mohsuggest);
+		ast_queue_unhold(p->subs[ANALOG_SUB_REAL].owner);
+
+		/* Unlock the call-waiting call that we swapped to real-call. */
+		ast_channel_unlock(p->subs[ANALOG_SUB_REAL].owner);
+	}
```
Allan-N, line 1743: Q? since every case has a call to ast_channel_unlock() I'm curious if all of the above calls can be removed with a single call added here?

channels/sig_analog.c
```diff
@@ -1627,6 +1782,50 @@ void analog_handle_dtmf(struct analog_pvt *p, struct ast_channel *ast, enum anal
...
+			 * the default action (swap) for the call waiting.
+			 *
+			 * dahdi_write will also drop audio if callwaitdeluxepending is set HIGH,
+			 * and also check if flashtime hits 1000, in which case it will set the flag LOW and then take the
+			 * default action, e.g. analog_callwaiting_deluxe(p, 0);
+			 */
+
+			/* Slightly less than 1000, so there's no chance of a race condition
+			 * between do_monitor when it sees flashtime hitting 1000 and us. */
+			if (mssinceflash > 990) {
```
Allan-N, line 1803: Do we really want to be sensitive to timing?

channels/sig_analog.h
```diff
@@ -346,6 +347,12 @@ struct analog_pvt {
 	 * gives a positive reply.
 	 */
 	unsigned int callwaitcas:1;
+
+	/*!
+	 * \brief TRUE if a Call Waiting Deluxe action is currently pending.
+	 */
+	unsigned int callwaitdeluxepending:1;
```
Allan-N, line 354: Above, you have callwaitingdelux but here you have callwaitdeluxepending.  Should this be callwaitingdeluxepending (adding the "ing")?

funcs/func_frame_trace.c
```diff
@@ -379,6 +379,9 @@ static void print_frame(struct ast_frame *frame)
 		case AST_CONTROL_RECORD_MUTE:
 			ast_verbose("SubClass: RECORD_MUTE\n");
 			break;
+		case AST_CONTROL_PLAYBACK_BEGIN:
```
Allan-N, line 382: Maybe keep the order of the "case" statements in line with the enum order (i.e. follow frame.h and move between AST_CONTROL_STREAM_FORWARD and AST_CONTROL_RECORD_CANCEL)?

include/asterisk/frame.h
```diff
@@ -335,6 +335,7 @@ enum ast_control_frame_type {
 	AST_CONTROL_STREAM_RESTART = 1002,	/*!< Indicate to a channel in playback to restart the stream */
 	AST_CONTROL_STREAM_REVERSE = 1003,	/*!< Indicate to a channel in playback to rewind */
 	AST_CONTROL_STREAM_FORWARD = 1004,	/*!< Indicate to a channel in playback to fast forward */
+	AST_CONTROL_PLAYBACK_BEGIN = 1005,	/*!< Indicate to a dialing interface that playback of an audio file should begin on the dialing channel. Currently only supported by app_dial. */
```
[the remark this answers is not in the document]
InterLinked1: No, the frames are not for a channel in playback, they're for `app_dial` which is dialing channel(s). It's not really like any type of frame that exists currently. I don't know that I thought about the wording that much but I don't think it's inaccurate. If you can suggest a better phrasing I'm happy to update it.

configs/samples/chan_dahdi.conf.sample
```diff
@@ -685,6 +685,31 @@ usecallerid=yes
...
+; The following extensions are also available in the dialplan to utilize this functionality:
+; - The CALLWAITDELUXEANNOUNCEMENT variable can be set on the incoming channel to control the
+;   announcement prompt played to the call waiting caller. Default is "one-moment-please".
+; - If the forward option is used, the call waiting channel will be forwarded to the 'f' extension
+;   in the channel's configured context. You can then use any dialplan mechanism to route the call.
+;
+; Default is 'no'.
+;
+;callwaitingdeluxe=yes
+;
```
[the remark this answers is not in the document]
InterLinked1: > I have not found any description of "Call Waiting Deluxe"? What is this?

https://en.wikipedia.org/wiki/Call_waiting#Call_waiting_deluxe

> Why 3, 7, 6/8, and 9 ?

These are the DTMF replies specified in the CWD standard. They are not arbitrary.

> I want to question anything that's driven by timing. With a 1 second trigger does this suggest that if I'm really quick to do something then I can get a different behavior? What happens if Asterisk (or the system it's hosted on) is faster/slower then expected, could that result in a change in behavior? What if I just can't flash+DTMF within this short window, do I miss out on being able to do something important? One should strive for very predictable results.

My recollection is 1,000 ms is the exact timing specified in the standard. It's been so long since I wrote this implementation that I no longer remember the source I used but I remember following it closely.

GR-416[1] 3.1.1.2 would be the ultimate authority on this. However, I don't have a copy of that spec off hand.

The reason it is not longer is that when taking the default action, the timer needs to expire. Waiting more than a second to switch to the call waiting would be a poor user experience. Generally, this feature is machine operated, it can be operated manually by flashing + dialing a digit by hand, but that can be done in less than 1 second.

[1] https://telecom-info.njdepot.ericsson.net/ido/AUX2/GR_416_TOC.i02.pdf

channels/sig_analog.h
```diff
@@ -331,6 +331,7 @@ struct analog_pvt {
 
 	/* XXX: All variables after this are internal */
 	unsigned int callwaiting:1;		/*!< TRUE if call waiting is enabled. (Active option) */
+	unsigned int callwaitingdeluxe:1;	/*!< TRUE if Call Waiting Deluxe options are available */
```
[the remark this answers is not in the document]
InterLinked1: Actually, I think the ordering here probably could be improved, though not exactly in the way you suggest. "callwaitingdeluxe" is a permanent setting and should be grouped with the permanent settings. 2 lines above, it says "All variables after this are internal", which that one is not. I meant to place it after the permanent call waiting flag, but that is called `permcallwaiting`. I will go ahead and adjust this.

channels/sig_analog.c
InterLinked1: They're all calling ast_channel_unlock, but not on the same sub channel, if you look carefully.

channels/sig_analog.c
InterLinked1: Yes, this is related to 1 second being the timeout specified for this feature.

funcs/func_frame_trace.c
InterLinked1: Yes, this switch does appear to be going in order so I'll go ahead and fix this, thanks!

channels/sig_analog.c
```diff
@@ -3310,6 +3511,20 @@ static struct ast_frame *__analog_handle_event(struct analog_pvt *p, struct ast_
 					goto winkflashdone;
 				}
 
+				/* If line has Call Waiting Deluxe, see what the user wants to do.
+				 * Only do this if this is an as yet unanswered call waiting, not an existing, answered SUB_CALLWAIT. */
+				if (ast_channel_state(p->subs[ANALOG_SUB_CALLWAIT].owner) == AST_STATE_RINGING) {
```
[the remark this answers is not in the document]
InterLinked1: I might be missing something here, but what would the difference be? Both of these should just be a field access. (ast_channel_state just returns chan->state)

channels/sig_analog.c
InterLinked1: Yes, this phrasing appears elsewhere in the file.... I'm just trying to make that fit in with those :)

InterLinked1:
@gtjoseph Is there any way to do a "recheck" on this? I'm not sure what this error is about but it's definitely irrelevant to the change. Thanks!

gtjoseph:
> Is there any way to do a "recheck" on this? I'm not sure what this error is about but it's definitely irrelevant to the change. Thanks!

Well, I wouldn't say "definitely". :)  The PR is based on a very old version of the "master" branch that doesn't have commit 64597bf7277d56e38c50244 which fixes the quoting of the 100rel column in contrib/ast-db-manage/config/versions/539f68bede2c_add_peer_supported_to_100rel.py.  If you rebase the PR on current master and force-push, the alembic issue should go away.

channels/sig_analog.c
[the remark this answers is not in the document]
Allan-N: The comments above the "if" say "If line has Call Waiting Deluxe..." but the first check you make is checking for ringing.  Does the order make a difference, not really.

channels/sig_analog.h
Allan-N: For consistency, "I" think naming the variables the same way is a plus.  I also don't see removing the "ing" as being at all helpful.  Lastly, I would put both of the variables together (e.g. these are all of the variables for the call waiting deluxe).

include/asterisk/frame.h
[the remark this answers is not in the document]
Allan-N: Well, if this is not for a channel in playback then the value (and any associated comments) should be split out into their own section. I would also suggest the enum value being in it's own range and not pick then next AST_CONTROL_STREAM_xxx value.  What I see is AST_CONTROL_STREAM_xxx are the 1000's, the AST_CONTROL_RECORD_xxx values are the 1100's.  Maybe set the AST_CONTROL_PLAYBACK_xxx's @ 1150's or 1200's

channels/sig_analog.h
InterLinked1: I don't feel strongly about the presence/absence of "ing" so I can add that in.

But the variables should not be grouped together regardless. One is a config setting and the other is internal. The existing call waiting variables are separated out by this distinction.

configs/samples/chan_dahdi.conf.sample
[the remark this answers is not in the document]
Allan-N: >These are the DTMF replies specified in the CWD standard. They are not arbitrary.

First, channels/sig_analog.c says that "6" is HOLD and "8" is ANNOUNCEMENT so I think that distinction should be made here.  And, for completeness, are there other DTMF replies defined in the CWD standard that should be noted here (or more importantly, in channels/sig_analog.c) ?

>Generally, this feature is machine operated, it can be operated manually by flashing + dialing a digit by hand, but that can be done in less than 1 second.

Did I see a comment somewhere suggesting that *76 might be used? Can a flash + * + 7 + 8 all be completed in less than 1 second?

InterLinked1:
> > Is there any way to do a "recheck" on this? I'm not sure what this error is about but it's definitely irrelevant to the change. Thanks!
> 
> @InterLinked1 Well, I wouldn't say "definitely". :) The PR is based on a very old version of the "master" branch

Yeah, if it hadn't been for the new commit after the GitHub migration, the commit would be from 2022 :)

> that doesn't have commit [64597bf](https://github.com/asterisk/asterisk/commit/64597bf7277d56e38c502442a4446245a5854a75) which fixes the quoting of the 100rel column in contrib/ast-db-manage/config/versions/539f68bede2c_add_peer_supported_to_100rel.py. If you rebase the PR on current master and force-push, the alembic issue should go away.

Okay, just gave that a try, hopefully that fixes it. Surprisingly, it rebased without any issues.

include/asterisk/frame.h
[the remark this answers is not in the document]
InterLinked1: Well, it is somewhat related to "manipulating streams on channels", only it's starting a stream rather than manipulating an existing one as the other ones do.

@jcolp commented in November "There is no existing control frame specifically for this. There are other playback stream control frames, provided that it is done in a generic fashion and documented a control frame could be added alongside these to start playback.", which is why I added it where I did. Maybe someone from the Asterisk team can chime in if it should be in a new category or with the other stream control frames?

configs/samples/chan_dahdi.conf.sample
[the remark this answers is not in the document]
InterLinked1: > > These are the DTMF replies specified in the CWD standard. They are not arbitrary.
> 
> First, channels/sig_analog.c says that "6" is HOLD and "8" is ANNOUNCEMENT so I think that distinction should be made here. And, for completeness, are there other DTMF replies defined in the CWD standard that should be noted here (or more importantly, in channels/sig_analog.c) ?

From a user perspective, there is no difference between 6 and 8 as those are treated the same in the code. Either code can be used with identical results.

There is a bit of confusion there, since they are named different things in the specification. However, common practice amongst CPE is to send 6 rather than 8, which is the reason for this: https://github.com/asterisk/asterisk/pull/272/files#diff-4569fb3f073200b66e3e14551c80f7b7819b4c92ff034e9093e607e1876df1a3R1673

The digit 8 is what is actually assigned for that function in the spec, so both digits receive the same treatment for compatibility.

To actually hold, no DTMF digit is used at all. One just flashes and waits 1 second to swap calls, just like with normal call waiting without call waiting deluxe (although there would be no 1 second delay in that case).

> > Generally, this feature is machine operated, it can be operated manually by flashing + dialing a digit by hand, but that can be done in less than 1 second.
> 
> Did I see a comment somewhere suggesting that *76 might be used? Can a flash + * + 7 + 8 all be completed in less than 1 second?

*76 isn't in play here, that's the vertical service code assigned for Advanced Call Waiting Deluxe, which I am not implementing as part of this change. But that code would never be dialed as part of feature operation regardless, the code is used to select a disposition to occur automatically. The most anyone would ever dial manually is a single DTMF digit. CWD-supporting CPE have dedicated buttons for each of these functions that flash and send the appropriate digit automatically, e.g. the CI-7112.

include/asterisk/frame.h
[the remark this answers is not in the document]
jcolp: AST_CONTROL_STREAM_* is probably misnamed realistically, as it's controlling playback on the channel and not really what we know of as media streams today. I don't have a problem with it being with the other stream control frames, they're all related to playback fundamentally.

Allan-N (approved):

asteriskteam:

InterLinked1:
@jcolp Seems GitHub thinks changes are requested from your last input on this, would you mind giving this another review when you have a moment? Thanks!

asteriskteam:

InterLinked1:
> This PR has been marked stale because it has been in "Changes Requested" or "submitter-action-required" state for 28 days or more. Please make the requested changes within 14 days or the PR will be closed.

@asteriskteam I have made the requested changes, could you please change the status of this? Thank you.

InterLinked1:
@jcolp Any chance you'll be able to take another look at this soon? Everything should be good to go now, but the UI still says it's pending a review from you.
Thanks!

jcolp:
Other reviews are higher on my list, so no idea.

InterLinked1:
> Other reviews are higher on my list, so no idea.

I realize there's no SLA for reviews, but this patch is for a core supported module, and dates back to 2022, making it by far the oldest patch in the review queue. You also additionally asked for a community review - something seldom if ever required for other patches - which in any case was obtained half a year ago now. In light of that, I would at least hope its priority could be reconsidered.

jcolp (approved):

gtjoseph:
Can you rebase on current master and force-push?  The manager checks are infinitely looping because your base branch doesn't have the updateconfig fixes you did a short time ago.

gtjoseph (dismissed):
Can you rebase on current master and force-push? The manager checks are infinitely looping because your base branch doesn't have the updateconfig fixes you did a short time ago.

InterLinked1:
> **Attention!** This pull request may contain issues that could prevent it from being accepted. Please review the checklist below and take the recommended action. If you believe any of these are not applicable, just add a comment and let us know.
> 
> * [ ]  The PR description does not match the commit message body. This can cause confusion for reviewers and future maintainers. GitHub doesn't automatically update the PR description when you update the commit message so if you've updated the commit with a force-push, please update the PR description to match the new commit message body.
> * [ ]  A change was detected to the sample configuration files in ./config/samples but no UserNote or UpgradeNote was found in the commit message. If this PR includes changes that contain new configuration parameters or a change to existing parameters, please include a UserNote in the commit message.  If the changes require some action from the user to preserve existing behavior, please include an UpgradeNote.
> 
> Documentation:
> 
> * [Asterisk Developer Documentation](https://docs.asterisk.org/Development/)
>   
>   * [Code Contribution](https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/)
>   * [Commit Messages](https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages)
>   * [Alembic Scripts](https://docs.asterisk.org/Development/Reference-Information/Other-Reference-Information/Alembic-Scripts/)

The "Imported from Gerrit" line is only in the PR description, I don't think that's needed/useful in the commit message.
I've added a UserNote.

asteriskteam:

Issue #269, opened 2023-08-22
[new-feature]: Allow wildcard CN's in TLS certs for SIP


dirkx:
### Feature Description

chan_sip currently does not accept wildcarded DN's when connecting to a TLS/SSL host.

Below tiny tweak to `tcptls.c` makes that work.

```
static int check_tcptls_cert_name(ASN1_STRING *cert_str, const char *hostname, const char *desc)
{
        unsigned char *str;
        int ret;

        ret = ASN1_STRING_to_UTF8(&str, cert_str);
        if (ret < 0 || !str) {
                return -1;
        }

        if (strlen((char *) str) != ret) {
                ast_log(LOG_WARNING, "Invalid certificate %s length (contains NULL bytes?)\n", desc);

                ret = -1;
        } else if (!strcasecmp(hostname, (char *) str)) {
                ret = 0;
        } else if (strlen(str) > 2 && str[0] == '*' && str[1] == '.' && strlen(str) - 2 <= strlen(hostname) && strcasecmp(hostname+strlen(hostname)-strlen(str)+2, str+2) == 0) {
                ast_log(LOG_WARNING,"Warning: allowing match on wildcard (%s =~ %s)\n", hostname, str);
                ret = 0;
        } else {
                ret = -1;
        }

        ast_debug(3, "SSL %s compare s1='%s' s2='%s'\n", desc, hostname, str);
        OPENSSL_free(str);

        return ret;
}
```

jcolp:
Inline patches are not accepted. All patches need to go through a PR request with a contributor license agreement signed.  If you want to maintain such a patch yourself it would need to be done in your own private fork.

Patches relating to chan_sip are also not accepted any longer. If this patch is generally applicable, then it would be evaluated as such.

jcolp:
@dirkx Do you plan on going through the code contribution process for this?

dirkx:
Absolutely happy to do so & then do a pull request. 

But also understood that ``_Patches relating to chan_sip are also not accepted any longer'_'. And I have no idea how generic tcptls.c is in the code base*.

So please advice.

*) We are highly reliant on cisco-kit and the wonderful https://usecallmanager.nz/  support -- so pjsip is not something I can test here.

PR #276, opened 2023-08-26
TLS in asterisk currently does not accept wildcarded DN's when connecting

dirkx:
TLS  currently does not accept wildcarded DN's when connecting to a TLS/SSL host. This tweak makes that work. Fixes/addresses issue  #269

### DIFF
--- a/main/tcptls.c
+++ b/main/tcptls.c
@@ -85,21 +85,21 @@ static void session_instance_destructor(void *obj)
 static int check_tcptls_cert_name(ASN1_STRING *cert_str, const char *hostname, const char *desc)
 {
 	unsigned char *str;
-	int ret;
+	int ret = -1,len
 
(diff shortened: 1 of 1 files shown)


InterLinked1 (changes requested):
chan_sip no longer accepts changes here. Is this applicable to anything else that would merit its inclusion? If so, the commit message should be framed as such.

This PR currently has 2 PRs that need to be squashed together.

There are some coding guideline violations that need to be addressed as well.

The existing commit message does not comply with the commit message guidelines.

seanbright (changes requested):
This code will allow `*.example.com` to match `any.number.of.subdomain.labels.example.com` which is incorrect.

asteriskteam:

asteriskteam:

PR #279, opened 2023-08-28
res_pjsip_session: Send Session Interval too small response

btriller:
Handle session interval lower than endpoint's configured minimum timer when sending first answer. Timer setting is checked during this step and needs to handled appropriately. Otherwise no response will be sent at all.

### DIFF
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -4053,15 +4053,20 @@ static int new_invite(struct new_invite *invite)
 	 * so let's go ahead and send a 100 Trying out to stop any
 	 * retransmissions.
 	 */
+	if (pjsip_inv_initial_answer(invite->session->inv_session, invite->rdata, 100, NULL, NULL, &tdata) != PJ_SUCCESS) {
+		if (tdata) {
+			pjsip_inv_send_msg(invite->session->inv_session, tdata);
+		} else {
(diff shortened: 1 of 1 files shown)


btriller:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

res/res_pjsip_session.c
```diff
@@ -4053,15 +4053,21 @@ static int new_invite(struct new_invite *invite)
 	 * so let's go ahead and send a 100 Trying out to stop any
 	 * retransmissions.
 	 */
+	if (pjsip_inv_initial_answer(invite->session->inv_session, invite->rdata, 100, NULL, NULL, &tdata) != PJ_SUCCESS) {
+		if (tdata == NULL) {
```
jcolp, line 4057: This needs to be updated according to [coding guidelines](https://docs.asterisk.org/Development/Policies-and-Procedures/Historical-Policies-and-Procedures/Code-Review/Coding-Guidelines/).

if (!tdata) {
} else {
}

jcolp (changes requested):
I think the commit message should be updated to include what previously happened, and what now happens, from a SIP response perspective in this scenario.

res/res_pjsip_session.c
jcolp: if (!tdata) {

jcolp (approved):

Issue #275, opened 2023-08-26
[bug]:Asterisk make now requires ASTCFLAGS='-std=gnu99 -Wdeclaration-after-statement'


AstersikRoss:
### Severity

Trivial

### Versions

18, 20, 21, master

### Components/Modules

app_voicemail

### Operating Environment

CentOS

### Frequency of Occurrence

Constant

### Issue Description

When compiling asterisk 18 from GIT the following errors are now encountered due to the compile flag 'c99' not being set.

This appears to be due to the fix for issue #170

### Relevant log output

```shell
app_voicemail.c: In function ‘show_mailbox_snapshot’:
app_voicemail.c:11594:2: error: ‘for’ loop initial declarations are only allowed in C99 mode
  for (int i = 0; i < mailbox_snapshot->folders; i++) {
  ^
app_voicemail.c:11594:2: note: use option -std=c99 or -std=gnu99 to compile your code
app_voicemail.c: In function ‘complete_voicemail_move_message’:
app_voicemail.c:11813:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
   for (int i = 0; i < ARRAY_LEN(mailbox_folders); i++) {
   ^
app_voicemail.c: In function ‘append_vmbox_info_astman’:
app_voicemail.c:13818:2: error: ‘for’ loop initial declarations are only allowed in C99 mode
  for (int i = 0; i < mailbox_snapshot->folders; i++) {
  ^
app_voicemail.c: At top level:
cc1: warning: unrecognized command line option "-Wno-format-truncation" [enabled by default]
make[1]: *** [app_voicemail.o] Error 1
make: *** [apps] Error 2
   [CCi] app_voicemail.i -> app_voicemail.o
app_voicemail.c: In function ‘show_mailbox_snapshot’:
app_voicemail.c:11594:2: error: ‘for’ loop initial declarations are only allowed in C99 mode
  for (int i = 0; i < mailbox_snapshot->folders; i++) {
  ^
app_voicemail.c:11594:2: note: use option -std=c99 or -std=gnu99 to compile your code
app_voicemail.c: In function ‘complete_voicemail_move_message’:
app_voicemail.c:11813:3: error: ‘for’ loop initial declarations are only allowed in C99 mode
   for (int i = 0; i < ARRAY_LEN(mailbox_folders); i++) {
   ^
app_voicemail.c: In function ‘append_vmbox_info_astman’:
[...]

PR #283, opened 2023-08-29
app_voicemail: Fix for loop declarations

mbradeen:
Resolve for loop initial declarations added in cli changes.

Resolves: #275

### DIFF
--- a/apps/app_voicemail.c
+++ b/apps/app_voicemail.c
@@ -11550,6 +11550,7 @@ static int show_mailbox_snapshot(struct ast_cli_args *a)
 	const char *context = a->argv[4];
 	struct ast_vm_mailbox_snapshot *mailbox_snapshot;
 	struct ast_vm_msg_snapshot *msg;
+	int i;
 
 	/* Take a snapshot of the mailbox and walk through each folder's contents */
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

InterLinked1:
I'm kind of curious, shouldn't 48c6e3fb1da06b8e76a73a4d9099f33664b4b372 have initially caused the build tests to fail? I thought they were built in developer mode, which would catch these types of things and make the warnings errors? Did the CI change when it moved to GitHub?

jcolp:
Even in developer mode it depends on the compiler version. For example, it's not a problem on my system.

gtjoseph:
> I'm kind of curious, shouldn't [48c6e3f](https://github.com/asterisk/asterisk/commit/48c6e3fb1da06b8e76a73a4d9099f33664b4b372) have initially caused the build tests to fail? I thought they were built in developer mode, which would catch these types of things and make the warnings errors? Did the CI change when it moved to GitHub?

I'm going to take a look at the test build options to see if we can catch these.

PR #284, opened 2023-08-29
res_pjsip_diversion: Add debugging to aid testsuite failures

gtjoseph:
Added many ast_debug statements to the Diversion path.  Not
so much to the HistoryInfo path.

Also added two new macros in res_pjsip.h to assist printing
URI user and host without having to allocate memory from the
pool and doing a "sprintf" on the uri to it.


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

gtjoseph:
converted to draft until we know what's going on with #288

PR #285, opened 2023-08-30
core,app_dial,pjsip: Implement Advanced Codec Negotiation (ACN)

### LINKED ISSUE #223: [new-feature]: Advanced Codec Negotiation
### Feature Description

This feature has long been talked about and a lot of excellent work to prepare this feature has been done by @gtjoseph (e.g. 8d1064eaafe899044adbd30a40cdaa15c315c463, 647c53c41fa36ab8e5b3e8824952aea01c76e37e, 9bd1d686a120db5233acb2d6f90c31d4f52464a0).

The pjsip configuration options for endpoints exist, what's left is implementing the functionality.

References:
- https://www.asterisk.org/asterisk-acn-advanced-codec-negotiation/
- https://docs.asterisk.org/Configuration/Channel-Drivers/SIP/Configuring-res_pjsip/PJSIP-Advanced-Codec-Negotiation/
- https://docs.asterisk.org/Development/Roadmap/Asterisk-18-Projects/AstriDevCon-2020-Advanced-Codec-Negotiation.pdf
- https://community.asterisk.org/t/advanced-codec-negotiation/92468

I have a working patch + testsuite test cases. Feel free to assign this issue to me.

maximilianfridrich:
Advanced Codec Negotiation is now implemented for the dial application, pjsip channels and local channels.

Resolves: #223

UpgradeNote: The two pjsip.conf options "incoming_call_offer_pref" and "outgoing_call_offer_pref" have been removed. Instead, the new options "codec_prefs_incoming_answer", "codec_prefs_incoming_offer", "codec_prefs_outgoing_answer", "codec_prefs_outgoing_offer" must be used. Even if the old settings were not used, this could break existing deployments as the default ACN options are now in use and they might behave differently in some call flows (e.g. by default, pending codecs are always preferred, not the endpoint configuration).

UserNote: The Advanced Codec Negotiation feature is now implemented.


maximilianfridrich:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

jcolp:
I haven't looked at the code, just the upgrade note, but the options can not be removed. They must continue to exist.

maximilianfridrich:
This PR must be tested with the corresponding testsuite PR which adds 48 tests to test the codec negotiation:
- https://github.com/asterisk/testsuite/pull/20

Further, 10 tests were adopted to reflect the new behavior. For those tests, `pjsip.conf` had to be adopted to use the new ACN settings. The default settings prefer the pending stream while these tests expect the configured settings to be preferred.

maximilianfridrich:
> the options can not be removed. They must continue to exist.

@jcolp And they must continue to work? Or silently be ignored like the ACN settings were? I'm not sure if I see a way for both settings to co-exist without interfering with each other.

jcolp:
Yes. We can not remove options in minor versions, or alter their behavior, unless there is a critical reason why - such as a security issue. This does not rise to that level.

jcolp:
For example, if the options are kept but default to empty in the code we can determine that they are actually unset. Default behavior should remain the same in that case even if the options were removed. If they are set then those settings take priority. If unset then the other options take priority.

maximilianfridrich:
> We can not remove options in minor versions, or alter their behavior

Makes sense, I probably should have checked with the maintainers first. So would it be an option for Asterisk 21?

jcolp:
Asterisk 21, once branched, can no longer receive breaking changes.

maximilianfridrich:
Ok, then I will look into your suggestion and try to find a way to not change the existing behavior and not remove any settings.

Is it possible to mark this PR as a draft while I do that?

jcolp:
I've converted it to a draft.

maximilianfridrich:
> 1. Default behavior with this change should match previous versions
> 2. If the incoming_* and outgoing_* options are set by the user, then they take priority

Consider the case where neither the "old" options `incoming_call_offer_pref` and `outgoing_call_offer_pref` nor the "new" `codec_prefs_*` options are set. As you described it, since the old options are unset, the new settings are used. And in order to not change the default behavior, I **must** change the defaults of two of the new ACN options:

```
codec_prefs_incoming_offer = prefer: configured, operation: intersect, keep: all, transcode: allow
codec_prefs_incoming_answer = prefer: configured, operation: intersect, keep: all, transcode: allow
```

They must be set to `prefer: configured`, this is the only way the requirements can be met. So far, the documentation (and code) had `prefer: pending` as default. The ACN settings had no effect so far, so I think this should be alright.

gtjoseph:
Thanks for picking this up!!!
I'm going to try and set some time aside to review in detail.

maximilianfridrich:
@gtjoseph Thank you! I am now incorporating Joshua's remarks and I noticed that some fax/pjsip tests failed (which I didn't check). So this is still a WIP, but we have confirmed that the ACN is working as expected (e.g. see the linked testsuite PR and we have done manual testing as well).

maximilianfridrich:
> If the incoming_* and outgoing_* options are set by the user, then they take priority.

To be more precise, I think the logic should be:

As soon as one of `incoming_call_offer_pref` or `outgoing_call_offer_pref` are set, the ACN options are ignored.

I think everything else would lead to maximum confusion. I will state this in the docs for the settings.

InterLinked1:
> @jcolp Is the removal of the `incoming_call_offer_pref` and `outgoing_call_offer_pref` an option for the next Asterisk major version release? If so, how can we go about this? Should we close this PR and re-open it at a better time?

There is a waiting-for-standard-release-development-cycle label for this purpose, that could be applied. But I'm not sure contributors can add it themselves.

jcolp:
I've added it.

btriller:
This leaks some memory. Following patch fixes it. Unsure, if this is the correct approach though. @maximilianfridrich 

```
diff --git a/channels/chan_pjsip.c b/channels/chan_pjsip.c
index 66ea75d0e0..e1d003180f 100644
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -2917,6 +2917,10 @@ static int request(void *obj)
 		SCOPE_EXIT_RTN_VALUE(-1, "Couldn't create session\n");
 	}
 
+	if (resolved_top) {
+		ast_stream_topology_free(resolved_top);
+	}
+
 	req_data->session = session;
 
 	SCOPE_EXIT_RTN_VALUE(0);
diff --git a/res/res_pjsip_session.c b/res/res_pjsip_session.c
index 282878adb8..0da14eca13 100644
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -941,6 +941,7 @@ static int handle_incoming_sdp(struct ast_sip_session *session, const pjmedia_sd
 					ast_trace(-1, "%s: No common codecs between incoming SDP offer and endpoint configuration.\n", ast_sip_session_get_name(session));
 				} else {
 					stream = ast_stream_clone(resolved_stream, stream_name);
+					ast_stream_free(resolved_stream);
 				}
 			} else {
 				/* This can happen when the stream is marked as removed.
```

maximilianfridrich:
Thank you for looking at it @btriller. That seems right. I declared both variables with the `RAII_VAR` macro now, that seems to be the best way. Not sure how this one slipped through and wasn't caught by our automated tests (which we ran with valgrind). I have updated the patch.

Also, please be aware that this patch as it is now breaks some of the fax/pjsip tests.

btriller:
There's another issue, that leads to no audio, if a client sends more streams than endpoint's capability. See #344 @maximilianfridrich

maximilianfridrich:
I'm not sure if I can follow. Could you please describe the scenario in more detail (endpoint config, and who answers/offers what)?

res/res_pjsip_session.c
```diff
@@ -5401,6 +5611,8 @@ static void session_inv_on_create_offer(pjsip_inv_session *inv, pjmedia_sdp_sess
 		SCOPE_EXIT_RTN("%s: create offer failed\n", ast_sip_session_get_name(session));
 	}
 
+	ast_sip_session_media_state_reset(session->pending_media_state);
```
btriller, line 5614: This seems to cause no audio with `all_codecs_on_empty_reinvite=yes` when a UA renegotiates a different codec. See [1] for reproducing this issue. Seems to be fixed with
```
diff --git a/res/res_pjsip_session.c b/res/res_pjsip_session.c
index 5486a97040..72b4130a40 100644
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -5594,25 +5594,23 @@ static void session_inv_on_create_offer(pjsip_inv_session *inv, pjmedia_sdp_sess
 		}
 	}
 
-	if (inv->neg) {
-		if (pjmedia_sdp_neg_was_answer_remote(inv->neg)) {
-			pjmedia_sdp_neg_get_active_remote(inv->neg, &previous_sdp);
-		} else {
-			pjmedia_sdp_neg_get_active_local(inv->neg, &previous_sdp);
-		}
-	}
-
 	if (ignore_active_stream_topology) {
 		offer = create_local_sdp(inv, session, NULL, 1);
 	} else {
+		if (inv->neg) {
+			if (pjmedia_sdp_neg_was_answer_remote(inv->neg)) {
+				pjmedia_sdp_neg_get_active_remote(inv->neg, &previous_sdp);
+			} else {
+				pjmedia_sdp_neg_get_active_local(inv->neg, &previous_sdp);
+			}
+		}
 		offer = create_local_sdp(inv, session, previous_sdp, 0);
+		ast_sip_session_media_state_reset(session->pending_media_state);
 	}
 	if (!offer) {
 		SCOPE_EXIT_RTN("%s: create offer failed\n", ast_sip_session_get_name(session));
 	}
 
-	ast_sip_session_media_state_reset(session->pending_media_state);
-
 	ast_queue_unhold(session->channel);
 
 	/*
```

[1] [acn_reinvite_without_sdp.patch](https://github.com/asterisk/asterisk/files/13270602/acn_reinvite_without_sdp.patch)

btriller:
There's another memory leak. Please check following patch. @maximilianfridrich 

```
diff --git a/res/res_pjsip_session.c b/res/res_pjsip_session.c
index 5486a97040..cbbf6d05c4 100644
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -1225,8 +1225,8 @@ static int handle_negotiated_sdp(struct ast_sip_session *session, const pjmedia_
 		for (index = 0; index < local->media_count; index++) {
 			RAII_VAR(struct ast_format_cap *, remote_caps, NULL, ao2_cleanup);
 			RAII_VAR(struct ast_stream_topology *, remote_top, NULL, ao2_cleanup);
+			RAII_VAR(struct ast_stream *, resolved_stream, NULL, ast_stream_free);
 			struct ast_stream *remote_ast_stream;
-			struct ast_stream *resolved_stream;
 			struct ast_stream *configured_stream = ast_stream_topology_get_stream(session->pending_media_state->topology, index);
 			int skip_stream_setting = 0;
 
```

btriller:
> There's another memory leak. Please check following patch. @maximilianfridrich
> 
> ```
> diff --git a/res/res_pjsip_session.c b/res/res_pjsip_session.c
> index 5486a97040..cbbf6d05c4 100644
> --- a/res/res_pjsip_session.c
> +++ b/res/res_pjsip_session.c
> @@ -1225,8 +1225,8 @@ static int handle_negotiated_sdp(struct ast_sip_session *session, const pjmedia_
>  		for (index = 0; index < local->media_count; index++) {
>  			RAII_VAR(struct ast_format_cap *, remote_caps, NULL, ao2_cleanup);
>  			RAII_VAR(struct ast_stream_topology *, remote_top, NULL, ao2_cleanup);
> +			RAII_VAR(struct ast_stream *, resolved_stream, NULL, ast_stream_free);
>  			struct ast_stream *remote_ast_stream;
> -			struct ast_stream *resolved_stream;
>  			struct ast_stream *configured_stream = ast_stream_topology_get_stream(session->pending_media_state->topology, index);
>  			int skip_stream_setting = 0;
>  
> ```

Former patch leads to crashes. Following seams more appropriate
```
diff --git a/res/res_pjsip_session.c b/res/res_pjsip_session.c
index 5486a97040..9bfe47df30 100644
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -1259,6 +1259,7 @@ static int handle_negotiated_sdp(struct ast_sip_session *session, const pjmedia_
 				&session->endpoint->media.codec_prefs_incoming_answer,
 				NULL);
 			if (!resolved_stream || ast_format_cap_empty(ast_stream_get_formats(resolved_stream))) {
+				ast_stream_free(resolved_stream); /* Can handle NULL */
 				if (session->endpoint->media.codec_prefs_incoming_answer.transcode == CODEC_NEGOTIATION_TRANSCODE_ALLOW) {
 					ast_log(LOG_DEBUG, "%s: No common codecs between channels and transcoding allowed.\n",
 						ast_sip_session_get_name(session));
```

maximilianfridrich:
> Former patch leads to crashes. Following seams more appropriate

That looks good, thank you for the find!

res/res_pjsip_session.c
maximilianfridrich: I don't doubt it, but the provided testcase passes for me even without your suggested patch. Is it testing exactly the scenario you described?

btriller: Negotiating codecs works fine without the patch but there's no audio (no RTP sent) after negotiating a new codec. I don't know how to verify that RTP is sent (maybe it's possible to check for RTP AMI events?). IIRC without the patch there's following log message
```
res_pjsip_sdp_rtp.c:487 set_caps: No joint capabilities for 'audio' media stream between our configuration((alaw)) and incoming SDP((g722))
```

maximilianfridrich: Makes sense, maybe I can find a way to test it. Thank you for clarifying.

SirDMZ:
Hey Maximilianfridrich,

Have you got a version of this patch for version 20 available?

maximilianfridrich:
@SirDMZ I tested the patches with Asterisk 20, they should work. A recent PR (by me) has now created a conflict but it should be very easy to resolve.

khoegh:
I have testet a few calls and it looks good.
If asterisk receives a call with no matching codec it crashes, instead of sending "188 Not Acceptable Here".
I have allow = !all,g722,alaw
If I send an INVITE containing g.729 and GSM asterisk dies.

Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000506e69 in ast_format_cap_append_from_cap (dst=0x7f8f5c00aac8, src=0x7f8f5c013ae0, type=AST_MEDIA_TYPE_AUDIO) at format_cap.c:276
276                     struct format_cap_framed *framed = AST_VECTOR_GET(&src->preference_order, idx);

or

Program terminated with signal 11, Segmentation fault.
#0  0x000000000055c025 in ast_rtp_codecs_payloads_destroy (codecs=0x1) at rtp_engine.c:1000
1000            for (idx = 0; idx < AST_VECTOR_SIZE(&codecs->payload_mapping_rx); ++idx) {

or

Program terminated with signal 11, Segmentation fault.
#0  0x000000000050256f in ast_format_cap_count (cap=0x0) at format_cap.c:397
397             return AST_VECTOR_SIZE(&cap->preference_order);

maximilianfridrich:
@khoegh Thank you for looking into it! Unfortunately, I currently don't really have time to look into this, but feel free to share your progress if you find the bug. If I continue working on this, I will post it here.

khoegh:
The following solves my crashes, by sending "488 Not Acceptable Here"

```
--- a/res/res_pjsip_session.c   2024-09-09 10:27:44.197588384 +0200
+++ b/res/res_pjsip_session.c   2024-09-20 10:17:05.487365310 +0200
@@ -938,7 +938,8 @@
                                if (!resolved_stream || ast_format_cap_empty(ast_stream_get_formats(resolved_stream))) {
                                        ast_stream_free(resolved_stream); /* Can handle NULL */
                                        stream = ast_stream_alloc(stream_name, type);
-                                       ast_trace(-1, "%s: No common codecs between incoming SDP offer and endpoint configuration.\n", ast_sip_session_get_name(session));
+                                       handled = 0;
+                                       SCOPE_EXIT_LOG_EXPR(goto end, LOG_ERROR, "No common codecs between incoming SDP offer and endpoint configuration.\n");
                                } else {
                                        stream = ast_stream_clone(resolved_stream, stream_name);
                                }
```

maximilianfridrich:
Thank you! @khoegh
I will hopefully find time to look into it soon. I want to make sure it doesn't break any other scenarios.

res/res_pjsip_session.c
```diff
@@ -1115,6 +1230,70 @@ static int handle_negotiated_sdp(struct ast_sip_session *session, const pjmedia_
...
+				ast_log(LOG_TRACE, "%s: Replacing pending stream: %s with resolved stream: %s\n", ast_sip_session_get_name(session),
+					ast_str_tmp(256, ast_stream_to_str(configured_stream, &STR_TMP)),
+					ast_str_tmp(256, ast_stream_to_str(resolved_stream, &STR_TMP)));
+				if (ast_stream_topology_set_stream(session->pending_media_state->topology, index, resolved_stream)) {
+					ast_stream_free(resolved_stream);
+					ast_log(LOG_WARNING, "%s: Couldn't set stream in topology.\n",
+						ast_sip_session_get_name(session));
+					return -1;
+				}
+			}
```
ipoddubny, line 1293: resolved_stream is never freed when skip_stream_setting is true or when resolved_stream is identical to configured_stream, there should be an else branch with ``ast_stream_free(resolved_stream);`` here.

channels/chan_pjsip.c
```diff
@@ -727,6 +747,91 @@ static int answer(void *data)
...
+				ast_log(LOG_WARNING, "Failed to create resolved topology. Still answering call.\n");
+				goto end;
+			}
+			if (ast_sip_push_task(channel->session->serializer, regenerate_sdp, regen_data)) {
+				ast_log(LOG_WARNING, "Error regenerating SDP %s on channel %s\n", regen_data->is_offer ? "offer" : "answer", ast_channel_name(ast));
+				ao2_cleanup(regen_data);
+				goto end;
+			}
+		} else {
+			ast_log(LOG_NOTICE, "No resolved topology or pending_media_state topology. Still answering call.\n");
```
ipoddubny, line 825: ``resolved_top`` is created but never freed in this case

ipoddubny:
Great work! It would be really awesome to see this PR merged one day.
I found a couple of leaks I found during testing (reported above), and also had a couple of crashes in indicate>session_inv_on_media_update (in different places), not sure if it's a use-after-free or a race condition, core dumps indicate that there's memory corruption, I'm trying to figure it out how it happened.

ipoddubny:
The crashes I had were caused by this code (in ``chan_pjsip_indicate`` and ``chan_pjsip_answer_with_stream_topology``):
```c
ast_stream_topology_free(channel->session->pending_media_state->topology);
channel->session->pending_media_state->topology = ast_stream_topology_clone(resolved_top);
```
``pending_media_state->topology`` is destroyed (and replaced) in the PBX thread, while `indicate` task is processed at the same time by a PJSIP thread, and it uses ``pending_media_state`` (in ``handle_negotiated_sdp``).

btriller:
> The crashes I had were caused by this code (in `chan_pjsip_indicate` and `chan_pjsip_answer_with_stream_topology`):
> 
> ```c
> ast_stream_topology_free(channel->session->pending_media_state->topology);
> channel->session->pending_media_state->topology = ast_stream_topology_clone(resolved_top);
> ```
> 
> `pending_media_state->topology` is destroyed (and replaced) in the PBX thread, while `indicate` task is processed at the same time by a PJSIP thread, and it uses `pending_media_state` (in `handle_negotiated_sdp`).

Seeing similar crashes.

ipoddubny:
> @ipoddubny have you tested offers w/ audio and video stream whereas matching endpoint does not allow video? w/o #344 there's no RTP as rtp instance shuts down. With patch suggested from [#285 (comment)](https://github.com/asterisk/asterisk/pull/285#issuecomment-2363277070) Asterisk responds w/ 488 as it should just reject video stream.

No, I haven't tried it with video streams.

res/res_pjsip_session.c
maximilianfridrich: Thank you! I have fixed the leak and refactored the logic to make it simpler. Please mark as resolved if it looks good to you.

channels/chan_pjsip.c
maximilianfridrich: Thank you! I have fixed the leak and refactored the logic to make it simpler. Please mark as resolved if it looks good to you.

maximilianfridrich:
> It seems that moving topology assignments into regenerate_sdp and running it synchronously (ast_sip_push_task_wait_servant) fixes the crashes.

@ipoddubny I have [implemented and pushed your suggested change](https://github.com/asterisk/asterisk/compare/e6fd114292a657dc32854d40fbef2d5e6c4344cc..6fd00e77fddbb5ac219caea4fb413193908ede4c) to this PR and it looks good! Thank you for finding this bug and finding a fix too! I would appreciate if you could report back if this fix works in your setup too.

maximilianfridrich:
Thank you! I will rely on your written comments regarding resolving comments and I will mark them as resolved.

Thanks again for the review, I have fixed your mentioned leak in `chan_pjsip_indicate`. If you have a SIPP scenario that you would like to share which triggered the crashes before, that could be helpful.

ipoddubny:
It doesn't crash and leak memory in my simple setup anymore.

eallovon:
Hi

Great work ! It would be really great to have ACN in asterisk.

I was able to test this PR against an asterisk 20 and it worked well for my use case [1].
I'm wondering what would be the next step for this PR.

AFAIU - since it introduces breaking change in the configuration - it seems it has no chance to be merged in any current asterisk version (even for asterisk 22 it seems too late ?).
Fair enough. But on the other hand I'm sad since :
- it seems to work pretty well on asterisk 20 (and probably below).
- and also it seems there is a way to make it backward compatible (as discussed [here](https://github.com/asterisk/asterisk/pull/285#issuecomment-1698949588) and [here](https://github.com/asterisk/asterisk/pull/285#issuecomment-1699055442))

Also there are the T.38 tests failing. I guess it would be a show stopper for this to be merged too.
BTW  @maximilianfridrich  did you have a look at what is failing for T.38 ? Or someone else ?

Anyway I understood that @maximilianfridrich  you're not going to invest more in this topic anytime soon.
So it seems this PR is a bit stuck.

So here's my questions : what are our options there to make this PR move forward ?
We may try to help on proposing a solution for the backward compatibility. Is this something interesting ?
Though for the T.38 I'm not sure we can help.


[1] What I tested with ACN is the following:
Given endpoints
```
[alice]
allow=!all,alaw,g729
incoming_call_offer_pref=remote
outgoing_call_offer_pref=remote_merge
[bob]
allow=!all,alaw,g729
incoming_call_offer_pref=remote
outgoing_call_offer_pref=remote_merge
```

Given alice calls bob offering g729
Then asterisk does call bob offering g729
Then bob answers only with g729
Then asterisk answers alice with g729 only
But then without clear cause, the alaw (probably from the endpoint configuration) comes back and asterisk stops the call because it can't transcode g729 to alaw (log is `channel.c:5796 set_format: Unable to find a codec translation path: (g729) -> (alaw)`).
This definitely seems to be because the `{incoming,outgoing}_call_answer_pref` options were not implemented.

By using the patch of this PR and using the new ACN option:
```
codec_prefs_incoming_answer= prefer: pending, operation: only_preferred, keep: first, transcode: allow
codec_prefs_outgoing_answer=prefer: pending, operation: only_preferred, keep: first, transcode: allow
codec_prefs_incoming_offer=prefer: pending, operation: intersect, keep: all, transcode: allow
codec_prefs_outgoing_offer=prefer: pending, operation: intersect, keep: all, transcode: allow
```

It works fine.

maximilianfridrich:
Thanks for the comment and for offering help! @eallovon

You're right, this PR is a bit stuck, mostly because it is not really a priority on our side anymore, so I don't get to invest time in this currently.

To move this PR forward, I would say the following must be done in this order (can be done by anyone):

1. Fix failing PJSIP + T.38 tests. Some might simply be failing because some of the new config options need different defaults for backward compatibility reasons - see [this comment](https://github.com/asterisk/asterisk/pull/285#issuecomment-1698949588), and some might need changes to the [Testsuite PR](https://github.com/asterisk/testsuite/pull/20).
2. Together with maintainers, figure out how to resolve backwards compatibility issues.
   E.g. If this PR goes into Asterisk 23/24, can we then drop the old config options if the defaults remain the same? If not, then we have to figure out how both sets of options can coexist (and then this PR can always be merged into the release branches). Since it's not urgent for us, removing the old options would be our preference, since it's simpler for us. A merge into master would have to be done [before mid August](https://docs.asterisk.org/About-the-Project/Asterisk-Versions/) for it to be in Asterisk 23.

If you can help with any of the above, that would be greatly appreciated. And we can also talk about how to correctly attribute the work.

PS:

> even for asterisk 22 it seems too late ?

With the breaking changes that we have currently, yes. If we can make this PR backwards compatible and make it coexist with the existing options, then it can be merged anytime. If the maintainers allow the removal of the old options for Asterisk 23/24, then it would have to be merged into master by mid August. See the bottom of the [Asterisk Versions](https://docs.asterisk.org/About-the-Project/Asterisk-Versions/) page.

> did you have a look at what is failing for T.38

Not in detail, no.

> But then without clear cause, the alaw (probably from the endpoint configuration) comes back and asterisk stops the call because it can't transcode g729 to alaw

I'm not sure I fully understood the scenario, but it seems like this scenario should work even without ACN. The way you described that sounds more like a bug to me, but I'm not sure.

eallovon:
Hi,

Thanks a lot for your detailed answer.

> > But then without clear cause, the alaw (probably from the endpoint configuration) comes back and asterisk stops the call because it can't transcode g729 to alaw  
  
> I'm not sure I fully understood the scenario, but it seems like this scenario should work even without ACN. The way you described it that sounds more like a bug to me, but I'm not sure.  

In the end I focused mainly on your comment here trying to figure out if it could be a bug.

## tl;dr
  it may be a bug, but it never really worked. Though, in the process of debugging it I found that setting `asymmetric_rtp_codec=true` fixed my issue. To be honest, therefore, our interest in this PR decreased quite a lot, and I'm not sure we would be able to help more then. I'm sorry for that.  

## FTR, the longer explanation

Given the following configuration
```
[test_codec](!)
allow = !all,alaw,g729
direct_media=no
incoming_call_offer_pref=remote
outgoing_call_offer_pref=remote_merge

[alice](test_codec)
type = endpoint
auth=alice
aors=alice

[bob](test_codec)
type = endpoint
auth=bob
aors=bob
```

As I said in my previous message, with current version of asterisk (20.11 for example), if caller sends an offer with g729>g722>alaw, asterisk correctly sends an offer to callee with g729>alaw (respecting caller order, filtered from codec unknown in caller/callee configuration).
And when callee answers with g729, asterisk correctly answers with g729 too.  
	    
But then the communication fails to establish with the message  
```
WARNING[70]: channel.c:5765 set_format: Unable to find a codec translation path: (alaw) -> (g729)
WARNING[70]: channel.c:5765 set_format: Unable to find a codec translation path: (g729) -> (alaw)
```

What I found is that the option `outgoing_call_offer_pref=remote_merge` never worked till asterisk 18.10.0 which has fix [res_pjsip_sdp_rtp: Preserve order of RTP codecs](https://github.com/asterisk/asterisk/commit/4e204db2bfdc9b58ac881fe16e27f900662f94e1)
Before asterisk 18.10.0, with the above setting, asterisk would offer alaw > g729 to the callee (so what was configured instead of something related to what was offered by the remote).
Since asterisk 18.10 it behaves the same with the configuration I provided : that is, the offer from asterisk correctly follows what is set in the `{incoming,outgoing}_call_offer_pref` options, but then the call fails to establish when there is no translation path.

What I discovered when adding some log in the code `set_caps` function of `res/res_pjsip_sdp_rtp.c` tracing the different codec capability computed, is that in fact the code was going inside the `!asymmetric_rtp_codec` part of the if-else which introduces a "best" capabilities (this option was introduced way before by [pjsip: Extend 'asymmetric_rtp_codec' option to include us changing.](https://github.com/asterisk/asterisk/commit/d3e951edf5517b9f508a7e1b474176ec2be9e18f). This is this "best" cap which was introducing back the alaw from 'nowhere'.

Therefore in our case, adding `asymmetric_rtp_codec=yes` in our configuration solved our issue: we still have the correct offer from asterisk and the call this time establishes without. Asterisk no longer tries to translate between alaw and g729 a call that was negotiated in g729 both sides.

As I said, I'm afraid that if we don't find another culprit our interest in this PR will fade.

khoegh:
I had a crash today.
I have lots of calls between the same endpoints which look similar without crashes.
Unfortunately I'm unable to fix the code.
Hope the following can help anyone interested.

#0  ast_stream_topology_get_first_stream_by_type (type=AST_MEDIA_TYPE_AUDIO, topology=0x7f9038a9d5a8) at stream.c:976
976                     if (stream->type == type
[Current thread is 1 (Thread 0x7f8fee524700 (LWP 3461580))]
(gdb) p stream
$1 = (struct ast_stream *) 0x7f9785dda96a
(gdb) p stream->type
Cannot access memory at address 0x7f9785dda96a

(gdb) bt
#0  ast_stream_topology_get_first_stream_by_type (type=AST_MEDIA_TYPE_AUDIO, topology=0x7f9038a9d5a8) at stream.c:976
#1  ast_stream_topology_create_resolved (pending_topology=pending_topology@entry=0x7f907cc439f8, configured_topology=0x7f9038a9d5a8, prefs=prefs@entry=0x157b30c, error_message=error_message@entry=0x7f8fee51eca0) at stream.c:1046
#2  0x00007f90b0298c69 in chan_pjsip_indicate (ast=0x7f903a1ee050, condition=14, data=0x7f907cc439f8, datalen=0) at chan_pjsip.c:1836
#3  0x00007f9104a6d138 in wait_for_answer (in=in@entry=0x7f903a1ee050, out_chans=out_chans@entry=0x7f8fee51fde0, to_answer=to_answer@entry=0x7f8fee51fda0, to_progress=to_progress@entry=0x7f8fee51fda4, 
    peerflags=peerflags@entry=0x7f8fee520938, opt_args=opt_args@entry=0x7f8fee520080, pa=0x7f8fee5202e0, num_in=0x7f8fee51fe00, result=0x7f8fee51fdac, dtmf_progress=<optimized out>, mf_progress=<optimized out>, 
    mf_wink=<optimized out>, sf_progress=<optimized out>, sf_wink=<optimized out>, hearpulsing=0, ignore_cc=1, forced_clid=0x7f8fee51fe50, stored_clid=0x7f8fee51fea0, config=<optimized out>) at app_dial.c:1611
#4  0x00007f9104a71b64 in dial_exec_full (chan=0x7f903a1ee050, data=<optimized out>, peerflags=peerflags@entry=0x7f8fee520938, continue_exec=continue_exec@entry=0x0) at app_dial.c:3052
#5  0x00007f9104a74179 in dial_exec (chan=<optimized out>, data=<optimized out>) at app_dial.c:3596
#6  0x000000000052d2c6 in pbx_exec (c=c@entry=0x7f903a1ee050, app=app@entry=0x121cde0, data=data@entry=0x7f8fee521a80 "PJSIP/23456789@sip-peer,180,iS(43200)b(Dial_Handler^23456789^1)")
    at /usr/local/src/asterisk/asterisk-18.25.0+acn/include/asterisk/strings.h:67
#7  0x0000000000522202 in pbx_extension_helper (c=c@entry=0x7f903a1ee050, context=0x7f903a1eea90 "Trunk", exten=exten@entry=0x7f903a1eeae0 "23456789", priority=priority@entry=11, label=label@entry=0x0, 
    callerid=callerid@entry=0x7f8fd41e4350 "98765432", action=E_SPAWN, found=0x7f8fee523b04, combined_find_spawn=1, con=0x0) at pbx.c:2948
#8  0x0000000000523ccb in ast_spawn_extension (combined_find_spawn=1, found=0x7f8fee523b04, callerid=0x7f8fd41e4350 "98765432", priority=11, exten=0x7f903a1eeae0 "23456789", context=<optimized out>, c=0x7f903a1ee050) at pbx.c:4203
#9  __ast_pbx_run (c=c@entry=0x7f903a1ee050, args=args@entry=0x0) at pbx.c:4377
#10 0x00000000005252bb in pbx_thread (data=data@entry=0x7f903a1ee050) at pbx.c:4701
#11 0x00000000005a6e29 in dummy_start (data=<optimized out>) at utils.c:1607
#12 0x00007f91097921ca in start_thread () from /usr/lib64/libpthread.so.0
#13 0x00007f91071568d3 in clone () from /usr/lib64/libc.so.6

maximilianfridrich:
Thanks for the report @khoegh! The hint about the 183 and 180 with SDP is very useful and likely has to do with it. Unfortunately, currently I won't have time to look into the details.

khoegh:
I've had 7 crashes in as many weeks, handling several hundred thousand calls pr day.
180 is always received less than a ms after the 183 which makes asterisk crash.
There are small variations, where the crash is.
chan_pjsip.c:1611 is in common, which is:
ast_channel_tech(in)->indicate(in, AST_CONTROL_PROGRESS, top, 0);
guarded by ast_channel_lock(in);

When 180/SDP is received, there is no channel lock, and it is using the "old"
ast_indicate(in, AST_CONTROL_RINGING);


> I had a crash today. I have lots of calls between the same endpoints which look similar without crashes. I think it is a timing issue. The called endpoint responded 183 followed by 180 after 0.000358 second. (Both had SDP) Unfortunately I'm unable to fix the code. Hope the following can help anyone interested.
> 
> #0 ast_stream_topology_get_first_stream_by_type (type=AST_MEDIA_TYPE_AUDIO, topology=0x7f9038a9d5a8) at stream.c:976 976 if (stream->type == type [Current thread is 1 (Thread 0x7f8fee524700 (LWP 3461580))] (gdb) p stream $1 = (struct ast_stream *) 0x7f9785dda96a (gdb) p stream->type Cannot access memory at address 0x7f9785dda96a
> 
> (gdb) bt #0 ast_stream_topology_get_first_stream_by_type (type=AST_MEDIA_TYPE_AUDIO, topology=0x7f9038a9d5a8) at stream.c:976 #1 ast_stream_topology_create_resolved (pending_topology=pending_topology@entry=0x7f907cc439f8, configured_topology=0x7f9038a9d5a8, prefs=prefs@entry=0x157b30c, error_message=error_message@entry=0x7f8fee51eca0) at stream.c:1046 #2 0x00007f90b0298c69 in chan_pjsip_indicate (ast=0x7f903a1ee050, condition=14, data=0x7f907cc439f8, datalen=0) at chan_pjsip.c:1836 #3 0x00007f9104a6d138 in wait_for_answer (in=in@entry=0x7f903a1ee050, out_chans=out_chans@entry=0x7f8fee51fde0, to_answer=to_answer@entry=0x7f8fee51fda0, to_progress=to_progress@entry=0x7f8fee51fda4, peerflags=peerflags@entry=0x7f8fee520938, opt_args=opt_args@entry=0x7f8fee520080, pa=0x7f8fee5202e0, num_in=0x7f8fee51fe00, result=0x7f8fee51fdac, dtmf_progress=, mf_progress=, mf_wink=, sf_progress=, sf_wink=, hearpulsing=0, ignore_cc=1, forced_clid=0x7f8fee51fe50, stored_clid=0x7f8fee51fea0, config=) at app_dial.c:1611

khoegh:
I've used the following patch for a couple of weeks (~5M calls) without crashes and haven't noticed any side effects.

```
--- a/apps/app_dial.c   2024-12-02 15:34:43.486312781 +0100
+++ b/apps/app_dial.c   2025-02-12 16:03:57.472812221 +0100
@@ -1574,7 +1574,11 @@
                                                        ast_channel_early_bridge(in, c);
                                                }
                                                if (!(pa->sentringing) && !ast_test_flag64(outgoing, OPT_MUSICBACK) && ast_strlen_zero(opt_args[OPT_ARG_RINGBACK])) {
-                                                       ast_indicate(in, AST_CONTROL_RINGING);
+                                                       struct ast_stream_topology *top = ao2_bump(ast_channel_get_stream_topology(c));
+                                                       ast_channel_lock(in);
+                                                       ast_channel_tech(in)->indicate(in, AST_CONTROL_RINGING, top, 0);
+                                                       ast_channel_unlock(in);
+                                                       ao2_ref(top, -1);
                                                        pa->sentringing++;
                                                }
                                                if (!sent_ring) {
```

> I've had 7 crashes in as many weeks, handling several hundred thousand calls pr day. 180 is always received less than a ms after the 183 which makes asterisk crash. There are small variations, where the crash is. chan_pjsip.c:1611 is in common, which is: ast_channel_tech(in)->indicate(in, AST_CONTROL_PROGRESS, top, 0); guarded by ast_channel_lock(in);
> 
> When 180/SDP is received, there is no channel lock, and it is using the "old" ast_indicate(in, AST_CONTROL_RINGING);
> 
> > I had a crash today. I have lots of calls between the same endpoints which look similar without crashes. I think it is a timing issue. The called endpoint responded 183 followed by 180 after 0.000358 second. (Both had SDP) Unfortunately I'm unable to fix the code. Hope the following can help anyone interested.
> > #0 ast_stream_topology_get_first_stream_by_type (type=AST_MEDIA_TYPE_AUDIO, topology=0x7f9038a9d5a8) at stream.c:976 976 if (stream->type == type [Current thread is 1 (Thread 0x7f8fee524700 (LWP 3461580))] (gdb) p stream $1 = (struct ast_stream *) 0x7f9785dda96a (gdb) p stream->type Cannot access memory at address 0x7f9785dda96a
> > (gdb) bt #0 ast_stream_topology_get_first_stream_by_type (type=AST_MEDIA_TYPE_AUDIO, topology=0x7f9038a9d5a8) at stream.c:976 #1 ast_stream_topology_create_resolved (pending_topology=pending_topology@entry=0x7f907cc439f8, configured_topology=0x7f9038a9d5a8, prefs=prefs@entry=0x157b30c, error_message=error_message@entry=0x7f8fee51eca0) at stream.c:1046 #2 0x00007f90b0298c69 in chan_pjsip_indicate (ast=0x7f903a1ee050, condition=14, data=0x7f907cc439f8, datalen=0) at chan_pjsip.c:1836 #3 0x00007f9104a6d138 in wait_for_answer (in=in@entry=0x7f903a1ee050, out_chans=out_chans@entry=0x7f8fee51fde0, to_answer=to_answer@entry=0x7f8fee51fda0, to_progress=to_progress@entry=0x7f8fee51fda4, peerflags=peerflags@entry=0x7f8fee520938, opt_args=opt_args@entry=0x7f8fee520080, pa=0x7f8fee5202e0, num_in=0x7f8fee51fe00, result=0x7f8fee51fdac, dtmf_progress=, mf_progress=, mf_wink=, sf_progress=, sf_wink=, hearpulsing=0, ignore_cc=1, forced_clid=0x7f8fee51fe50, stored_clid=0x7f8fee51fea0, config=) at app_dial.c:1611

eallovon:
Hi @maximilianfridrich 

I'm back again on this PR.
Last [time I wrote](https://github.com/asterisk/asterisk/pull/285#issuecomment-2607348011)
> our interest in this PR decreased quite a lot, and I'm not sure we would be able to help more then. I'm sorry for that.

In fact after more test the workaround I had found did not work for all cases so we had to add this ACN patch to our asterisk.

After some discussion with our customer he is willing to help on this ACN work in order for this PR to - hopefully - be included in asterisk. So we have now some time to work on this.

As you said last time :

> To move this PR forward, I would say the following must be done in this order (can be done by anyone):
> 
>     1. Fix failing PJSIP + T.38 tests. Some might simply be failing because some of the new config options need different defaults for backward compatibility reasons - see [this comment](https://github.com/asterisk/asterisk/pull/285#issuecomment-1698949588), and some might need changes to the [Testsuite PR](https://github.com/asterisk/testsuite/pull/20).
> 
>     2. Together with maintainers, figure out how to resolve backwards compatibility issues.
>        E.g. If this PR goes into Asterisk 23/24, can we then drop the old config options if the defaults remain the same? If not, then we have to figure out how both sets of options can coexist (and then this PR can always be merged into the release branches). Since it's not urgent for us, removing the old options would be our preference, since it's simpler for us. A merge into master would have to be done [before mid August](https://docs.asterisk.org/About-the-Project/Asterisk-Versions/) for it to be in Asterisk 23.


We are going in the next couple of weeks to try to address the first point, namely *Fix failing PJSIP + T.38 tests*.

Two questions: 
- first to take over the work we thought that the easiest would probably that we work on our side and maybe re-open PR (asterisk and testsuite) based on yours, mentioning your work ; if that's ok for you ?
- second, for the collaboration, if we have any question on the current implementation are you still willing to help on this PR ? do you have any sort of time to perhaps review our proposal ? etc.

looking forward to reading you

maximilianfridrich:
@eallovon That sounds great, thank you for picking this up!

- That works for me, please either just keep my commits in your PR and add more commits or add me as a [Co-Author](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) to the commit in your PR (or add yourself as a Co-Author to my commit :wink:).
- Yes, I will certainly make time for review and I can also help if there's questions. I will likely won't have much time to actively code, but I will happily try out your changes and review/make suggestions.

btriller:
@eallovon I added some changes to https://github.com/asterisk/asterisk/compare/master...btriller:asterisk:acn

eallovon:
Hi @maximilianfridrich 

So we were able to take some time to work on the T.38 issue.

Our analysis seem to indicate that there is some kind of desynchronization between the media session and the T.38 part which leads asterisk to refuse the re-INVITE.
(As a reminder a T.38 session is established via a re-INVITE containing the T.38 in the SDP after a first INVITE (with usually alaw codec) was accepted between caller and callee).

From what we understood this desynchronization seem to come from [these modifications in the PR](https://github.com/asterisk/asterisk/pull/285/files#diff-7374bf50d8247b70fc29965fa7a22bd75b58e9be3b3afc8f53bc04ac8301799bL2660-R3029) which pushes to the stream handling only a copy of the media_state.
This is what leads the T.38 not to be aware of some modifications.

From our analysis this copy does not really seem necessary (?) and from our tests, going backward, on this media_state copy restores the T.38 scenarios without breaking the ACN as far as we tried.

Regarding the asterisk testsuite it seems to pass all the T.38 related test (except the `directmedia_reinvite_t38` scenario).
Before looking deeper in this scenario we would be interested by your feedback on this patch which, in fact, only go backward on [these changes](https://github.com/asterisk/asterisk/pull/285/files#diff-7374bf50d8247b70fc29965fa7a22bd75b58e9be3b3afc8f53bc04ac8301799bL2660-R3029).

Here is the proposed patch. Please don't hesitate to put it in your PR (we don't really care of being mentioned etc.) so that the testsuite could be rerun and we could check everything's fine :
```diff
Index: src/asterisk/res/res_pjsip_session.c
===================================================================
--- src.orig/asterisk/res/res_pjsip_session.c
+++ src/asterisk/res/res_pjsip_session.c
@@ -2870,11 +2870,7 @@ static int sdp_requires_deferral(struct
                /* As this is only called on an incoming SDP offer before processing it is not possible
                 * for streams and their media sessions to exist.
                 */
-               media_state = ast_sip_session_media_state_clone(session->pending_media_state);
-               if (!media_state) {
-                       return -1;
-               }
-               if (ast_stream_topology_set_stream(media_state->topology, i, stream)) {
+               if (ast_stream_topology_set_stream(session->pending_media_state->topology, i, stream)) {
                        ast_stream_free(stream);
                        return -1;
                }
@@ -2887,7 +2883,7 @@ static int sdp_requires_deferral(struct
                        }
                }
 
-               session_media = ast_sip_session_media_state_add(session, media_state, ast_media_type_from_str(media), i);
+               session_media = ast_sip_session_media_state_add(session, session->pending_media_state, ast_media_type_from_str(media), i);
                if (!session_media) {
                        return -1;
                }
@@ -3027,8 +3023,6 @@ static pj_bool_t session_reinvite_on_rx_
                return PJ_FALSE;
        }
 
-       ast_sip_session_media_state_reset(session->pending_media_state);
-
        pjsip_rx_data_clone(rdata, 0, &session->deferred_reinvite);
 
        return PJ_TRUE;
```

Any feedback would be appreciated, thanks

eallovon:
hi @maximilianfridrich 

We continued to look into the T.38 tests and specifically the `directmedia_reinvite_t38` which was the last failing.

In the end this second patch fix also the `directmedia_reinvite_t38` test
```diff
Index: asterisk/channels/chan_pjsip.c
===================================================================
--- asterisk.orig/channels/chan_pjsip.c
+++ asterisk/channels/chan_pjsip.c
@@ -208,10 +208,16 @@ static enum ast_rtp_glue_result chan_pjs
                return AST_RTP_GLUE_RESULT_FORBID;
        }
 
+       /*
+        * Forbid native RTP only when T.38 is actually in use or negotiating.
+        */
        datastore = ast_sip_session_get_datastore(channel->session, "t38");
        if (datastore) {
+               enum ast_sip_session_t38state t38state = channel->session->t38state;
                ao2_ref(datastore, -1);
-               return AST_RTP_GLUE_RESULT_FORBID;
+               if (t38state != T38_DISABLED) {
+                       return AST_RTP_GLUE_RESULT_FORBID;
+               }
        }
 
        endpoint = channel->session->endpoint;
```

So normally with these two patches it should fix the T.38 scenario and the first task we were trying to accomplish to move this PR forward.

Did you have any time to look at the first patch we suggested ? Any comments ?

Again don't hesitate to integrate it in your PR to see if it actually fixes the testsuite.

Or do you prefer that I open a new PR ?

maximilianfridrich:
Hi @eallovon. Sorry for the late response, I'm a little swamped these days. The patches look good! I just added them as a commit to this branch and triggered the CI testsuite runs here on GitHub. Let's see what they report.

eallovon:
hi @maximilianfridrich 

I had a quick look at the failure. I searched for `incoming_call_offer_pref` in the repo and if I understand well the failure is du to the inclusion present in `res/res_pjsip/pjsip_manager.xml`.

This diff should do the trick (as pjsip_manager.xml already references the new options) :

```diff
diff --git a/res/res_pjsip/pjsip_manager.xml b/res/res_pjsip/pjsip_manager.xml
index 9f2cf634c5..50861ade33 100644
--- a/res/res_pjsip/pjsip_manager.xml
+++ b/res/res_pjsip/pjsip_manager.xml
@@ -638,9 +638,6 @@
                                <parameter name="Ignore183WithoutSdp">
                                        <para><xi:include xpointer="xpointer(/docs/configInfo[@name='res_pjsip']/configFile[@name='pjsip.conf']/configObject[@name='endpoint']/configOption[@name='ignore_183_without_sdp']/synopsis/node())"/></para>
                                </parameter>
-                               <parameter name="IncomingCallOfferPref">
-                                       <para><xi:include xpointer="xpointer(/docs/configInfo[@name='res_pjsip']/configFile[@name='pjsip.conf']/configObject[@name='endpoint']/configOption[@name='incoming_call_offer_pref']/synopsis/node())"/></para>
-                               </parameter>
                                <parameter name="IncomingMwiMailbox">
                                        <para><xi:include xpointer="xpointer(/docs/configInfo[@name='res_pjsip']/configFile[@name='pjsip.conf']/configObject[@name='endpoint']/configOption[@name='incoming_mwi_mailbox']/synopsis/node())"/></para>
                                </parameter>
```

maximilianfridrich:
multiple-commits: interim

maximilianfridrich:
Thank you for your efforts @eallovon. Currently, I still don't have the capacity to work on this unfortunately. I wish I could be of more help.

eallovon:
hello

Just for information we're still investigating why our suggested patches failed quite some tests in the testsuite.
It seems that:
1. the patch we sent was done on top of other patches we have on the T.38 which make the whole work together
2. and also that our patch create some problem with some re-INVITE scenario

Sorry for this and sorry for the "no news" since a few weeks.
We're still trying to figure this out and I'm also working on quite some other topics so it's a bit on the side that we're able to move this forward.

I hope to be able give more news in a not to far future ... :sweat_smile:

eallovon:
hello,

So I'm back with a new patch that should fix most of the failing tests.

Attached is a new patch that should apply on top of your branch.
[fix-ACN-reinvite-for-upstream.patch](https://github.com/user-attachments/files/24617658/fix-ACN-reinvite-for-upstream.patch)

It mainly fixes the re-INVITE scenario where it was needed to distinguish whether it was a T.38 scenario or not when refreshing the session.


I re-ran the testsuite (but mostly on the tests that failed last time) and here are the results compared between your initial PR, your PR with our first patches (current state of your PR) and then with this new attached patch (see table at the end).
So in the end we have:
- tests that were passing on your PR now pass again (like the blind_transfer)
- at least 3 new t38 tests should pass
- and then some tests don't pass at all (like dtmf_sdp_48k_audio_8k_digits_bob) (did not investigate for the moment)


Could you please integrate this new patch in your PR so that we could see what the *real* testsuite says.

I hope the real testsuite will give us the same results.
Then there are still some T38 case to fix. We may be able to invest a bit more time on this (and on the backward compatibility) but we will see.

Regards,


## Appendix 

Test result I had :

| Test | pr285 [mfridrich](https://github.com/asterisk/asterisk/commit/35521e2bd62703dbb4d1e36744366c75cdf0c818) | pr285 mfridrich + patch1-2 ([Fix ACN for T.38](https://github.com/asterisk/asterisk/commit/1ad5259d5f3384ecf4e82798b9ff4adf8ea6ef01)) | pr285 mfridrich + patch1-3|
| --- | --- | --- | ---|
| apps/dial/preserve_top| ✅️| ❌️ | ✅️|
| channels/pjsip/all_codecs_on_empty_reinvite/early-offer| ✅️| ❌️ | ✅️|
| channels/pjsip/all_codecs_on_empty_reinvite/late-offer| ✅️| ❌️ | ✅️|
| channels/pjsip/basic_calls/two_parties/nominal/add_remove_sources| ✅️| ❌️ | ✅️|
| channels/pjsip/basic_calls/two_parties/nominal/video_bidirectional_to_unidirectional| ✅️| ❌️ | ✅️|
| channels/pjsip/basic_calls/two_parties/nominal/video_unidirectional_to_bidirectional| ✅️| ❌️ | ✅️|
| channels/pjsip/connected_line/connected_line_allow| ✅️| ❌️ | ✅️|
| channels/pjsip/connected_line/connected_line_no_send| ✅️| ❌️ | ✅️|
| channels/pjsip/connected_line/connected_line_no_trust| ✅️| ❌️ | ✅️|
| channels/pjsip/dtmf_sdp/dtmf_sdp_generation/dtmf_sdp_48k_audio_8k_digits_bob | ❌️| ❌️ | ❌️|
| channels/pjsip/dtmf_sdp/dtmf_sdp_generation/dtmf_sdp_48k_priority_diff | ❌️| ❌️ | ❌️|
| channels/pjsip/dtmf_sdp/dtmf_sdp_generation/dtmf_sdp_8k_priority_diff | ❌️| ❌️ | ❌️|
| channels/pjsip/dtmf_sdp/dtmf_sdp_negotiation_alice | ❌️| ❌️ | ❌️|
| channels/pjsip/dtmf_sdp/dtmf_sdp_negotiation_alice_asym | ❌️| ❌️ | ❌️|
| channels/pjsip/dtmf_sdp/dtmf_sdp_negotiation_bob | ❌️| ❌️ | ❌️|
| channels/pjsip/hold| ✅️| ❌️ | ✅️|
| channels/pjsip/hold_ice| ✅️| ❌️ | ✅️|
| channels/pjsip/hold_inactive| ✅️| ❌️ | ✅️|
| channels/pjsip/moh_passthru_inactive| ✅️| ❌️ | ✅️|
| channels/pjsip/nat/rewrite_contact/route_set_reinvite| ✅️| ❌️ | ✅️|
| channels/pjsip/rtp/timeout_hold| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/callee_local| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/callee_local_app |  SKIPPED| SKIPPED | SKIPPED|
| channels/pjsip/transfers/attended_transfer/nominal/callee_local_blonde| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/callee_local_direct_media |  SKIPPED| SKIPPED | SKIPPED|
| channels/pjsip/transfers/attended_transfer/nominal/callee_local_semi_attended_transfer_record_route| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/callee_remote| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/caller_local| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/caller_local_app| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/caller_local_blonde| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/attended_transfer/nominal/caller_local_direct_media |  SKIPPED| SKIPPED | SKIPPED|
| channels/pjsip/transfers/attended_transfer/nominal/caller_remote| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/blind_transfer/callee_refer_only| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/blind_transfer/callee_with_hold| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/blind_transfer/caller_refer_only| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/blind_transfer/caller_with_hold| ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/blind_transfer/caller_with_hold_481 |  ✅️| ❌️ | ✅️|
| channels/pjsip/transfers/blind_transfer/caller_with_hold_drop_options| ✅️| ❌️ | ✅️|
| fax/pjsip/directmedia_reinvite_t38 | ❌️| ✅️ | ✅️|
| fax/pjsip/t38 | ❌️| ✅️ | ✅️|
| fax/pjsip/t38_1xx_response| ✅️| ✅️ | ✅️|
| fax/pjsip/t38_enabled_two_reinvite_on_a_leg |  SKIPPED| SKIPPED | ❌️|
| fax/pjsip/t38_enabled_two_reinvite_on_b_leg |  SKIPPED| SKIPPED | ❌️|
| fax/pjsip/t38_fast_reject | ❌️| ❌️ | ❌️|
| fax/pjsip/t38_initial_offer | ❌️| ❌️ | ❌️|
| fax/pjsip/t38_with_auth | ❌️| ✅️ | ✅️|
| fax/pjsip/t38_zero_port| ✅️| ✅️ | ✅️|
| manager/bridge_actions| ✅️| ❌️ | ✅️|

maximilianfridrich:
Thank you very much @eallovon! I applied your patch to this branch by amending the last commit (with minor restructuring of your patch to fix "ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]" errors). Let's see what the CI-run tests report.

eallovon:
Thank you for the update @maximilianfridrich  (and for the fix in our patch).

In the end the test result looks even better than what I had on my side.
If I read correctly:
- all T38 tests passed (except t38_initial_offer but that may be an easy one)
- and now we are left with some failures which I think mainly consist of new tests in the testsuite that were added after the PR you did in the testsuite so it's perhaps more a matter of configuration (and of this backward compatibility issue)

silentindark:
@maximilianfridrich Could you please also update testsuite, and re-run tests?

NickBouwhuis:
Hi @maximilianfridrich, @eallovon,

I dug into the remaining CI failures on this PR and I believe I have root causes for all of them, plus a patch (attached below) that I verified against the testsuite (current testsuite **master**, Debian 13, `--with-pjproject-bundled --enable-dev-mode`). With it, all 9 tests from the last PRCheck run that are actually related to this PR pass, and a sweep of previously-passing codec-sensitive tests stays green.

## Root cause analysis of the failing gates

### master-pjs1 (geolocation ×2, message_redirect ×2) — not caused by this PR

These are simply the branch being ~250 commits behind master. The testsuite tests were updated/added for features that landed in Asterisk master after this branch's base:

- `res_pjsip`: 3xx redirect module (`2bd999ae5f`, `d8caa664a8`) → `message_redirect`, `message_redirect_multiple`
- `res_geolocation`/`res_pjsip_geolocation` updates (`5c16412dc4`, `2c97fd3ea4`) → `geolocation/incoming/*`

A rebase fixes these four. I test-merged current master: only `res/res_pjsip/pjsip_configuration.c` conflicts, and trivially.

### master-pjs2 (dtmf_sdp ×6) and master-fax (t38_initial_offer) — default behavior regression

All of these tests assert the pre-ACN default negotiation behavior:

- Answers must follow the **endpoint's configured codec order**, not the offer's order. E.g. `t38_initial_offer` offers `RTP/AVP 9 8 0` against `allow=!all,g722,ulaw,alaw` and expects the answer `9 0 8`. With `codec_prefs_incoming_offer = prefer: pending` the answer comes back `9 8 0` and the test fails. Same story for the `dtmf_sdp` priority tests.
- Outgoing offers must contain **exactly the endpoint's allow list**. The `_bob` tests fail because with `codec_prefs_outgoing_offer = prefer: pending, operation: union`, the slin/L16 format that the core requests for an AMI Originate gets unioned into the wire offer: Asterisk offers `m=audio ... RTP/AVP 10 0 101` (L16 first!) where legacy behavior — and the test — expects `0 101`.

This is exactly the backwards-compatibility point @jcolp raised back in 2023 ("default behavior with this change should match previous versions"), and matches @maximilianfridrich's own 2023 analysis that the incoming defaults would have to become `prefer: configured`. The patch changes the defaults to:

| option | old default | new default |
|---|---|---|
| `codec_prefs_incoming_offer` | `prefer: pending, operation: intersect` | `prefer: configured, operation: intersect` |
| `codec_prefs_incoming_answer` | `prefer: pending, operation: intersect` | `prefer: configured, operation: intersect` |
| `codec_prefs_outgoing_offer` | `prefer: pending, operation: union` | `prefer: configured, operation: only_preferred` |
| `codec_prefs_outgoing_answer` | `prefer: pending, operation: intersect` | (unchanged) |

`prefer: configured, operation: only_preferred` for the outgoing offer reproduces the legacy "offer the endpoint's allow list, in configured order" behavior (the resolved stream is exactly the configured caps). Anyone who wants the new pending-preferred behavior can still opt in per endpoint — the point is only that *unset* options behave like pre-ACN Asterisk.

### master-pjs5 (codec_mismatch, hold_declined) — real bugs in `handle_incoming_sdp()`

Three compounding problems in the no-common-codec path of `handle_incoming_sdp()`:

1. **Whole-SDP abort instead of per-stream decline.** When *any single* stream resolves to empty caps, the code does `goto end` with `handled = 0`, i.e. 488s the entire INVITE. The `codec-mismatch` test's audio-mismatch scenario offers a bad audio stream + a good video stream and expects `200 OK` with `m=audio 0` and live video (which is what pre-ACN Asterisk does via the SDP handler returning 0). The patch declines just the affected stream via the same path as the existing port-0 decline; if *no* stream ends up accepted, `handled` stays 0 and the 488 still happens naturally.
2. **Codec resolution runs on streams the remote already declined.** `hold-declined`'s initial offer contains a valid audio stream plus a declined video stream (`m=video 0 RTP/AVP 99 34`) with codecs that don't match the endpoint. Resolution runs on it *before* the port-0 check and kills the whole call. The patch skips resolution for `desc.port == 0` streams.
3. **Double-free.** `resolved_stream` is declared as `RAII_VAR(..., ast_stream_free)`, but the empty-resolution path also called `ast_stream_free(resolved_stream)` explicitly before `goto end` — so a non-NULL stream with empty formats got freed twice. (Possibly related to the crashes @khoegh reported.) The patch drops the explicit free.

## Test results

Everything run against current testsuite master. Baseline is this branch as-is (`d659d4c92c`), reproducing the CI failures 1:1; "patched" is with the attached patch:

| Test | baseline | patched |
|---|---|---|
| `sdp_offer_answer/.../audio-video/codec-mismatch` | ❌ | ✅ |
| `sdp_offer_answer/.../audio-video/hold-declined` | ❌ | ✅ |
| `fax/pjsip/t38_initial_offer` | ❌ | ✅ |
| `dtmf_sdp/dtmf_sdp_generation/dtmf_sdp_48k_audio_8k_digits_bob` | ❌ | ✅ |
| `dtmf_sdp/dtmf_sdp_generation/dtmf_sdp_48k_priority_diff` | ❌ | ✅ |
| `dtmf_sdp/dtmf_sdp_generation/dtmf_sdp_8k_priority_diff` | ❌ | ✅ |
| `dtmf_sdp/dtmf_sdp_negotiation_alice` | ❌ | ✅ |
| `dtmf_sdp/dtmf_sdp_negotiation_alice_asym` | ❌ | ✅ |
| `dtmf_sdp/dtmf_sdp_negotiation_bob` | ❌ | ✅ |

Regression sweep of previously-passing codec-sensitive tests, all still ✅ with the patch: `apps/dial/preserve_top`, `all_codecs_on_empty_reinvite/early-offer`, `all_codecs_on_empty_reinvite/late-offer`, `dtmf_sdp_48k_priority_match`, `dtmf_sdp_48k_priority_match_8k_digits`, `dtmf_sdp_8k_priority_match`, `sdp_offer_answer/.../audio/basic`, `sdp_offer_answer/.../decline/basic`, `channels/pjsip/hold`, `fax/pjsip/t38`.

## Remaining work beyond this patch

- **Rebase onto master** (fixes the 4 pjs1 failures; one small conflict).
- The 2023 review blocker still stands: `incoming_call_offer_pref`/`outgoing_call_offer_pref` are removed and the Alembic migration drops their columns. With the compat defaults above, restoring the old options (old options take priority when set, ACN options otherwise, as @jcolp suggested) becomes much more tractable, and the migration could be dropped entirely.
- The PR checklist bot wants an UpgradeNote containing "alembic"/"database"/"schema" — moot if the migration is dropped.
- With these defaults, most of the pjsip.conf adaptations in the testsuite PR (asterisk/testsuite#20) to *pre-existing* tests should no longer be needed and could be reverted, which also addresses the "please update testsuite" request.

<details>
<summary><b>Patch</b> (applies on top of <code>d659d4c92c</code>)</summary>

```diff
diff --git a/res/res_pjsip/pjsip_config.xml b/res/res_pjsip/pjsip_config.xml
index 960fed1f8b..df0817a46c 100644
--- a/res/res_pjsip/pjsip_config.xml
+++ b/res/res_pjsip/pjsip_config.xml
@@ -106,8 +106,8 @@
 								<para>
 								</para>
 								<enumlist>
-									<enum name="pending"><para>The codec list from the caller. (default)</para></enum>
-									<enum name="configured"><para>The codec list from the endpoint.</para></enum>
+									<enum name="pending"><para>The codec list from the caller.</para></enum>
+									<enum name="configured"><para>The codec list from the endpoint. (default)</para></enum>
 								</enumlist>
 							</enum>
 							<enum name="operation : &lt; intersect | only_preferred | only_nonpreferred &gt;">
@@ -170,17 +170,17 @@
 								<para>
 								</para>
 								<enumlist>
-									<enum name="pending"><para>The codec list from the core. (default)</para></enum>
-									<enum name="configured"><para>The codec list from the endpoint.</para></enum>
+									<enum name="pending"><para>The codec list from the core.</para></enum>
+									<enum name="configured"><para>The codec list from the endpoint. (default)</para></enum>
 								</enumlist>
 							</enum>
 							<enum name="operation : &lt; union | intersect | only_preferred | only_nonpreferred &gt;">
 								<para>
 								</para>
 								<enumlist>
-									<enum name="union"><para>Merge the lists with the preferred codecs first. (default)</para></enum>
-									<enum name="intersect"><para>Only common codecs with the preferred codecs first. (default)</para></enum>
-									<enum name="only_preferred"><para>Use only the preferred codecs.</para></enum>
+									<enum name="union"><para>Merge the lists with the preferred codecs first.</para></enum>
+									<enum name="intersect"><para>Only common codecs with the preferred codecs first.</para></enum>
+									<enum name="only_preferred"><para>Use only the preferred codecs. (default)</para></enum>
 									<enum name="only_nonpreferred"><para>Use only the non-preferred codecs.</para></enum>
 								</enumlist>
 							</enum>
@@ -235,8 +235,8 @@
 								<para>
 								</para>
 								<enumlist>
-									<enum name="pending"><para>The codec list in the received SDP answer. (default)</para></enum>
-									<enum name="configured"><para>The codec list from the endpoint.</para></enum>
+									<enum name="pending"><para>The codec list in the received SDP answer.</para></enum>
+									<enum name="configured"><para>The codec list from the endpoint. (default)</para></enum>
 								</enumlist>
 							</enum>
 							<enum name="operation : &lt; union | intersect | only_preferred | only_nonpreferred &gt;">
diff --git a/res/res_pjsip/pjsip_configuration.c b/res/res_pjsip/pjsip_configuration.c
index 5f6fa62b3f..7e45c11889 100644
--- a/res/res_pjsip/pjsip_configuration.c
+++ b/res/res_pjsip/pjsip_configuration.c
@@ -1269,15 +1269,15 @@ static int codec_prefs_handler(const struct aco_option *opt,
 			return -1;
 		}
 		option_prefs = &endpoint->media.codec_prefs_incoming_offer;
-		default_prefer = CODEC_NEGOTIATION_PREFER_PENDING;
+		default_prefer = CODEC_NEGOTIATION_PREFER_CONFIGURED;
 		default_operation = CODEC_NEGOTIATION_OPERATION_INTERSECT;
 	} else if (strcmp(var->name, "codec_prefs_outgoing_offer") == 0) {
 		option_prefs = &endpoint->media.codec_prefs_outgoing_offer;
-		default_prefer = CODEC_NEGOTIATION_PREFER_PENDING;
-		default_operation = CODEC_NEGOTIATION_OPERATION_UNION;
+		default_prefer = CODEC_NEGOTIATION_PREFER_CONFIGURED;
+		default_operation = CODEC_NEGOTIATION_OPERATION_ONLY_PREFERRED;
 	} else if (strcmp(var->name, "codec_prefs_incoming_answer") == 0) {
 		option_prefs = &endpoint->media.codec_prefs_incoming_answer;
-		default_prefer = CODEC_NEGOTIATION_PREFER_PENDING;
+		default_prefer = CODEC_NEGOTIATION_PREFER_CONFIGURED;
 		default_operation = CODEC_NEGOTIATION_OPERATION_INTERSECT;
 	} else if (strcmp(var->name, "codec_prefs_outgoing_answer") == 0) {
 		option_prefs = &endpoint->media.codec_prefs_outgoing_answer;
@@ -2226,13 +2226,13 @@ int ast_res_pjsip_initialize_configuration(void)
 	ast_sorcery_object_field_register(sip_sorcery, "endpoint", "suppress_q850_reason_headers", "no", OPT_BOOL_T, 1, FLDSET(struct ast_sip_endpoint, suppress_q850_reason_headers));
 	ast_sorcery_object_field_register(sip_sorcery, "endpoint", "ignore_183_without_sdp", "no", OPT_BOOL_T, 1, FLDSET(struct ast_sip_endpoint, ignore_183_without_sdp));
 	ast_sorcery_object_field_register_custom(sip_sorcery, "endpoint", "codec_prefs_incoming_offer",
-		"prefer: pending, operation: intersect, keep: all, transcode: allow",
+		"prefer: configured, operation: intersect, keep: all, transcode: allow",
 		codec_prefs_handler, incoming_offer_codec_prefs_to_str, NULL, 0, 0);
 	ast_sorcery_object_field_register_custom(sip_sorcery, "endpoint", "codec_prefs_outgoing_offer",
-		"prefer: pending, operation: union, keep: all, transcode: allow",
+		"prefer: configured, operation: only_preferred, keep: all, transcode: allow",
 		codec_prefs_handler, outgoing_offer_codec_prefs_to_str, NULL, 0, 0);
 	ast_sorcery_object_field_register_custom(sip_sorcery, "endpoint", "codec_prefs_incoming_answer",
-		"prefer: pending, operation: intersect, keep: all",
+		"prefer: configured, operation: intersect, keep: all",
 		codec_prefs_handler, incoming_answer_codec_prefs_to_str, NULL, 0, 0);
 	ast_sorcery_object_field_register_custom(sip_sorcery, "endpoint", "codec_prefs_outgoing_answer",
 		"prefer: pending, operation: intersect, keep: all",
diff --git a/res/res_pjsip_session.c b/res/res_pjsip_session.c
index da0bfe3d69..bcc8cf80b3 100644
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -877,6 +877,7 @@ static int handle_incoming_sdp(struct ast_sip_session *session, const pjmedia_sd
 		RAII_VAR(struct sdp_handler_list *, handler_list, NULL, ao2_cleanup);
 		struct ast_sip_session_media *session_media = NULL;
 		int res;
+		int no_common_codecs = 0;
 		enum ast_media_type type;
 		struct ast_stream *stream = NULL;
 		pjmedia_sdp_media *remote_stream = sdp->media[i];
@@ -939,16 +940,19 @@ static int handle_incoming_sdp(struct ast_sip_session *session, const pjmedia_sd
 			}
 
 			remote_ast_stream = ast_stream_topology_get_first_stream_by_type(remote_topology, type);
-			if (remote_ast_stream) {
+			if (remote_ast_stream && remote_stream->desc.port) {
 				/* Here we always have a remote offer, so we create a resolved stream using the remote SDP offer
 				 * and the configured endpoint settings codec_prefs_incoming_offer. */
 				configured_stream = ast_stream_topology_get_first_stream_by_type(session->endpoint->media.topology, type);
 				resolved_stream = ast_stream_create_resolved(remote_ast_stream, configured_stream, &session->endpoint->media.codec_prefs_incoming_offer, NULL);
 				if (!resolved_stream || ast_format_cap_empty(ast_stream_get_formats(resolved_stream))) {
-					ast_stream_free(resolved_stream); /* Can handle NULL */
+					/* Decline just this stream below instead of failing the whole
+					 * negotiation. If no stream ends up accepted, the negotiation
+					 * still fails as a whole. resolved_stream is freed by RAII_VAR. */
 					stream = ast_stream_alloc(stream_name, type);
-					handled = 0;
-					SCOPE_EXIT_LOG_EXPR(goto end, LOG_ERROR, "No common codecs between incoming SDP offer and endpoint configuration.\n");
+					no_common_codecs = 1;
+					ast_trace(-1, "%s: No common codecs between incoming SDP offer and endpoint configuration, declining stream\n",
+						ast_sip_session_get_name(session));
 				} else {
 					stream = ast_stream_clone(resolved_stream, stream_name);
 				}
@@ -1007,8 +1011,9 @@ static int handle_incoming_sdp(struct ast_sip_session *session, const pjmedia_sd
 				 ast_sip_session_get_name(session));
 		}
 
-		/* If this stream is already declined mark it as such, or mark it as such if we've reached the limit */
-		if (!remote_stream->desc.port || is_stream_limitation_reached(type, session->endpoint, type_streams)) {
+		/* If this stream is already declined mark it as such, or mark it as such if we've reached
+		 * the limit or no codecs were in common with the endpoint configuration */
+		if (no_common_codecs || !remote_stream->desc.port || is_stream_limitation_reached(type, session->endpoint, type_streams)) {
 			remove_stream_from_bundle(session_media, stream);
 			SCOPE_EXIT_EXPR(continue, "%s: Declining incoming SDP media stream %s'\n",
 				ast_sip_session_get_name(session), ast_str_tmp(128, ast_stream_to_str(stream, &STR_TMP)));
```

</details>

Issue #289, opened 2023-08-31
[new-feature]: Add support for deleting channel and global variables


jcolp:
### Feature Description

Adds two new dialplan functions to allow the deletion of channel and global variables.

PR #290, opened 2023-08-31
variables: Add additional variable dialplan functions.

jcolp:
Using the Set dialplan application does not actually
    delete channel or global variables. Instead the
    variables are set to an empty value.
    
    This change adds two dialplan functions,
    GLOBAL_DELETE and DELETE which can be used to
    delete global and channel variables instead
    of just setting them to empty.
    
    There is also no ability within the dialplan to
    determine if a global or channel variable has
    actually been set or not.
    
    This change also adds two dialplan functions,
    GLOBAL_EXISTS and VARIABLE_EXISTS which can be
    used to determine if a global or channel variable
    has been set or not.
    
    Resolves: #289
    
    UserNote: Four new dialplan functions have been added.
    GLOBAL_DELETE and DELETE have been added which allows
    the deletion of global and channel variables.
    GLOBAL_EXISTS and VARIABLE_EXISTS have been added
    which checks whether a global or channel variable has
    been set.

### DIFF
--- a/funcs/func_global.c
+++ b/funcs/func_global.c
@@ -53,6 +53,40 @@
 			<para>Set or get the value of a global variable specified in <replaceable>varname</replaceable></para>
 		</description>
 	</function>
+	<function name="GLOBAL_DELETE" language="en_US">
+		<synopsis>
+			Deletes a specified global variable.
+		</synopsis>
+		<syntax>
+			<parameter name="varname" required="true">
+				<para>Global variable name</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Delete the global variable specified in <replaceable>varname</replaceable>.
+			Will succeed if the global variable exists or not.</para>
+		</description>
+		<see-also>
+			<ref type="function">GLOBAL</ref>
+			<ref type="function">DELETE</ref>
+		</see-also>
+	</function>
+	<function name="GLOBAL_EXISTS" language="en_US">
+		<synopsis>
+			Check if a global variable exists or not.
+		</synopsis>
+		<syntax>
+			<parameter name="varname" required="true">
+				<para>Global variable name</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Returns <literal>1</literal> if global variable exists or <literal>0</literal> otherwise.</para>
+		</description>
+		<see-also>
+			<ref type="function">VARIABLE_EXISTS</ref>
+		</see-also>
+	</function>
 	<function name="SHARED" language="en_US">
 		<synopsis>
 			Gets or sets the shared variable specified.
@@ -143,6 +177,33 @@ static struct ast_custom_function global_function = {
 	.name = "GLOBAL",
 	.read = global_read,
 	.write = global_write,
+};
+
+static int global_delete_write(struct ast_channel *chan, const char *cmd, char *data, const char *value)
+{
+	pbx_builtin_setvar_helper(NULL, data, NULL);
+
+	return 0;
+}
+
+static struct ast_custom_function global_delete_function = {
+	.name = "GLOBAL_DELETE",
+	.write = global_delete_write,
+};
+
+static int global_exists_read(struct ast_channel *chan, const char *cmd, char *data,
+		  char *buf, size_t len)
+{
+	const char *var = pbx_builtin_getvar_helper(NULL, data);
+
+	strcpy(buf, var ? "1" : "0");
+
+	return 0;
+}
+
+static struct ast_custom_function global_exists_function = {
+	.name = "GLOBAL_EXISTS",
+	.read = global_exists_read,
 };
 
 static int shared_read(struct ast_channel *chan, const char *cmd, char *data, char *buf, size_t len)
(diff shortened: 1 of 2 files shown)


jcolp:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

funcs/func_global.c
```diff
@@ -53,6 +53,19 @@
...
+		<synopsis>
+			Deletes a specified global variable.
+		</synopsis>
+		<syntax>
+			<parameter name="varname" required="true">
+				<para>Global variable name</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Delete the global variable specified in <replaceable>varname</replaceable></para>
```
InterLinked1, line 66: To use the same verb tense as other descriptions, I think "Deletes" would be better than "Delete"
Also might be good to state that it will work (always return 0) regardless of whether the variable existed or not, I think some people might wonder about that.

main/pbx_variables.c
```diff
@@ -105,6 +105,19 @@
...
+		<synopsis>
+			Deletes a specified channel variable.
+		</synopsis>
+		<syntax>
+			<parameter name="varname" required="true">
+				<para>Channel variable name</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Delete the channel variable specified in <replaceable>varname</replaceable></para>
```
InterLinked1, line 118: Same suggestion about tense, "Deletes" vs "Delete"

funcs/func_global.c
```diff
@@ -145,6 +158,18 @@ static struct ast_custom_function global_function = {
 	.write = global_write,
 };
 
+static int global_delete_write(struct ast_channel *chan, const char *cmd, char *data, const char *value)
+{
+	pbx_builtin_setvar_helper(NULL, data, NULL);
```
InterLinked1, line 163: This function returns an int, usually 0, so could be simplified to just:

`return pbx_builtin_setvar_helper(NULL, data, NULL);`

funcs/func_global.c
```diff
@@ -53,6 +53,19 @@
...
+		</synopsis>
+		<syntax>
+			<parameter name="varname" required="true">
+				<para>Global variable name</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Delete the global variable specified in <replaceable>varname</replaceable></para>
+		</description>
+	</function>
```
InterLinked1, line 68: There should be a see-also for VARIABLE_DELETE (and maybe GLOBAL)

main/pbx_variables.c
```diff
@@ -105,6 +105,19 @@
...
+		</description>
+	</function>
```
InterLinked1, line 120: There should be a see-also linking to GLOBAL_DELETE (and maybe SET)

InterLinked1 (changes requested):

funcs/func_global.c
jcolp: I am amused as the verb tense matches that of the existing dialplan function this was copied from.

InterLinked1: > I am amused as the verb tense matches that of the existing dialplan function this was copied from.

Hmm, I was looking at the Gets/Sets nearby... in that case just leave it as is.

gtjoseph (changes requested):
I think the VARIABLE_DELETE stuff should go in func_logic since that's where the SET function is implemented.  I'd also shorten the name to just DELETE.

funcs/func_global.c
```diff
@@ -53,6 +53,24 @@
...
+				<para>Global variable name</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Delete the global variable specified in <replaceable>varname</replaceable>.
+			Will succeed if the global variable exists or not.</para>
+		</description>
+		<see-also>
+			<ref type="function">GLOBAL</ref>
+			<ref type="function">VARIABLE_DELETE</ref>
```
InterLinked1, line 71: This reference also needs to be updated to `DELETE`

InterLinked1 (changes requested):

InterLinked1 (approved):

seanbright:
Tests?

jcolp:
> Tests?

I looked at doing test coverage but the problem is that there isn't a difference between a variable existing and not existing when it comes to actually using variables so checking whether it worked in a test mechanism. I guess I could just assume that if an empty string comes back it was deleted.

seanbright:
> > Tests?
> 
> I looked at doing test coverage but the problem is that there isn't a difference between a variable existing and not existing when it comes to actually using variables so checking whether it worked in a test mechanism isn't precise. I guess I could just assume that if an empty string comes back it was deleted.

Could create a "variable exists" function if TEST_FRAMEWORK is defined and use that from a test? Could just make a "variable exists" function as part of this PR? Could do neither of those things?

There is already an `EXISTS()` function, but that seems to return false for the empty string which I don't think would help here.

jcolp:
> > > Tests?
> > 
> > 
> > I looked at doing test coverage but the problem is that there isn't a difference between a variable existing and not existing when it comes to actually using variables so checking whether it worked in a test mechanism isn't precise. I guess I could just assume that if an empty string comes back it was deleted.
> 
> Could create a "variable exists" function if TEST_FRAMEWORK is defined and use that from a test? Could just make a "variable exists" function as part of this PR? Could do neither of those things?
> 
> There is already an `EXISTS()` function, but that seems to return false for the empty string which I don't think would help here.

I thought the functionality may be generally useful so didn't hide it behind TEST_FRAMEWORK

gtjoseph (approved):

PR #292, opened 2023-09-02
func_groupcount.c: Adding Group Variables + additional Group functions

### LINKED ISSUE #291: [new-feature]: Group Variables (For real now)
### Feature Description

    func_groupcount.c: Adding Group Variables + additional Group functions

    DumpGroups
    -------------------
    * New application.  This will dump all channel group membership and associated
      variables

    Groups and Group Variables
    ------------------
     * Group variables can be set on a group once the group is created
       When a group is destroyed, all variables on that group are also destroyed

       A group variable is somewhat like a global variable, but it's on a per-group
       basis.

       GroupSet - Adds functionality to the manager to be able to set a GROUP()
                  on a channel.
       GroupsShowChannels - Show each channel and it's associated groups
                            (a channel will be repeated for each group@category
                             it's a member of)
       GroupsShowVariables - Show variables in each group@category, one event per
                             group, all variables are contained in each
                             group@category event
       GroupVarSet - Set a group variable (the group must already exist)
       GroupVarGet - Get a group variable

     * New Manager events:
       ------------------
       GroupCreate - Event is fired any time a group is made,
                     ie: Set(GROUP=x) or Set(GROUP()=x@y).
                     This event is only sent on when a channel is added to a group

intellasoft:
DumpGroups
-------------------
* New application.  This will dump all channel group membership and associated
  variables

Groups and Group Variables
------------------
 * Group variables can be set on a group once the group is created
   When a group is destroyed, all variables on that group are also destroyed

   A group variable is somewhat like a global variable, but it's on a per-group
   basis.

   GroupSet - Adds functionality to the manager to be able to set a GROUP()
	      on a channel.
   GroupsShowChannels - Show each channel and it's associated groups
                        (a channel will be repeated for each group@category
                         it's a member of)
   GroupsShowVariables - Show variables in each group@category, one event per
                         group, all variables are contained in each
                         group@category event
   GroupVarSet - Set a group variable (the group must already exist)
   GroupVarGet - Get a group variable

 * New Manager events:
   ------------------
   GroupCreate - Event is fired any time a group is made,
                 ie: Set(GROUP=x) or Set(GROUP()=x@y).
                 This event is only sent on when a channel is added to a group
                 that did not exist previously.
   GroupChannelAdd - Event is fired any time a channel is added to a group
   GroupChannelRemove - Event is fired any time a channel is removed from a
                        group
   GroupDestroy - Event is fired when there are no longer any channels assigned
                  to the group
   GroupVarSet - Event is fired when any group variable is changed

 * New CLI Command
   ---------------
   group show variables

 * New Application
   ---------------
   DumpGroups() - Show groups and group assigments (similar to DumpChan)

Resolves: #291

### DIFF
--- a/funcs/func_groupcount.c
+++ b/funcs/func_groupcount.c
@@ -27,11 +27,16 @@
 
 #include "asterisk.h"
 
+#include <regex.h>
+
 #include "asterisk/module.h"
 #include "asterisk/channel.h"
 #include "asterisk/pbx.h"
 #include "asterisk/utils.h"
 #include "asterisk/app.h"
+#include "asterisk/manager.h"
+#include "asterisk/strings.h"
+#include "asterisk/cli.h"
 
 /*** DOCUMENTATION
 	<function name="GROUP_COUNT" language="en_US">
@@ -54,6 +59,233 @@
 			channel's current group if not specified (and non-empty).</para>
 		</description>
 	</function>
+	<function name="GROUP_MATCH_LIST" language="en_US">
+		<synopsis>Find groups by regular expression</synopsis>
+		<syntax argsep="@">
+			<parameter name="group_regex"/>
+			<parameter name="category_regex"/>
+		</syntax>
+		<description>
+			<para>Search for groups matching group@category. This search will look at all known groups and filter by the regex(s) provided.</para>
+			<para>If only the group_regex is specified, then return group@category results that only match the group name against the group_regex</para>
+			<para>If only the category_regex is specified, then return group@category results that only match the category name against the category_regex.</para>
+			<para>If the search regex parameters are entirely empty, all group@category will be returned.</para>
+			<para>Uses standard regular expression matching (see regex(7)).</para>
+			<example title="Find groups containing the string 'foo'">
+				On Channel 1:
+					Set(GROUP()=groupName)
+				On Channel 2:
+					Set(GROUP()=foobarbaz)
+				On Channel 3:
+					Set(GROUP()=foobarbill)
+				On Channel 4:
+					Set(GROUP()=somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(groups_found=${GROUP_LIST(.*foo.*)})
+				Given the above:
+					This will find the groups 'foobarbaz' and 'foobarbill' since they both match .*foo.* regex against the group name
+                                        Variable groups_found will be set to: foobarbaz,foobarbill
+			</example>
+			<example title="Find groups containing category name of 'bar'">
+				On Channel 1:
+					Set(GROUP()=groupName@categoryName)
+				On Channel 2:
+					Set(GROUP()=foo@barbaz)
+				On Channel 3:
+					Set(GROUP()=foo@barbill)
+				On Channel 4:
+					Set(GROUP()=anything@somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(groups_found=${GROUP_LIST(@.*bar.*)})
+
+				Given the above:
+					This will find the groups 'foo@barbaz' and 'foo@barbill' since they both match .*bar.* regex against the category name
+                                        Variable groups_found will be set to: foo@barbaz,foo@barbill
+			</example>
+			<example title="Find groups containing 'foo' and category name of 'bar'">
+				On Channel 1:
+					Set(GROUP()=groupName@categoryName)
+				On Channel 2:
+					Set(GROUP()=foo@barbaz)
+				On Channel 3:
+					Set(GROUP()=foo@barbill)
+				On Channel 4:
+					Set(GROUP()=anything@somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(groups_found=${GROUP_LIST(.*foo.*@.*bar.*)})
+
+				Given the above:
+					This will find the groups 'foo@barbaz' and 'foo@barbill' since they both match .*foo*. against the group name and .*bar.* regex against the category name
+                                        Variable groups_found will be set to: foo@barbaz,foo@barbill
+			</example>
+		</description>
+		<see-also>
+			<ref type="function">GROUP</ref>
+			<ref type="function">GROUP_CHANNEL_LIST</ref>
+		</see-also>
+	</function>
+	<function name="GROUP_CHANNEL_LIST" language="en_US">
+		<synopsis>
+			Find channels that are members of groups/categories by exact match
+		</synopsis>
+		<syntax argsep="@">
+			<parameter name="group"/>
+			<parameter name="category"/>
+		</syntax>
+		<description>
+			<para>Search for groups matching group@category and return the channels that are members of those groups, the group/category must match exactly</para>
+			<para>If only the group is specified, then return group@category results that only match against the group name</para>
+			<para>If only the category is specified, then return group@category results that only match the against the category name</para>
+			<para>If the search parameters are entirely empty, all channels having any kind of group membership will be returned.</para>
+			<para>After GROUP_CHANNEL_LIST is complete.  Channel variables may be set with information related to the function call.</para>
+			<variablelist>
+				<variable name="GROUP_CHANNEL_LIST_ERROR">
+					<value name="TRUNCATED">
+						The full result is not complete and was truncated
+					</value>
+				</variable>
+			</variablelist>
+			<example>
+				Find channels that are members of the group 'foobarbaz':
+
+				On Channel PJSIP/1234-0000001:
+					Set(GROUP()=groupName)
+				On Channel PJSIP/1234-0000003:
+					Set(GROUP()=foobarbaz)
+				On Channel PJSIP/1234-0000004:
+					Set(GROUP()=foobarbill)
+				On Channel PJSIP/1234-0000005:
+					Set(GROUP()=somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(channels_found=${GROUP_CHANNEL_LIST(foobarbaz)}
+				Given the above:
+					This will find the channels that are in the group 'foobarbaz'
+                                        Variable channels_found will be set to: PJSIP/1234-0000003
+			</example>
+			<example>
+				Find channels that have the category name of 'barbaz'
+
+				On Channel PJSIP/1234-0000001:
+					Set(GROUP()=groupName@categoryName)
+				On Channel PJSIP/1234-0000002:
+					Set(GROUP()=foo@barbaz)
+                                On Channel PJSIP/1234-0000003:
+					Set(GROUP()=oof@barbaz)
+                                On Channel PJSIP/1234-0000004:
+					Set(GROUP()=anything@somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(channels_found=${GROUP_CHANNEL_LIST(@barbaz)})
+
+				Given the above:
+					This will find the channels that are in group 'foo@foobarbaz' and 'oof@barbaz' since they both have the category barbaz
+                                        Variable channels_found will be set to: PJSIP/1234-0000002,PJSIP/1234-0000003
+			</example>
+			<example>
+				Find groups containing 'foo' and category name of 'bar'
+
+                                On Channel PJSIP/1234-0000001:
+					Set(GROUP()=groupName@categoryName)
+                                On Channel PJSIP/1234-0000002:
+					Set(GROUP()=foo@bar)
+                                On Channel PJSIP/1234-0000003:
+					Set(GROUP()=foo@bar)
+                                On Channel PJSIP/1234-0000004:
+					Set(GROUP()=anything@somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(channels_found=${GROUP_MATCH_CHANNEL_LIST(.*foo.*@.*bar.*)})
+
+				Given the above:
+					This will find the channels that are in the group@category assignment of foo@bar
+                                        Variable channels_found will be set to: PJSIP/1234-0000002,PJSIP/1234-0000003
+			</example>
+		</description>
+		<see-also>
+			<ref type="function">GROUP</ref>
+			<ref type="function">GROUP_MATCH_CHANNEL_LIST</ref>
+		</see-also>
+	</function>
+	<function name="GROUP_MATCH_CHANNEL_LIST" language="en_US">
+		<synopsis>
+			Find channels that are members of groups by regular expression
+		</synopsis>
+		<syntax argsep="@">
+			<parameter name="group_regex"/>
+			<parameter name="category_regex"/>
+		</syntax>
+		<description>
+			<para>Search for groups matching group@category and return the channels that are members of those groups, filtering known groups by the regex(s) provided.</para>
+			<para>If only the group_regex is specified, then return group@category results that only match the group name against the group_regex</para>
+			<para>If only the category_regex is specified, then return group@category results that only match the category name against the category_regex.</para>
+			<para>If the search regex parameters are entirely empty, all channels having any kind of group membership will be returned.</para>
+			<para>Uses standard regular expression matching (see regex(7)).</para>
+			<example>
+				Find channels containing the string 'foo'
+
+				On Channel PJSIP/1234-0000001:
+					Set(GROUP()=groupName)
+				On Channel PJSIP/1234-0000003:
+					Set(GROUP()=foobarbaz)
+				On Channel PJSIP/1234-0000004:
+					Set(GROUP()=foobarbill)
+				On Channel PJSIP/1234-0000005:
+					Set(GROUP()=somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(channels_found=${GROUP_MATCH_CHANNEL_LIST(.*foo.*)})
+				Given the above:
+					This will find the channels that are in groups 'foobarbaz' and 'foobarbill' since they both match .*foo.* regex against the group name
+                                        Variable channels_found will be set to: PJSIP/1234-0000003,PJSIP/1234-0000004
+			</example>
+			<example>
+				Find channels in groups containing category name of 'bar'
+
+				On Channel PJSIP/1234-0000001:
+					Set(GROUP()=groupName@categoryName)
+				On Channel PJSIP/1234-0000002:
+					Set(GROUP()=foo@barbaz)
+                                On Channel PJSIP/1234-0000003:
+					Set(GROUP()=foo@barbill)
+                                On Channel PJSIP/1234-0000004:
+					Set(GROUP()=anything@somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(channels_found=${GROUP_MATCH_CHANNEL_LIST(@.*bar.*)})
+
+				Given the above:
+					This will find the channels that are in groups 'foo@barbaz' and 'foo@barbill' since they both match .*bar.* regex against the category name
+                                        Variable channels_found will be set to: PJSIP/1234-0000002,PJSIP/1234-0000003
+			</example>
+			<example>
+				Find channels in groups where they contain 'foo' and have the matching category contain the name of 'bar'
+
+                                On Channel PJSIP/1234-0000001:
+					Set(GROUP()=groupName@categoryName)
+                                On Channel PJSIP/1234-0000002:
+					Set(GROUP()=foo@bar)
+                                On Channel PJSIP/1234-0000003:
+					Set(GROUP()=foo@barbill)
+                                On Channel PJSIP/1234-0000004:
+					Set(GROUP()=anything@somethingelse)
+				On any Channel:
+					Assuming that channels still exist that have the above associated GROUP() assignents
+					Set(channels_found=${GROUP_MATCH_CHANNEL_LIST(.*foo.*@.*bar.*)}
+
+				Given the above:
+					This will find the channels that are in groups 'foo@barbaz' and 'foo@barbill' since they both match .*foo*. against the group name and .*bar.* regex against the category name
+                                        Variable channels_found will be set to: PJSIP/1234-0000002,PJSIP/1234-0000003
+			</example>
+		</description>
+		<see-also>
+			<ref type="function">GROUP</ref>
+			<ref type="function">GROUP_CHANNEL_LIST</ref>
+		</see-also>
+	</function>
 	<function name="GROUP_MATCH_COUNT" language="en_US">
 		<since>
 			<version>1.2.0</version>
@@ -62,17 +294,17 @@
 			Counts the number of channels in the groups matching the specified pattern.
 		</synopsis>
 		<syntax argsep="@">
-			<parameter name="groupmatch" required="true">
+			<parameter name="group_match">
 				<para>A standard regular expression used to match a group name.</para>
 			</parameter>
-			<parameter name="category">
+			<parameter name="category_match">
 				<para>A standard regular expression used to match a category name.</para>
 			</parameter>
 		</syntax>
 		<description>
 			<para>Calculates the group count for all groups that match the specified pattern.
 			Note: category matching is applied after matching based on group.
-			Uses standard regular expression matching on both (see regex(7)).</para>
+			Uses standard regular expression matching on both parameters (see regex(7)).</para>
 		</description>
 	</function>
 	<function name="GROUP" language="en_US">
@@ -82,15 +314,61 @@
 		<synopsis>
 			Gets or sets the channel group.
 		</synopsis>
-		<syntax>
+		<syntax argsep="@">
+			<parameter name="group">
+				<para>Group name.</para>
+			</parameter>
 			<parameter name="category">
 				<para>Category name.</para>
 			</parameter>
 		</syntax>
 		<description>
 			<para><replaceable>category</replaceable> can be employed for more fine grained group management. Each channel
-			can only be member of exactly one group per <replaceable>category</replaceable>.</para>
+			can only be member of exactly one group per <replaceable>category</replaceable>.  Once a group is assigned to
+			a channel, per-group variables can then be assigned via GROUP_VAR()</para>
 		</description>
+		<see-also>
+			<ref type="function">GROUP_VAR</ref>
+		</see-also>
+	</function>
+	<function name="GROUP_VAR" language="en_US">
+		<synopsis>
+			Gets or sets a variable on a channel group.
+		</synopsis>
+		<syntax>
+			<parameter name="category">
+				<para>Category name.</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>
+				Once a GROUP() is assinged to a channel, variables can then be attached to that group.	As long as the
+				group exists, these variables will exist. When the group no longer exists, the variables will be cleaned up
+				automatically.
+			</para>
+			<para>
+				These variables can be considered 'globals for a group of channels'.  These variables are
+				like a SHARED() variable but for a collection of channels.
+			</para>
+			<example>
+				On Channel 1
+				Set(GROUP()=foo)
+				Set(GROUP_VAR(foo,savethis)=123)
+
+				On Channel 2 -- Assuming Channel1 is still up and Channel1 is still a member of GROUP(foo)
+                                We can now join the same Group and read a Group Variable
+				Set(GROUP()=foo)
+				NoOp(${GROUP_VAR(foo,savethis) &lt;-- This prints '123'
+
+			        On Channel 3 -- Assuming Channel1 is still up and Channel1 is still a member of GROUP(foo)
+                                We do not need to join the group in order to read a Group Variable
+				NoOp(${GROUP_VAR(foo,savethis) &lt;-- This prints '123'
+			</example>
+		</description>
+		<see-also>
+			<ref type="function">GROUP</ref>
+			<ref type="function">SHARED</ref>
+		</see-also>
 	</function>
 	<function name="GROUP_LIST" language="en_US">
 		<since>
@@ -104,15 +382,179 @@
 			<para>Gets a list of the groups set on a channel.</para>
 		</description>
 	</function>
+	<manager name="GroupSet" language="en_US">
+		<synopsis>
+			Add channel group assignments
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="channel">
+				<para>Channel to operate on.</para>
+			</parameter>
+			<parameter name="group">
+				<para>Group name to set.</para>
+			</parameter>
+			<parameter name="category">
+				<para>Category name to set.</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>For more information, see the dialplan function GROUP()</para>
+		</description>
+	</manager>
+	<manager name="GroupRemove" language="en_US">
+		<synopsis>
+			Remove channel group assignments
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="channel">
+				<para>Channel to operate on.</para>
+			</parameter>
+			<parameter name="group">
+				<para>Group name to remove.</para>
+			</parameter>
+			<parameter name="category">
+				<para>Category name to remove.</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>For more information, see the dialplan function GROUP()</para>
+		</description>
+	</manager>
+	<application name="DumpGroups" language="en_US">
+		<synopsis>
+			Dump all group information to the console
+		</synopsis>
+		<description>
+			<para>When executed, this will show all group assignments and group variables
+			in the console</para>
+		</description>
+	</application>
+	<manager name="GroupVarGet" language="en_US">
+		<synopsis>
+			Get channel group variables.
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="Group" required="false" />
+			<parameter name="Category" required="false" />
+			<parameter name="Variable" required="true" />
+		</syntax>
+		<description>
+			<para>
+				At a minimum, either group or category must be provided.
+				For more information, see the dialplan function GROUP_VAR().
+			</para>
+		</description>
+	</manager>
+	<manager name="GroupVarSet" language="en_US">
+		<synopsis>
+			Set channel group variables.
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="Group" required="false" />
+			<parameter name="Category" required="false" />
+			<parameter name="Variable" required="true" />
+			<parameter name="Value" required="false" />
+		</syntax>
+		<description>
+			<para>
+				At a minimum, either group or category must be provided.
+				For more information, see the dialplan function GROUP_VAR().
+			</para>
+		</description>
+	</manager>
+	<manager name="GroupsShow" language="en_US">
+		<synopsis>
+			Show channel groups.  With optional filtering
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="Group" required="false">
+				<para>The exact group name to find, or to use a regular expression: set this parameter to: /regex/</para>
+			</parameter>
+			<parameter name="Category" required="false">
+				<para>The exact category to find, or to use a regular expression: set this parameter to: /regex/</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>
+				This will return a list of channel groups that are in use.
+				For more information, see the dialplan function GROUP().
+			</para>
+		</description>
+	</manager>
+	<manager name="GroupsShowChannels" language="en_US">
+		<synopsis>
+			Show group channel assignments. With optional filtering
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="Group" required="false">
+				<para>The exact group name to find, or to use a regular expression: set this parameter to: /regex/</para>
+			</parameter>
+			<parameter name="Category" required="false">
+				<para>The exact category to find, or to use a regular expression: set this parameter to: /regex/</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>
+				This will return a list the channels that are within each group.
+				For more information, see the dialplan function GROUP().
+			</para>
+		</description>
+	</manager>
+	<manager name="GroupsShowVariables" language="en_US">
+		<synopsis>
+			Show group channel variable assignments.
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+		</syntax>
+		<description>
+			<para>
+				This will return a list of groups and the variables assigned in each group.
+				For more information, see the dialplan function GROUP_VAR().
+			</para>
+		</description>
+	</manager>
+	<managerEvent language="en_US" name="GroupVarGetResponse">
+		<managerEventInstance class="EVENT_FLAG_REPORTING">
+			<synopsis>
+				Raised in response to a a GroupVarGet command
+			</synopsis>
+			<syntax>
+				<parameter name="ActionID">
+					<para>ActionID (if any) that was passed into the GroupVarGet request</para>
+				</parameter>
+				<parameter name="Group">
+					<para>Name of the group that the variable is a part of</para>
+				</parameter>
+				<parameter name="Category">
+					<para>Name of the category that the variable is a part of</para>
+				</parameter>
+				<parameter name="Variable">
+					<para>The variable that was requested</para>
+				</parameter>
+				<parameter name="Value">
+					<para>The value of the variable</para>
+				</parameter>
+			</syntax>
+		</managerEventInstance>
+	</managerEvent>
 
  ***/
+
+static const char *dumpgroups_app = "DumpGroups";
 
 static int group_count_function_read(struct ast_channel *chan, const char *cmd,
 				     char *data, char *buf, size_t len)
 {
 	int ret = -1;
 	int count = -1;
-	char group[80] = "", category[80] = "";
+	char group[MAX_GROUP_LEN] = "", category[MAX_CATEGORY_LEN] = "";
 
 	if (!chan) {
 		ast_log(LOG_WARNING, "No channel was provided to %s function.\n", cmd);
@@ -128,15 +570,18 @@ static int group_count_function_read(struct ast_channel *chan, const char *cmd,
 
 		ast_app_group_list_rdlock();
 		for (gi = ast_app_group_list_head(); gi; gi = AST_LIST_NEXT(gi, group_list)) {
-			if (gi->chan != chan)
+			if (gi->chan != chan) {
 				continue;
-			if (ast_strlen_zero(category) || (!ast_strlen_zero(gi->category) && !strcasecmp(gi->category, category)))
+			}
+			if (ast_strlen_zero(category) || (!ast_strlen_zero(gi->category) && !strcasecmp(gi->category, category))) {
 				break;
+			}
 		}
 		if (gi) {
 			ast_copy_string(group, gi->group, sizeof(group));
-			if (!ast_strlen_zero(gi->category))
+			if (!ast_strlen_zero(gi->category)) {
 				ast_copy_string(category, gi->category, sizeof(category));
+			}
 		}
 		ast_app_group_list_unlock();
 	}
@@ -161,8 +606,8 @@ static int group_match_count_function_read(struct ast_channel *chan,
 					   const char *cmd, char *data, char *buf,
 					   size_t len)
 {
-	char group[80] = "";
-	char category[80] = "";
+	char group[MAX_GROUP_LEN] = "";
+	char category[MAX_CATEGORY_LEN] = "";
 
 	ast_app_group_split_group(data, group, sizeof(group), category,
 				  sizeof(category));
@@ -198,12 +643,15 @@ static int group_function_read(struct ast_channel *chan, const char *cmd,
 	ast_app_group_list_rdlock();
 
 	for (gi = ast_app_group_list_head(); gi; gi = AST_LIST_NEXT(gi, group_list)) {
-		if (gi->chan != chan)
+		if (gi->chan != chan) {
 			continue;
-		if (ast_strlen_zero(data))
+		}
+		if (ast_strlen_zero(data)) {
 			break;
-		if (!ast_strlen_zero(gi->category) && !strcasecmp(gi->category, data))
+		}
+		if (!ast_strlen_zero(gi->category) && !strcasecmp(gi->category, data)) {
 			break;
+		}
 	}
 
 	if (gi) {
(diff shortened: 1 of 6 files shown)


doc/CHANGES-staging/func_groupcount_group_variables.txt
```diff
@@ -0,0 +1,47 @@
+Subject: func_groupcount
```
InterLinked1, line 1: The CHANGES file is no longer needed. This should go in UserNote: in the commit message (and probably be condensed, usage explanations should not be in the commit message, but in the XML documentation).

funcs/func_groupcount.c
```diff
@@ -51,37 +56,200 @@
...
+			<para>It is possible to only specify @category_regex, and it is also possible to pass in nothing for the search.</para>
+			<para>If only the category_regex is specified, then it will only return group@category that matches this category search.</para>
+			<para>If the search regex is entirely empty, all groups will be returned.</para>
+			<para>The return value is the number of groups found. Calling GROUP_MATCH_LIST_NEXT() more times than there
+			are matches will result in an empty value being returned.</para>
+			<para>This function is useful to use instead of GROUP_MATCH_LIST if you would prefer to process the results one by one.
+			If your group/category names contain a comma, then this is better to avoid parse issues</para>
+			<para>GROUP_CHANNEL_LIST does not support regex, but GROUP_MATCH_LIST_START does</para>
+			<para>Note: the search is executed and stored at the time of calling this function.
+			If GROUP() assignments change during successive calls to the _NEXT function, you will need to call _START again
```
InterLinked1, line 77: Instead of _NEXT and _START, use the full names.
I still don't think this function is necessary, anyone have any other thoughts?

funcs/func_groupcount.c
```diff
@@ -51,37 +56,200 @@
...
+		<description>
+			<para>Once GROUP_MATCH_LIST_START is executed with a search, then each call to GROUP_MATCH_LIST_NEXT()
+			will return a matched group.  This search will look at all known groups and filter by the regex provided.</para>
+			<para>It is possible to only specify @category_regex, and it is also possible to pass in nothing for the search.</para>
+			<para>If only the category_regex is specified, then it will only return group@category that matches this category search.</para>
+			<para>If the search regex is entirely empty, all groups will be returned.</para>
+			<para>The return value is the number of groups found. Calling GROUP_MATCH_LIST_NEXT() more times than there
+			are matches will result in an empty value being returned.</para>
+			<para>This function is useful to use instead of GROUP_MATCH_LIST if you would prefer to process the results one by one.
+			If your group/category names contain a comma, then this is better to avoid parse issues</para>
```
InterLinked1, line 74: *parsing issues*

funcs/func_groupcount.c
```diff
@@ -237,6 +611,539 @@ static struct ast_custom_function group_function = {
...
+
+	if (!group) {
+		group = "";
+	}
+
+	if (!category) {
+		category = "";
+	}
+
+	group_category = ast_malloc(strlen(group) + strlen(category) + 2); /* 1 for @, 1 for NULL*/
```
InterLinked1, line 710: You need to check for malloc failure

main/cli.c
```diff
@@ -1894,6 +1894,18 @@ char *ast_complete_channels(const char *line, const char *word, int pos, int sta
 	return ret;
 }
 
+int ast_app_group_all_callback(ao2_callback_fn *callback, int flags);
+
+struct group_list_entry {
+	struct ast_str *group;                         /*!< A group assignment is group@category, this is the group part  */
+	struct ast_str *category;                      /*!< This is the category part */
+	int transfer_fixup:1;                          /*!< Whether we will run a fixup on masquerade */
```
InterLinked1, line 1902: Is this still needed? It's not used anywhere so should be removed.

main/cli.c
```diff
@@ -1948,6 +1960,62 @@ static char *group_show_channels(struct ast_cli_entry *e, int cmd, struct ast_cl
...
+		e->usage =
+			"Usage: group show variables [group@[category]]\n"
+			"       Lists all currently active groups and their variables.\n"
+			"       Optional group, or group@category can be used to only show a specific group\n";
+		return NULL;
+	case CLI_GENERATE:
+		return NULL;
+	}
+
+	if (a->argc < 3 || a->argc > 4)
```
InterLinked1, line 1987: Braces needed

InterLinked1 (changes requested):

InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

funcs/func_groupcount.c
kobaz: My thoughts are redoing this to return a hash object.

InterLinked1: But what does this permit users to do, that cannot already be accomplished using existing dialplan applications/functions?
I don't really think individual constructs all need their own methods of iterating over them, maybe we should improve the existing ones (separately) in that case.

funcs/func_groupcount.c
```diff
@@ -51,37 +56,200 @@
...
+			<para>It is possible to only specify @category_regex, and it is also possible to pass in nothing for the search.</para>
+			<para>If only the category_regex is specified, then it will only return group@category that matches this category search.</para>
+			<para>If the search regex is entirely empty, all groups will be returned.</para>
+			<para>The return value is the number of groups found. Calling GROUP_MATCH_LIST_NEXT() more times than there
+			are matches will result in an empty value being returned.</para>
+			<para>Note: the search is executed and stored at the time of calling this function.
+			If GROUP() assignments change during successive calls to the _NEXT function, you will need to call _START again
+			to see the new groups. An empty string is returned to indicate the end of the list.</para>
+			<para>Uses standard regular expression matching (see regex(7)).</para>
+			<example>Find groups containing 'foo': ${GROUP_MATCH_LIST_START(.*foo.*)}</example>
```
InterLinked1, line 167: I'm pretty sure I raised this issue in the Gerrit review but this has not been fixed. These are not valid examples and this documentation is malformed. There should be titles, e.g.

```
<example title="foo bar">
some example
</example>
```

```
PHP Fatal error:  Uncaught ErrorException: Undefined array key "title" in /usr/src/asterisk-20.4.0/astdocgen.php:431
Stack trace:
#0 /usr/src/asterisk-20.4.0/astdocgen.php(431): exception_error_handler()
#1 {main}
  thrown in /usr/src/asterisk-20.4.0/astdocgen.php on line 431
```

InterLinked1 (changes requested):

funcs/func_groupcount.c
kobaz: Thanks... this is getting rewritten anyway as mentioned previously.  The new version of this will return a hash.

The biggest issue with the current return types from most functions is they are not data-agnostic.  For example if you return a comma delimited list and your group names have commas, it's not possible to parse it properly.  The problem is that in dialplan, data arguments are not isolated from each other like in a traditional language.  So you would have this sort of issue no matter what delimiter you used (if the delimiter was contained within the data)

InterLinked1: Okay, fair enough, I think the only problems were in this part of the documentation. I just manually deleted the `<example>` lines in the core-doc file so I could build my docs.

I don't use commas or other possible delimiters in my group names so I guess I haven't seen this kind of issue... usually they are something pretty simple like [unique identifier]@[some property of relevance]. At this point most of my group usage is actually in C, not in dialplan, typically GROUP_COUNT or GROUP_VARs now :)

I think you should put out a feeler on the -dev list about this, as I think there should be a more generic mechanism, if at all, that is not specific to groups. To me it feels like if somebody needs this type of thing, it would be best done in a C module where you can directly iterate over what you want in a structured programming language with a proper iterator, rather than trying to deal with parsing to begin with.

funcs/func_groupcount.c
```diff
@@ -273,21 +1169,229 @@ static int group_list_function_read(struct ast_channel *chan, const char *cmd,
...
+	struct ast_str *foundgroup_str = ast_str_create(MAX_GROUP_LEN + MAX_CATEGORY_LEN);
+
+	buf[0] = '\0';
+
+	if (!foundgroup_str) {
+		ast_log(LOG_ERROR, "Unable to allocate new variable.  Cannot proceed with GROUP_MATCH_LIST()\n");
+		return -1;
+	}
+
+	char groupmatch[MAX_GROUP_LEN] = "";
```
InterLinked1, line 1185: This code *still* does not compile properly... *please* make sure it compiles in developer mode before pushing for review.

```
func_groupcount.c: In function 'group_match_list_function_read':
func_groupcount.c:1185:9: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
 1185 |         char groupmatch[MAX_GROUP_LEN] = "";
```

If you can fix this quickly, that'd be appreciated, that way I don't need to write a temporary diff to fix compilation.

InterLinked1 (changes requested):
See inline comment:

funcs/func_groupcount.c
kobaz: Fixed!

InterLinked1: It still doesn't seem to be fixed when I look at the latest version.

InterLinked1: > It still doesn't seem to be fixed when I look at the latest version.

@gtjoseph Also, I think this came up before, but I remember there were previous changes the CI didn't catch this same compiler issue with (mixed declarations and code), and then it caused problems on when it was merged and compiled out in the wild.

The strange thing is it's actually causing compilation failures with my GitHub CI, so I'm not sure why the Asterisk one isn't getting tripped. Perhaps it's the gcc version? Just wanted to get your thoughts.

kobaz:
Whoops.  I just realized my push was the same code.  I forgot to 'git add'.  This one should be all set.

InterLinked1:
Incidentally, what happened to `GROUP_CHANNEL_LIST`? That doesn't seem to be available on my system anymore.
I see references to it in the documentation but the function itself doesn't seem to exist.

kobaz:
> Incidentally, what happened to `GROUP_CHANNEL_LIST`? That doesn't seem to be available on my system anymore. I see references to it in the documentation but the function itself doesn't seem to exist.

It got renamed
GROUP_MATCH_CHANNEL_LIST

to be consistent with
GROUP_MATCH_LIST

InterLinked1:
> > Incidentally, what happened to `GROUP_CHANNEL_LIST`? That doesn't seem to be available on my system anymore. I see references to it in the documentation but the function itself doesn't seem to exist.
> 
> It got renamed GROUP_MATCH_CHANNEL_LIST
> 
> to be consistent with GROUP_MATCH_LIST
> 
> I'll update the docs

I'm confused. The docs say it takes a regular expression, but in the code it doesn't do any such thing. It just returns the full list (which is what I want). The `MATCH` bit to me feels confusing.

kobaz:
Hmm... Something got lost in the redo.  I'll get the regex matching back in.  Thanks.

InterLinked1:
> Hmm... Something got lost in the redo. I'll get the regex matching back in. Thanks.

That's my point though, I don't *want* the regex matching to begin with. I like the current implementation, but the name doesn't agree with it.

Maybe you can keep the MATCH one with regex if you really want, but restore the plain GROUP_CHANNEL_LIST that does no matching, as it is currently?

kobaz:
> > Hmm... Something got lost in the redo. I'll get the regex matching back in. Thanks.
> 
> That's my point though, I don't _want_ the regex matching to begin with. I like the current implementation, but the name doesn't agree with it.
> 
> Maybe you can keep the MATCH one with regex if you really want, but restore the plain GROUP_CHANNEL_LIST that does no matching, as it is currently?

Got it!

InterLinked1:
> > > Hmm... Something got lost in the redo. I'll get the regex matching back in. Thanks.
> > 
> > 
> > That's my point though, I don't _want_ the regex matching to begin with. I like the current implementation, but the name doesn't agree with it.
> > Maybe you can keep the MATCH one with regex if you really want, but restore the plain GROUP_CHANNEL_LIST that does no matching, as it is currently?
> 
> Got it!

If you can do it in just one function without a performance overhead (like trying to parse regex), that might be fine, too.

Also, GROUP_MATCH_CHANNEL_LIST still adds a trailing comma, so might want to clean that up while you're looking into all this.

Side note, do you mind putting your hangup callbacks patch up again when you get a moment?

main/app.c
```diff
@@ -2227,6 +2349,111 @@ int ast_app_group_set_channel(struct ast_channel *chan, const char *data)
...
+
+		AST_LIST_INSERT_HEAD(headp, newvariable, entries);
+		manager_event(EVENT_FLAG_DIALPLAN, "GroupVarSet",
+			"Channel: %s\r\n"
+			"Category: %s\r\n"
+			"Group: %s\r\n"
+			"Variable: %s\r\n"
+			"Value: %s\r\n"
+			"Uniqueid: %s\r\n",
+			chan ? ast_channel_name(chan) : "none",
```
InterLinked1, line 2400: This could be simplified using S_OR, same with two lines below

main/cli.c
```diff
@@ -1948,6 +1949,63 @@ static char *group_show_channels(struct ast_cli_entry *e, int cmd, struct ast_cl
...
+		return CLI_SHOWUSAGE;
+	}
+
+	if (a->argc == 4) {
+		if (ast_app_group_split_group(a->argv[3], group, sizeof(group), category, sizeof(category))) {
+			return NULL;
+		}
+	}
+
+	ast_cli(a->fd, "Group   Category\n   Variables\n");
```
InterLinked1, line 1986: The title format does not line up width-wise with the contents. Make Category line up with the categories.
Variables makes sense if there are any variables but otherwise it might be confusing.

```
ess*CLI> group show variables
Group   Category
   Variables
PJSIP/gxw9            deviceinuse
1                     lineinuse
1                     oe
     master=PJSIP/gxw9-0000000c
1                     oeall
1                     cdpr
5 active groups
```

funcs/func_groupcount.c
```diff
@@ -51,37 +56,307 @@
...
+		</syntax>
+		<description>
+			<para>Search for groups matching group@category and return the channels that are members of those groups, the group/category must match exactly</para>
+			<para>If only the group is specified, then return group@category results that only match against the group name</para>
+			<para>If only the categoryis specified, then return group@category results that only match the against the category name</para>
+			<para>If the search parameters are entirely empty, all channels having any kind of group membership will be returned.</para>
+			<example>
+				Find channels that are members of the group 'foobarbaz':
+
+				On Channel SIP/1234-0000001:
```
InterLinked1, line 148: Don't use the SIP technology for an example, using DAHDI, PJSIP, IAX2, or some core supported driver.

funcs/func_groupcount.c
```diff
@@ -51,37 +56,307 @@
...
+			Find channels that are members of groups/categories by exact match
+		</synopsis>
+		<syntax argsep="@">
+			<parameter name="group"/>
+			<parameter name="category"/>
+		</syntax>
+		<description>
+			<para>Search for groups matching group@category and return the channels that are members of those groups, the group/category must match exactly</para>
+			<para>If only the group is specified, then return group@category results that only match against the group name</para>
+			<para>If only the categoryis specified, then return group@category results that only match the against the category name</para>
```
InterLinked1, line 143: category is

funcs/func_groupcount.c
```diff
@@ -51,37 +56,307 @@
...
+					Set(channels_found=${GROUP_MATCH_CHANNEL_LIST(.*foo.*@.*bar.*)})
+
+				Given the above:
+					This will find the channels that are in the group@category assignment of foo@bar
+                                        Variable channels_found will be set to: SIP/1234-0000002,SIP/1234-0000003
+			</example>
+		</description>
+		<see-also>
+			<ref type="function">GROUP</ref>
+			<ref type="function">GROUP_MATCH_LIST</ref>
```
InterLinked1, line 204: This should be GROUP_MATCH_CHANNEL_LIST, not GROUP_MATCH_LIST, no, these are the similar ones? Or at least list both of them otherwise.

funcs/func_groupcount.c
InterLinked1: You're still not checking for malloc failure here.
Either way, don't use ast_malloc, use ast_asprintf.

funcs/func_groupcount.c
```diff
@@ -237,6 +676,539 @@ static struct ast_custom_function group_function = {
...
+		astman_send_error(s, m, "No group specified.");
+		return 0;
+	}
+
+	if (ast_strlen_zero(variable)) {
+		astman_send_error(s, m, "No variable specified.");
+		return 0;
+	}
+
+	if (!ast_strlen_zero(id))
```
InterLinked1, line 824: need braces

InterLinked1 (changes requested):

InterLinked1 (changes requested):
Another minor whitespace issue flagged by patch:

Applying PR 292 unconditionally
/tmp/292.pr.diff:314: space before tab in indent.
                                On Channel 1

gtjoseph:
closing and re-opening just to kick off the proper tests.

funcs/func_groupcount.c
kobaz: Fixed

funcs/func_groupcount.c
kobaz: Fixed

funcs/func_groupcount.c
kobaz: Fixed

funcs/func_groupcount.c
kobaz: Fixed

main/cli.c
kobaz: Fixed

main/app.c
kobaz: I think this is just syntactic sugar here.  There's no simplification between chan ? ... and S_SOR(x,y,z)

gtjoseph:
@kobaz @InterLinked1 Can you guys resolve all the review comments so I know where this PR stands?

funcs/func_groupcount.c
```diff
@@ -273,21 +1266,287 @@ static int group_list_function_read(struct ast_channel *chan, const char *cmd,
...
+
+	if (ast_app_group_split_group(args.groupcategory, group, sizeof(group), category, sizeof(category))) {
+		return -1;
+	}
+
+	ast_app_group_list_rdlock();
+	gi = ast_app_group_list_head();
+	while (gi) {
+		if (!strcasecmp(gi->group, group) && (ast_strlen_zero(category) || (!ast_strlen_zero(gi->category) && !strcasecmp(gi->category, category)))) {
+			ast_str_append(&out, 0, "%s,", ast_channel_name(gi->chan));
```
InterLinked1, line 1436: No comma after last one

InterLinked1 (changes requested):

funcs/func_groupcount.c
```diff
@@ -273,21 +1266,287 @@ static int group_list_function_read(struct ast_channel *chan, const char *cmd,
...
+	ast_app_group_list_unlock();
+
+	out_len = strlen(ast_str_buffer(out));
+
+	if (out_len) {
+		if (out_len <= len) {
+		  len -= 1; /* rid , only if we didn't go over the buffer limit */
+		}
+
+		strncpy(buf, ast_str_buffer(out), len);
```
InterLinked1, line 1450: Don't use strncpy, use ast_copy_string.
But don't use ast_copy_string, we already have a buffer, use that directly

InterLinked1 (changes requested):

funcs/func_groupcount.c
kobaz: I think it's just nitpicking here on this comment, for no functional gain.  It sounds like more 'c purist' than actually making any actual improvements in the code itself.

Yes... we do have a buffer to work with, but as usual 'the work has to be done somewhere'.  So either keep track of the write position of *buf with pointer math, or use something that *already* has smarts of position.

Current version:
- build a string, copy it into buf (ever-so-slightly wasteful of a buffer)

What you are suggesting:
- Write a more delicate method to accomplish the same thing... using a buffer directly with... what exactly?  sprintf() or strncat.  Now we need to keep track of where we are and not overwrite the end

new code would be something like...
loop
  strncat, with max of (length - written-so-far).. 
  bump the write pointer for the next strncat
  pos+= additional length written

  and then repeated calls to strncat will of course need the length of what's being copied... 

My vote is go with safe-and-effective string utilities to do what they are designed to do, rather than hand-code delicate string/pointer code to build a string for something that I don't personally really think needs that kind of level of optimization.  And considering we need to keep track of lengths written _anyway_ , there's literally no computational advantage of doing it 'by hand' versus using strings.h because internally, ast_str keeps track of length for you.

main/cli.c
```diff
@@ -1948,6 +1949,64 @@ static char *group_show_channels(struct ast_cli_entry *e, int cmd, struct ast_cl
...
+		}
+	}
+
+	ast_cli(a->fd, "Group\n    Category\nVariables\n");
+	ast_cli(a->fd, "------------------------------\n");
+
+	/* Print group variables */
+	ast_app_group_meta_rdlock();
+	gmi = ast_app_group_meta_head();
+	while (gmi) {
```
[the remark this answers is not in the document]
intellasoft: I think the suggested move would violate historic layout.  No other component implements CLI functions in app.c and no other component implements any kind of command in app.c, (no cli, no dialplan applications, no dialplan functions)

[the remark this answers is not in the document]
kobaz: I think the suggested move would violate historic layout. No other component implements CLI functions in app.c and no other component implements any kind of command in app.c, (no cli, no dialplan applications, no dialplan functions)

funcs/func_groupcount.c
InterLinked1: This one is not nitpicking, there is a big performance difference.
Don't use strncat, that's also a wasteful function. Just use snprintf each time, it's not that complicated.

```
len = snprintf(pos, left, "%s", foobar);
pos += len;
left -= len;
```

snprintf already returns bytes copied for free (unless truncation occurs).

The advantage is that `ast_str` involves allocation, so if you are going to do that, either use the ast_str version as Sean said or do a zero-copy version into the buffer already provided.

main/cli.c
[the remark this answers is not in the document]
InterLinked1: Okay, fair enough - I don't really like the ast_app_group_meta functions existing - is there another way we could get rid of them? If not, I guess it's fine.

funcs/func_groupcount.c
```diff
@@ -1433,21 +1433,34 @@ static int group_channel_list_function_read(struct ast_channel *chan, const char
...
-	out_len = strlen(ast_str_buffer(out));
-
 	if (out_len) {
 		if (out_len <= len) {
-		  len -= 1; /* rid , only if we didn't go over the buffer limit */
+			out_len -= 1; /* rid , only if we didn't go over the buffer limit */
 		}
 
-		strncpy(buf, ast_str_buffer(out), len);
+		strncpy(buf, ast_str_buffer(out), out_len); /* We already check for overrun above, so we'll never have > len here */
```
InterLinked1, line 1463: Do not use strncpy, use ast_copy_string!

InterLinked1 (changes requested):

funcs/func_groupcount.c
kobaz: Ah okay.  Makes sense. No problem.

main/cli.c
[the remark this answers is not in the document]
kobaz: There's already public ast_app_group_list_wrlock/ast_app_group_list_rdlock that's existed for many years.  But also, just because it's there, doesn't mean it's a good design.

Another approach would be implement an iterator, but then you still want an obvious way to grab the lock... I don't thing the locking should be 'hidden' / embedded in the api, because just like channels, you still have to explicitly grab a channel lock and then do your thing to the channel.  And then implementing all this iterator stuff  sounds like re-inventing ao2 to me.

So speaking of ao2... this would require a bit of an overhaul to convert to ao2.   If anything it would add overhead without much in terms of functional gain considering groups and group variables are fairly simple.

funcs/func_groupcount.c
```diff
@@ -1442,29 +1443,23 @@ static int group_channel_list_function_read(struct ast_channel *chan, const char
 				break;
```
InterLinked1, line 1443: I think there should be a warning emitted if all the channels don't fit into the buffer, rather than silently truncating.
Other than that, looks good!

InterLinked1 (changes requested):

funcs/func_groupcount.c
kobaz: Thanks!

This is why I was thinking about having support for a structured data return like a hash.  We should be able to put an 'unlimited' amount of things into a hash in theory.

But of course, HASHKEYS is going to have the same limit that any other dialplan return is going to have, which is the fixed-size of buf

InterLinked1: I don't actually think it's likely to be an issue... nobody is going to have 1,000 channels in a group... however, just to cover all bases, there should be a warning in case something weird happens, is all I care about.

kobaz: This is why I previously implemented these functions as iterator-style where you have a START/NEXT/END and then you can keep asking for the next one until you run out.  Thereby avoiding any kind of buf-size limit... but you didn't like that style of functions because nothing else in asterisk-dialplan behaves like this.  With > 400 channels (or just really long channel names like Local/ channels can be)... you can easily run into hitting the upper limit on the size of buf

InterLinked1: Right, again, I'm saying that I don't think it's a real issue - all I want is a warning in the case that should never be hit anyways :)

kobaz: I think adding HASHKEYS_ITERATOR and adding an option to all the things that return a list to be able to also return a HASH would solve this issue

InterLinked1: Look, my issue is not that this might be truncated. I don't have any use case for retrieving so many channels the buffer would be exhausted. I just think there needs to be a warning there, that's all.

Please don't over-complicate this... just add a warning so I can approve this PR

kobaz: Yup, I'll add in a warning and play with adding a HASHKEYS_ITERATOR.. it's something i've wanted anyway (separate project)

kobaz:
```
    -- Executing [3333@services:1] Set("Local/3333@services-000002bb;2", "GROUP()=foo") in new stack
    -- Executing [3333@services:2] NoOp("Local/3333@services-000002bb;2", "groups=GROUP_CHANNEL_LIST(foo)") in new stack
[2024-03-18 11:21:17.074-0400] WARNING[943161][C-000004b0]: func_groupcount.c:1445 group_channel_list_function_read: GROUP_CHANNEL_LIST(foo) Channel membership list too large for buffer, truncating.
    -- Executing [3333@services:3] Set("Local/3333@services-000002bb;2", "groups=Local/3333@services-000001f4;2,Local/3333@services-000001f5;2,Local/3333@services-000001f6;2,Local/3333@services-000001f7;2,Local/3333@services-000001f8;2,Local/3333@services-000001f9;2,Local/3333@services-000001fa;2,Local/3333@services-000001fb;2,Local/3333@services-000001fc;2,Local/3333@services-000001fd;2,Local/3333@services-000001fe;2,Local/3333@services-000001ff;2,Local/3333@services-00000200;2,Local/3333@services-00000201;2,Local/3333@services-00000202;2,Local/3333@services-00000203;2,Local/3333@services-00000204;2,Local/3333@services-00000205;2,Local/3333@services-00000206;2,Local/3333@services-00000207;2,Local/3333@services-00000208;2,Local/3333@services-00000209;2,Local/3333@services-0000020a;2,Local/3333@services-0000020b;2,Local/3333@services-0000020c;2,Local/3333@services-0000020d;2,Local/3333@services-0000020e;2,Local/3333@services-0000020f;2,Local/3333@services-00000210;2,Local/3333@services-00000211;2,Local/3333@services-00000212;2,Local/3333@services-00000213;2,Local/3333@services-00000214;2,Local/3333@services-00000215;2,Local/3333@services-00000216;2,Local/3333@services-00000217;2,Local/3333@services-00000218;2,Local/3333@services-00000219;2,Local/3333@services-0000021a;2,Local/3333@services-0000021b;2,Local/3333@services-0000021c;2,Local/3333@services-0000021d;2,Local/3333@services-0000021e;2,Local/3333@services-0000021f;2,Local/3333@services-00000220;2,Local/3333@services-00000221;2,Local/3333@services-00000222;2,Local/3333@services-00000223;2,Local/3333@services-00000224;2,Local/3333@services-00000225;2,Local/3333@services-00000226;2,Local/3333@services-00000227;2,Local/3333@services-00000228;2,Local/3333@services-00000229;2,Local/3333@services-0000022a;2,Local/3333@services-0000022b;2,Local/3333@services-0000022c;2,Local/3333@services-0000022d;2,Local/3333@services-0000022e;2,Local/3333@services-0000022f;2,Local/3333@services-00000230;2,Local/3333@services-00000231;2,Local/3333@services-00000232;2,Local/3333@services-00000233;2,Local/3333@services-00000234;2,Local/3333@services-00000235;2,Local/3333@services-00000236;2,Local/3333@services-00000237;2,Local/3333@services-00000238;2,Local/3333@services-00000239;2,Local/3333@services-0000023a;2,Local/3333@services-0000023b;2,Local/3333@services-0000023c;2,Local/3333@services-0000023d;2,Local/3333@services-0000023e;2,Local/3333@services-0000023f;2,Local/3333@services-00000240;2,Local/3333@services-00000241;2,Local/3333@services-00000242;2,Local/3333@services-00000243;2,Local/3333@services-00000244;2,Local/3333@services-00000245;2,Local/3333@services-00000246;2,Local/3333@services-00000247;2,Local/3333@services-00000248;2,Local/3333@services-00000249;2,Local/3333@services-0000024a;2,Local/3333@services-0000024b;2,Local/3333@services-0000024c;2,Local/3333@services-0000024d;2,Local/3333@services-0000024e;2,Local/3333@services-0000024f;2,Local/3333@services-00000250;2,Local/3333@services-00000251;2,Local/3333@services-00000252;2,Local/3333@services-00000253;2,Local/3333@services-00000254;2,Local/3333@services-00000255;2,Local/3333@services-00000256;2,Local/3333@services-00000257;2,Local/3333@services-00000258;2,Local/3333@services-00000259;2,Local/3333@services-0000025a;2,Local/3333@services-0000025b;2,Local/3333@services-0000025c;2,Local/3333@services-0000025d;2,Local/3333@services-0000025e;2,Local/3333@services-0000025f;2,Local/3333@services-00000260;2,Local/3333@services-00000261;2,Local/3333@services-00000262;2,Local/3333@services-00000263;2,Local/3333@services-00000264;2,Local/3333@services-00000265;2,Local/3333@services-00000266;2,Local/3333@services-00000267;2,Local/3333@services-00000268;2,Local/3333@services-00000269;2,Local/3333@services-0000026a;2,Local/3333@services-0000026b;2,Local/3333@services-0000026c;2,Local/3333@services-0000026d;2,Local/3333@services-0000026e;2,Local/3333@services-0000026f;2,Local/3333@services-00000270;2,Local/3333@services-00000271;2,Local/3333@services-00000272;2,Local/3333@services-00000273;2,Local/3333@services-00000274;2,Local/3333@services-00000275;2,Local/3333@services-00000276;2,Local/3333@services-00000277;2") in new stack
    -- Executing [3333@services:4] NoOp("Local/3333@services-000002bb;2", "GROUP_CHANNEL_LIST_ERROR=TRUNCATED") in new stack

```

funcs/func_groupcount.c
```diff
@@ -142,6 +142,8 @@
 			<para>If only the group is specified, then return group@category results that only match against the group name</para>
 			<para>If only the category is specified, then return group@category results that only match the against the category name</para>
 			<para>If the search parameters are entirely empty, all channels having any kind of group membership will be returned.</para>
+			<para>Upon error, the channel variable GROUP_CHANNEL_LIST_ERROR will be set</para>
+			<para>Possible GROUP_CHANNEL_LIST_ERROR: TRUNCATED</para>
```
InterLinked1, line 146: This should be an enumlist,

funcs/func_groupcount.c
```diff
@@ -1440,6 +1442,8 @@ static int group_channel_list_function_read(struct ast_channel *chan, const char
 			int channel_name_len = strlen(ast_channel_name(gi->chan));
 
 			if ((out_len + channel_name_len + 1) > len) { /* +1 for null */
+				ast_log(LOG_WARNING, "GROUP_CHANNEL_LIST(%s) Channel membership list too large for buffer, truncating.\n", args.groupcategory);
+				pbx_builtin_setvar_helper(chan, "GROUP_CHANNEL_LIST_ERROR", "TRUNCATED");
```
InterLinked1, line 1446: delete the variable at the beginning of the func to prevent stale values

InterLinked1 (changes requested):

kobaz:
multiple-commits: standalone

InterLinked1 (approved):

main/app.c
```diff
@@ -2306,16 +2541,219 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
 	return 0;
 }
 
+/* Remove a channel from a group meta assignment */
+/* Right now this just removes all the group metadata for a group@category if this is the last channel in the group@category */
+/* Ideally we would have direct pointers from the channel group assignments into the metadata struct, so we don't have to search */
+int ast_app_group_remove_channel(struct ast_channel *chan, char *group, char *category)
```
bkford, line 2547: The name of this function is misleading since it doesn't actually remove the channel, it just does some extra logic tied to channel removal. I would either add the channel removal logic into this function, or rename it.

main/app.c
```diff
@@ -2306,16 +2541,219 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+			"Category: %s\r\n"
+			"Group: %s\r\n",
+			category, group);
+	}
+
+	return 1;
+}
+
+/* Add a channel to a group meta assignment, create a group meta item if it doesn't exist */
+int ast_app_group_add_channel(struct ast_channel *chan, char *group, char *category)
```
bkford, line 2600: Same as ast_app_group_remove_channel. I would rename or refactor.

main/app.c
```diff
@@ -2306,16 +2541,219 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+		ast_app_group_remove_channel(gi->chan, gi->group, gi->category);
+		AST_RWLIST_REMOVE_CURRENT(group_list);
+		ast_free(gi);
+
+		channels_found++;
+	}
+	AST_RWLIST_TRAVERSE_SAFE_END;
+	AST_RWLIST_UNLOCK(&groups);
+
+	return !(channels_found > 0);
```
bkford, line 2676: Is there a reason this is returning a value other than the number of channels removed? `group` should also be null checked and bail early if it is with a warning.

main/app.c
```diff
@@ -2306,16 +2541,219 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+}
+
+int ast_app_group_rename(const char *old_group, const char *old_category, const char *new_group, const char *new_category)
+{
+	struct ast_group_info *gi = NULL;
+	struct ast_group_meta *gmi = NULL; /*!< Group metadatas	  */
+
```
bkford, line 2683: `old_group` and `new_group` should be null checked here as well.

main/app.c
```diff
@@ -2306,16 +2541,219 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+	AST_RWLIST_WRLOCK(&groups_meta);
+	AST_RWLIST_TRAVERSE_SAFE_BEGIN(&groups_meta, gmi, group_meta_list) {
+		if (strcasecmp(gmi->group, old_group) || strcasecmp(gmi->category, old_category)) {
+			/* Need to match group@category exactly */
+			continue;
+		}
+
+		strncpy(gmi->group,    new_group,    MAX_GROUP_LEN    - 1);
+		strncpy(gmi->category, new_category, MAX_CATEGORY_LEN - 1);
+		channels_found++;
```
bkford, line 2720: This shouldn't increment the channel count since it's for a group, not a channel.

main/app.c
```diff
@@ -2306,16 +2541,219 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+			"OldCategory: %s\r\n"
+			"NewGroup: %s\r\n"
+			"NewCategory: %s\r\n",
+			old_group,
+			old_category,
+			new_group,
+			new_category);
+	}
+
+	return !(channels_found > 0);
```
bkford, line 2737: I think this function should return either success or failure in different places depending on:
1. Is `new_group` null?
2. Does the group@category we were looking for exist?

Maybe a warning message or debug message would be nice if it doesn't exist.

bkford (dismissed):

asteriskteam:

main/app.c
```diff
@@ -2227,6 +2349,111 @@ int ast_app_group_set_channel(struct ast_channel *chan, const char *data)
...
+
+		headp = &gmi->varshead;
+
+		AST_LIST_TRAVERSE(headp, ast_var, entries) {
+			variable = ast_var_name(ast_var);
+
+			if (!strcasecmp(variable, name)) {
+				/* found it */
+				AST_RWLIST_UNLOCK(&groups_meta);
+				return ast_var_value(ast_var);
```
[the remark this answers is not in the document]
kobaz: Fixed

main/app.c
```diff
@@ -2227,6 +2349,111 @@ int ast_app_group_set_channel(struct ast_channel *chan, const char *data)
...
+			continue;
+		}
+
+		headp = &gmi->varshead;
+
+		AST_LIST_TRAVERSE(headp, newvariable, entries) {
+			if (strcasecmp(ast_var_name(newvariable), name) == 0) {
+				/* there is already such a variable, delete it */
+				AST_LIST_REMOVE(headp, newvariable, entries);
+				ast_var_delete(newvariable);
```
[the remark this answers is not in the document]
kobaz: Fixed

main/app.c
kobaz: Fixed

main/app.c
kobaz: Fixed... Makes sense to return total found, if the caller wants something, rather than returning less information.

main/app.c
kobaz: Fixed

main/app.c
kobaz: Fixed

main/app.c
kobaz: Fixed

asteriskteam:

intellasoft:
Hey all.  Why is this marked stale?  I made the requested changes...

intellasoft:
Okay hmm.  Looks like there were buttons for marking the change as completed.  Not sure why they didn't show up previously.  But it's all submitted now.

include/asterisk/app.h
```diff
@@ -1196,9 +1196,54 @@ struct ast_group_info;
 /*! \brief Split a group string into group and category, returning a default category if none is provided. */
 int ast_app_group_split_group(const char *data, char *group, int group_max, char *category, int category_max);
 
+/* Create group meta if it doesn't exist  */
```
InterLinked1, line 1199: Missing doxygen syntax (`\brief`)

include/asterisk/app.h
```diff
@@ -1196,9 +1196,54 @@ struct ast_group_info;
...
+int ast_app_group_add_meta_ifneeded(const char *group, const char *category);
+
+/* This removes the the group metadata for a group@category if there are no more channels in the group@category */
```
InterLinked1, line 1202: doxygen + rephrase to imperative ("Remove the group metadata...")

include/asterisk/app.h
```diff
@@ -1196,9 +1196,54 @@ struct ast_group_info;
...
+int ast_app_group_remove_meta_ifneeded(const char *group, const char *category);
+
+/*! \brief Remove channel assignments for the specified group  */
+int ast_app_group_remove_all_channels(const char *group, const char *category);
+
+/*! \brief Rename a group@category while retaining all the channel memberships
+ * \retval -1 on Error
+ * \retval 0 On zero rename operations performed (nothing found to rename)
+ * \retval >=1 On 1 or more rename operations performed
```
InterLinked1, line 1211: This one might be clearer as `\return Number of rename operations performed`. It would also cover the 0 case

include/asterisk/app.h
```diff
@@ -1196,9 +1196,54 @@ struct ast_group_info;
...
+ * \param group     group to get the variable from
+ * \param category  category to get the variable from
+ * \param name      name of the variable to get
+ *
+ * \retval NOT NULL On success return  char* of variable contents
+ * \retval NULL On failure (variable in group@category not found)
+
+ * \post caller must ast_free() the returned value
+ */
+char* ast_app_group_get_var(const char *group, const char *category, const char *name);
```
InterLinked1, line 1245: The `*` should be on the function name

include/asterisk/app.h
```diff
@@ -1196,9 +1196,54 @@ struct ast_group_info;
...
+int ast_app_group_set_var(struct ast_channel *chan, const char *group, const char *category, const char *name, const char *value);
+
+/*!
+ * \brief Get a group variable for a group@category
+ *
+ * \param group     group to get the variable from
+ * \param category  category to get the variable from
+ * \param name      name of the variable to get
+ *
+ * \retval NOT NULL On success return  char* of variable contents
```
InterLinked1, line 1240: This one could just be `\return Allocated string containing variable contents`

include/asterisk/chanvars.h
```diff
@@ -33,27 +33,158 @@ struct ast_var_t {
 
 AST_LIST_HEAD_NOLOCK(varshead, ast_var_t);
 
+
+/*!
+ * \brief Create a new variables list and initalize to empty
+ *
+ * \retval NULL on error
+ * \retval returns the pointer to the newly created list on success
```
InterLinked1, line 41: "returns the" is not needed
Do a find and replace on the rest of the file too for "\retval returns "

include/asterisk/chanvars.h
```diff
@@ -33,27 +33,158 @@ struct ast_var_t {
...
+ * \param value     value of the variable to set
+ *
+ * \retval NULL on error
+ * \retval returns a populated struct ast_var_t
+ */
 #define ast_var_assign(name, value) _ast_var_assign(name, value, __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
+/*!
+ * \brief Replace the value of an existing variable
+ * \pre Caller assumes all responsibility for handling locking
```
InterLinked1, line 68: Locking of what? The variable list? The channel?

include/asterisk/chanvars.h
```diff
@@ -33,27 +33,158 @@ struct ast_var_t {
 
 AST_LIST_HEAD_NOLOCK(varshead, ast_var_t);
 
+
```
InterLinked1, line 36: The additions of these variable functions doesn't appear related to group vars at all... maybe pull that out as a separate improvement change?

main/app.c
```diff
@@ -2180,51 +2296,244 @@ int ast_app_group_split_group(const char *data, char *group, int group_max, char
...
+
+        AST_RWLIST_UNLOCK(&groups);
+
+        return res;
+}
+
+int ast_app_group_set_var(struct ast_channel *chan, const char *group, const char *category, const char *name, const char *value)
+{
+	struct ast_group_meta *gmi = NULL;
+
```
InterLinked1, line 2418: why the empty line?

main/app.c
```diff
@@ -2180,51 +2296,244 @@ int ast_app_group_split_group(const char *data, char *group, int group_max, char
...
 
-	AST_RWLIST_UNLOCK(&groups);
+	return 0;
+}
 
-	return res;
+char *ast_app_group_get_var(const char *group, const char *category, const char *name)
+{
+	struct ast_group_meta *gmi = NULL;
+
```
InterLinked1, line 2488: empty line

main/app.c
```diff
@@ -2306,16 +2623,194 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+	AST_RWLIST_UNLOCK(&groups_meta);
+
+	return 1;
+}
+
+/* Remove all channels from the given group */
+int ast_app_group_remove_all_channels(const char *group, const char *category)
+{
+	struct ast_group_info *gi = NULL;
+
```
InterLinked1, line 2669: empty line

main/app.c
```diff
@@ -2306,16 +2623,194 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+	int group_meta_found = 0;
+
+	if (!group) {
+		ast_log(LOG_WARNING, "Group Remove: At the minimum, group must be non-NULL\n");
```
InterLinked1, line 2673: If this is for the user, this should be rephrased to something like a group must be provided rather than non-NULL, since the dialplan doesn't have the concept of NULL

main/app.c
```diff
@@ -2306,16 +2623,194 @@ int ast_app_group_update(struct ast_channel *old, struct ast_channel *new)
...
+	}
+
+	if (!old_category) {
+		old_category = "";
+	}
+
+	if (!new_category) {
+		new_category = "";
+	}
+
```
InterLinked1, line 2735: I think you could remove the empty lines between the preceding blocks and eliminate this empty line

main/chanvars.c
```diff
@@ -52,6 +52,18 @@ struct ast_var_t *_ast_var_assign(const char *name, const char *value, const cha
 	return var;
 }
 
+int ast_var_set(const struct ast_var_t *var, const char *new_value)
+{
+
```
InterLinked1, line 57: empty line / same comment re variable changes

InterLinked1 (changes requested):

asteriskteam:

include/asterisk/app.h
intellasoft: Fixed

include/asterisk/app.h
intellasoft: Fixed

include/asterisk/app.h
intellasoft: Fixed

include/asterisk/app.h
intellasoft: Fixed

include/asterisk/app.h
intellasoft: Fixed

main/app.c
intellasoft: Fixed

main/app.c
intellasoft: This error is for a developer.  ast_app_group_remove_all_channels would never be called from user-code with NULL

The only usage of this is via AMI manager_group_remove
```
static int manager_group_remove(struct mansession *s, const struct message *m)
{
        const char *group    = astman_get_header(m, "Group");
        const char *category = astman_get_header(m, "Category");

        if (ast_app_group_remove_all_channels(group, category) == 0) {
```



group/category will never be null here.

main/app.c
intellasoft: Empty line removed, but.. I think this is just a syle-preference at this point.

Is there anything in the coding style guidelines that requires blocks to be smushed together?  My eyes prefer them split apart!

main/chanvars.c
intellasoft: Fixed

main/app.c
intellasoft: Fixed

include/asterisk/app.h
```diff
@@ -1196,9 +1196,54 @@ struct ast_group_info;
 /*! \brief Split a group string into group and category, returning a default category if none is provided. */
 int ast_app_group_split_group(const char *data, char *group, int group_max, char *category, int category_max);
 
+/* \brief Create group meta if it doesn't exist  */
+int ast_app_group_add_meta_ifneeded(const char *group, const char *category);
+
+/* Removes the the group metadata for a group@category if there are no more channels in the group@category */
```
InterLinked1, line 1202: Should have `\brief`

include/asterisk/channel.h
```diff
@@ -2919,6 +2924,19 @@ struct ast_group_info {
 	AST_LIST_ENTRY(ast_group_info) group_list;
 };
 
+/*! \brief list of groups currently in use, with a pointer to a list of channels within the groups
+  */
```
InterLinked1, line 2928: The comment close could just be at the end of the preceding line here

include/asterisk/channel.h
```diff
@@ -3133,6 +3151,35 @@ struct ast_channel *ast_channel_get_by_name_prefix(const char *name, size_t name
...
+ * \brief Find a channel by a regex string
+ *
+ * \arg regex_string the regex pattern to search for
+ *
+ * Return a channel that where the regex pattern matches the channel name
+ *
+ * \retval a channel that matches the regex pattern
+ * \retval NULL if no channel was found or pattern is bad
+ *
+ * \since 18
```
InterLinked1, line 3164: This isn't going into 18 at this point.

include/asterisk/channel.h
```diff
@@ -3133,6 +3151,35 @@ struct ast_channel *ast_channel_get_by_name_prefix(const char *name, size_t name
...
+ * \brief Find a channel by a regex pattern
+ *
+ * \arg regex the compiled regex pattern to search for
+ *
+ * Return a channel that where the regex pattern matches the channel name
+ *
+ * \retval a channel that matches the regex pattern
+ * \retval NULL if no channel was found
+ *
+ * \since 18
```
InterLinked1, line 3178: Ditto

main/app.c
```diff
@@ -2180,51 +2296,243 @@ int ast_app_group_split_group(const char *data, char *group, int group_max, char
...
-	struct ast_group_info *gi = NULL;
-	size_t len = 0;
+	struct ast_group_meta *gmi = NULL; /*!< Group metadatas	  */
 
-	if (ast_app_group_split_group(data, group, sizeof(group), category, sizeof(category))) {
-		return -1;
+	struct varshead *headp;
+	struct ast_var_t *vardata;
+
+	int destroy = 0;
```
InterLinked1, line 2307: Technically, this variable is not needed, gmi being non-NULL after the loop provides the same guarantee

include/asterisk/app.h
```diff
@@ -1196,9 +1196,54 @@ struct ast_group_info;
 /*! \brief Split a group string into group and category, returning a default category if none is provided. */
 int ast_app_group_split_group(const char *data, char *group, int group_max, char *category, int category_max);
 
+/* \brief Create group meta if it doesn't exist  */
+int ast_app_group_add_meta_ifneeded(const char *group, const char *category);
```
InterLinked1, line 1200: Does this need to be public? It seems to only be used in app.c?

main/cli.c
InterLinked1, line 1992: This while loop might be a little bit more elegant as a for

main/chanvars.c
```diff
@@ -52,6 +52,17 @@ struct ast_var_t *_ast_var_assign(const char *name, const char *value, const cha
 	return var;
 }
 
+int ast_var_set(const struct ast_var_t *var, const char *new_value)
+{
+	if (var->value) {
+		ast_free(var->value);
+	}
+
```
InterLinked1, line 60: This doesn't seem right... you're potentially freeing `var->value` so it's no longer valid, then trying to copy a string to that location?

InterLinked1 (changes requested):

main/chanvars.c
intellasoft: Moot now. ast_var_set() has been removed

include/asterisk/app.h
intellasoft: Fixed

include/asterisk/channel.h
intellasoft: Fixed

include/asterisk/channel.h
intellasoft: Fixed

include/asterisk/channel.h
intellasoft: Fixed

main/app.c
intellasoft: Fixed

include/asterisk/app.h
intellasoft: Fixed, now static.

funcs/func_groupcount.c
```diff
@@ -51,37 +56,315 @@
...
+			<parameter name="group_regex"/>
+			<parameter name="category_regex"/>
+		</syntax>
+		<description>
+			<para>Search for groups matching group@category. This search will look at all known groups and filter by the regex(s) provided.</para>
+			<para>If only the group_regex is specified, then return group@category results that only match the group name against the group_regex</para>
+			<para>If only the category_regex is specified, then return group@category results that only match the category name against the category_regex.</para>
+			<para>If the search regex parameters are entirely empty, all group@category will be returned.</para>
+			<para>Uses standard regular expression matching (see regex(7)).</para>
+			<example>
```
InterLinked1, line 71: The example tag should have a title attribute, move "Find groups containing the string foo" to the title

funcs/func_groupcount.c
```diff
@@ -51,37 +56,315 @@
...
+			<parameter name="group_regex"/>
+			<parameter name="category_regex"/>
+		</syntax>
+		<description>
+			<para>Search for groups matching group@category and return the channels that are members of those groups, filtering known groups by the regex(s) provided.</para>
+			<para>If only the group_regex is specified, then return group@category results that only match the group name against the group_regex</para>
+			<para>If only the category_regex is specified, then return group@category results that only match the category name against the category_regex.</para>
+			<para>If the search regex parameters are entirely empty, all channels having any kind of group membership will be returned.</para>
+			<para>Uses standard regular expression matching (see regex(7)).</para>
+			<example>
```
InterLinked1, line 229: Move the following line to title attribute in example tag

funcs/func_groupcount.c
```diff
@@ -51,37 +56,315 @@
...
+			<para>
+				Once a GROUP() is assinged to a channel, variables can then be attached to that group.	As long as the
+				group exists, these variables will exist. When the group no longer exists, the variables will be cleaned up
+				automatically.
+			</para>
+			<para>
+				These variables can be considered 'globals for a group of channels'.  These variables are
+				like a SHARED() variable but for a collection of channels.
+			</para>
+			<example>
```
InterLinked1, line 349: Add a title attribute to the example tag

InterLinked1 (changes requested):

asteriskteam:

funcs/func_groupcount.c
intellasoft: Fixed

funcs/func_groupcount.c
intellasoft: Fixed

funcs/func_groupcount.c
intellasoft: Fixed

main/chanvars.c
```diff
@@ -52,6 +52,13 @@ struct ast_var_t *_ast_var_assign(const char *name, const char *value, const cha
 	return var;
 }
 
+int ast_var_set(const struct ast_var_t *var, const char *new_value)
+{
+	ast_copy_string(var->value, new_value, strlen(new_value));
```
seanbright, line 57: The 3rd argument to ast_copy_string should be the size of the destination buffer.

InterLinked1 (dismissed):
This patch no longer applies for me, I guess it's changed enough it no longer patches out of tree nicely.

In any case, you should address these warnings, one of which has been present for a long time:

```
/tmp/292.pr.diff:317: space before tab in indent.
 			        On Channel 1
/tmp/292.pr.diff:2373: trailing whitespace.
                  No groups_found++ here, since we already found it earlier. 
error: patch failed: funcs/func_groupcount.c:51
error: funcs/func_groupcount.c: patch does not apply
```

asteriskteam:

asteriskteam:

intellasoft:
Have you tried the rebase?  If it's not patching against your private branch, then I would think that's more of an issue with your branch being out of sync with master?

> This patch no longer applies for me, I guess it's changed enough it no longer patches out of tree nicely.
>

InterLinked1:
> Have you tried the rebase? If it's not patching against your private branch, then I would think that's more of an issue with your branch being out of sync with master?
> 
> > This patch no longer applies for me, I guess it's changed enough it no longer patches out of tree nicely.

I'm not using my own branch forked from master. It could have just been the patch got stale at the time, I hadn't tried a manual rebase, that might have worked.

I have one system where I am still using group vars which I am leaving alone for this and other reasons, but otherwise I'm going to wait until this patch is merged instead of patching it in to other installs at the moment.

intellasoft:
> > Have you tried the rebase? If it's not patching against your private branch, then I would think that's more of an issue with your branch being out of sync with master?
> > > This patch no longer applies for me, I guess it's changed enough it no longer patches out of tree nicely.
> 
> I'm not using my own branch forked from master. It could have just been the patch got stale at the time, I hadn't tried a manual rebase, that might have worked.
> 
> I have one system where I am still using group vars which I am leaving alone for this and other reasons, but otherwise I'm going to wait until this patch is merged instead of patching it in to other installs at the moment.

You're going to want to patch in the latest after my next push.  This fixes a possible memory overwrite/overrun when replacing a group variable with a new value.

main/chanvars.c
intellasoft: Fixed in ca26d3b93e673e419e5c36a3dd41b214e049a9b5

asteriskteam:

kobaz:
All changes have been completed.

InterLinked1 (changes requested):
> All changes have been completed.

Your patch doesn't apply - it needs to be rebased (GitHub is also indicating there's a merge conflict):

```
/tmp/292.pr.diff:385: space before tab in indent.
                                On Channel 1
/tmp/292.pr.diff:3030: trailing whitespace.
                  No groups_found++ here, since we already found it earlier.
/tmp/292.pr.diff:3082: trailing whitespace.

/tmp/292.pr.diff:3519: trailing whitespace.

error: patch failed: include/asterisk/channel.h:3179
error: include/asterisk/channel.h: patch does not apply
error: patch failed: include/asterisk/channel.h:2970
error: include/asterisk/channel.h: patch does not apply
```

InterLinked1 (changes requested):
It applies now, but there are still warnings:

```
/tmp/292.pr.diff:381: space before tab in indent.
                                On Channel 1
/tmp/292.pr.diff:2172: trailing whitespace.

/tmp/292.pr.diff:2423: trailing whitespace.
                  No groups_found++ here, since we already found it earlier.
warning: 3 lines add whitespace errors.
```

intellasoft:
> It applies now, but there are still warnings:
> 
> ```
> /tmp/292.pr.diff:381: space before tab in indent.
>                                 On Channel 1
> /tmp/292.pr.diff:2172: trailing whitespace.
> 
> /tmp/292.pr.diff:2423: trailing whitespace.
>                   No groups_found++ here, since we already found it earlier.
> warning: 3 lines add whitespace errors.
> ```

How are you testing/patching?

InterLinked1:
> > It applies now, but there are still warnings:
> > ```
> > /tmp/292.pr.diff:381: space before tab in indent.
> >                                 On Channel 1
> > /tmp/292.pr.diff:2172: trailing whitespace.
> > 
> > /tmp/292.pr.diff:2423: trailing whitespace.
> >                   No groups_found++ here, since we already found it earlier.
> > warning: 3 lines add whitespace errors.
> > ```
> 
> How are you testing/patching?

In a build, I'm downloading and applying this PR as a patch. I'm sure you'll see those warnings if you try it yourself.

intellasoft:
Hi Interlinked,

I'm still not seeing what your seeing.

If I rebase my commit on master as of
  7c52e1404223ecdf7a5863a1d8511def1a9f4bc6 Wed Feb 18 10:04:13 2026 -0500
  
  It rebases clean, no warnings on configure or compile
  
  If I git checkout 7c52e1404223ecdf7a5863a1d8511def1a9f4bc6, and apply my commit as a patch, it's also clean.

Can you list out your steps to reproduce the issue?

Also, I've rebased and pushed.

intellasoft:
Also, @All 

Is there anything actively blocking this from merge any more?

Thanks!

jcolp:
What you mean is "is there anything preventing this from being reviewed?" and I don't think so, just noone has done so.

intellasoft:
Thanks... Just making sure nothing is waiting on me.

InterLinked1:
> Hi Interlinked,
> 
> I'm still not seeing what your seeing.
> 
> If I rebase my commit on master as of [7c52e14](https://github.com/asterisk/asterisk/commit/7c52e1404223ecdf7a5863a1d8511def1a9f4bc6) Wed Feb 18 10:04:13 2026 -0500
> 
> It rebases clean, no warnings on configure or compile
> 
> If I git checkout [7c52e14](https://github.com/asterisk/asterisk/commit/7c52e1404223ecdf7a5863a1d8511def1a9f4bc6), and apply my commit as a patch, it's also clean.
> 
> Can you list out your steps to reproduce the issue?
> 
> Also, I've rebased and pushed.

Funny you mention this, I just installed a new system yesterday with the group vars patch, and noticed the same errors at the time.

Unless you addressed them since yesterday, the issue still remains.

There aren't any compiler warnings. The warnings are from `git apply` itself, warning about trailing whitespace. Look at what I posted and you should be able to find the lines and rectify the issue.

InterLinked1:
> Thanks... Just making sure nothing is waiting on me.

There are still changes requested from me, about the whitespace issues that `git apply` is whining about that. I'll update my review if you fix that.

intellasoft:
> > Thanks... Just making sure nothing is waiting on me.
> 
> There are still changes requested from me, about the whitespace issues that `git apply` is whining about that. I'll update my review if you fix that.

Thanks.  I never knew how you were getting errors... I never saw this since I never use git apply.

Fixed.

InterLinked1 (approved):

InterLinked1:
FYI, this is a crash I encountered on a newer system (22.8.2) that seems to be caused by this patch.

If I recall, previously I was using an older version of this patch that may not have had that function... so more recent versions of this patch seem to have a regression that the original version did not.

```
Thread 1 (Thread 0x7f511e4a06c0 (LWP 187960)):
#0  0x00007f51615af95c in ??? () at /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007f516155acc2 in raise () at /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007f51615434ac in abort () at /lib/x86_64-linux-gnu/libc.so.6
#3  0x0000559f88a16ec7 in ast_do_crash () at utils.c:2841
#4  0x0000559f88846036 in my_do_crash () at astmm.c:230
#5  0x0000559f88846ac1 in __ast_free (ptr=0x7f512801bd70, file=0x559f88ad6e40 "chanvars.c", lineno=57, func=0x559f88ad6e50 <__PRETTY_FUNCTION__.3> "ast_var_delete") at astmm.c:488
        reg = 0x0
#6  0x0000559f888ba4e9 in ast_var_delete (var=0x7f512801bd70) at chanvars.c:57
#7  0x0000559f88827848 in ast_app_group_remove_meta_ifneeded (group=0x7f515c7391f0 "224", category=0x7f515c7391f4 "oe") at app.c:2368
        __list_head = 0x559f88b91100 <groups_meta>
        __list_next = 0x7f514404cbc0
        __list_prev = 0x0
        __list_current = 0x7f514404c980
        gmi = 0x7f514404c980
        headp = 0x7f514404c988
        vardata = 0x7f512801bd70
#8  0x0000559f8882933a in ast_app_group_discard (chan=0x7f515c1825b0) at app.c:2808
        __list_head = 0x559f88b910a0 <groups>
        __list_next = 0x7f5154046600
        __list_prev = 0x0
        __list_current = 0x7f515c7391d0
        gi = 0x7f515c7391d0
#9  0x0000559f88892bf2 in ast_channel_destructor (obj=0x7f515c1825b0) at channel.c:2285
```

InterLinked1:
Here's another crash due to some other kind of hanky panky going on:

There is definitely some kind of regression in the latest version of this patch. It worked fine on the old version for years, but the new version results in several crashes a day. This needs to be addressed.

```
Thread 1 (Thread 0x7f7941c206c0 (LWP 197292)):
#0  0x000055dd48542519 in ast_var_name (var=0xdeaddeaddeaddead) at chanvars.c:67
        name = 0xdeaddeaddeaddebd <error: Cannot access memory at address 0xdeaddeaddeaddebd>
#1  0x000055dd484b03cf in ast_app_group_get_var (group=0x7f7941c1c470 "218", category=0x7f798acfb0fc "oeall", name=0x7f798acf2289 "toneblock") at app.c:2553
        gmi = 0x7f79780d1750
        headp = 0x7f79780d1758
        variable = 0x7f7980115fa0 "\255", 'ޭ' <repeats 35 times>, "\336\272\355\376"
        return_value = 0xffffffff48890940 <error: Cannot access memory at address 0xffffffff48890940>
        ast_var = 0xdeaddeaddeaddead
        len = 32633
```

InterLinked1 (changes requested):
Marking changes requested due to crashes caused by the current version of this patch

InterLinked1:
> Here's another crash due to some other kind of hanky panky going on:
> 
> There is definitely some kind of regression in the latest version of this patch. It worked fine on the old version for years, but the new version results in several crashes a day. This needs to be addressed.
> 
> ```
> Thread 1 (Thread 0x7f7941c206c0 (LWP 197292)):
> #0  0x000055dd48542519 in ast_var_name (var=0xdeaddeaddeaddead) at chanvars.c:67
>         name = 0xdeaddeaddeaddebd <error: Cannot access memory at address 0xdeaddeaddeaddebd>
> #1  0x000055dd484b03cf in ast_app_group_get_var (group=0x7f7941c1c470 "218", category=0x7f798acfb0fc "oeall", name=0x7f798acf2289 "toneblock") at app.c:2553
>         gmi = 0x7f79780d1750
>         headp = 0x7f79780d1758
>         variable = 0x7f7980115fa0 "\255", 'ޭ' <repeats 35 times>, "\336\272\355\376"
>         return_value = 0xffffffff48890940 <error: Cannot access memory at address 0xffffffff48890940>
>         ast_var = 0xdeaddeaddeaddead
>         len = 32633
> ```

@kobaz Do you have a current version of the patch handy as a diff by chance? I think I'm going to have to do some rollbacks due to the instability here. I could probably extract it with some effort from a previous system, but if you have a copy that would help. One of the (many) things I miss about Gerrit...

kobaz:
Got it. 
 
 
 
 
 Will investigate. 
 
 
 
 
 
 
 
 
 InterLinked1  left a comment  (asterisk/asterisk#292) 
 
 Here's another crash due to some other kind of hanky panky going on: 
 There is definitely some kind of regression in the latest version of this patch. It worked fine on the old version for years, but the new version results in several crashes a day. This needs to be addressed. 
 Thread 1 (Thread 0x7f7941c206c0 (LWP 197292)):
#0  0x000055dd48542519 in ast_var_name (var=0xdeaddeaddeaddead) at chanvars.c:67
        name = 0xdeaddeaddeaddebd <error: Cannot access memory at address 0xdeaddeaddeaddebd>
#1  0x000055dd484b03cf in ast_app_group_get_var (group=0x7f7941c1c470 "218", category=0x7f798acfb0fc "oeall", name=0x7f798acf2289 "toneblock") at app.c:2553
        gmi = 0x7f79780d1750
        headp = 0x7f79780d1758
        variable = 0x7f7980115fa0 "\255", 'ޭ' <repeats 35 times>, "\336\272\355\376"
        return_value = 0xffffffff48890940 <error: Cannot access memory at address 0xffffffff48890940>
        ast_var = 0xdeaddeaddeaddead
        len = 32633

InterLinked1:
Here's another instance... seems like memory corruption somewhere given the stack traces are different every time but there seems to be access to freed memory?

```
Thread 1 (Thread 0x7f32d37ce6c0 (LWP 198699)):
#0  0x000055900ab0d519 in ast_var_name (var=0xdeaddeaddeaddead) at chanvars.c:67
        name = 0xdeaddeaddeaddebd <error: Cannot access memory at address 0xdeaddeaddeaddebd>
#1  0x000055900aa7b3cf in ast_app_group_get_var (group=0x7f32d37ca470 "218", category=0x7f331d1780fc "oeall", name=0x7f331d16f289 "toneblock") at app.c:2553
        gmi = 0x7f32e815e6f0
        headp = 0x7f32e815e6f8
        variable = 0x7f32f0114410 "\255", 'ޭ' <repeats 35 times>, "\336\272\355\376\3602\177"
        return_value = 0xffffffff0ae5b940 <error: Cannot access memory at address 0xffffffff0ae5b940>
        ast_var = 0xdeaddeaddeaddead
        len = 32563
```

include/asterisk/chanvars.h
InterLinked1: Following up on this - pull all the changes to this file out of this patch, they belong in a separate PR

InterLinked1:
FYI, I confirmed this older version of your patch (that I assembled, updated, and rebased from some old source code) doesn't have this issue: https://github.com/InterLinked1/phreakscript/blob/master/patches/groupvars_old.diff

With this patch, the issues do not happen and things are stable.

I would suggest comparing what is different between that version and the current version as one of those changes seems to be the culprit.

InterLinked1:
FYI, here is another compilation errors with newer gcc:

```
gcc -o /dev/null -c func_groupcount.i -pthread -I/usr/src/asterisk-22.9.0/include   -I/usr/include/libxml2 -pipe -Wall -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declarations   -g3 -Werror -Wunused -Wdeclaration-after-statement -Wtrampolines -Wundef -Wmissing-format-attribute -Wformat=2 -O0  -march=native -fPIC -DAST_MODULE=\"func_groupcount\" -DAST_MODULE_SELF_SYM=__internal_func_groupcount_self   -Wno-format-truncation  -O3 -fno-partial-inlining
func_groupcount.c: In function ‘dumpgroups_exec’:
func_groupcount.c:1250:13: error: variable ‘numgroups’ set but not used [-Werror=unused-but-set-variable=]
 1250 |         int numgroups = 0;
      |             ^~~~~~~~~
```

In fairness, this is against an older version of this patch, since the new version isn't stable. I'll patch a fix for the older version for now but maybe it exists in the new one too - haven't checked.

kobaz:
Thanks for your work on this.  I'll do some bisecting and locate the issue.

kobaz:
I'm going to dedicate time to get this completed!  Very very sorry for the long timeframe here

I just pushed the fix

InterLinked1:
> I'm going to dedicate time to get this completed! Very very sorry for the long timeframe here

Today is the 3-year anniversary of this PR :)

> I just pushed the fix

Thanks, I will test it out when I get a chance. But given the nature of that last issue, have you put together any tests for this, either unit test or test suite? I think that would have uncovered the issue pretty quickly without anyone testing.

InterLinked1 (changes requested):
There are still 2 changes requested outstanding:
- Remove numgroups variable from dumpgroups_exec to avoid a compiler warning about an unused variable
- Remove the documentation changes to chanvars.h and put them in another PR. This is big enough already without unrelated changes.

Issue #294, opened 2023-09-04
[improvement]: chan_dahdi: Improve call pickup documentation


InterLinked1:
### Improvement Description

Minor improvements to chan_dahdi and sample config clarifying the scope of pickup capabilities.

PR #295, opened 2023-09-04
chan_dahdi: Clarify scope of callgroup/pickupgroup.

InterLinked1:
Internally, chan_dahdi only applies callgroup and
pickupgroup to FXO signalled channels, but this is not documented anywhere. This is now documented in the sample config, and a warning is emitted if a
user tries configuring these settings for channel
types that do not support these settings, since they will not have any effect.

Resolves: #294


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #298, opened 2023-09-05
[improvement]: chan_rtp: Implement RTP glue


maximilianfridrich:
### Improvement Description

Currently, `chan_rtp` does not implement an RTP glue. This would be desirable for example to be able to directly pass-through media packages through native bridges.

One Limitation is that the `update_peer` callback of `chan_rtp` will not be able to actually update the peer, but that is obvious when using `chan_rtp`.

PR #300, opened 2023-09-05
chan_rtp: Implement RTP glue for UnicastRTP channels

maximilianfridrich:
Resolves: #298

UserNote: The dial string option 'g' was added to the UnicastRTP channel which enables RTP glue and therefore native RTP bridges with those channels.

### DIFF
--- a/channels/chan_rtp.c
+++ b/channels/chan_rtp.c
@@ -249,6 +249,7 @@ failure:
 enum {
 	OPT_RTP_CODEC =  (1 << 0),
 	OPT_RTP_ENGINE = (1 << 1),
+	OPT_RTP_GLUE = (1 << 2),
 };
 
 enum {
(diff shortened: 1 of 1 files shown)


maximilianfridrich:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (changes requested):
I think this needs a UserNote[1] in the commit message, and I think it should be explicitly enabled by passing it as an option in the dial string to each. This relies on other endpoints behaving as expected (including for multicast) and with how the negotiation is non-existent, and how dumb the RTP handling of the remote side can be in multicast mode I'm not comfortable having this just start potentially happening for people without it being enabled, plus with no ability to disable it except for purposely sabotaging things (unlike PJSIP where you can disable direct media on a per-endpoint basis). If you need this functionality to occur, then you have to know exactly what you're doing and have tested/tried everything involved.

[1] https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages

maximilianfridrich:
Good point, done. I have removed it for the MulticastRTP channels, those are not quite as simple as I thought.

jcolp (approved):

Issue #308, opened 2023-09-09
[bug]: chan_console: Deadlock when hanging up console channels


InterLinked1:
### Severity

Major

### Versions

20.4.0

### Components/Modules

chan_console

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Imported from JIRA: https://issues-archive.asterisk.org/ASTERISK-30481

chan_console deadlocks when channels are hung up, due to the read thread being cancelled while it is holding locks.

This redoes the disconnection more cleanly to avoid using pthread_cancel and uncleanly terminating threads.

### Relevant log output

```shell
[2023-09-09 10:59:45.148] WARNING[210230]: chan_console.c:621 console_write: before write lock
[2023-09-09 10:59:45.174] WARNING[210231]: chan_console.c:286 stream_monitor: done read lock
[2023-09-09 10:59:45.174] WARNING[210230]: chan_console.c:623 console_write: got write lock
[2023-09-09 10:59:45.174] WARNING[210230]: chan_console.c:626 console_write: done write lock
[2023-09-09 10:59:45.174] WARNING[210230]: chan_console.c:621 console_write: before write lock
[2023-09-09 10:59:45.174] WARNING[210230]: chan_console.c:623 console_write: got write lock
[2023-09-09 10:59:45.175] WARNING[210230]: chan_console.c:626 console_write: done write lock
[2023-09-09 10:59:45.175] WARNING[210231]: chan_console.c:281 stream_monitor: before read lock
[2023-09-09 10:59:45.175] WARNING[210231]: chan_console.c:283 stream_monitor: got read lock
[2023-09-09 10:59:45.175] WARNING[210231]: chan_console.c:286 stream_monitor: done read lock
[2023-09-09 10:59:45.175] WARNING[210231]: chan_console.c:281 stream_monitor: before read lock
[2023-09-09 10:59:45.175] WARNING[210231]: chan_console.c:283 stream_monitor: got read lock
[2023-09-09 10:59:45.188] WARNING[210230]: chan_console.c:621 console_write: before write lock
debian*CLI> channel request hangup all
Requested Hangup on channel 'Console/default'
[2023-09-09 10:59:45.217] WARNING[210231]: chan_console.c:286 stream_monitor: done read lock
[2023-09-09 10:59:45.217] WARNING[210230]: chan_console.c:623 console_write: got write lock
[2023-09-09 10:59:45.217] WARNING[210230]: chan_console.c:626 console_write: done write lock
[...]

PR #309, opened 2023-09-09
chan_console: Fix deadlock caused by unclean thread exit.

InterLinked1:
To terminate a console channel, stop_stream causes pthread_cancel to make stream_monitor exit. However, commit 5b8fea93d106332bc0faa4b7fa8a6ea71e546cac added locking to this function which results in deadlock due to the stream_monitor thread being killed while it's holding the pvt lock.

To resolve this, a flag is now set and read to indicate abort, so the use of pthread_cancel and pthread_kill can be avoided altogether.

Resolves: #308

### DIFF
--- a/channels/chan_console.c
+++ b/channels/chan_console.c
@@ -152,6 +152,8 @@ static struct console_pvt {
 	struct ast_frame fr;
 	/*! Running = 1, Not running = 0 */
 	unsigned int streamstate:1;
+	/*! Abort stream processing? */
+	unsigned int abort:1;
 	/*! On-hook = 0, Off-hook = 1 */
 	unsigned int hookstate:1;
 	/*! Unmuted = 0, Muted = 1 */
@@ -275,18 +277,19 @@ static void *stream_monitor(void *data)
 	};
 
 	for (;;) {
-		pthread_testcancel();
 		console_pvt_lock(pvt);
 		res = Pa_ReadStream(pvt->stream, buf, sizeof(buf) / sizeof(int16_t));
 		console_pvt_unlock(pvt);
-		pthread_testcancel();
 
-		if (!pvt->owner) {
+		if (!pvt->owner || pvt->abort) {
 			return NULL;
 		}
 
-		if (res == paNoError)
+		if (res == paNoError) {
 			ast_queue_frame(pvt->owner, &f);
+		} else {
+			ast_log(LOG_WARNING, "Console ReadStream failed: %s\n", Pa_GetErrorText(res));
+		}
 	}
 
 	return NULL;
@@ -401,8 +404,9 @@ static int stop_stream(struct console_pvt *pvt)
 	if (!pvt->streamstate || pvt->thread == AST_PTHREADT_NULL)
 		return 0;
 
-	pthread_cancel(pvt->thread);
-	pthread_kill(pvt->thread, SIGURG);
+	pvt->abort = 1;
+	/* Wait for pvt->thread to exit cleanly, to avoid killing it while it's holding a lock. */
+	pthread_kill(pvt->thread, SIGURG); /* Wake it up if needed, but don't cancel it */
 	pthread_join(pvt->thread, NULL);
 
 	console_pvt_lock(pvt);


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

channels/chan_console.c
```diff
@@ -401,8 +404,8 @@ static int stop_stream(struct console_pvt *pvt)
 	if (!pvt->streamstate || pvt->thread == AST_PTHREADT_NULL)
 		return 0;
 
-	pthread_cancel(pvt->thread);
```
jcolp, line 404: These should still happen to cause the thread to wake up if needed.

jcolp (changes requested):

channels/chan_console.c
InterLinked1: `pthread_cancel` was the whole reason for the deadlock in the first place, that's why I'd prefer to avoid it. It didn't seem to be necessary since the read eventually expires and it returns on its own.

It might work if `pthread_setcancelstate` is used to disable cancellability in the critical section, but is that the place you're thinking the thread would need to be woken up?

seanbright:
Isn’t this what `pthread_cleanup_push()` is for? That would be cleaner IMO.

InterLinked1:
> Isn’t this what `pthread_cleanup_push()` is for? That would be cleaner IMO.

Maybe cleaner than setting the cancellability, but it's a little messy and using `pthread_cancel` is never very "clean", in my opinion, so if it can be avoided that's probably more ideal.

I restored the `pthread_kill` so if the read is blocking for some reason, it should still wake up, and exit after it releases the mutex, and then the thread can join.

seanbright:
> using `pthread_cancel` is never very "clean", in my opinion

And why is that?

InterLinked1:
> > using `pthread_cancel` is never very "clean", in my opinion
> 
> And why is that?

Because the thread that's cancelled doesn't get to control how it exits. You can add a cleanup handler but now you'd probably need to keep track of whether it's holding a mutex, for example, and then release it only then.

This guy here has a good explanation that says it better than I can: https://stackoverflow.com/a/3438576

> I subscribe to the theory that each thread should totally control its own resources, including its execution lifetime. I've always found that to be the best way to avoid deadlock.

`pthread_cancel` was fine before the commit that added locking, but now that there are resources fine, it's cleaner to not use `pthread_cancel` and have the thread control where it exits.

seanbright:
Is `Pa_ReadStream()` a cancellation point? I don’t see how calling `pthread_cancel()` here could leave `pvt` locked.

InterLinked1:
> Is `Pa_ReadStream()` a cancellation point? I don’t see how calling `pthread_cancel()` here could leave `pvt` locked.

Internally, probably (internally, I imagine it would call `read` somewhere), because it was exiting while holding the pvt lock. There's an example in the linked issue.

seanbright:
> Because the thread that's cancelled doesn't get to control how it exits.

Of course it does. `pthread_testcancel()` and `pthread_cancel()` are cooperative.

InterLinked1:
> > Because the thread that's cancelled doesn't get to control how it exits.
> 
> Of course it does. `pthread_testcancel()` and `pthread_cancel()` are cooperative.

Yes, but if `PA_readstream` internally calls `pthread_testcancel`, which I'm almost positive it does (perhaps via `read` or `poll`), then you're screwed because it will exit at the wrong cancellation point. That's why cancellability would need to be disabled while it's holding the mutex, to ensure it never exits in the middle.

I believe by default all threads are cancellable so without this there isn't a guarantee it won't be cancelled in the middle. I originally tested it this way before changing it to a flag. This should also work, if you strongly prefer this approach to setting a flag and not using `pthread_cancel`, I could change it back, they should both do the same thing.

jcolp (approved):

seanbright:
My preference would be to use cancellation but I am not writing the code. If you've confirmed this fixes the issue, then it's fine.

Issue #307, opened 2023-09-08
[bug]: Deadlock related to dial/pjsip/import


kobaz:
### Severity

Major

### Versions

18.18.0

### Components/Modules

dial,func_import,pjsip

### Operating Environment

Debian
Completely Vanilla Asterisk 18.18.0

### Frequency of Occurrence

Constant

### Issue Description

I have a system locking up about a dozen times a day with this deadlock.

### Relevant log output

```shell
=======================================================================
=== 18.18.0
=== Currently Held Locks
=======================================================================
===
=== <pending> <lock#> (<file>): <lock type> <line num> <function> <lock name> <lock addr> (times locked)
===
=== Thread ID: 0x7f46b3820700 LWP:8100 (worker_start         started at [ 1111] threadpool.c worker_thread_start())
=== ---> Lock #0 (channel.c): MUTEX 920 __ast_channel_alloc_ap tmp 0x7f46e4119fe0 (1)
        [0x564c39cb3194] asterisk lock.c:273 __ast_pthread_mutex_lock()
        [0x564c39bf1705] asterisk astobj2.c:241 __ao2_lock()
        [0x564c39c2f612] asterisk channel.c:922 __ast_channel_alloc_ap()
        [0x564c39c2f923] asterisk channel.c:966 __ast_channel_alloc()
        [0x7f46b52482de] chan_pjsip.so chan_pjsip.c:553 chan_pjsip_new()
        [0x7f46b5251c34] chan_pjsip.so chan_pjsip.c:2992 chan_pjsip_incoming_request()
        [0x7f46b52ca1b3] res_pjsip_session.so res_pjsip_session.c:4467 handle_incoming_request()
        [0x7f46b52c80ba] res_pjsip_session.so res_pjsip_session.c:4080 new_invite()
        [0x7f46b52c866c] res_pjsip_session.so res_pjsip_session.c:4172 handle_new_invite_request()
        [0x7f46b52c9136] res_pjsip_session.so res_pjsip_session.c:4284 session_on_rx_request()
        [0x7f46f8882cb5] libasteriskpj.so.2 sip_endpoint.c:929 pjsip_endpt_process_rx_data()
        [0x7f46b532665d] res_pjsip.so pjsip_distributor.c:956 distribute()
        [0x564c39d685a6] asterisk taskprocessor.c:1304 ast_taskprocessor_execute()
        [0x564c39d7226b] asterisk threadpool.c:1350 execute_tasks()
[...]

PR #311, opened 2023-09-09
channels: Avoid deadlocks when locking ao2_container *channels

intellasoft:
One situation where deadlocks occur is when:
   -> res_pjsip_sdp_rtp is checking for rtp timeouts -> Which locks
      ao2_container channels, and then waits for a lock on a channel
      that's currently running IMPORT
 At the same time
   -> Dialplan asks for IMPORT(${some_channel},...) -> Which grabs a
      channel lock and then waits on ao2_container channels

So in this situation, res_pjsip_sdp_rtp is waiting on a channel lock
that will never release, because res_pjsip_sdp_rtp is currently
holding ao2_container channels, that the IMPORT needs after having
already locked the channel that res_pjsip_sdp_rtp is processing next.

Resolves: #307

cherry-pick-to: 19
cherry-pick-to: 20
cherry-pick-to: master


main/channel.c
```diff
@@ -712,12 +718,23 @@ static int does_id_conflict(const char *uniqueid)
 {
 	struct ast_channel *conflict;
 	size_t length = 0;
+	struct ast_channel_callback_data callback_data = {.search = uniqueid, .final_match_result = 0};
```
InterLinked1, line 721: Can we put each element on its own line? Not sure if that's explicitly in the coding guidelines but that's easier to read.

main/channel.c
```diff
@@ -712,12 +718,23 @@ static int does_id_conflict(const char *uniqueid)
...
 
 	if (ast_strlen_zero(uniqueid)) {
 		return 0;
 	}
 
-	conflict = ast_channel_callback(ast_channel_by_uniqueid_cb, (char *) uniqueid, &length, OBJ_NOLOCK);
+  try_again:
```
InterLinked1, line 727: I don't think goto is necessary here, this could be a do while or for (;;) with break

main/channel.c
```diff
@@ -1363,14 +1404,23 @@ struct ast_channel_iterator *ast_channel_iterator_by_exten_new(const char *exten
...
 	}
 
+  try_again:
 	i->active_iterator = (void *) ast_channel_callback(ast_channel_by_exten_cb,
-		l_context, l_exten, OBJ_MULTIPLE);
+		&callback_data, l_exten, OBJ_MULTIPLE);
+
+	if (callback_data.final_match_result == CMP_RETRY_NEEDED) {
+		/* We ran into a possible deadlock, try again */
+		ast_debug(3, "ast_channel_iterator_by_exten_new() Avoid deadlock trying to find channel by extension@context: %s@%s\n", exten, context);
```
InterLinked1, line 1419: ast_debug includes the function name, no need to repeat it.

This could all be simplified to:

`ast_debug(3, "Avoided deadlock trying to find channel by %@%s\n", exten, context);

main/channel.c
```diff
@@ -1429,20 +1479,41 @@ struct ast_channel *ast_channel_get_by_name_prefix(const char *name, size_t name
...
 	char *l_name = (char *) name;
+	struct ast_channel_callback_data callback_data = {.search = name, .final_match_result = 0};
 
 	if (ast_strlen_zero(l_name)) {
 		/* We didn't have a name to search for so quit. */
 		return NULL;
 	}
 
-	chan = ast_channel_callback(ast_channel_by_name_cb, l_name, &name_len,
+  try_name_again:
```
InterLinked1, line 1489: Same thing here, you can use 2 loops here and goto isn't needed.

main/channel.c
```diff
@@ -1429,20 +1479,41 @@ struct ast_channel *ast_channel_get_by_name_prefix(const char *name, size_t name
...
 
+  try_uniqueid_again:
 	/* Now try a search for uniqueid. */
-	return ast_channel_callback(ast_channel_by_uniqueid_cb, l_name, &name_len, 0);
+
+	chan = ast_channel_callback(ast_channel_by_uniqueid_cb, &callback_data, &name_len, 0);
+
+	if (callback_data.final_match_result == CMP_RETRY_NEEDED) {
+		/* We ran into a possible deadlock, try again */
+		ast_debug(3, "ast_channel_get_by_name_prefix() Avoid deadlock trying to find channel by uniqueid: %s\n", name);
```
InterLinked1, line 1511: Same comment about log format

InterLinked1 (changes requested):
Remove the cherry pick lines from the commit message and make a comment with those instead.

kobaz:
cherry-pick-to: 19
    cherry-pick-to: 20
    cherry-pick-to: master

main/channel.c
kobaz: Fixed

main/channel.c
kobaz: Fixed

main/channel.c
kobaz: Fixed

InterLinked1:
> ```
> cherry-pick-to: 19
> cherry-pick-to: 20
> cherry-pick-to: master
> ```

PRs should be raised against master, not 18 (unlike Gerrit, it makes a difference now).
And it should be cherry picked to 18, 20, and 21, not 19.

kobaz:
master is slightly different.  It will not cleanly cherry-pick to branches from master.  Is the process still go to to master first? Make separate PRs for each branch?

main/channel.c
kobaz: Are goto's that bad?  It reduces nesting.  The only thing that adding a loop does is introduce an extra level of nesting.

InterLinked1:
> master is slightly different. It will not cleanly cherry-pick to branches from master. Is the process still go to to master first? Make separate PRs for each branch?

Ah, okay, then it's more complicated, then you need to do them separately since GitHub can only cherry pick verbatim, might want to double check with George.

main/channel.c
InterLinked1: They're not inherently bad, this just doesn't seem like a necessary or appropriate use for them, since a loop works perfectly fine. I don't think there's a problem with adding a level of nesting, it contains the logic in its own scope.

main/channel.c
intellasoft: Fixed

main/channel.c
intellasoft: Fixed

kobaz:
> > master is slightly different. It will not cleanly cherry-pick to branches from master. Is the process still go to to master first? Make separate PRs for each branch?
> 
> Ah, okay, then it's more complicated, then you might need to do multiple sets since GitHub can only cherry pick verbatim, might want to double check with George.

master no longer has support for macros, so the checks in ast_channel_by_exten_cb() will not cherry-pick clean with the change as-is.

main/channel.c
```diff
@@ -1363,14 +1415,31 @@ struct ast_channel_iterator *ast_channel_iterator_by_exten_new(const char *exten
...
+			/* We ran into a possible deadlock, try again */
+			ast_debug(3, "Avoid deadlock trying to find channel by extension@context: %s@%s\n", exten, context);
+			usleep(100);
+			continue;
+		}
+
+		/* We found a match or the search fully completed through all channels */
+		break;
+	}
+	while (1);
```
InterLinked1, line 1441: Same here.

main/channel.c
```diff
@@ -1429,20 +1515,55 @@ struct ast_channel *ast_channel_get_by_name_prefix(const char *name, size_t name
...
+			/* We ran into a possible deadlock, try again */
+			ast_debug(3, "Avoid deadlock trying to find channel by uniqueid: %s\n", name);
+			usleep(100);
+			continue;
+		}
+
+		/* We found a match or the search fully completed through all channels */
+		break;
+	}
+	while (1);
```
InterLinked1, line 1563: ditto

InterLinked1 (changes requested):

kobaz:
Update: Been running this in the wild with mixed results.

Some days it's been completely solid with no deadlocks (better than before with 3-5 deadlocks per day or more).

Some days it actually almost immediately deadlocked after handling just a few calls.

jcolp:
Due to the last comment I am transitioning this to a draft as it does not appear to be in a state for review as it does not resolve the underlying issue. If this is untrue and it should actually be reviewed, then it can be published.

kobaz:
Fixed previous issue where it did not fully fix the problem.  Small refactor to avoid breaking assumptions about value of 'arg' with respect to find channel callbacks.

It's now passing my tests regarding the original deadlock situation.

I would nominate this ready for review now.

main/channel.c
```diff
@@ -1278,67 +1308,92 @@ struct ast_channel *ast_channel_callback(ao2_callback_data_fn *cb_fn, void *arg,
...
 	int ret = CMP_MATCH;
 
 	if (ast_strlen_zero(name)) {
 		ast_log(LOG_ERROR, "BUG! Must supply a channel name or partial name to match!\n");
 		return CMP_STOP;
 	}
 
-	ast_channel_lock(chan);
+	if (ast_channel_trylock(chan) != 0) {
+		ast_verb(2, "ast_channel_by_name_cb: Deadlock-Avoid RETRY NEEDED for: %s\n", name);
```
InterLinked1, line 1322: This should NOT be a verbose and there should NOT be a function name prefix in the log message.

main/channel.c
```diff
@@ -1383,14 +1456,33 @@ struct ast_channel_iterator *ast_channel_iterator_by_name_new(const char *name,
...
+	};
 
 	if (!(i = ast_calloc(1, sizeof(*i)))) {
 		return NULL;
 	}
 
-	i->active_iterator = (void *) ast_channel_callback(ast_channel_by_name_cb,
-		l_name, &name_len,
-		OBJ_MULTIPLE | (name_len == 0 /* match the whole word, so optimize */ ? OBJ_KEY : 0));
+	for (;;) {
```
InterLinked1, line 1468: This logic seems duplicated several times, could it be refactored?

InterLinked1 (changes requested):

main/channel.c
kobaz: I remembered that after I committed. I was using verb for testing so I didn't have to core debug to see it.  That'll get fixed.

main/channel.c
kobaz: I'll see what I can do!

include/asterisk/channel.h
```diff
@@ -3076,14 +3082,49 @@ struct ast_channel *ast_channel_iterator_next(struct ast_channel_iterator *i);
...
- * system.  The channel is provided as an argument to the function.
+ * system.  The channel (or thing) to be searched for is provided as an argument (arg)
+ * to the function.
  *
+ * \param arg Must be a char *, Typically either a channel name or a channel uniqueid.
+ *   In some cases exten@context is used.
+ *   Will be passed to cb_fn as its arg param
+ * \param data Arbitrary data to be passed to cb_fn as its data param
+ *
+ * \note This function SHOULD NOT be called unlessfully understanding the implications.
```
InterLinked1, line 3093: unlessfully?

include/asterisk/channel.h
```diff
@@ -3076,14 +3082,49 @@ struct ast_channel *ast_channel_iterator_next(struct ast_channel_iterator *i);
  *
  * \details
  * This function executes a callback one time for each active channel on the
- * system.  The channel is provided as an argument to the function.
+ * system.  The channel (or thing) to be searched for is provided as an argument (arg)
+ * to the function.
  *
+ * \param arg Must be a char *, Typically either a channel name or a channel uniqueid.
```
InterLinked1, line 3088: typically should not be capitalized

main/channel.c
```diff
@@ -708,16 +708,75 @@ static void ast_channel_destructor(void *obj);
...
+static struct ast_channel *ast_channel_callback_internal(ao2_callback_data_fn *cb_fn, void *arg,
+		void *data, int ao2_flags)
+{
+	return ao2_callback_data(channels, ao2_flags, cb_fn, arg, data);
+}
+
+struct ast_channel *ast_channel_callback(ao2_callback_data_fn *cb_fn, void *arg,
+		void *data, int ao2_flags)
+{
+	ast_log(LOG_WARNING, "Using ast_channel_callback() directly may lead to deadlocks.  Please update your module.\n");
```
InterLinked1, line 720: This is an existing function, right? Won't that start spewing warnings all over the place? If this really was a proper fix, then everything else should be updated at the same time too. But I would hold off on that for now until that is determined.

main/channel.c
```diff
@@ -708,16 +708,75 @@ static void ast_channel_destructor(void *obj);
...
+		break;
+	}
+
+	return channel;
+}
+
 static int does_id_conflict(const char *uniqueid)
 {
 	struct ast_channel *conflict;
+        char *l_uniqueid = (char *) uniqueid;
```
InterLinked1, line 771: You're pointlessly casting to the same thing here

main/channel.c
```diff
@@ -1269,76 +1328,95 @@ void ast_channel_undefer_dtmf(struct ast_channel *chan)
...
 	return ret;
 }
 
 static int ast_channel_by_exten_cb(void *obj, void *arg, void *data, int flags)
 {
 	struct ast_channel *chan = obj;
+	struct ast_channel_callback_data *callback_data = data;
 	char *context = arg;
-	char *exten = data;
+	char *exten = (char *) callback_data->data_additional;
```
InterLinked1, line 1367: Redundant cast

main/channel.c
```diff
@@ -1359,18 +1437,39 @@ struct ast_channel_iterator *ast_channel_iterator_destroy(struct ast_channel_ite
...
-	char *l_exten = (char *) exten;
-	char *l_context = (char *) context;
 
 	if (!(i = ast_calloc(1, sizeof(*i)))) {
 		return NULL;
 	}
 
-	i->active_iterator = (void *) ast_channel_callback(ast_channel_by_exten_cb,
-		l_context, l_exten, OBJ_MULTIPLE);
+	i->active_iterator = (void *) channel_find_by_exten(exten, context, OBJ_MULTIPLE);
```
InterLinked1, line 1471: You don't need to cast to void

main/channel.c
```diff
@@ -1388,9 +1487,10 @@ struct ast_channel_iterator *ast_channel_iterator_by_name_new(const char *name,
 		return NULL;
 	}
 
-	i->active_iterator = (void *) ast_channel_callback(ast_channel_by_name_cb,
-		l_name, &name_len,
+	i->active_iterator = (void *) ast_channel_callback_safe(ast_channel_by_name_cb,
```
InterLinked1, line 1490: Ditto

apps/app_directed_pickup.c
```diff
@@ -380,15 +393,18 @@ static struct ast_channel *find_by_part(struct ast_channel *chan, const char *pa
...
 	if (target) {
 		return target;
 	}
 
 	/* Now try a search for uniqueid. */
 	pickup_args.name = part;
 	pickup_args.len = 0;
-	return ast_channel_callback(find_by_uniqueid, NULL, &pickup_args, 0);
+
+	return ast_channel_callback_safe(find_by_uniqueid, "channel uniqueid, where pickup is possible", pickup_args.name, NULL, &pickup_args, 0);
```
InterLinked1, line 407: The description is identical to line 229. Shouldn't they be different if the goal is to be unique?
Additionally I think automatically include the filename and line number is more robust than hardcoding descriptions. In which case, there should be a macro that injects these automatically.

InterLinked1 (changes requested):

include/asterisk/channel.h
kobaz: Haha... yeah I noticed that after I sent the last push.  Should be 'unless fully'

main/channel.c
kobaz: It's dropping the const qualifier so that you don't have to cast it for the callback.

apps/app_directed_pickup.c
kobaz: Fixed

main/channel.c
kobaz: So, I was thinking about this.  Yes the void * cast is not required, but it does add code clarity.  Normally ast_channel_callback returns a channel, but in this case with OBJ_MULTIPLE it returns an iterator containing the list of matched channels so (void *) is more applicable here for syntactic clarity.

main/channel.c
kobaz: This is a public api function.  All internal usages across the core and modules have been updated in this patch.  People who may have written 3rd party modules that iterate through channels with ao2_channel_callback would need to be rewritten to use the safe version (which needs to be trylock aware) so there's no 'drop-in' replacement for ast_channel_callback.

3rd party modules would get this warning.  I was thinking about it that there should be an option to turn the warning off, but I believe avoiding deadlocks is important enough it should be pretty loud (with the ability to mute).

main/channel.c
kobaz: Looking at this some more, the void * is required.

kobaz:
Pushed a new one.. fixes channel reference leaks (introduced in the previous patch) when OBJ_MULTIPLE is used to return an iterator

kobaz:
While trying to extract out components from our full platform into a standalone deadlock reproduction test, I've uncovered an additional deadlock related to ast_sip_push_task_wait() also related to IMPORT import_helper() (posted to #307).  I realized this is not the same deadlock that this #311 pull is addressing.  So I'll work on another test to trigger this specific one as well.

This is related, but probably should be split into a new issue since this other newly uncovered deadlock is not related to ao2_container *channels.

gtjoseph:
I'm converting this back to a draft again since there are still outstanding changes requested and the cherry-pick lines still aren't correct.

Speaking of which, are you saying this PR will cherry-pick cleanly from 18 to master but not the other way around?

kobaz:
Correct, I don't think it goes the other way, but I can do some tests.

gtjoseph:
@intellasoft Is this PR still relevant since the channel storage backends work was merged?

intellasoft:
That's a really good question.  This can probably be put on hold for now... I'll need to validate if this still happens with the new backend.

PR #312, opened 2023-09-11
sip_transport.c: Make sure transport not shutdown before message finish.

Tinet-mucw:
From the gdb information, ast_websocket_read read message successfully, then function transport_read is called in serializer. during to the executing of pjsip_transport_down, the ws_session->stream->fd is closed; transport_read encountered an error and exited the while loop. After executing transport_shutdown, the transport's reference count became 0, causing a crash when sending SIP messages. In the original design, pjsip_transport_dec_ref executed earlier than pjsip_rx_data_clone lead to this issue.

Resolves: asterisk#299

### DIFF
--- a/third-party/pjproject/patches/0011-Make-sure-transport-not-shutdown-before-message-finish.patch
+++ b/third-party/pjproject/patches/0011-Make-sure-transport-not-shutdown-before-message-finish.patch
@@ -1,0 +1,46 @@
+diff --git a/pjsip/src/pjsip/sip_transport.c b/pjsip/src/pjsip/sip_transport.c
+index 4f483fa..13b7c5b 100644
+--- a/pjsip/src/pjsip/sip_transport.c
++++ b/pjsip/src/pjsip/sip_transport.c
+@@ -770,16 +770,28 @@ PJ_DEF(pj_status_t) pjsip_rx_data_clone( const pjsip_rx_data *src,
+     pj_pool_t *pool;
+     pjsip_rx_data *dst;
+     pjsip_hdr *hdr;
++    pj_status_t clone_status;
++    pjsip_transport *src_tp = (pjsip_transport*)src->tp_info.transport;
+ 
+     PJ_ASSERT_RETURN(src && flags==0 && p_rdata, PJ_EINVAL);
+ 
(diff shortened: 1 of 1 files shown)


jcolp:
Did this patch come from PJSIP? Has it been provided to them, since it is a patch against it?

Tinet-mucw:
i'm sorry. The problem has been identified, and I'm looking forward to your more reasonable solution.
Furthermore, pjsip_rx_data_clone should ensure that the transport is available, either by internal assurance or external assurance.

jcolp:
You've created a patch against PJSIP itself. Are you expecting this to be the solution to the problem? If so, it should be provided to PJSIP first for review and inclusion, before being included in Asterisk.

silentindark:
@Tinet-mucw Could you please upload your patch to https://github.com/pjsip/pjproject/pulls ?

Tinet-mucw:
I have already submitted a PR: https://github.com/pjsip/pjproject/pull/3694
My English proficiency is limited, and the description may not be very clear.

Tinet-mucw:
The reply is as follows:

The pjsip_tpmgr_receive_packet() is normally invoked from ioqueue read callback which should be proceeded by add-ref to the transport group lock (shared group lock with ioqueue key), code is [here](https://github.com/pjsip/pjproject/blob/ad152c2f42c3e067b8b280f18572a009a13d32d4/pjlib/src/pj/ioqueue_select.c#L1088C18-L1088C18), so theoretically the group lock ref-count should never reach 0 there. The tp->ref_cnt may reach 0 but the transport pool/atomic/etc will not be released until group lock ref-count reaches 0, so it should not cause crash.

Currently all built-in PJSIP SIP transports are equipped with group lock, shared with its lower level socket/ioqueue-key. Any custom SIP transports are strongly adopt the approach as otherwise there may be other synchronization issues (this case seems to be one of them). See also https://github.com/pjsip/pjproject/issues/1773.

referring to the group lock and adding an transport's ao2_lock in Asterisk. This can resolve the pjsip_transport_shutdown initiated by Asterisk.

https://github.com/pjsip/pjproject/commit/6cbe3651b59c4747ebba409d65c9e00cf5b7df94

PR #327, opened 2023-09-13
res_pjsip_transport_websocket: Prevent transport from being destroyed before message finishes.

Tinet-mucw:
From the gdb information, ast_websocket_read reads a message successfully,
then transport_read is called in the serializer. During execution of pjsip_transport_down,
ws_session->stream->fd is closed; ast_websocket_read encounters an error and exits the while loop.
After executing transport_shutdown, the transport's reference count becomes 0, causing a crash when sending SIP messages.
This was due to pjsip_transport_dec_ref executing earlier than pjsip_rx_data_clone, leading to this issue.
In websocket_cb executeing pjsip_transport_add_ref, this we now ensure the transport is not destroyed while in the loop.

Resolves: asterisk#299

### DIFF
--- a/res/res_pjsip_transport_websocket.c
+++ b/res/res_pjsip_transport_websocket.c
@@ -85,14 +85,26 @@ static pj_status_t ws_send_msg(pjsip_transport *transport,
 static pj_status_t ws_destroy(pjsip_transport *transport)
 {
 	struct ws_transport *wstransport = (struct ws_transport *)transport;
+
+	ao2_ref(wstransport, -1);
+
+	return PJ_SUCCESS;
+}
+
+/*!
+ * \brief Shut down the pjsip transport.
+ *
+ * Called by pjsip transport manager.
+ */
+static pj_status_t ws_shutdown(pjsip_transport *transport)
+{
+	struct ws_transport *wstransport = (struct ws_transport *)transport;
 	int fd = ast_websocket_fd(wstransport->ws_session);
 
 	if (fd > 0) {
 		ast_websocket_close(wstransport->ws_session, 1000);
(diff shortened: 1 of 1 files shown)


Tinet-mucw:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

res/res_pjsip_transport_websocket.c
```diff
@@ -83,6 +83,20 @@ static pj_status_t ws_send_msg(pjsip_transport *transport,
...
+{
+	struct ws_transport *wstransport = (struct ws_transport *)transport;
+
+	ao2_ref(wstransport, -1);
+
+	return PJ_SUCCESS;
+}
+
+/*!
+ * \brief shutdown the pjsip transport.
```
InterLinked1, line 95: Change "shutdown" to "Shut down"

InterLinked1 (changes requested):
The commit message should be wrapped, right now it's all on one line.

Additionally it could be clearer to read, I've rephrased your original here based on my understanding of this:

```
res_pjsip_transport_websocket: Prevent transport from being destroyed before message finishes.

From the gdb information, ast_websocket_read reads a message successfully,
then transport_read is called in the serializer. During execution of pjsip_transport_down,
ws_session->stream->fd is closed; transport_read encounters an error and exits the while loop.
After executing transport_shutdown, the transport's reference count becomes 0, causing a crash when sending SIP messages.
This was due to pjsip_transport_dec_ref executing earlier than pjsip_rx_data_clone, leading to this issue.
In websocket_cb executeing pjsip_transport_add_ref, this we now ensure the transport is not destroyed while in the loop.
```

PR #329, opened 2023-09-13
make_buildopts_h, et. al.  Allow adding all cflags to buildopts.h

gtjoseph:
The previous behavior of make_buildopts_h was to not add the
non-ABI-breaking MENUSELECT_CFLAGS like DETECT_DEADLOCKS,
REF_DEBUG, etc. to the buildopts.h file because "it caused
ccache to invalidate files and extended compile times". They're
only defined by passing them on the gcc command line with '-D'
options.   In practice, including them in the include file rarely
causes any impact because the only time ccache cares is if you
actually change an option so the hit occurrs only once after
you change it.

OK so why would we want to include them?  Many IDEs follow the
include files to resolve defines and if the options aren't in an
include file, it can cause the IDE to mark blocks of "ifdeffed"
code as unused when they're really not.

So...

* Added a new menuselect compile option ADD_CFLAGS_TO_BUILDOPTS_H
  which tells make_buildopts_h to include the non-ABI-breaking
  flags in buildopts.h as well as the ABI-breaking ones. The default
  is disabled to preserve current behavior.  As before though,
  only the ABI-breaking flags appear in AST_BUILDOPTS and only
  those are used to calculate AST_BUILDOPT_SUM.
  A new AST_BUILDOPT_ALL define was created to capture all of the
  flags.

* make_version_c was streamlined to use buildopts.h and also to
  create asterisk_build_opts_all[] and ast_get_build_opts_all(void)

* "core show settings" now shows both AST_BUILDOPTS and
  AST_BUILDOPTS_ALL.

UserNote: The "Build Options" entry in the "core show settings"
CLI command has been renamed to "ABI related Build Options" and
a new entry named "All Build Options" has been added that shows
both breaking and non-breaking options.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -377,7 +377,7 @@ $(MOD_SUBDIRS_MENUSELECT_TREE):
 	+@$(SUBMAKE) -C $(@:-menuselect-tree=) SUBDIR=$(@:-menuselect-tree=) moduleinfo
 	+@$(SUBMAKE) -C $(@:-menuselect-tree=) SUBDIR=$(@:-menuselect-tree=) makeopts
 
-$(SUBDIRS): makeopts .lastclean main/version.c include/asterisk/build.h include/asterisk/buildopts.h defaults.h
+$(SUBDIRS): makeopts .lastclean main/version.c include/asterisk/build.h defaults.h
 
 ifeq ($(findstring $(OSARCH), mingw32 cygwin ),)
 main: third-party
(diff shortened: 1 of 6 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #321, opened 2023-09-11
[bug]: Performance suffers unnecessarily when debugging deadlocks


gtjoseph:
### Severity

Minor

### Versions

18,20,21,master

### Components/Modules

lock.c

### Operating Environment

all

### Frequency of Occurrence

Occasional

### Issue Description

There could be improvements in the locking code when debugging deadlocks.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #334, opened 2023-09-15
lock.c: Separate DETECT_DEADLOCKS from DEBUG_THREADS

gtjoseph:
Previously, DETECT_DEADLOCKS depended on DEBUG_THREADS.
Unfortunately, DEBUG_THREADS adds a lot of lock tracking overhead
to all of the lock lifecycle calls whereas DETECT_DEADLOCKS just
causes the lock calls to loop over trylock in 200us intervals until
the lock is obtained and spits out log messages if it takes more
than 5 seconds.  From a code perspective, the only reason they were
tied together was for logging.  So... The ifdefs in lock.c were
refactored to allow DETECT_DEADLOCKS to be enabled without
also enabling DEBUG_THREADS.

Resolves: #321

### DIFF
--- a/build_tools/cflags-devmode.xml
+++ b/build_tools/cflags-devmode.xml
@@ -9,7 +9,6 @@
 			<support_level>extended</support_level>
 		</member>
 		<member name="DETECT_DEADLOCKS" displayname="Detect Deadlocks">
-			<depend>DEBUG_THREADS</depend>
 			<support_level>extended</support_level>
 		</member>
 		<member name="DUMP_SCHEDULER" displayname="Dump Scheduler Contents for Debugging">
--- a/include/asterisk/lock.h
+++ b/include/asterisk/lock.h
@@ -134,14 +134,15 @@ struct ast_lock_track_flags {
  */
 struct ast_mutex_info {
 	pthread_mutex_t mutex;
-#if !defined(DEBUG_THREADS) && !defined(DEBUG_THREADS_LOOSE_ABI)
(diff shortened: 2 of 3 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

gtjoseph:
> Will people have to update their compilation options now? If so, there should at least be a UserNote stating the new option that's required.

There are no new options.  Both DEBUG_THREADS and DETECT_DEADLOCKS have existed for a long time.

InterLinked1:
> > Will people have to update their compilation options now? If so, there should at least be a UserNote stating the new option that's required.
> 
> There are no new options. Both DEBUG_THREADS and DETECT_DEADLOCKS have existed for a long time.

But if `DEBUG_THREADS` is no longer required for `DETECT_DEADLOCKS`, people currently compiling with that to use `DETECT_DEADLOCKS` may want to remove the `DEBUG_THREADS` option.

gtjoseph:
> > > Will people have to update their compilation options now? If so, there should at least be a UserNote stating the new option that's required.
> > 
> > 
> > There are no new options. Both DEBUG_THREADS and DETECT_DEADLOCKS have existed for a long time.
> 
> But if `DEBUG_THREADS` is no longer required for `DETECT_DEADLOCKS`, people currently compiling with that to use `DETECT_DEADLOCKS` may want to remove the `DEBUG_THREADS` option.

Yeah, true.  Standby.

jcolp (approved):

Issue #335, opened 2023-09-15
[bug]: res_pjsip_pubsub: The bad_event unit test causes a SEGV in build_resource_tree


gtjoseph:
### Severity

Trivial

### Versions

18,20,21,master

### Components/Modules

res_pjsip_pubsub

### Operating Environment

any

### Frequency of Occurrence

None

### Issue Description

The  build_resource_tree() function was refactored in commit 659f2aae3af to use handler->body_type to test for whether a resource tree needs children.  The test_handler used in the module's unit tests however, were not updated to set body_type to anything so it was left as NULL.  The result is a SEGV.

### Relevant log output

```shell
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __strcmp_avx2_rtm () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:283
283		VMOVU	(%rdi), %ymm0
[Current thread is 1 (Thread 0x7f514b83d840 (LWP 3709485))]
Missing separate debuginfos, use: dnf debuginfo-install glib2-2.76.5-1.fc38.x86_64 libqb-2.0.8-1.fc38.x86_64 python3-libs-3.11.4-1.fc38.x86_64
(gdb) bt
#0  __strcmp_avx2_rtm () at ../sysdeps/x86_64/multiarch/strcmp-avx2.S:283
#1  0x00007f5148a117be in build_resource_tree
    (endpoint=0x0, handler=0x7f5148a2ad80 <test_handler>, resource=0x7f5148a24859 "foo", tree=0x29b9ed0, has_eventlist_support=1, rdata=0x0)
    at res_pjsip_pubsub.c:1168
#2  0x00007f5148a202e9 in bad_event (info=0x281c960, cmd=TEST_EXECUTE, test=0x281c960) at res_pjsip_pubsub.c:5750
#3  0x00000000005e60eb in test_execute (test=0x281c960) at test.c:297
#4  0x00000000005e757c in test_execute_multiple (name=0x0, category=0x21edb26 "/res/res_pjsip_pubsub/", cli=0x7fff1eede040) at test.c:688
#5  0x00000000005e8b9b in test_cli_execute_registered (e=0x775528 <test_cli+200>, cmd=-4, a=0x7fff1eede040) at test.c:1156
#6  0x00000000004d27d0 in ast_cli_command_full (uid=-1, gid=-1, fd=1, s=0x23cfac0 "test execute category /res/res_pjsip_pubsub/ ") at cli.c:3018
[...]

PR #336, opened 2023-09-15
res_pjsip_pubsub: Add body_type to test_handler for unit tests

gtjoseph:
The ast_sip_subscription_handler "test_handler" used for the unit
tests didn't set "body_type" so the NULL value was causing
a SEGV in build_subscription_tree().  Itr's now set to "".

Resolves: #335


InterLinked1 (changes requested):
There's an extra 'r' in the commit message.

gtjoseph:
cherry-pick-to: 21

Issue #337, opened 2023-09-15
[bug]: asterisk.c: The CLI history file is written to the wrong directory in some cases


gtjoseph:
### Severity

Trivial

### Versions

18,20,21,master

### Components/Modules

asterisk.c

### Operating Environment

any

### Frequency of Occurrence

Frequent

### Issue Description

The CLI .asterisk_history file is read from/written to the directory specified by the HOME environment variable.  If the root user starts asterisk with the -U/-G options, or with runuser/rungroup set in asterisk.conf, the asterisk process is started as root but then it calls setuid/setgid to set the new user/group.  This does NOT reset the HOME environment variable to the new user's home directory though so it's still left as "/root".  In this case, the new user will almost certainly NOT have access to read from or write to the history file.  We should be getting the new home directory by calling getpwuid(geteuid()) and using pw->dir.

If you're not a developer running asterisk in the foreground, this probably doesn't apply to you.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #338, opened 2023-09-15
asterisk.c: Use the euid's home directory to read/write cli history

gtjoseph:
The CLI .asterisk_history file is read from/written to the directory
specified by the HOME environment variable. If the root user starts
asterisk with the -U/-G options, or with runuser/rungroup set in
asterisk.conf, the asterisk process is started as root but then it
calls setuid/setgid to set the new user/group. This does NOT reset
the HOME environment variable to the new user's home directory
though so it's still left as "/root". In this case, the new user
will almost certainly NOT have access to read from or write to the
history file.

* Added function process_histfile() which calls
  getpwuid(geteuid()) and uses pw->dir as the home directory
  instead of the HOME environment variable.
* ast_el_read_default_histfile() and ast_el_write_default_histfile()
  have been modified to use the new process_histfile()
  function.

Resolves: #337

### DIFF
--- a/main/asterisk.c
+++ b/main/asterisk.c
@@ -3172,26 +3172,41 @@ static int ast_el_read_history(const char *filename)
 	return history(el_hist, &ev, H_LOAD, filename);
 }
 
+static void process_histfile(int (*readwrite)(const char *filename))
+{
+	struct passwd *pw = getpwuid(geteuid());
+	int ret = 0;
+	char *name = NULL;
+
+	if (!pw || ast_strlen_zero(pw->pw_dir)) {
+		ast_log(LOG_ERROR, "Unable to determine home directory.  History read/write disabled.\n");
+		return;
+	}
+
+	ret = ast_asprintf(&name, "%s/.asterisk_history", pw->pw_dir);
+	if (ret <= 0) {
+		ast_log(LOG_ERROR, "Unable to create history file name.  History read/write disabled.\n");
+		return;
+	}
+
+	ret = readwrite(name);
+	if (ret < 0) {
+		ast_log(LOG_ERROR, "Unable to read or write history file '%s'\n", name);
+	}
+
+	ast_free(name);
+
+	return;
+}
+
 static void ast_el_read_default_histfile(void)
 {
-	char histfile[80] = "";
-	const char *home = getenv("HOME");
-
-	if (!ast_strlen_zero(home)) {
-		snprintf(histfile, sizeof(histfile), "%s/.asterisk_history", home);
-		ast_el_read_history(histfile);
-	}
+	process_histfile(ast_el_read_history);
 }
 
 static void ast_el_write_default_histfile(void)
 {
-	char histfile[80] = "";
-	const char *home = getenv("HOME");
-
-	if (!ast_strlen_zero(home)) {
-		snprintf(histfile, sizeof(histfile), "%s/.asterisk_history", home);
-		ast_el_write_history(histfile);
-	}
+	process_histfile(ast_el_write_history);
 }
 
 static void ast_remotecontrol(char *data)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

InterLinked1 (approved):
Would be nice if the common logic were combined, but seems simple enough...

jcolp (dismissed):

main/asterisk.c
```diff
@@ -3174,22 +3174,22 @@ static int ast_el_read_history(const char *filename)
...
 {
-	char histfile[80] = "";
-	const char *home = getenv("HOME");
+	char histfile[PATH_MAX] = "";
+	struct passwd *pw = getpwuid(geteuid());
 
-	if (!ast_strlen_zero(home)) {
-		snprintf(histfile, sizeof(histfile), "%s/.asterisk_history", home);
+	if (!ast_strlen_zero(pw->pw_dir)) {
+		snprintf(histfile, sizeof(histfile), "%s/.asterisk_history", pw->pw_dir);
```
seanbright, line 3181: This is one of those times where I think we _should_ care if truncation occurs.

main/asterisk.c
```diff
@@ -3174,22 +3174,22 @@ static int ast_el_read_history(const char *filename)
...
```
seanbright, line 3192: Idem

seanbright (changes requested):

gtjoseph:
> Would be nice if the common logic were combined, but seems simple enough...

Done :)

main/asterisk.c
```diff
@@ -3172,26 +3172,43 @@ static int ast_el_read_history(const char *filename)
...
+	size_t bufsize = 0;
+	char *name = NULL;
+
+	if (ast_strlen_zero(pw->pw_dir)) {
+		return NULL;
+	}
+
+	/* pw_dir + '/' + fn + NULL */
+	bufsize = strlen(pw->pw_dir) + 1 + strlen(fn) + 1;
+	name = ast_malloc(bufsize);
```
InterLinked1, line 3188: What if malloc fails?
I'm not sure I like the idea of adding an allocation rather than taking a buffer, is this mainly to deal with the truncation? I guess it's only done once per session so not really a big deal.

InterLinked1 (changes requested):

main/asterisk.c
InterLinked1: Personally, my preference would be to use ast_asprintf instead of ast_malloc + NULL check + snprintf. It's not used as much but there is an ast_ API for it so that would simplify a bit.
At the very least, snprintf is unnecessary since we know the buffer is large enough, sprintf can be safely used.

main/asterisk.c
```diff
@@ -3172,26 +3172,46 @@ static int ast_el_read_history(const char *filename)
 	return history(el_hist, &ev, H_LOAD, filename);
 }
 
+static char *get_default_histfile_name(void)
+{
+	const char *fn = ".asterisk_history";
+	struct passwd *pw = getpwuid(geteuid());
```
seanbright, line 3178: `getpwuid()` can return `NULL`

main/asterisk.c
seanbright: Yes, use `ast_asprintf()` instead of manually allocating.

main/asterisk.c
```diff
@@ -3172,26 +3172,46 @@ static int ast_el_read_history(const char *filename)
 	return history(el_hist, &ev, H_LOAD, filename);
 }
 
+static char *get_default_histfile_name(void)
```
seanbright, line 3175: Just a thought - have this take a function pointer (`int (*readwrite)(const char *filename)`) and pass in either `ast_el_read_history` or `ast_el_write_history` depending on where you call it from. Then you can put all the memory allocation & freeing in one place rather than relying on the caller to do it. Something like:

```
/* rename "get_default_histfile_name" to "process_histfile" or something like that */

static void ast_el_read_default_histfile(void)
{
    process_histfile(ast_el_read_history);
}

static void ast_el_write_default_histfile(void)
{
    process_histfile(ast_el_write_history);
}
```

seanbright (dismissed):

seanbright (approved):

PR #339, opened 2023-09-19
Fix null pointer comparison in stasis_topic_pool_delete_topic

Yashyk456:
This pull request addresses an issue in the stasis_topic_pool_delete_topic function, where a null pointer comparison was causing unexpected behavior. Specifically, the comparison was performed on a potentially null pointer, pool_topic_name, which needed to be checked for null before invoking strncmp.

To fix this issue, I have added a null check for pool_topic_name before the strncmp comparison. This ensures that the comparison only occurs when pool_topic_name is not null, preventing any potential null pointer dereference issues.

This change ensures the code functions correctly and avoids undefined behavior when pool_topic_name is null. Please review and merge this pull request to improve the reliability and stability of the codebase.

REMINDER: This PR should be cherry-picked to the following branches:
- cherry-pick-to: branch-18
- cherry-pick-to: branch-20
- cherry-pick-to: branch-21
- cherry-pick-to: master

Please consider the inclusion of this pull request in the mentioned branches. Thank you.


jcolp:
This can not be reviewed yet as the contributor license agreement has not been signed. It would also be nice to know under which scenario this would actually occur., because underlying topic creation already does checks to make sure that the topic name is non-NULL.

seanbright (commented):
There is other code in this function (see the call to `strlen()` a few lines before your change) that assumes `pool_topic_name` will not be `NULL`.

I would look at call sites and try to figure out why `NULL` is being passed instead.

InterLinked1 (changes requested):
You need to squash the commits together into a single one (get rid of the merge commit), and the commit message does not follow the commit message guidelines.

Additionally, the cherry pick statements need to be in a comment as outlined in the docs.

Yashyk456:
@jcolp The crash occurred while the program was trying to delete a topic from the Stasis topic pool. It seems that the topic name passed to the function stasis_topic_pool_delete_topic() was an invalid memory address , which caused the program to crash when the __strncmp_avx2() function was called to compare the topic name with the names of topics in the pool.

Basically when we get two simultaneous requests to delete the same bridge then there is a possibility .


Even if the scenario has not come  we can't compare null in strncmp. We got the case when same request came 2 time


Also @jcolp let me know from where i can sign the contributor license agreement.


@InterLinked1 squashing commit won't be possible as the last and the current commit have 151 commit in between

jcolp:
This is not the correct fix then, and is just a bandaid trying to fix another issue. The real underlying issue should be resolved.

jcolp:
The link to sign the contributor license agreement is in the list of checks:

https://oss-cla.sangoma.com/asterisk/asterisk?pullRequest=339

I am against accepting this change though because it doesn't address the underlying true issue.

Yashyk456:
> This is not the correct fix then, and is just a bandaid trying to fix another issue. The real underlying issue should be resolved. The code has a contract that the name is not NULL, and this should not be violated - which is why it doesn't do a NULL check. If it's NULL then something else is behaving incorrectly and THAT should be fixed.




@jcolp Thanks for your feedback. I'd like to address your concern regarding the null check for pool_topic_name before comparing it to the stasis_topic_pool_delete_topic function.

In a multithreaded environment, there's a possibility that different threads may attempt to delete the same bridge simultaneously. This could lead to a scenario where pool_topic_name becomes invalid or null during execution, which is why I added the null check.

While I understand that there is a contract assuming pool_topic_name should not be null, real-world scenarios can sometimes deviate from ideal conditions. Ensuring robust code means accounting for unexpected situations and handling them gracefully.

By adding this null check, we can ensure that the code is robust in various situations, including those where pool_topic_name might unexpectedly become null. It provides an extra layer of safety to prevent potential issues and crashes.

In conclusion, while the contract assumes that pool_topic_name should not be null, real-world scenarios can be unpredictable. It's good practice to include defensive programming measures like this null check to enhance the stability and reliability of the codebase.

Yashyk456:
> The link to sign the contributor license agreement is in the list of checks:
> 
> https://oss-cla.sangoma.com/asterisk/asterisk?pullRequest=339
> 
> I am against accepting this change though because it doesn't address the underlying true issue.

ALready Signed it

jcolp:
> In a multithreaded environment, there's a possibility that different threads may attempt to delete the same bridge simultaneously. This could lead to a scenario where pool_topic_name becomes invalid or null during execution, which is why I added the null check.

And in this scenario it should not be possible for pool_topic_name to become invalid or NULL and ultimately crash here. A NULL check doesn't help if it's invalid, and if two threads are operating on the bridge at the same time it may or may not solve the crash depending on the timing. The comment @seanbright gave is also applicable - there is code further up that assumes non-NULL and should have crashed. The behavior is undefined at this point and defensive code can only go so far, and is not a true fix for the underlying problem. Something else could then crash because of an assumption based on the contract.

I like defensive code to a point, but there are times where it's not the answer and the contract has to be kept.

jcolp:
Other people can chime in and give their own opinion, they may be disagree.

Yashyk456:
> > In a multithreaded environment, there's a possibility that different threads may attempt to delete the same bridge simultaneously. This could lead to a scenario where pool_topic_name becomes invalid or null during execution, which is why I added the null check.
> 
> And in this scenario it should not be possible for pool_topic_name to become invalid or NULL and ultimately crash here. A NULL check doesn't help if it's invalid, and if two threads are operating on the bridge at the same time it may or may not solve the crash depending on the timing. The comment @seanbright gave is also applicable - there is code further up that assumes non-NULL and should have crashed. The behavior is undefined at this point and defensive code can only go so far, and is not a true fix for the underlying problem. Something else could then crash because of an assumption based on the contract.
> 
> I like defensive code to a point, but there are times where it's not the answer and the contract has to be kept.



@jcolp,

Thanks for your review and feedback; I appreciate it.

I share your concern about finding the root cause of this issue. I'll dig deeper to investigate it further. If you're also looking into it, your insights would be valuable.

Additionally, I encourage anyone else who has thoughts on this matter to join the discussion and share their perspectives.

jcolp:
I'm not actively working on such an issue.

InterLinked1:
> @InterLinked1 squashing commits won't be possible as the last and the current commit have 151 commits in between please let me know what can i do.

Then you're not using Git correctly. The easiest way is to make sure you have a current upstream and apply your patch against that. You can also probably do it in more complicated ways, but that's by far the easiest way. This patch is trivial enough you can just start over and force push again.

And to reiterate, you need to add an actual commit message, the PR description is irrelevant.

> Other people can chime in and give their own opinion, they may be disagree.

I agree with Josh, this would just mask the problem further and make it less likely to get fixed in the future, and there could be other paths affected by whatever it is. Better to fix it properly now. If the real issue isn't known, that's what should be investigated.

If anything, I think an assertion could maybe be added here, rather than doing a NULL check. But...

> The code has a contract that the name is not NULL, and this should not be violated - which is why it doesn't do a NULL check.

@jcolp ...any reason we couldn't formalize this by adding the `nonnull` gcc attribute, at least for versions that support it? Could be helpful in catching misusage, either at compile time or with static analysis.

jcolp:
> ..any reason we couldn't formalize this by adding the `nonnull` gcc attribute, at least for versions that support it? Could be helpful in catching misusage, either at compile time or with static analysis.

The change in question is in regards to a field on a structure. The attribute is for function parameters, and thus would not be applicable to this.

InterLinked1:
> > ..any reason we couldn't formalize this by adding the `nonnull` gcc attribute, at least for versions that support it? Could be helpful in catching misusage, either at compile time or with static analysis.
> 
> The change in question is in regards to a field on a structure. The attribute is for function parameters, and thus would not be applicable to this.

Ah, yes, I conflated `topic_name` with `pool_topic_name`, my bad.
In that case, I guess an assertion is probably all we could do, though it'll crash either way, so it wouldn't really give us anything new.

jcolp:
I'm actually just going to close this pull request out since now myself and someone else have agreed this is not the correct path forward. Discussion can still occur here if needed.

PR #342, opened 2023-09-21
func_json: Fix crashes for some types

btriller:
This commit fixes crashes in JSON_DECODE() for types null, true, false and real numbers.

In addition it ensures that a path is not deeper than 32 levels.

Also allow root object to be an array.

Add unit tests for above cases.


btriller:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

InterLinked1 (approved):

Issue #303, opened 2023-09-06
[bug]: SpeechBackground never exits


wphowell:
### Severity

Minor

### Versions

18.19.0

### Components/Modules

app_speech_utils

### Operating Environment

Debian 11

### Frequency of Occurrence

Occasional

### Issue Description

If the WebSocket connect terminates while in SpeechBackground, it will never exit -- doesn't honour set timeout.  A channel is left open, the dial plan doesn't continue and an Asterisk restart is necessary to clear the issue.

### Relevant log output

```shell
2023-09-05 15:43:43.156 - res_http_websocket.c: WebSocket connection to '127.0.0.1:9099' closed

"core show channels concise"
PJSIP/anonymous-00000d19!wait-audio!s!2!Up!SpeechBackground!/audio/optimhearing/7d6ab972-6d5b-4846-a9cd-e4c911fdd16d/prod/4bb4de9e-b27e-4cdd-956b-bc7be650df1d,6!16083163537!d2439c79-2af6-42fb-bf11-5f8aee4b85d2!d2439c79-2af6-42fb-bf11-5f8aee4b85d2!3!35505!!p002-1693928603.6677
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

wphowell:
This can also occur in SpeechDestroy as well with the channel being hung forever.

[2023-09-06 18:16:58.989] VERBOSE[1328307][C-00002397] pbx.c: Executing [fcc4b571-aed8-4ae8-945d-d735491961ef@speech-disa:47] SpeechDestroy("PJSIP/anonymous-000023a7", "") in new stack
[2023-09-06 18:16:58.989] VERBOSE[1328307][C-00002397] res_http_websocket.c: WebSocket connection to '127.0.0.1:9099' closed

"core show channels concise"
PJSIP/anonymous-000023a7!speech-disa!fcc4b571-aed8-4ae8-945d-d735491961ef!47!Up!SpeechDestroy!!19192149275!fcc4b571-aed8-4ae8-945d-d735491961ef!fcc4b571-aed8-4ae8-945d-d735491961ef!3!10089!!p001-1694024215.18208

PR #343, opened 2023-09-21
res_speech_aeap: add aeap error handling

mbradeen:
res_speech_aeap previously did not register an error handler
with aeap, so it was not notified of a disocnnect. This resulted
in SpeechBackground never exiting upon a websocket disconnect.

Resolves: #303


mbradeen:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (changes requested):
Commit message should be "disconnect", one is misspelled.

jcolp (approved):

Issue #345, opened 2023-09-22
[improvement]: Increase pj_sip Realm Size to 255 Characters for Improved Functionality


pchero:
### Improvement Description

Currently, the endpoint and relevant ID length for the pj_sip real-time database are confined to a restrictive limit of 40 characters. This limitation presents various challenges, including the inability to fully specify domain length and constraints on the practical utility of the realm.

To remedy these issues and enhance functionality, we recommend expanding the realm size from its current 40-character limit to a more generous 255 characters.

PR #346, opened 2023-09-22
pjsip: Increase pj_sip Realm Size

pchero:
The pjsip's realm size for pj_sip is set at a restrictive limit of 40 characters.
This limitation poses several drawbacks, such as preventing the full qualification of domain length and limiting the practical use of the realm.
To address these issues and enhance functionality, we propose increasing the realm size from its current limit of 40 characters to a more accommodating size of 255 characters.

Resolves: #345

### DIFF
--- a/res/res_pjsip/pjsip_distributor.c
+++ b/res/res_pjsip/pjsip_distributor.c
@@ -42,7 +42,7 @@ static pjsip_module distributor_mod = {
 struct ast_sched_context *prune_context;
 
 /* From the auth/realm realtime column size */
-#define MAX_REALM_LENGTH 40
+#define MAX_REALM_LENGTH 255
 static char default_realm[MAX_REALM_LENGTH + 1];
 
 #define DEFAULT_SUSPECTS_BUCKETS 53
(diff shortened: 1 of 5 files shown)


res/res_pjsip/pjsip_distributor.c
```diff
@@ -42,7 +42,7 @@ static pjsip_module distributor_mod = {
 struct ast_sched_context *prune_context;
 
 /* From the auth/realm realtime column size */
-#define MAX_REALM_LENGTH 40
+#define MAX_REALM_LENGTH 255
```
InterLinked1, line 45: I know these are duplicated to begin with, but could we have a single definition in a header file, that could then be used in all 3 places, to ensure they all match?

InterLinked1 (changes requested):
Commit message header and first sentence should use 'res_pjsip' as the prefix, not 'pjsip'.

Also, this commit does not propose an increase, it increases it, so it would be more direct to say "this increases the realm size" as opposed to "we propose increasing the realm size".

res/res_pjsip/pjsip_distributor.c
pchero: Hi InterLiked1,

Thanks, btw, I think I forked from the old one. I will sync the repository and will submit the PR again.

Issue #349, opened 2023-09-25
[improvement]: Add libjwt to third-party


gtjoseph:
### Improvement Description

The current STIR/SHAKEN implementation is not currently usable due to encryption issues.  Rather than trying to futz with OpenSSL and the the current code, we can take advantage of the existing capabilities of libjwt but we first need to add it to the third-party infrastructure already in place for jansson and pjproject.

seanbright:
Please describe the “encryption issues.”

gtjoseph:
> Please describe the “encryption issues.”

I'm not going to get this exactly right but here goes...

STIR/SHAKEN depends JSON Web Tokens for encapsulation and on EC/ecdsa for signatures.  Normally a EC signature produced by OpenSSL has the (R,S) values prepended to the signature but S/S requires a "raw" signature.  Stripping the (R,S) from the signature is not a simple matter of chopping the first 8 bytes off though. To get it right, we'd have to implement the same code that's already in [benmcollins/libjwt](https://github.com/benmcollins/libjwt).  We also did our own implementation of JWT for res_stir_shaken which is somewhat fragile so moving to libjwt kills two birds with one stone.  libjwt is also used by [libstirshaken](https://github.com/signalwire/libstirshaken) which is used by freeswitch.  libstirshaken is basically a plumbing layer a UA can use but since we already have the plumbing in place, we don't need that layer.


References:
https://stackoverflow.com/questions/44807170/openssl-ecdsa-signatures-longer-than-expected
https://www.google.com/search?q=ecdsa+signatue+64+vs+71+bytes&sourceid=chrome&ie=UTF-8

PR #350, opened 2023-09-25
Add libjwt to third-party

gtjoseph:
The current STIR/SHAKEN implementation is not currently usable due
to encryption issues. Rather than trying to futz with OpenSSL and
the the current code, we can take advantage of the existing
capabilities of libjwt but we first need to add it to the
third-party infrastructure already in place for jansson and
pjproject.

A few tweaks were also made to the third-party infrastructure as
a whole.  The jansson "dest" install directory was renamed "dist"
to better match convention, and the third-party Makefile was updated
to clean all product directories not just the ones currently in
use.

Resolves: #349


gtjoseph:
This isn't going to cherry-pick cleanly to the other branches because of the changes to configure.  Will cherry-pick manually after approval.

seanbright (dismissed):
Just want to preemptively reject until my questions on the issue is answered. Please do not rush this in.

gtjoseph:
> Just want to preemptively reject until my questions on the issue is answered. Please do not rush this in.

Answered in https://github.com/asterisk/asterisk/issues/349#issuecomment-1734316804

Here's the text:

I'm not going to get this exactly right but here goes...

STIR/SHAKEN depends JSON Web Tokens for encapsulation and on EC/ecdsa for signatures.  Normally a EC signature produced by OpenSSL has the (R,S) values prepended to the signature but S/S requires a "raw" signature.  Stripping the (R,S) from the signature is not a simple matter of chopping the first 8 bytes off though. To get it right, we'd have to implement the same code that's already in [benmcollins/libjwt](https://github.com/benmcollins/libjwt).  We also did our own implementation of JWT for res_stir_shaken which is somewhat fragile so moving to libjwt kills two birds with one stone.  libjwt is also used by [libstirshaken](https://github.com/signalwire/libstirshaken) which is used by freeswitch.  libstirshaken is basically a plumbing layer a UA can use but since we already have the plumbing in place, we don't need that layer.


References:
https://stackoverflow.com/questions/44807170/openssl-ecdsa-signatures-longer-than-expected
https://www.google.com/search?q=ecdsa+signatue+64+vs+71+bytes&sourceid=chrome&ie=UTF-8

gtjoseph:
Actually, just thought of something else...  We only implemented the EVP_sha256 algo but libjwt supports sha384 and sha512 as well, so we get naother benefit.

gtjoseph:
I get scared when you respond with a single word let alone a single letter. 😊 Does "K" really mean you fine with the explanation or does it mean you're not OK and will be responding later?

seanbright:
Im good. Thanks for the explanation.

gtjoseph:
Moving to draft because one of the configure checks isn't working correctly.

gtjoseph:
actually it was working fine but i made a small update so that we don't have to hard code the libjwt version in both ./configure.ac and third-party/versions.mak.

jcolp (approved):
This needs review by someone else with more build experience than I as well.

Issue #352, opened 2023-09-26
[bug]: Update qualify_timeout documentation to include DNS note


mbradeen:
### Severity

Trivial

### Versions

all

### Components/Modules

res_pjsip

### Operating Environment

all using bundled pjsip

### Frequency of Occurrence

None

### Issue Description

The qualify_timeout on a pjsip endpoint includes time for any required DNS lookups.  This should be explicit in the documentation.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #353, opened 2023-09-26
res_pjsip: update qualify_timeout documentation with DNS note

mbradeen:
The documentation on qualify_timeout does not explicitly state that the timeout
includes any time required to perform any needed DNS queries on the endpoint.

If the OPTIONS response is delayed due to the DNS query, it can still render an
endpoint as Unreachable if the net time is enough for qualify_timeout to expire.

Resolves: #352


mbradeen:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

InterLinked1 (approved):

jcolp (approved):

Issue #356, opened 2023-09-27
[new-feature]: app_directory: Add ADSI support.


InterLinked1:
### Feature Description

This adds optional ADSI support to the Directory application, which allows callers with supporting CPE to navigate the application and find the desired entry significantly faster than is possible purely using the existing audio prompts.

PR #357, opened 2023-09-27
app_directory: Add ADSI support to Directory.

InterLinked1:
This adds optional ADSI support to the Directory
application, which allows callers with ADSI CPE
to navigate the Directory system significantly
faster than is possible using the audio prompts.
Callers can see the directory name (and optionally extension) on their screenphone and confirm or
reject a match immediately rather than waiting
for it to be spelled out, enhancing usability.

Resolves: #356

### DIFF
--- a/apps/app_directory.c
+++ b/apps/app_directory.c
@@ -39,6 +39,7 @@
 #include "asterisk/say.h"
 #include "asterisk/app.h"
 #include "asterisk/utils.h"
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #360, opened 2023-10-02
[improvement]: Update documentation for CHANGES/UPGRADE files


kobaz:
### Improvement Description

Problem: Traditionally modifications to CHANGES or UPGRADE would each go into their respective directories as an individual text file.  Current CHANGES/UPGRADE files include old instructions on how to modify said files.

Solution:
Update the documentation header in these two files to point to the wiki

jcolp:
> FYI this change broke the link referenced in the following wiki page: https://docs.asterisk.org/Asterisk_20_Documentation/WhatsNew/

For documentation things, please open an issue in the documentation repo referencing this.

PR #361, opened 2023-10-02
Remove UPGRADE.txt and CHANGES

intellasoft:
Fixes: #360


kobaz:
cherry-pick-to: 20
cherry-pick-to: 18

gtjoseph (dismissed):
Actually, these 2 files need to be deleted as they are no longer even updated. :)

kobaz:
For historical/handyness purposes would it be applicable to add back in the updates to those files instead?

jcolp:
What updates are you referring to? The release process creates a CHANGES.md in releases.

kobaz:
I was referring to the base file called CHANGES which has instructions that point to a file that no longer exists as well as UPGRADE.txt which has instructions that point to a file that no longer exists


There is no CHANGES.md file so far that I can see that's part of the git repo in master.  And the only changes related files that exist, appear to be the old 'CHANGES" file.  There is however a CHANGES.md file in branch 18 for example.

jcolp:
Okay, but what does this mean:

"For historical/handyness purposes would it be applicable to add back in the updates to those files instead?"

Add back in what updates to what files

jcolp:
The files themselves are no longer used as George mentioned, and this is a git repo so we can always go back to examine past history if need be. I don't see why they need to remain.

seanbright:
This PR should be updated to delete the files as mentioned by @gtjoseph or it should be closed

kobaz:
Done

jcolp (approved):

gtjoseph:
Git doesn't like cherry-picking to delete files when when file contents differ.  Once the master and 21 versions merge, I'll manually merge the 20 and 18 versions.

Issue #362, opened 2023-10-02
[improvement]: Speed up ARI command processing


mbradeen:
### Improvement Description

res_statsis currently has a polling interval that is not escaped when new commands are issued.  The suggested change is to signal the sleeping thread when new commands are enqueued so it can wake and process the command immediately.

PR #363, opened 2023-10-02
res_stasis: signal when new command is queued

mbradeen:
res_statsis's app loop sleeps for up to .2s waiting on input
to a channel before re-checking the command queue. This can
cause delays between channel setup and bridge.

This change is to send a SIGURG on the sleeping thread when
a new command is enqueued. This exits the sleeping thread out
of the ast_waitfor() call triggering the new command being
processed on the channel immediately.

Resolves: #362

UserNote: Call setup times should be significantly improved
when using ARI.

### DIFF
--- a/res/stasis/control.c
+++ b/res/stasis/control.c
@@ -93,6 +93,10 @@ struct stasis_app_control {
 	 */
 	char *next_app;
 	/*!
+	 * The thread currently blocking on the channel.
+	 */
+	pthread_t control_thread;
+	/*!
 	 * The list of arguments to pass to StasisStart when moving to another app.
 	 */
 	AST_VECTOR(, char *) next_app_args;
@@ -156,6 +160,8 @@ struct stasis_app_control *control_create(struct ast_channel *channel, struct st
 	control->next_app = NULL;
 	AST_VECTOR_INIT(&control->next_app_args, 0);
 
+	control_set_thread(control, AST_PTHREADT_NULL);
+
 	return control;
 }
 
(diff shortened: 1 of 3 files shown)


mbradeen:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

gtjoseph (dismissed):

res/stasis/control.c
```diff
@@ -295,6 +306,12 @@ static struct stasis_app_command *exec_command_on_condition(
 	ast_cond_signal(&control->wait_cond);
 	ao2_unlock(control->command_queue);
 
+	if(control->control_thread != AST_PTHREADT_NULL) {
```
InterLinked1, line 309: Needs space after if

InterLinked1 (changes requested):

res/stasis/control.c
mbradeen: thanks, fixed.

res/stasis/control.c
```diff
@@ -185,6 +191,11 @@ static void app_control_unregister_rule(
 	ao2_unlock(control->command_queue);
 }
 
+void control_set_thread(struct stasis_app_control *control, pthread_t threadid)
+{
+	control->control_thread = threadid;
```
jcolp, line 196: I think this should be protected by the command_queue lock.

jcolp (changes requested):

res/stasis/control.c
```diff
@@ -295,6 +308,12 @@ static struct stasis_app_command *exec_command_on_condition(
 	ast_cond_signal(&control->wait_cond);
 	ao2_unlock(control->command_queue);
 
+	if (control->control_thread != AST_PTHREADT_NULL) {
```
mbradeen, line 311: @jcolp should the command_queue lock above be extended through this if?

jcolp: Yes

jcolp (approved):

PR #369, opened 2023-10-07
ari/stasis: Indicate progress before playback on a bridge

zecke:
Make it possible to start a playback and the calling party to receive audio on a bridge before the call is connected.

Model the implementation after play_on_channel and deliver a AST_CONTROL_PROGRESS before starting the playback.

For a pjsip channel this will result in sending a SIP 183 Session Progress.

### DIFF
--- a/res/ari/resource_bridges.c
+++ b/res/ari/resource_bridges.c
@@ -425,6 +425,12 @@ static void ari_bridges_play_new(const char **args_media,
 	struct bridge_channel_control_thread_data *thread_data;
 	pthread_t threadid;
 
+	struct ast_frame prog = {
+		.frametype = AST_FRAME_CONTROL,
+		.subclass.integer = AST_CONTROL_PROGRESS,
+	};
+
+
 	if (!(play_channel = prepare_bridge_media_channel("Announcer"))) {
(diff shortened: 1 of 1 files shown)


zecke:
cherry-pick-to: 18
cherry-pick-to: 20

InterLinked1 (commented):
I'm not sure what this change has to do with PJSIP specifically, it seems completely channel tech agnostic to me. But if you want to keep that, might want to capitalize PJSIP.

Also is this not applicable to 21, since you didn't cherry pick to it?

zecke:
> I'm not sure what this change has to do with PJSIP specifically, it seems completely channel tech agnostic to me. But if you want to keep that, might want to capitalize PJSIP.

Updated. The driver for this change is to play audio on a bridge with a PJSIP channel before the call is connected. By sending the SIP 183 progress with the SDP file we provide an initial answer to what has been offered.

After looking at the simple bridge, topology changed information, how this is done for plain channels. It seemed that this is an appropriate place.

 
> Also is this not applicable to 21, since you didn't cherry pick to it?

Done.

jcolp (approved):

PR #371, opened 2023-10-09
res_crypto: Added OpenSSL convenience wrappers

gtjoseph:
Added wrappers to register and retrieve locally registered X.509
certificate extensins.

Added wrappers to create an X.509 Certitifate store and to load it
from a file or directory.

Added wrappers to load certificates from files or memory, validate
dates and trust, and to get extension data.

### DIFF
--- a/include/asterisk/crypto.h
+++ b/include/asterisk/crypto.h
@@ -27,6 +27,11 @@
 extern "C" {
 #endif
 
+#ifdef HAVE_OPENSSL
+#include "openssl/x509.h"
+#include "openssl/x509_vfy.h"
+#endif
+
 #include "asterisk/optional_api.h"
 #include "asterisk/logger.h"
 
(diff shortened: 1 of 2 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

res/res_crypto.c
```diff
@@ -797,6 +792,383 @@ int AST_OPTIONAL_API_NAME(ast_aes_decrypt)(const unsigned char *in, unsigned cha
...
+	text = ast_malloc(len + 1);
+	if (!text) {
+		BIO_free(bio);
+		return NULL;
+	}
+	memcpy(text, buf, len);
+	text[len] = '\0';
+	BIO_free(bio);
+	return text;
+}
```
seanbright, line 817: I'm certainly biased but I'm partial to this implementation:

https://github.com/asterisk/asterisk/blob/87e053f6d1809366704de2e4a8a2338cd621d35c/main/tcptls.c#L111-L130

include/asterisk/crypto.h
```diff
@@ -188,6 +197,144 @@ AST_OPTIONAL_API(int, ast_crypto_loaded, (void), { return 0; });
...
+	const char *path);
+
+/*!
+ * \brief Check if the reftime is within the cert's valid dates
+ * \param cert The cert to check
+ * \param reftime to use or 0 to use current time
+ * \retval 1 Cert is valid
+ * \retval 0 Cert is not valid
+ */
+int ast_crypto_is_cert_valid(X509*cert, time_t reftime);
```
seanbright, line 320: Space before the `*`

include/asterisk/crypto.h
```diff
@@ -188,6 +197,144 @@ AST_OPTIONAL_API(int, ast_crypto_loaded, (void), { return 0; });
...
+ * \param nid The NID of the extension
+ *            (0 to search by short_name)
+ * \param short_name The short name of the extension
+ * \note Either nid or short_name may be supplied.  If both are,
+ * nid takes precedence.
+ * \returns The definition of the extension or NULL if not found
+ * \warning Do not attempt to free or otherwise manipulate the
+ * structure returned or its members.
+ */
+struct ast_X509_Extension * ast_crypto_get_registered_extension(
```
seanbright, line 224: No space after the `*`

res/res_crypto.c
```diff
@@ -797,6 +792,383 @@ int AST_OPTIONAL_API_NAME(ast_aes_decrypt)(const unsigned char *in, unsigned cha
...
+	}
+
+	return ext->nid;
+}
+
+ASN1_OCTET_STRING *ast_crypto_get_cert_extension_data(X509 *cert,
+	int nid, const char *short_name)
+{
+	int ex_idx;
+	X509_EXTENSION* ex;
```
seanbright, line 944: `X509_EXTENSION *ex;`

seanbright (dismissed):

include/asterisk/crypto.h
seanbright, line 320: Maybe name this `ast_crypto_is_cert_time_valid(...)` or something? `..._is_cert_valid` might be confusing.

include/asterisk/crypto.h
```diff
@@ -188,6 +197,144 @@ AST_OPTIONAL_API(int, ast_crypto_loaded, (void), { return 0; });
...
+	int nid, const char *short_name);
+
+/*!
+ * \brief Check if an extensions is already locally registered
```
bkford, line 228: extension

include/asterisk/crypto.h
gtjoseph: yeah, went back and forth on this. i'll add "time" or "time_range" back in.

res/res_crypto.c
gtjoseph: we have openssl stuff spread out in too many places :(.

it'd be better if you could pass it a message as well.  maybe i can just make it a va_arg.

gtjoseph: made a new function: ast_log_openssl.

gtjoseph:
> It would also be nice if these were available even if `res_crypto.so` wasn't loaded but I don't know how painful that would be for you.

I debated using tcptls.c vs creating a new file in main vs a new module vs res_crypto.  I could easily put them in tcptls but that's more focused on I/O than certs and encryption.  A new file in main is just as easy though so if you think a new file is better I can move them easily.  BTW,I didn't see this comment until after my last force push. :)

gtjoseph:
I've had a change of heart and am going to move this stuff back in res_crypto.  Having to ifdef the whole source file in case openssl isn't used isn't sitting right.  If we required openssl like we require jansson, it'd make more sense but we don't.

seanbright:
> I've had a change of heart and am going to move this stuff back in res_crypto. 

Yeah that is probably the right call.

gtjoseph:
@jcolp @seanbright 
So should I abandon this PR as well or does it meet the Bright-Colp&reg; standard for usefullness?

Issue #379, opened 2023-10-19
[bug]: Orphaned taskprocessors cause shutdown delays


gtjoseph:
### Severity

Trivial

### Versions

18,20,21,master

### Components/Modules

res_pjsip_exten_state.c,res_pjsip_mwi.c

### Operating Environment

all

### Frequency of Occurrence

Constant

### Issue Description

Commit f66f77fa4f88dbae59761534e last year prevents the res_pjsip_exten_state and res_pjsip_mwi modules from unloading due to possible pjproject asserts if the modules are reloaded.  A side effect of the implementation is that the taskprocessors these modules use aren't being released.  When asterisk is doing a graceful shutdown, it waits AST_TASKPROCESSOR_SHUTDOWN_MAX_WAIT seconds for all taskprocessors to stop but since those 2 modules don't release theirs, the shutdown hangs for that amount of time.  It's not a long wait but it can be annoying when you're doing things like using the leak sanitizer since that requires a graceful shutdown.

### Relevant log output

```shell
ERROR[1993568]: taskprocessor.c:338 tps_shutdown: Asertion may occur, the following taskprocessors are still runing:
ERROR[1993568]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/exten_state'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-00000068'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-00000069'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-0000006a'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-0000006b'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-0000006c'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-0000006d'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-0000006e'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'pjsip/mwi-0000006f'
ERROR[1983902]: taskprocessor.c:349 tps_shutdown: taskprocessor 'stasis/p:mwi:all/1171@default-00000079'
[...]

PR #380, opened 2023-10-19
res_pjsip_exten_state,res_pjsip_mwi: Allow unload on shutdown

gtjoseph:
Commit f66f77f last year prevents the res_pjsip_exten_state and
res_pjsip_mwi modules from unloading due to possible pjproject
asserts if the modules are reloaded. A side effect of the
implementation is that the taskprocessors these modules use aren't
being released. When asterisk is doing a graceful shutdown, it
waits AST_TASKPROCESSOR_SHUTDOWN_MAX_WAIT seconds for all
taskprocessors to stop but since those 2 modules don't release
theirs, the shutdown hangs for that amount of time.

This change allows the modules to be unloaded and their resources to
be released when ast_shutdown_final is true.

Resolves: #379

### DIFF
--- a/res/res_pjsip_exten_state.c
+++ b/res/res_pjsip_exten_state.c
@@ -973,33 +973,35 @@ static int publisher_stop(struct ast_sip_outbound_publish_client *client)
 
 static int unload_module(void)
 {
-#if 0
-	ast_sip_unregister_event_publisher_handler(&dialog_publisher);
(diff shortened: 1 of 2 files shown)


jcolp (approved):

gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

PR #381, opened 2023-10-22
ARI/stasis: Indicate the diversion target and make it available in hangup request

zecke:
Add the `caller_rdnis` attribute to the channel in ARI. Update the channel snapshot before publishing the hangup request to have the latest information.

### DIFF
--- a/main/channel.c
+++ b/main/channel.c
@@ -9119,6 +9119,7 @@ void ast_channel_set_redirecting(struct ast_channel *chan, const struct ast_part
 	ast_channel_lock(chan);
 	ast_party_redirecting_set(ast_channel_redirecting(chan), redirecting, update);
 	ast_channel_snapshot_invalidate_segment(chan, AST_CHANNEL_SNAPSHOT_INVALIDATE_CALLER);
+	ast_channel_publish_snapshot(chan);
 	ast_channel_unlock(chan);
 }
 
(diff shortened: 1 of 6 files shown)


zecke:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

zecke:
> What information was not up to date on the snapshot in the various cases?

The (new) caller_rdnis didn't make it into the ChannelHangupRequest. pjsip calls ast_channel_set_redirecting which sets the bit but the snapshot was not be updated until stasis end.

jcolp:
> > What information was not up to date on the snapshot in the various cases?
> 
> The (new) caller_rdnis didn't make it into the ChannelHangupRequest. pjsip calls ast_channel_set_redirecting which sets the bit but the snapshot was not be updated until stasis end.

The snapshot should be published at the point the value is changed, not at hangup. Doing it at hangup results in unnecessary work on all cases, instead of only when the rdnis value is changed.

zecke:
> > > What information was not up to date on the snapshot in the various cases?
> > 
> > 
> > The (new) caller_rdnis didn't make it into the ChannelHangupRequest. pjsip calls ast_channel_set_redirecting which sets the bit but the snapshot was not be updated until stasis end.
> 
> The snapshot should be published at the point the value is changed, not at hangup. Doing it at hangup results in unnecessary work on all cases, instead of only when the rdnis value is changed.

Thank you. I guessed we want to defer the update as long as possible. Will revise to something like:

```
        ast_channel_snapshot_invalidate_segment(chan, AST_CHANNEL_SNAPSHOT_INVALIDATE_CALLER);
+       ast_channel_publish_snapshot(chan);
```

jcolp (commented):
This should also include test coverage in the testsuite, which should be possible by manually setting the RDNIS using the dialplan function.

jcolp (approved):

Issue #388, opened 2023-10-26
[bug]: Crash in app_followme.c due to not acquiring a reference to nativeformats


creslin287:
### Severity

Trivial

### Versions

master, 21, 20, 18

### Components/Modules

app_followme.c

### Operating Environment

CentOS 7.x

### Frequency of Occurrence

One Time

### Issue Description

When doing an ast_request in app_followme, a reference isn't properly acquired when calling ast_channel_nativeformats(), causing a crash due to nativeformats being changed in the middle of the ast_request call.

### Relevant log output

```shell
(gdb) bt
#0  0x00000000004fa398 in ast_format_get_type (format=0x90) at format.c:356
#1  0x00000000004ffd2e in ast_format_cap_append_from_cap (dst=0x7fb5047b4068, src=0x7fb53884a1e8, type=AST_MEDIA_TYPE_AUDIO) at format_cap.c:278
#2  0x00000000005b6160 in ast_stream_topology_create_from_format_cap (cap=0x7fb53884a1e8) at stream.c:874
#3  0x00000000004ac3ef in request_channel (type=0x7fb4ff3dcc84 "Local", request_cap=0x7fb53884a1e8, topology=0x0, assignedids=0x0, requestor=0x7fb539037350, addr=0x7fb504941328 "202@inside-extensions/n", cause=0x7fb4cc22418c) at channel.c:6336
#4  0x00000000004aca94 in ast_request (type=0x7fb4ff3dcc84 "Local", request_cap=0x7fb53884a1e8, assignedids=0x0, requestor=0x7fb539037350, addr=0x7fb504941328 "202@inside-extensions/n", cause=0x7fb4cc22418c) at channel.c:6434
#5  0x00007fb4ff3d96f4 in findmeexec (tpargs=0x7fb504c24b30, caller=0x7fb539037350) at app_followme.c:1100
#6  0x00007fb4ff3db227 in app_exec (chan=0x7fb539037350, data=0x7fb4cc225900 "201-WestmorelandHealthRehab,ldNb(set-FollowMe,s,1(WestmorelandHealthRehab,,0,,PJSIP/101-WestmorelandHealthRehab-00052b5a))") at app_followme.c:1490
[...]

PR #389, opened 2023-10-26
app_followme.c: Grab reference on nativeformats before using it

creslin287:
Fixes a crash due to a lack of proper reference on the nativeformats
object before passing it into ast_request().  Also found potentially
similar use case bugs in app_chanisavail.c, bridge.c, and bridge_basic.c

Fixes: #388

### DIFF
--- a/apps/app_chanisavail.c
+++ b/apps/app_chanisavail.c
@@ -117,6 +117,7 @@ static int chanavail_exec(struct ast_channel *chan, const char *data)
 	struct ast_str *tmp_availcause = ast_str_alloca(2048);
 	struct ast_channel *tempchan;
 	struct ast_custom_function *cdr_prop_func = ast_custom_function_find("CDR_PROP");
+	struct ast_format_cap *caps = NULL;
 	AST_DECLARE_APP_ARGS(args,
 		AST_APP_ARG(reqchans);
 		AST_APP_ARG(options);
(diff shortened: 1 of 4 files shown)


creslin287:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

apps/app_chanisavail.c
```diff
@@ -175,9 +180,11 @@ static int chanavail_exec(struct ast_channel *chan, const char *data)
 			status = inuse = ast_device_state(trychan);
 		}
 		ast_str_append(&tmp_availstat, 0, "%s%d",
-			ast_str_strlen(tmp_availstat) ? "&" : "", status);
+		ast_str_strlen(tmp_availstat) ? "&" : "", status);
```
jcolp, line 183: Indentation is off here now

apps/app_followme.c
```diff
@@ -1097,8 +1098,15 @@ static struct ast_channel *findmeexec(struct fm_args *tpargs, struct ast_channel
...
-			outbound = ast_request("Local", ast_channel_nativeformats(caller), NULL, caller,
-				tmpuser->dialarg, &dg);
+			/* Capture nativeformats reference in case it gets changed */
+			ast_channel_lock(caller);
+			caps = ao2_bump(ast_channel_nativeformats(caller));
+			ast_channel_unlock(caller);
+
+			outbound = ast_request("Local", caps, NULL, caller, tmpuser->dialarg, &dg);
+
+			ao2_ref(caps, -1);
```
jcolp, line 1108: This should use ao2_cleanup, it can technically be NULL albeit unlikely

jcolp (changes requested):

apps/app_chanisavail.c
creslin287: It looked broken before (like the indentation was incorrect) so I just updated it since I was already touching the code around here.  I can revert it though if you'd like.

creslin287: Oh, I see it now.  That second function was an arg to the first one.  I'm going to try to just make that one line if you're ok with it so that it's not confusing as it was before.

seanbright (approved):

Issue #384, opened 2023-10-24
[bug]: Unnecessary re-INVITE after answer


berjon9:
### Severity

Minor

### Versions

20.5.0, 20.4.0, 20.0.1, 18.20.0, 18.17.0, 18.16.0

### Components/Modules

res_pjsip

### Operating Environment

The software is run on a container image based on Debian 11.6 using Docker.

### Frequency of Occurrence

Constant

### Issue Description

I've been trying to simulate a call using `SIPp` for both server and client on a very simple scenario: 
```
A        Asterisk    B
  ----> Invite ---->
  <---- 100   <----
  <---- 180   <----
  <---- 200   <----
   Pause
  -----> Bye ----->
  <----- 200 <-----
```

This has been working for a long time using vesion `18.15.0`, however after an upgrade to version `20.4.0` I started noticing that B wasn't properly managing the BYE message, even if the scenario seemed to be successful on A.
After picking up some traces I realized that B was reciving the BYE message okay, just like it did on the previous version. 

However, right after the call establishment, Asterisk would be sending towards B a re-INVITE, which apparently was just like the original invite:
  - There is no change on the SDP except for the session version on the `o=` line
  - The timer has not expired yet

Given that the extra message is not covered in the simulation (it shouldn't be there), the call would fail.

I could also appreciate that this wouldn't happen if the call had media flowing on it, (on the simulation I cannot inject media). This kind of reminded me of this issue on `Asterisk 16` where an unnecessary re-INVITE was generated: [issue #28871](https://issues-archive.asterisk.org/ASTERISK-28871)

I run the test on several versions, being the most recent one in which it worked `Asterisk 18.15.0`. This situation took place in all other versions mentioned on the issue.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
You need to show the actual configuration, as well as the SIP trace. The re-INVITE can also occur for other reasons, such as connected line updates.

berjon9:
Sure. 

Here are the traces. I suggest using a `sip || http` filter to get all valuable information.
[traces.zip](https://github.com/asterisk/asterisk/files/13163763/traces.zip)

About the config:

Here is the `pjsip.conf`:
```
[simpletrans]
type=transport
protocol=udp
bind=0.0.0.0:port_sip
external_signaling_address=signal_IP
external_media_address=media_IP

[to_b]
type=aor
contact=sip:A1352472@192.168.0.8:4090

[...]

PR #403, opened 2023-10-31
bridge_simple: Suppress unchanged topology change requests

gtjoseph:
In simple_bridge_join, we were sending topology change requests
even when the new and old topologies were the same.  In some
circumstances, this can cause unnecessary re-invites and even
a re-invite flood.  We now suppress those.

Resolves: #384


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

seanbright (approved):

PR #405, opened 2023-10-31
chan_pjsip: Add PJSIPHangup dialplan app and manager action

gtjoseph:
See UserNote below.

Exposed the existing Hangup AMI action in manager.c so we can use
all of it's channel search and AMI protocol handling without
duplicating that code in dialplan_functions.c.

Added a lookup function to res_pjsip.c that takes in the
string represenation of the pjsip_status_code enum and returns
the actual status code.  I.E.  ast_sip_str2rc("DECLINE") returns
603.  This allows the caller to specify PJSIPHangup(decline) in
the dialplan, just like Hangup(call_rejected).

Also extracted the XML documentation to its own file since it was
almost as large as the code itself.

UserNote: A new dialplan app PJSIPHangup and AMI action allows you
to hang up an unanswered incoming PJSIP call with a specific SIP
response code in the 400 -> 699 range.


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

gtjoseph:
oops...forgot there needs to be a companion to this that sets a Reason header on the next response.  will do that tomorrow.

channels/pjsip/dialplan_functions.c
```diff
@@ -204,6 +204,25 @@
...
+			<para>SIP response code between 400 and 699.</para>
+		</parameter>
+	</syntax>
+	<description>
+		<para>Hangs up an incoming PJSIP channel and returns the
+		specified SIP response code to the caller.</para>
+		<warning><para>
+		This function must be called BEFORE anything that
+		might cause any other final (non 1XX) response to be sent.
+		</para></warning>
```
InterLinked1, line 222: An example would be nice.

e.g. `exten => s,1,Set(PJSIP_HANGUP()=404)`

And I think people might want to know if they can do other stuff after calling this, will dialplan execution end immediately after?

Are there examples of things that might cause a final response to be sent?

channels/pjsip/dialplan_functions.c
```diff
@@ -1784,3 +1803,65 @@ int pjsip_acf_session_refresh_write(struct ast_channel *chan, const char *cmd, c
...
+{
+	struct ast_sip_channel_pvt *channel;
+	struct hangup_data hdata;
+
+	if (!chan) {
+		ast_log(LOG_WARNING, "No channel was provided to %s function.\n", cmd);
+		return -1;
+	}
+
+	if (sscanf(value, "%30d", &hdata.response_code) != 1) {
```
InterLinked1, line 1835: Isn't ast_str_to_int is preferred over sscanf, for int parsing?

InterLinked1 (approved):

channels/pjsip/dialplan_functions.c
```diff
@@ -1784,3 +1228,71 @@ int pjsip_acf_session_refresh_write(struct ast_channel *chan, const char *cmd, c
...
+	int response_code;
+};
+
+static int pjsip_hangup(void *obj)
+{
+	struct hangup_data *hdata = obj;
+	pjsip_tx_data *packet = NULL;
+
+	if ((hdata->session->inv_session->state != PJSIP_INV_STATE_DISCONNECTED) &&
+		(pjsip_inv_answer(hdata->session->inv_session, hdata->response_code, NULL, NULL, &packet) == PJ_SUCCESS)) {
```
jcolp, line 1243: What's the difference between using pjsip_inv_answer here and pjsip_inv_end_session?

channels/pjsip/dialplan_functions.c
```diff
@@ -1784,3 +1228,71 @@ int pjsip_acf_session_refresh_write(struct ast_channel *chan, const char *cmd, c
...
+struct hangup_data {
+	struct ast_sip_session *session;
+	int response_code;
+};
+
+static int pjsip_hangup(void *obj)
+{
+	struct hangup_data *hdata = obj;
+	pjsip_tx_data *packet = NULL;
+
```
jcolp, line 1241: The hangup cause on the channel should be a conversion from SIP response code to ISDN cause code for a bit of consistency

channels/pjsip/dialplan_functions.xml
```diff
@@ -0,0 +1,591 @@
...
+			Hangup an incoming PJSIP channel with a SIP response code
+		</synopsis>
+		<syntax>
+			<parameter name="response_code" required="true">
+				<para>SIP response code between 400 and 699.</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>
+			Hangs up an incoming PJSIP channel and returns the
```
jcolp, line 189: It should be documented what the behavior of the ISDN cause code is with this

jcolp (dismissed):

channels/pjsip/dialplan_functions.c
gtjoseph: pjsip_inv_end_session just calls pjsip_inv_answer when we're a UAS and the state is PJSIP_INV_STATE_INCOMING (which we've already tested for).

channels/pjsip/dialplan_functions.c
gtjoseph: It's set by chan_pjsip_session_end when the session ends.

channels/pjsip/dialplan_functions.xml
gtjoseph: Done

jcolp (approved):

jcolp:
This should also have test coverage in the testsuite.

gtjoseph:
> This should also have test coverage in the testsuite.

Yep.  In progress.

Issue #377, opened 2023-10-16
[bug]: PJSIP Crash during SendDTMF from AMI


sim1camporeale:
### Severity

Critical

### Versions

13.38.3

### Components/Modules

chan_pjsip

### Operating Environment

Ubuntu 14.04.4 LTS

### Frequency of Occurrence

Occasional

### Issue Description

I'm using AMI to interact with phone calls. In my scenario I integrated some safety related analog intercom using some ATA. This intercoms has some inner timeout that hangs up the call if they didn't receive proper DTMF tone when call is established. 

In some instances (one a week) occurs that chan_pjsip crashes.

I think that this is similar to [ASTERISK-28086](https://issues-archive.asterisk.org/ASTERISK-28086) but it occurs when sending DTMF in chan_pjsip_digit_start.

I'm trying to add in chan_pjsip_digit_start same ASTERISK-28086 patch that i found in chan_pjsip_digit_end.

I know that this version of asterisk is unsupported and also operating system is unsupported. I didn't checked if this issue can be present also in more recent version of asterisk. I'm asking you if can help me understing if the patch i'm preparing for asterisk 13.38.3 can be useful for other branches.

 

### Relevant log output

```shell
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Core was generated by `/usr/sbin/asterisk'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fe0e81af7f8 in ?? () from /usr/lib/asterisk/modules/chan_pjsip.so
(gdb) bt
#0  0x00007fe0e81af7f8 in ?? () from /usr/lib/asterisk/modules/chan_pjsip.so
#1  0x00000000004af3be in ast_senddigit_external ()
#2  0x00007fe0e655828b in ?? () from /usr/lib/asterisk/modules/app_senddtmf.so
#3  0x000000000053cc3e in ?? ()
#4  0x0000000000541626 in ?? ()
#5  0x00000000005d07ba in ?? ()
#6  0x00000000005df198 in ?? ()
#7  0x00007fe14fc21182 in start_thread (arg=0x7fe109f79700) at pthread_create.c:312
#8  0x00007fe14f1fd47d in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#9  0x0000000000000000 in ?? ()
(gdb) Quit
(gdb) exit


[...]

PR #411, opened 2023-11-03
chan_pjsip: handle crash if channel disapper while calling dtmf_begin

sim1camporeale:
Check if channel is still available in digit_begin like what we do in digit_end

Resolves issue #377


jcolp:
Changes can not be merged into 13.38.3, and a contributor license agreement needs to be signed. Instructions are on the docs site: https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/

PR #414, opened 2023-11-06
chan_iax2: Add log message for rejected calls.

InterLinked1:
Add a log message for a path that currently silently drops IAX2 frames without indicating that anything is wrong.

### DIFF
--- a/channels/chan_iax2.c
+++ b/channels/chan_iax2.c
@@ -10406,10 +10406,13 @@ static int socket_process_helper(struct iax2_thread *thread)
 		}
 
 		if (!(fr->callno = find_callno(ntohs(mh->callno) & ~IAX_FLAG_FULL, dcallno, &addr, new, fd, check_dcallno))) {
+			ast_debug(1, "Received frame without existent call number (%d)\n", ntohs(mh->callno) & ~IAX_FLAG_FULL);
 			if (f.frametype == AST_FRAME_IAX && f.subclass.integer == IAX_COMMAND_NEW) {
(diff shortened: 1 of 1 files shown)


seanbright:
“Possible network issue?” 

How is this useful?

InterLinked1:
> “Possible network issue?”
> 
> How is this useful?

It's something to go off of, and there are more comments in the code that can be used if needed.
I didn't want to get too specific like "Check /etc/hosts for non-canonical loopback addresses" because it's not guaranteed that the issue is there, that's why I tried to keep it generic.

InterLinked1:
> This is why I suggested actually digging in more to truly understand the scenario and what is going on to provide a better and more helpful message in the correct situation.

If chan_iax2 isn't listening on the particular loopback address, the packets are being sent somewhere and not actually getting processed, hence the lack of a valid call number at this point. Why it is that someone is trying to dial a loopback address that isn't being listened on is really their issue to resolve, but this should clue them in on that hopefully.

jcolp:
On further looking, they were actually being received so my initial analysis was incorrect - why things went awry I don't know, but it was seemingly due to not explicitly binding. I don't think it's actually understood why it's doing what it is doing, beyond that binding explicitly causes it to work.

seanbright (changes requested):
This should not be marked as fixing the associated issue. That issue should remain open.

seanbright (approved):

gtjoseph:
@InterLinked1 Cherry-pick?

InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

PR #415, opened 2023-11-06
chan_iax2.c: Ensure all IEs are displayed when dumping frame contents.

seanbright:
When IAX2 debugging was enabled (`iax2 set debug on`), if the last IE in a frame was one that may not have any data - such as the CALLTOKEN IE in an NEW request - it was not getting displayed.

### DIFF
--- a/channels/iax2/parser.c
+++ b/channels/iax2/parser.c
@@ -424,7 +424,7 @@ static void dump_ies(unsigned char *iedata, int len)
 
 	if (len < 2)
 		return;
-	while(len > 2) {
+	while(len >= 2) {
 		ie = iedata[0];
 		ielen = iedata[1];
 		if (ielen + 2> len) {


seanbright:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

channels/iax2/parser.c
```diff
@@ -260,6 +260,15 @@ static void dump_versioned_codec(char *output, int maxlen, void *value, int len)
 	}
 }
 
+static void dump_call_token(char *output, int maxlen, void *value, int len)
+{
+	if (len == 0) {
+		strcpy(output, "Present");
```
InterLinked1, line 266: Why do we output "Present" when the length is 0? Just trying to make sense of how I should be interpreting this.

seanbright: Because the IE is present.

InterLinked1: > Because the IE is present.

To me, this makes it sound like the call token itself is present (and just not shown), not just the IE - it's clear that's present just from looking at the debug, or it wouldn't appear in the first place.

Personally, I think "Empty" would be a bit clearer to disambiguate "IE present, but no actual call token" from "IE present, with valid call token"... "Present" could imply the call token is there, just not dumped, unless someone were familiar with the output. That's my only suggestion.

jcolp: I agree that Empty would be clearer.

jcolp (approved):

PR #418, opened 2023-11-07
app.c: Allow ampersands in playback lists to be escaped.

### LINKED ISSUE #172: [bug]: Playback() does not support ampersand in URLs
### Severity

Trivial

### Versions

18.18.0

### Components/Modules

app_playback

### Operating Environment

NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
aarch64

### Frequency of Occurrence

Constant

### Issue Description

Playback supports URLs, but not if they contain an ampersand character -- which is valid in a URL.  S3 playback assets have many arguments that must be passed for authentication, but Playback will try to play each component separated by an ampersand independently; thereby playing no assets.

The strsep() function appears to not check for any escape character before the token, nor does app_playback.c -- (front = strsep(&back, "&"))); so there isn't any way to pass an S3 URL, for example.

### Relevant log output

_No response_

### Asterisk Issue Guidelines

seanbright:
Any function or application that accepts a `&`-separated list of
filenames can now include a literal `&` in a filename by wrapping the
entire filename in single quotes, e.g.:

```
exten = _X.,n,Playback('https://example.com/sound.cgi?a=b&c=d'&hello-world)
```

Fixes #172

UpgradeNote: Ampersands in URLs passed to the `Playback()`,
`Background()`, `SpeechBackground()`, `Read()`, `Authenticate()`, or
`Queue()` applications as filename arguments can now be escaped by
single quoting the filename. Additionally, this is also possible when
using the `CONFBRIDGE` dialplan function, or configuring various
features in `confbridge.conf` and `queues.conf`.

### DIFF
--- a/apps/app_authenticate.c
+++ b/apps/app_authenticate.c
@@ -95,8 +95,17 @@ static const char app[] = "Authenticate";
 				maxdigits have been entered (without requiring the user to press the <literal>#</literal> key).
 				Defaults to 0 - no limit - wait for the user press the <literal>#</literal> key.</para>
 			</parameter>
-			<parameter name="prompt" required="false">
-				<para>Override the agent-pass prompt file.</para>
+			<parameter name="prompt" required="false" argsep="&amp;">
+				<para>Override the &quot;agent-pass&quot; sound file. Can be
+				an ampersand separated list of filenames. If the filename
+				is a relative filename (it does not begin with a slash), it
+				will be searched for in the Asterisk sounds directory. If the
+				filename is able to be parsed as a URL, Asterisk will
+				download the file and then begin playback on it. To include a
+				literal <literal>&amp;</literal> in the URL you can enclose
+				the URL in single quotes.</para>
+				<argument name="prompt" required="true" />
+				<argument name="prompt2" multiple="true" />
 			</parameter>
 		</syntax>
 		<description>
--- a/apps/app_confbridge.c
+++ b/apps/app_confbridge.c
@@ -3035,7 +3035,7 @@ static int action_playback(struct ast_bridge_channel *bridge_channel, const char
 	char *file_copy = ast_strdupa(playback_file);
 	char *file = NULL;
 
-	while ((file = strsep(&file_copy, "&"))) {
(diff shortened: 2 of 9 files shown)


include/asterisk/app.h
```diff
@@ -123,6 +123,8 @@ enum ast_timelen {
 */
 int ast_ivr_menu_run(struct ast_channel *c, struct ast_ivr_menu *menu, void *cbdata);
 
+char *ast_app_filename_list_iterate(char **restrict stringp);
```
InterLinked1, line 126: Documentation would be nice...
Also XML documentation for escaping to the dialplan applications (at least Playback), like you commented on the issue.

seanbright:
This is a draft. No reviews are necessary at this time.

seanbright:
Yes, I know this will have to be squashed. Yes, I know there should be a UserNote in the commit message. Yes, I know there should be a cherry pick comment on this PR.

seanbright:
cherry-pick-to: none

include/asterisk/app.h
```diff
@@ -123,6 +123,25 @@ enum ast_timelen {
...
+ *   - Removes the backslash character from the returned string if it is
+ *     escaping an ampersand
+ *
+ * \param stringp A pointer to the \c \0 terminated string to traverse.
+ *                Like \c strsep, the \c stringp argument is modified by
+ *                this function.
+ * \return A non-\c NULL pointer to the next filename from \c stringp,
+ *         or \c NULL if there are no remaining filenames.
+ */
+char *ast_app_filename_list_iterate(char **restrict stringp);
```
seanbright, line 143: I'm starting to think that it might be better to make this function more generic and take the separator as an argument. Basically a `strsep(...)` that allows you to escape the separator and doesn't return empty elements.

jcolp (changes requested):
Commit message says "can not" when it should say "can now".

Issue #84, opened 2023-05-11
[bug]: codec_ilbc:  Fails to build with ilbc version 3.0.4


gtjoseph:
### Severity

Minor

### Versions

18,20,master

### Components/Modules

codec_ilbc

### Operating Environment

Fedora 37
ilbc-3.0.4-3.fc37.x86_64

### Frequency of Occurrence

Constant

### Issue Description

ilbc 3.0 has api changes that prevent codec_ilbc from building

### Relevant log output

```shell
[CC] codec_ilbc.c -> codec_ilbc.o
codec_ilbc.c: In function ‘lintoilbc_new’:
codec_ilbc.c:79:9: error: implicit declaration of function ‘initEncode’ [-Werror=implicit-function-declaration]
   79 | initEncode(&tmp->enc, mode);
      | ^~~~~~~~~~
codec_ilbc.c: In function ‘ilbctolin_framein’:
codec_ilbc.c:130:17: error: implicit declaration of function ‘initDecode’ [-Werror=implicit-function-declaration]
  130 | initDecode(&tmp->dec, mode, USE_ILBC_ENHANCER);
      | ^~~~~~~~~~
codec_ilbc.c:139:17: error: implicit declaration of function ‘iLBC_decode’ [-Werror=implicit-function-declaration]
  139 | iLBC_decode(tmpf, plc_mode ? f->data.ptr + x : NULL, &tmp->dec, plc_mode);
      | ^~~~~~~~~~~
codec_ilbc.c: In function ‘lintoilbc_frameout’:
codec_ilbc.c:184:17: error: implicit declaration of function ‘iLBC_encode’ [-Werror=implicit-function-declaration]
  184 | iLBC_encode((ilbc_bytes *) pvt->outbuf.BUF_TYPE, tmpf, &tmp->enc);
      | ^~~~~~~~~~~
cc1: all warnings being treated as errors
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #419, opened 2023-11-07
codec_ilbc: Disable system ilbc if version >= 3.0.0

gtjoseph:
Fedora 37 started shipping ilbc 3.0.4 which we don't yet support.
configure.ac now checks the system for "libilbc < 3" instead of
just "libilbc".  If true, the system version of ilbc will be used.
If not, the version included at codecs/ilbc will be used.

Resolves: #84


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #425, opened 2023-11-09
[improvement]: configs: Improve documentation for bandwidth in iax.conf.sample


InterLinked1:
### Improvement Description

This adds additional documentation to iax.conf.sample for the bandwidth setting and also adjusts the sample default to be sane out of the box.

PR #426, opened 2023-11-09
configs: Improve documentation for bandwidth in iax.conf.

InterLinked1:
This improves the documentation for the bandwidth setting in iax.conf by making it clearer what the ramifications of this setting are. It also changes the sample default from low to high, since only high is compatible with good codecs that people will want to use in the vast majority of cases, and this is a common gotcha that trips up new users.

Resolves: #425


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

PR #427, opened 2023-11-09
IAX2 Parser Cleanups

seanbright:
There is a lot more that could be done but I want the review to be reviewable. There are 3 commits here that will be squashed if/when this is approved (although it's still not clear to me why the squash can't just happen at merge time, but regardless) but I think it is easier to review them as separate commits rather than a single diff across all 3.

### DIFF
--- a/channels/iax2/parser.c
+++ b/channels/iax2/parser.c
@@ -74,18 +74,34 @@ struct iax_frames {
 #define FRAME_CACHE_MAX_SIZE	20
 #endif
 
-static void internaloutput(const char *str)
+static __attribute__((format(printf, 1, 2)))
+void internaloutput(const char *fmt, ...)
 {
-	fputs(str, stdout);
+	va_list ap;
+
+	va_start(ap, fmt);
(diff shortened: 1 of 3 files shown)


seanbright:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

channels/iax2/parser.c
```diff
@@ -1311,8 +1284,8 @@ void iax_frame_free(struct iax_frame *fr)
 
 #if !defined(NO_FRAME_CACHE)
 	if (!fr->cacheable
-		|| !ast_opt_cache_media_frames
-		|| !(iax_frames = ast_threadstorage_get(&frame_cache, sizeof(*iax_frames)))) {
+	   || !ast_opt_cache_media_frames
```
InterLinked1, line 1287: Did you mean to change from tabs to spaces here?

seanbright: No I did not.

Issue #437, opened 2023-11-10
[new-feature]: sig_analog: Add Last Number Redial


InterLinked1:
### Feature Description

This adds a simple redial feature to the simple switch in sig_analog

PR #438, opened 2023-11-10
sig_analog: Add Last Number Redial feature.

InterLinked1:
This adds the Last Number Redial feature to
simple switch.

UserNote: Users can now redial the last number
called if the lastnumredial setting is set to yes.

Resolves: #437

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -13071,6 +13071,7 @@ static struct dahdi_pvt *mkintf(int channel, const struct dahdi_chan_conf *conf,
 		tmp->permhidecallerid = conf->chan.hidecallerid;
 		tmp->hidecalleridname = conf->chan.hidecalleridname;
 		tmp->callreturn = conf->chan.callreturn;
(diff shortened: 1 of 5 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp:
Why do this here in code instead of dialplan?

InterLinked1:
> Why do this here in code instead of dialplan?

If immediate=no, you can't easily capture the number dialed in the dialplan.
I do this in my own code stack normally but on a "normal" Asterisk system this is really the only elegant way to do it.

jcolp:
So yes it is possible.

InterLinked1:
> So yes it is possible.

No, not really, unless you have immediate=yes, which is the scenario to which I was referring. Otherwise this is too low level to be doable in the dialplan.

jcolp:
I disagree, I believe it is reasonably possible for the usage of most users and wouldn't surprise me if others have done so. That's my opinion though and I won't stop this, others are free to join in and review this if they wish. I may eventually get around to it.

Issue #442, opened 2023-11-11
[bug]: func_channel: Some channel options are not settable


InterLinked1:
### Severity

Major

### Versions

21.0.0

### Components/Modules

func_channel

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Some channel options are not settable, which prevents the user from being able to control them and means certain code is currently unreachable.

This adds support for setting and querying options that make sense to be exposed to users, which are not currently controllable in other ways. Documentation is also improved to describe how some of these options relate to other options.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #443, opened 2023-11-11
func_channel: Expose previously unsettable options.

InterLinked1:
Certain channel options are not set anywhere or
exposed in any way to users, making them unusable. This exposes some of these options which make sense for users to manipulate at runtime.

Resolves: #442


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #445, opened 2023-11-13
[bug]: ast_coredumper isn't figuring out file locations properly in all cases


gtjoseph:
### Severity

Minor

### Versions

18,20,21,master

### Components/Modules

scripts

### Operating Environment

all

### Frequency of Occurrence

Occasional

### Issue Description

ast_coredumper isn't figuring out things like the asterisk modules directory correctly when non-standard locations are being used.  For instance, on most RedHat based 64 bit systems, LIBDIR should be /usr/lib64 and the modules should be in /usr/lib64/asterisk/modules.  In some older distros, the logic autoconf uses to determine that is broken resulting in a LIBDIR of /usr/lib instead and ast_coredumper isn't always smart enough to figure that out.  Another difficult situation is when asterisk is installed using a DESTDIR.  In that case, all bets are off.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

InterLinked1:
@gtjoseph I'm not sure if you've seen this before or if you think a new issue should be opened for this, but sometimes even when `ast_coredumper` finds the core dump, it refuses to process it. I've been noticing this consistently on one system.

It has do with the ELF part where it looks inside the core dumper to see if it's an Asterisk core dump or not. As written it rejects the core dump, but if I comment out that check then everything works fine. Here you can see that it initially refuses to process the core dump, then I modify it to ignore the ELF check, and then it works as expected:

```
root@debian:/usr/src/asterisk-21.0.0# /var/lib/asterisk/scripts/ast_coredumper core
Not an asterisk core dump:

Displaying notes found at file offset 0x000261e8 with length 0x0003b7a8:
  Owner                Data size        Description
  CORE                 0x00000150       NT_PRSTATUS (prstatus structure)
  CORE                 0x00000088       NT_PRPSINFO (prpsinfo structure)
  CORE                 0x00000080       NT_SIGINFO (siginfo_t data)
  CORE                 0x00000150       NT_AUXV (auxiliary vector)
  CORE                 0x0002a2f5       NT_FILE (mapped files)
    Cannot decode 64-bit note in 32-bit build
  CORE                 0x00000200       NT_FPREGSET (floating point registers)
  CORE                 0x00000150       NT_PRSTATUS (prstatus structure)
  CORE                 0x00000200       NT_FPREGSET (floating point registers)
  CORE                 0x00000150       NT_PRSTATUS (prstatus structure)
[...]

PR #446, opened 2023-11-13
ast_coredumper: Increase reliability

gtjoseph:
Instead of searching for the asterisk binary and the modules in the
filesystem, we now get their locations, along with libdir, from
the coredump itself...

For the binary, we can use `gdb -c <coredump> ... "info proc exe"`.
gdb can print this even without having the executable and symbols.

Once we have the binary, we can get the location of the modules with
`gdb ... "print ast_config_AST_MODULE_DIR`

If there was no result then either it's not an asterisk coredump
or there were no symbols loaded.  Either way, it's not usable.

For libdir, we now run "strings" on the note0 section of the
coredump (which has the shared library -> memory address xref) and
search for "libasteriskssl|libasteriskpj", then take the dirname.

Since we're now getting everything from the coredump, it has to be
correct as long as we're not crossing namespace boundaries like
running asterisk in a docker container but trying to run
ast_coredumper from the host using a shared file system (which you
shouldn't be doing).

There is still a case for using --asterisk-bin and/or --libdir: If
you've updated asterisk since the coredump was taken, the binary,
libraries and modules won't match the coredump which will render it
useless.  If you can restore or rebuild the original files that
match the coredump and place them in a temporary directory, you can
use --asterisk-bin, --libdir, and a new --moddir option to point to
them and they'll be correctly captured in a tarball created
with --tarball-coredumps.  If you also use --tarball-config, you can
use a new --etcdir option to point to what normally would be the
/etc/asterisk directory.

Also addressed various "shellcheck" issues.

Resolves: #445

### DIFF
--- a/contrib/scripts/ast_coredumper
+++ b/contrib/scripts/ast_coredumper
@@ -1,59 +1,60 @@
-#!/usr/bin/env bash
+#!/bin/bash
+
 # Turn on extended globbing
 shopt -s extglob
 shopt -s nullglob
 # Bail on any error
 set -e
 
-prog=$(basename $0)
+prog=$(basename "$0")
 
 # NOTE: <(cmd) is a bash construct that returns a temporary file name
 # from which the command output can be read.  In this case, we're
 # extracting the block of text delimited by '#@@@FUNCSSTART@@@'
 # and '#@@@FUNCSEND@@@' from this file and 'source'ing it to
 # get some functions.
-source <(sed -n -r -e "/^#@@@FUNCSSTART@@@/,\${p;/^#@@@FUNCSEND@@@/q}" $0 | sed '1d;$d')
-
-# The "!(*.txt)" is a bash construct that excludes files ending with .txt
-# from the glob match.
-declare -a COREDUMPS=( /tmp/core!(*.txt) )
+# shellcheck disable=SC1090
+source <(sed -n "/^#@@@FUNCSSTART@@@/,/^#@@@FUNCSEND@@@/ p" "$0" | sed '1d;$d')
 
 # A line starting with ': ' is a POSIX construct that makes the shell
 # perform the operation but ignore the result.  This is an alternative to
 # having to do RUNNING=${RUNNING:=false} to set defaults.
 
-: ${ASTERISK_BIN:=$(which asterisk)}
-: ${DATEOPTS='-u +%FT%H-%M-%SZ'}
-: ${DELETE_COREDUMPS_AFTER:=false}
-: ${DELETE_RESULTS_AFTER:=false}
-: ${DRY_RUN:=false}
-: ${GDB:=$(which gdb)}
-: ${HELP:=false}
-: ${LATEST:=false}
-: ${OUTPUTDIR:=/tmp}
-: ${PROMPT:=true}
-: ${RUNNING:=false}
-: ${RENAME:=true}
-: ${TARBALL_CONFIG:=false}
-: ${TARBALL_COREDUMPS:=false}
-: ${TARBALL_RESULTS:=false}
+: "${DATEOPTS=-u +%FT%H-%M-%SZ}"
+: "${DELETE_COREDUMPS_AFTER:=false}"
+: "${DELETE_RESULTS_AFTER:=false}"
+: "${DRY_RUN:=false}"
+: "${GDB:=$(which gdb)}"
+: "${HELP:=false}"
+: "${LATEST:=false}"
+: "${OUTPUTDIR:=/tmp}"
+: "${PROMPT:=true}"
+: "${RUNNING:=false}"
+: "${RENAME:=true}"
+: "${TARBALL_CONFIG:=false}"
+: "${TARBALL_COREDUMPS:=false}"
+: "${TARBALL_RESULTS:=false}"
+: "${MODDIR:=}"
+: "${LIBDIR:=}"
+: "${ETCDIR:=}"
 
 COMMANDLINE_COREDUMPS=false
 
 # Read config files from most important to least important.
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

gtjoseph:
Found a few things debian/ubuntu related.  need to fix.

InterLinked1:
> Found a few things debian/ubuntu related. need to fix.

This is with the current core dumper, not with your change, but somehow it seems to fail on Debian sometimes, even with the standard paths:

```
root@debian:/usr/src/asterisk-21.0.0# ls -la core
-rw------- 1 root root 180887552 Nov 16 16:49 core
root@debian:/usr/src/asterisk-21.0.0# /var/lib/asterisk/scripts/ast_coredumper core
No coredumps found
root@debian:/usr/src/asterisk-21.0.0# ls -la core
-rw------- 1 root root 180887552 Nov 16 16:49 core
```

gtjoseph:
@InterLinked1 Try this version and see if it resolves your issues in various environments.

contrib/scripts/ast_coredumper
```diff
@@ -1,73 +1,78 @@
...
+prog=$(basename "$0")
 
 # NOTE: <(cmd) is a bash construct that returns a temporary file name
 # from which the command output can be read.  In this case, we're
 # extracting the block of text delimited by '#@@@FUNCSSTART@@@'
 # and '#@@@FUNCSEND@@@' from this file and 'source'ing it to
 # get some functions.
-source <(sed -n -r -e "/^#@@@FUNCSSTART@@@/,\${p;/^#@@@FUNCSEND@@@/q}" $0 | sed '1d;$d')
+# shellcheck disable=SC1090
+source <(sed -n -r -e "/^#@@@FUNCSSTART@@@/,\${p;/^#@@@FUNCSEND@@@/q}" "$0" | sed '1d;$d')
```
seanbright, line 17: And I might be wrong but I feel like this could be simplified to:

    source <(sed -n "/^#@@@FUNCSSTART@@@/,/^#@@@FUNCSEND@@@/ p" "$0" | sed '1d;$d')

contrib/scripts/ast_coredumper
```diff
@@ -411,7 +458,7 @@ find_pid() {
 	# Now that we have the pids, let's get the command and
 	# its args. We'll add them to an array indexed by pid.
 	declare -a candidates
-	while read LINE ; do
+	while read -r LINE ; do
```
seanbright, line 461: Could potentially avoid the regex with:

    while read -r pid prog args ; do

contrib/scripts/ast_coredumper
gtjoseph: Yep, that seems to work fine.

contrib/scripts/ast_coredumper
gtjoseph: I did it this way so if for some reason the result didn't match the regex, we'd skip that line and continue otherwise we won't be sure that we got a valid line.

InterLinked1:
> @InterLinked1 Try this version and see if it resolves your issues in various environments.

Still no bueno:

```
root@debian:/usr/src/asterisk-21.0.0# ls -la core
-rw------- 1 root root 226508800 Nov 29 12:49 core
root@debian:/usr/src/asterisk-21.0.0# /var/lib/asterisk/scripts/ast_coredumper core
No coredumps found
```

gtjoseph:
> root@debian:/usr/src/asterisk-21.0.0# /var/lib/asterisk/scripts/ast_coredumper core
> readlink: missing operand
> Try 'readlink --help' for more information.
> Examining core
>     Does appear to be an asterisk coredump
>     Coredump indicates executable 'asterisk'
>     Searching for asterisk module directory
>     Found asterisk module directory '/usr/lib/asterisk/modules'
> Processing /usr/src/asterisk-21.0.0/core
>     ASTBIN: asterisk
>     MODDIR: /usr/lib/asterisk/modules
>     ETCDIR: /etc/asterisk
>     LIBDIR: /usr/lib
>     Renaming /usr/src/asterisk-21.0.0/core to /usr/src/asterisk-21.0.0/core-asterisk-2023-11-29T17-49-18Z
>     Creating /tmp/core-asterisk-2023-11-29T17-49-18Z-thread1.txt
>     Creating /tmp/core-asterisk-2023-11-29T17-49-18Z-brief.txt
>     Creating /tmp/core-asterisk-2023-11-29T17-49-18Z-full.txt
>     Creating /tmp/core-asterisk-2023-11-29T17-49-18Z-locks.txt
>     Creating /tmp/core-asterisk-2023-11-29T17-49-18Z-info.txt

The `readlink: missing operand` should have been suppressed since you specified a core file on the command line.  Just pushed up a fix for that.

jcolp:
@InterLinked1 Is this now in a working state for you?

InterLinked1:
> @InterLinked1 Is this now in a working state for you?

Sorry - yup, it is, and that other spurious warning is gone now, too.
I'm not sure if there was some recent regression but ast_coredumper was previously not working on any of my machines recently (it always had mostly in the past). With this PR though all is well again.

jcolp (approved):

InterLinked1:
One more thing I just noticed now:

```
root@ess:/usr/src/asterisk-21.0.0# /var/lib/asterisk/scripts/ast_coredumper --RUNNING
Found a single asterisk instance running as process 848952
Dumping running asterisk process to /tmp/core-asterisk-running-2023-12-06T16-39-34Z
Dump is complete.
Couldn't get module directory from coredump!
```

However, it seems the backtrace was successfully extracted, so whatever happened didn't impede it from working.

PR #453, opened 2023-11-20
res_rtp_asterisk.c: Update for OpenSSL 3+.

seanbright:
In 5ac5c2b0 we defined `OPENSSL_SUPPRESS_DEPRECATED` to silence deprecation warnings. This commit switches over to using non-deprecated API.

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -47,7 +47,6 @@
 #include <math.h>
 
 #ifdef HAVE_OPENSSL
-#define OPENSSL_SUPPRESS_DEPRECATED 1
 #include <openssl/opensslconf.h>
 #include <openssl/opensslv.h>
 #if !defined(OPENSSL_NO_SRTP) && (OPENSSL_VERSION_NUMBER >= 0x10001000L)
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

res/res_rtp_asterisk.c
```diff
@@ -1914,6 +1913,27 @@ struct dtls_cert_info {
...
+static int apply_dh_params(SSL_CTX *ctx, BIO *bio)
+{
+	int res = 0;
+
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+	EVP_PKEY *dhpkey = PEM_read_bio_Parameters(bio, NULL);
+	if (dhpkey) {
+		res = EVP_PKEY_is_a(dhpkey, "DH") && SSL_CTX_set0_tmp_dh_pkey(ctx, dhpkey);
+	}
+	EVP_PKEY_free(dhpkey);
```
alex2grad, line 1925: EVP_PKEY_free should be inside if (dhpkey) {}

res/res_rtp_asterisk.c
```diff
@@ -1914,6 +1913,27 @@ struct dtls_cert_info {
...
+#else
+	DH *dh = PEM_read_bio_DHparams(bio, NULL, NULL, NULL);
+	if (dh) {
+		res = SSL_CTX_set_tmp_dh(ctx, dh);
+	}
+	DH_free(dh);
```
alex2grad, line 1931: DH_free should be inside if (dh) {}

res/res_rtp_asterisk.c
seanbright: [From `man 3 EVP_PKEY_free`](https://www.openssl.org/docs/man3.0/man3/EVP_PKEY_free.html):

> EVP_PKEY_free() decrements the reference count of key and, if the reference count is zero, frees it up. If key is NULL, nothing is done.

res/res_rtp_asterisk.c
seanbright: [From `man 3 DH_free`](https://www.openssl.org/docs/man3.0/man3/DH_free.html):

> DH_free() frees the DH structure and its components. The values are erased before the memory is returned to the system.  If dh is NULL nothing is done.

jcolp (approved):

Issue #458, opened 2023-11-24
[bug]: Memory leak in chan_dahdi when mwimonitor=yes on FXO


InterLinked1:
### Severity

Major

### Versions

21.0.0

### Components/Modules

chan_dahdi

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

If `mwimonitor=yes` is enabled for an FXO port in `chan_dahdi.conf`, when the FXO port receives an incoming call, the following log messages will appear indefinitely:

```
[2023-11-23 21:35:30.621] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.621] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.621] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.621] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.622] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.622] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.622] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.622] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.623] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[2023-11-23 21:35:30.623] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...
[...]

PR #459, opened 2023-11-24
sig_analog: Fix channel leak when mwimonitor is enabled.

InterLinked1:
When mwimonitor=yes is enabled for an FXO port,
the do_monitor thread will launch mwi_thread if it thinks there could be MWI on an FXO channel, due to the noise threshold being satisfied. This, in turns, calls
analog_ss_thread_start in sig_analog. However, unlike all other instances where __analog_ss_thread is called in sig_analog, this call path does not properly set pvt->ss_astchan to the Asterisk channel, which means that the Asterisk channel is NULL when __analog_ss_thread starts executing. As a result, the thread exits and the channel is never properly cleaned up by calling ast_hangup.

This caused issues with do_monitor on incoming calls, as it would think the channel was still owned even while receiving events, leading to an infinite barrage of warning messages; additionally, the channel would persist improperly.

To fix this, the assignment is added to the call path where it is missing (which is only used for mwi_thread). A warning message is also added since previously there was no indication that __analog_ss_thread was exiting abnormally. This resolves both the channel leak and the condition that led to the warning messages.

Resolves: #458

### DIFF
--- a/channels/sig_analog.c
+++ b/channels/sig_analog.c
@@ -1764,10 +1764,7 @@ static void *__analog_ss_thread(void *data)
 
 	ast_debug(1, "%s %d\n", __FUNCTION__, p->channel);
 
-	if (!chan) {
-		/* What happened to the channel? */
-		goto quit;
-	}
+	ast_assert(chan != NULL);
 
 	if ((callid = ast_channel_callid(chan))) {
 		ast_callid_threadassoc_add(callid);
@@ -2756,6 +2753,7 @@ int analog_ss_thread_start(struct analog_pvt *p, struct ast_channel *chan)
 {
 	pthread_t threadid;
 
+	p->ss_astchan = chan;
 	return ast_pthread_create_detached(&threadid, NULL, __analog_ss_thread, p);
 }
 


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

channels/sig_analog.c
```diff
@@ -1766,6 +1766,7 @@ static void *__analog_ss_thread(void *data)
 
 	if (!chan) {
 		/* What happened to the channel? */
+		ast_log(LOG_WARNING, "No Asterisk channel provided for DAHDI channel %d?\n", p->channel);
```
seanbright, line 1769: This is not something the user can fix so it should be a debug message

InterLinked1: I disagree... this means that something is wrong, and I feel it would be best to make some noise about this, since this should never happen. If there had been a warning here in the first place, this issue would have been resolved much more quickly.

I could make it a `LOG_ERROR` instead of a `LOG_WARNING`, since it really is an error of something wrong and not a warning of user misconfig.

`[2023-11-23 21:35:30.623] WARNING[1131310]: chan_dahdi.c:12011 do_monitor: Whoa....  I'm owned by DAHDI/12-1 but found (29) in read...` isn't something the user can fix either, but that's what clued me into this problem in the first place. Making it a debug message would just hide something that needs to be dealt with.

seanbright: Original comment still applies

InterLinked1: And I still disagree. Most `LOG_ERROR`'s in Asterisk are signs of something wrong with Asterisk itself, not an indication of something the user did or could do something about. This is a case where something is wrong with Asterisk itself. This is being consistent with existing logging practices, I see no reason to deviate from that.

If someone else still wants it to be a debug, I'll change it, but I don't think that's the right thing to do.

seanbright: It’s not going in unless it’s a debug.

jcolp: I agree with Sean, as well in the past log messages were not really thought about, reviewed, or considered. For log messages that are presented to users they should be informative to them and provide guidance if possible. This log message as it exists now is logging out an implementation detail aspect for developers and would be confusing for users.

channels/sig_analog.c
```diff
@@ -1766,6 +1766,7 @@ static void *__analog_ss_thread(void *data)
 
 	if (!chan) {
 		/* What happened to the channel? */
+		ast_debug(1, "Huh? No Asterisk channel provided for DAHDI channel %d?\n", p->channel);
```
seanbright, line 1769: If `chan` being `NULL` here is a programming error then this should probably be an `ast_assert()`. If it’s a runtime error then it’s fine as an `ast_debug()`.

seanbright (approved):

Issue #462, opened 2023-11-28
[new-feature]: app_dial: Add new option to preserve initial stream topology of caller


maximilianfridrich:
### Feature Description

Currently, when `app_dial` creates the outgoing channel it clones the stream topology of the incoming channel and then calls `ast_request_with_stream_topology` using that topology.

If the incoming channel has been in a call before, its stream topology might have changed in the meantime (e.g. codecs, stream directions). Sometimes this might not be desired and instead the original intention (topology) of the caller should be preserved and this topology be used for the outgoing channel.

PR #464, opened 2023-11-30
app_dial: Add option "j" to preserve initial stream topology of caller

maximilianfridrich:
Resolves: #462

UserNote: The option "j" is now available for the Dial application which uses the initial stream topology of the caller to create the outgoing channels.

### DIFF
--- a/apps/app_dial.c
+++ b/apps/app_dial.c
@@ -242,6 +242,10 @@
 					<para>Asterisk will ignore any connected line update requests or any redirecting party
 					update requests it may receive on this dial attempt.</para>
 				</option>
+				<option name="j">
+					<para>Use the initial stream topology of the caller for outgoing channels, even if the caller topology has changed.</para>
+					<para>NOTE: For this option to work, it has to be present in all invocations of Dial that the caller channel goes through.</para>
+				</option>
 				<option name="k">
 					<para>Allow the called party to enable parking of the call by sending
 					the DTMF sequence defined for call parking in <filename>features.conf</filename>.</para>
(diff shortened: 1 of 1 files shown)


maximilianfridrich:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

maximilianfridrich:
There's an accompanying testsuite PR that tests this new feature:

- https://github.com/asterisk/testsuite/pull/24

apps/app_dial.c
```diff
@@ -2627,7 +2643,20 @@ static int dial_exec_full(struct ast_channel *chan, const char *data, struct ast
 		 */
 		ast_party_connected_line_copy(&tmp->connected, ast_channel_connected(chan));
 
-		topology = ast_stream_topology_clone(ast_channel_get_stream_topology(chan));
+		topology_ds = ast_channel_datastore_find(chan, &topology_ds_info, NULL);
+
+		if (!topology_ds && (topology_ds = ast_datastore_alloc(&topology_ds_info, NULL))) {
```
jcolp, line 2648: I'm hesitant to incur the extra memory usage in all cases just for the few times someone may want to use the 'j' option. Is there a reason not to put it behind 'j'?

maximilianfridrich: It is done deliberately because the channel might go through multiple `Dial` invocations and only the last one might include the `j` option. If we didn't save the topology in the previous invocations, then we would not have access to it.

I understand your point though. Here's an alternative: We do it only for the `j` option and document that for this option to work, the `j` has to be used on all invocations of `Dial`  that the channel goes through.

jcolp (approved):

Issue #465, opened 2023-11-30
[improvement]: Change res_odbc connection pool request logic to not lock around blocking operations


creslin2877:
### Improvement Description

There are valid scenarios where res_odbc's connection pool might have some dead or stuck connections while others are healthy (imagine network elements/firewalls/routers silently timing out connections to a single DB and a single IP address, or a heterogeneous connection pool connected to potentially multiple IPs/instances of a replicated DB using a DNS front end for load balancing and one replica fails).

In order to time out those unhealthy connections without blocking access to other parts of Asterisk that may attempt access to the connection pool, it would be beneficial to not lock/block access around the entire pool in _ast_odbc_request_obj2 while doing potentially blocking operations on connection pool objects such as the connection_dead() test, odbc_obj_connect(), or by dereferencing a struct odbc_obj for the last time and triggering a odbc_obj_disconnect().

This would facilitate much quicker and concurrent timeout of dead connections via the connection_dead() test, which could block potentially for a long period of time depending on odbc.ini or other odbc connector specific timeout settings.

This also would make rapid failover (in the clustered DB scenario) much quicker.

creslin2877:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

PR #466, opened 2023-11-30
res_odbc.c: Allow concurrent access to request odbc connections

creslin287:
There are valid scenarios where res_odbc's connection pool might have some dead
or stuck connections while others are healthy (imagine network
elements/firewalls/routers silently timing out connections to a single DB and a
single IP address, or a heterogeneous connection pool connected to potentially
multiple IPs/instances of a replicated DB using a DNS front end for load
balancing and one replica fails).

In order to time out those unhealthy connections without blocking access to
other parts of Asterisk that may attempt access to the connection pool, it would
be beneficial to not lock/block access around the entire pool in
_ast_odbc_request_obj2 while doing potentially blocking operations on connection
pool objects such as the connection_dead() test, odbc_obj_connect(), or by
dereferencing a struct odbc_obj for the last time and triggering a
odbc_obj_disconnect().

This would facilitate much quicker and concurrent timeout of dead connections
via the connection_dead() test, which could block potentially for a long period
of time depending on odbc.ini or other odbc connector specific timeout settings.

This also would make rapid failover (in the clustered DB scenario) much quicker.

This patch changes the locking in _ast_odbc_request_obj2() to not lock around
odbc_obj_connect(), _disconnect(), and connection_dead(), while continuing to
lock around truly shared, non-immutable state like the connection_cnt member and
the connections list on struct odbc_class.

Fixes: #465

### DIFF
--- a/res/res_odbc.c
+++ b/res/res_odbc.c
@@ -915,12 +915,16 @@ struct odbc_obj *_ast_odbc_request_obj2(const char *name, struct ast_flags flags
 		return NULL;
 	}
 
-	ast_mutex_lock(&class->lock);
-
 	while (!obj) {
+		ast_mutex_lock(&class->lock);
+
 		obj = AST_LIST_REMOVE_HEAD(&class->connections, list);
 
+		ast_mutex_unlock(&class->lock);
+
 		if (!obj) {
+			ast_mutex_lock(&class->lock);
+
 			if (class->connection_cnt < class->maxconnections) {
 				/* If no connection is immediately available establish a new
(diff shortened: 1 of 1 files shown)


seanbright (approved):
LGTM

jcolp (approved):

gtjoseph:
@creslin287 This should be cherry-picked to the other branches.  Can you verify it's cherry-pickable and add the "cherry-pick-to" comment?

creslin287:
Deepest apologies - thought I already had.  (remember doing it, but maybe I never hit comment)

PR #471, opened 2023-12-02
pbx_config.c: Don't crash when unloading module.

### LINKED ISSUE #470: [bug]: Crash when unloading pbx_config
### Severity

Trivial

### Versions

21.0.0, master

### Components/Modules

pbx_config

### Operating Environment

```
Distributor ID: Ubuntu
Description:    Ubuntu 23.10
Release:        23.10
Codename:       mantic
```

### Frequency of Occurrence

Constant

### Issue Description

Don't load pbx_config at startup:

modules.conf:

```
...
noload = pbx_config
...
```

And at runtime load and then unload pbx_config:

```
*CLI> module load pbx_config
*CLI> module unload pbx_config
```

Asterisk crashes. This is due to dd171a4 not unsubscribing before module unload, but only when the FullyBooted event is fired which will not occur if the module is loaded/unloaded after startup.

### Relevant log output

_No response_

### Asterisk Issue Guidelines

seanbright:
`pbx_config` subscribes to manager events to capture the `FullyBooted` event but fails to unsubscribe if the module is loaded after that event fires. If the module is unloaded, a crash occurs the next time a manager event is raised.

We now unsubscribe when the module is unloaded if we haven't already unsubscribed.

Fixes #470

### DIFF
--- a/pbx/pbx_config.c
+++ b/pbx/pbx_config.c
@@ -104,6 +104,8 @@ static int clearglobalvars_config = 0;
 static int extenpatternmatchnew_config = 0;
 static char *overrideswitch_config = NULL;
 
+static struct stasis_subscription *fully_booted_subscription;
+
 AST_MUTEX_DEFINE_STATIC(save_dialplan_lock);
 
 AST_MUTEX_DEFINE_STATIC(reload_lock);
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 21

jcolp (approved):

PR #495, opened 2023-12-14
res_rtp_asterisk.c: Check DTLS packets against ICE candidate list

mbradeen:
When ICE is in use, we can prevent a possible DOS attack by allowing
DTLS protocol messages (client hello, etc) only from sources that
are in the active remote candidates list.

Resolves: GHSA-hxj9-xwr8-w8pq


mbradeen:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

mbradeen:
cherry-pick-test

gtjoseph (approved):

PR #497, opened 2023-12-14
res_pjsip_header_funcs: Duplicate new header value, don't copy.

mbradeen:
When updating an existing header the 'update' code incorrectly
just copied the new value into the existing buffer. If the
new value exceeded the available buffer size memory outside
of the buffer would be written into, potentially causing
a crash.

This change makes it so that the 'update' now duplicates
the new header value instead of copying it into the existing
buffer.

### DIFF
--- a/res/res_pjsip_header_funcs.c
+++ b/res/res_pjsip_header_funcs.c
@@ -676,6 +676,7 @@ static int add_header(void *obj)
 static int update_header(void *obj)
 {
 	struct header_data *data = obj;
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

mbradeen:
cherry-pick-test

gtjoseph (approved):

PR #498, opened 2023-12-14
res_pjsip: disable raw bad packet logging

mbradeen:
Add patch to split the log level for invalid packets received on the signaling port.
    The warning regarding the packet will move to level 2 so that it can still be displayed,
    while the raw packet will be at level 4.


mbradeen:
cherry-pick-to: none

gtjoseph (approved):

Issue #505, opened 2023-12-18
[bug]: res_pjproject: ast_sockaddr_cmp() always fails on sockaddrs created by ast_sockaddr_from_pj_sockaddr()


gtjoseph:
### Severity

Trivial

### Versions

18,20,21,master

### Components/Modules

res_pjproject

### Operating Environment

any

### Frequency of Occurrence

Constant

### Issue Description

If using ast_sockaddr_cmp() and either or both of the input sockaddrs were created with ast_sockaddr_from_pj_sockaddr(), the comparison will always be non-zero (not equal) even if both inputs have the exactly same family, address and port.  This is caused by ast_sockaddr_from_pj_sockaddr() not zeroing out the sin->sin_zero array in the destination socket.  With possible garbage in there, ast_sockaddr_cmp's memcmp of the entire sin structure will most likely return not-equal even if all other structure members are in fact equal.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

A1bi:
The issue persists on BSD platforms. For the past few days I have been struggling with Asterisk on FreeBSD dropping all DTLS packets. This is the first time I tried out connecting from a browser via WebRTC and I was surprised I couldn’t successfully make single call while all other kinds of connections to my server have been working flawlessly for years.

After some debugging I found the culprit to be another field only present on some platforms (mainly BSD derivatives) in `sockaddr_in6` / `sockaddr_in` not set by `ast_sockaddr_from_pj_sockaddr()`.

I have created a fix and will post a PR shortly.

Issue #503, opened 2023-12-15
[bug]: The res_rtp_asterisk DTLS check against ICE candidates fails when it shouldn't


gtjoseph:
### Severity

Blocker

### Versions

18.20.1,20.5.1,21.0.1

### Components/Modules

res_rtp_asterisk

### Operating Environment

Any

### Frequency of Occurrence

Frequent

### Issue Description

The rtp->ice_active_remote_candidates container used to validate the source of incoming DTLS packets doesn't contain peer reflexive candidates discovered during negotiation. This is causing the check to fail where it shouldn't.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

Issue #500, opened 2023-12-15
[bug regression]: res_rtp_asterisk doesn't build if pjproject isn't used


walter59:
### Severity

Blocker

### Versions

20.5.1 / 21.0.1

### Components/Modules

compile

### Operating Environment

linux 6.5.12 / pjproject 2.13.1 oor pjproject-2.14

### Frequency of Occurrence

None

### Issue Description

res_rtp_asterisk.c:3218:83: error: âPJ_ICE_MAX_CANDâ undeclared (first use in this function)
 3218 |                         while ((candidate = ao2_iterator_next(&i)) && (cand_cnt < PJ_ICE_MAX_CAND)) {
      |                                                                                   ^~~~~~~~~~~~~~~
res_rtp_aster

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
Is this with unbundled PJSIP?

gtjoseph:
@walter59 Was the error message in the issue description the only error, or were there others preceding it?  Since you mention pjproject versions, we're trying to confirm whether you're not using pjproject at all or using an externally built version.

If you weren't using it at all, you'd see the following...
```
   [CC] res_rtp_asterisk.c -> res_rtp_asterisk.o
res_rtp_asterisk.c: In function ‘__rtp_recvfrom’:
res_rtp_asterisk.c:3217:24: error: ‘struct ast_rtp’ has no member named ‘ice’
 3217 |                 if (rtp->ice) {
      |                        ^~
res_rtp_asterisk.c:3234:33: error: ‘struct ast_rtp’ has no member named ‘ice_media_started’
 3234 |                         if (!rtp->ice_media_started) {
      |                                 ^~
res_rtp_asterisk.c:3245:50: error: ‘struct ast_rtp’ has no member named ‘ice_active_remote_candidates’
 3245 |                         i = ao2_iterator_init(rtp->ice_active_remote_candidates, 0);
      |                                                  ^~
res_rtp_asterisk.c:3246:83: error: ‘PJ_ICE_MAX_CAND’ undeclared (first use in this function)
 3246 |                         while ((candidate = ao2_iterator_next(&i)) && (cand_cnt < PJ_ICE_MAX_CAND)) {
      |                                                                                   ^~~~~~~~~~~~~~~
```

Can you confirm?

PR #506, opened 2023-12-18
res_rtp_asterisk: Fix regression issues with DTLS client check

gtjoseph:
* Since ICE candidates are used for the check and pjproject is
  required to use ICE, res_rtp_asterisk was failing to compile
  when pjproject wasn't available.  The check is now wrapped
  with an #ifdef HAVE_PJPROJECT.

* The rtp->ice_active_remote_candidates container was being
  used to check the address on incoming packets but that
  container doesn't contain peer reflexive candidates discovered
  during negotiation. This was causing the check to fail
  where it shouldn't.  We now check against pjproject's
  real_ice->rcand array which will contain those candidates.

* Also fixed a bug in ast_sockaddr_from_pj_sockaddr() where
  we weren't zeroing out sin->sin_zero before returning.  This
  was causing ast_sockaddr_cmp() to always return false when
  one of the inputs was converted from a pj_sockaddr, even
  if both inputs had the same address and port.

Resolves: #500
Resolves: #503
Resolves: #505

### DIFF
--- a/include/asterisk/res_pjproject.h
+++ b/include/asterisk/res_pjproject.h
@@ -115,4 +115,17 @@ int ast_sockaddr_to_pj_sockaddr(const struct ast_sockaddr *addr, pj_sockaddr *pj
  */
 int ast_sockaddr_from_pj_sockaddr(struct ast_sockaddr *addr, const pj_sockaddr *pjaddr);
 
+/*!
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

bkford (approved):

Issue #341, opened 2023-09-20
[bug]: app_if.c : nested EndIf incorrectly exits parent If


asolovjov:
### Severity

Trivial

### Versions

18

### Components/Modules

app_if.c

### Operating Environment

It doesn't matter.
`Asterisk 18.19.0 built by builder @ builder.lo on a x86_64 running Linux on 2023-09-20 09:17:45`

### Frequency of Occurrence

None

### Issue Description

If nested if block condition is true then EndIf exits the parent if block. (BTW, If condition is false, it works correctly. ExitIf also works correctly.)
Example simple dialplan:
```
exten => _iftest,1,Noop
    same => n,If(1)
        same => n,Noop(inside main if)
        same => n,If(0)
            same => n,Noop()
        same => n,Endif
        same => n,Noop(after nested false if)
        same => n,If(1)
            same => n,Noop(inside true nested if)
        same => n,Endif
        same => n,Noop(after nested true if)
    same => n,Endif
    same => n,Noop(after main if)
    same => n,Hangup

> dialplan show iftest@test                                                                                                                                                                                                
[ Context 'test' created by 'pbx_config' ]
  '_iftest' =>      1. Noop()                                     [extensions_custom.conf:487]
                    2. If(1)                                      [extensions_custom.conf:488]
                    3. Noop(inside main if)                       [extensions_custom.conf:489]
                    4. If(0)                                      [extensions_custom.conf:490]
                    5. Noop()                                     [extensions_custom.conf:491]
                    6. Endif()                                    [extensions_custom.conf:492]
                    7. Noop(after nested false if)                [extensions_custom.conf:493]
                    8. If(1)                                      [extensions_custom.conf:494]
                    9. Noop(inside true nested if)                [extensions_custom.conf:495]
                    10. Endif()                                   [extensions_custom.conf:496]
                    11. Noop(after nested true if)                [extensions_custom.conf:497]
                    12. Endif()                                   [extensions_custom.conf:498]
                    13. Noop(after main if)                       [extensions_custom.conf:499]
                    14. Hangup()                                  [extensions_custom.conf:500]

```
We should see "inside true nested if" and then "after nested true if". Instead we jump to priority 12:
```
    -- Executing [iftest@test:1] NoOp("SIP/y1103-00000015", "") in new stack
    -- Executing [iftest@test:2] If("SIP/y1103-00000015", "1") in new stack
[...]

PR #512, opened 2023-12-23
app_if: Fix faulty EndIf branching.

InterLinked1:
This fixes faulty branching logic for the
EndIf application. Instead of computing
the next priority, which should be done
for false conditionals or ExitIf, we should
simply advance to the next priority.

Resolves: #341


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

InterLinked1:
Associated tests have also been submitted here: https://github.com/asterisk/testsuite/pull/26

Issue #513, opened 2023-12-26
[bug]: manager.c: Crash due to regression using wrong free function when built with MALLOC_DEBUG


InterLinked1:
### Severity

Critical

### Versions

21.0.2

### Components/Modules

manager.c

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Commit 424be345639d75c6cb7d0bd2da5f0f407dbd0bd5 (which ironically was a hotfix for another hotfix) introduced a regression causing a crash by calling `ast_free` on memory not allocated by Asterisk. This fixes that by using the correct free function.

### Relevant log output

```shell
Thread 1 (Thread 0x7f0c59b6f6c0 (LWP 2053099)):
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
        tid = <optimized out>
        ret = 0
        pd = <optimized out>
        old_mask = {__val = {0}}
        ret = <optimized out>
#1  0x00007f0c85c1ed9f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007f0c85bcff32 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
        ret = <optimized out>
#3  0x00007f0c85bba472 in __GI_abort () at ./stdlib/abort.c:79
        save_stage = 1
        act = {__sigaction_handler = {sa_handler = 0x20, sa_sigaction = 0x20}, sa_mask = {__val = {0 <repeats 13 times>, 139691021492016, 0, 0}}, sa_flags = -2050589549, sa_restore>
#4  0x000056452f3a1387 in ast_do_crash () at utils.c:2803
#5  0x000056452f1e0797 in my_do_crash () at astmm.c:230
#6  0x000056452f1e1222 in __ast_free (ptr=0x564531a0ce90, file=0x56452f49f8af "manager.c", lineno=3766, func=0x56452f4a5460 <__PRETTY_FUNCTION__.143> "_dtor_real_path") at astmm.c:>
        reg = 0x0
#7  0x000056452f4049a0 in _dtor_real_path (v=0x7f0c59b6e040) at manager.c:3766
        __PRETTY_FUNCTION__ = "_dtor_real_path"
        stripped_filename = 0x56452f49f860 "prefix != NULL"
[...]

PR #514, opened 2023-12-26
manager.c: Fix regression due to using wrong free function.

InterLinked1:
Commit 424be345639d75c6cb7d0bd2da5f0f407dbd0bd5 introduced a regression by calling ast_free on memory allocated by realpath. This causes Asterisk to abort when executing this function. Since the memory is allocated by glibc, it should be freed using ast_std_free.

Resolves: #513


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #525, opened 2024-01-10
[bug]: say.c: Money announcements off by one cent due to floating point rounding


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

say.c

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Floating point rounding error can cause certain money announcements to be wrong.

For example:
`SAYFILES(2.05,money)` will result in playing the prompts for $2.04.

This patch avoids the use of floating point to ensure the cents value is correct.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #526, opened 2024-01-10
say.c: Fix cents off-by-one due to floating point rounding.

InterLinked1:
Some of the money announcements can be off by one cent, due to the use of floating point in the money calculations, which is bad for obvious reasons.

This replaces floating point with simple string parsing to ensure the cents value is converted accurately.

Resolves: #525

### DIFF
--- a/main/say.c
+++ b/main/say.c
@@ -56,6 +56,7 @@
 #include "asterisk/utils.h"
 #include "asterisk/app.h"
 #include "asterisk/test.h"
+#include "asterisk/cli.h" /* use ESS */
 
 /* Forward declaration */
 static int wait_file(struct ast_channel *chan, const char *ints, const char *file, const char *lang);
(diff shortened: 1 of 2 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

main/say.c
```diff
@@ -353,26 +354,35 @@ static int say_digit_str_full(struct ast_channel *chan, const char *str, const c
...
-		amt = dollars * 100;
+	/* Don't use sscanf because floating point rounding
+	 * could distort the cents units. Just parse as string. */
+	dollars = atoi(str); /* atoi will stop at the period, if there is one */
+	period = strchr(str, '.');
+	if (period) {
+		/* Cents are after the decimal point, if present */
+		period++;
+		if (!ast_strlen_zero(period)) {
+			cents = atoi(period);
```
alex2grad, line 380: I think it'll be incorrect if the amount string has more then 2 digits after the period,
for example SAYFILES(2.051,money)

alex2grad (changes requested):

main/say.c
InterLinked1: Good catch, 1 digit was also not handled correctly

main/say.c
```diff
@@ -353,26 +354,45 @@ static int say_digit_str_full(struct ast_channel *chan, const char *str, const c
...
 
-	if (sscanf(str, "%30lf", &dollars) != 1) {
-		amt = 0;
-	} else { /* convert everything to cents */
-		amt = dollars * 100;
+	/* Don't use sscanf because floating point rounding
+	 * could distort the cents units. Just parse as string. */
+	dollars = atoi(str); /* atoi will stop at the period, if there is one */
+	period = strchr(str, '.');
+	if (period) {
```
seanbright, line 376: The decimal point would need to be just after the initial integer sequence, otherwise this would accept something like `2non-digit-stuff.05` as `2.05` which seems wrong (and regardless of FP accuracy is not consistent with the current behavior).

Using `strtol(...)` instead of `atoi(...)` would give you a pointer to the first character after any initial digits and you could just check that directly against `'.'`.

seanbright: And now that I think about it I think this would accept `blahblah.05` as `0.05` which also seems incorrect.

gtjoseph (changes requested):
Please address @seanbright's comments.

asteriskteam:

main/say.c
InterLinked1: I think I was overthinking this big time, I redid the patch with simpler logic that is more correct

main/say.c
```diff
@@ -353,26 +354,29 @@ static int say_digit_str_full(struct ast_channel *chan, const char *str, const c
...
+	filenames = ast_str_create(20);
 	if (!filenames) {
 		return NULL;
 	}
 	ast_str_reset(filenames);
 
-	if (sscanf(str, "%30lf", &dollars) != 1) {
+	/* Don't use %f because floating point rounding
+	 * could distort the cents units. Just parse as string. */
+	if (sscanf(str, "%d.%02d", &dollars, &cents) < 1) {
```
seanbright, line 373: Should be allowed: `.25`
Should fail: `1.-25` and `1.    25`

InterLinked1: Thanks, covered these in some additional test cases

asteriskteam:

gtjoseph (approved):

Issue #527, opened 2024-01-10
[bug]: app_voicemail_odbc no longer working after removal of macrocontext.


mbradeen:
### Severity

Major

### Versions

21, master

### Components/Modules

app_voicemail_odbc

### Operating Environment

all

### Frequency of Occurrence

Constant

### Issue Description

app_voicemail_odbc is no longer able to add or access the voicemail_messages table.

This is due to the macrocontext being incompletely removed when app_macro was removed from source.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #528, opened 2024-01-10
app_voicemail_odbc: remove macrocontext from voicemail_messages table

mbradeen:
When app_macro was deprecated, the macrocontext column was removed from
the INSERT statement but the binds were not renumbered. This broke the
insert.

This change removes the macrocontext column via alembic and re-numbers
the existing columns in the INSERT.

Fixes: #527

UserNote: The fix requires removing the macrocontext column from the
voicemail_messages table in the voicemail database via alembic upgrade.

UpgradeNote: The fix requires that the voicemail database be upgraded via
alembic. Upgrading to the latest voicemail database via alembic will
remove the macrocontext column from the voicemail_messages table.


mbradeen:
cherry-pick-to: 21

jcolp (approved):

gtjoseph (changes requested):
I had to go look at the code to understand what happened.  This might be a better commit description: 

When app_macro was deprecated, the macrocontext column was removed from the INSERT statement and while the macrocontext parameter bind was removed, the binds weren't renumbered.  This broke the insert.

Also, I think the UserNote content should be the first sentence of the UpgradeNote.

jcolp (approved):

gtjoseph (approved):

Issue #530, opened 2024-01-12
[bug]: bridge_channel.c: Stream topology change amplification with multiple layers of Local channels


InterLinked1:
### Severity

Major

### Versions

21.0.0

### Components/Modules

bridge_channel.c

### Operating Environment

Debian 12

### Frequency of Occurrence

Constant

### Issue Description

I dug into a specific instance of the infamous `Extremely long queue length queuing...` warnings that I was seeing and found that this is due to an amplification of stream topology change control frames (`AST_CONTROL_STREAM_TOPOLOGY_CHANGED`) being propagated back and forth between the channels. By this, I mean if I analyze how many frames are queued on the channel, it grows to oblivion, quickly reaching thousands of queued frames within seconds and this will keep growing as long as the channels are bridged.

As far as these kinds of things go, they make the bridge completely unusable, and result in a generally degraded system due to high CPU usage - and left unchecked, this is technically a resource leak until channels are hung up since frames will accumulate forever - faster than they are being consumed. If users don't do that manually, then I suspect Asterisk would, in time, use all available memory. Based on my experience, I suspect a number of instances of this error message might have the same underlying issue.

I've put together a sample dialplan that reproduces this as well as included logs from running against that, with some debug messages added to show the continuous stream topology change frames. This happens precisely when dialing a queue agent that is via a local channel and the agent does an attended transfer, and when the channel gets bridged.

The issue only happens when using a local channel; however, `core_unreal` isn't constantly triggering the stream topology change frames, that might be happening just in the bridging core.

(Either `Queue` or `Dial` can be used, though Dial is simpler for reproduction purposes).

Caller A calls B, and B initiates an attended transfer, e.g. using *2 (starting at `test-operator-transfer,1,1`)

The issue happens with both `Wait` and `BridgeWait`, i.e. regardless of if the channel is masqueraded.

Example below where party A dials extension 123, triggering the issue when B (`PJSIP/Polycom5` in this example) initiates the attended transfer, more precisely 3 seconds afterwards, when the channels are bridged.

I went ahead and reproduced it using the following on a second system, using DAHDI instead of PJSIP for both endpoints, without modifications to log bridging changes but indeed I do get hundreds of these messages, so this issue is very easy to reproduce consistently:

`DEBUG[136258][C-000001e4]: chan_dahdi.c:9344 dahdi_indicate: Requested indication 36 on channel DAHDI/18-1`

```
features.conf:

[featuremap]
atxfer => *2 

Dialplan:

[...]

PR #532, opened 2024-01-12
Revert "core & res_pjsip: Improve topology change handling."

gtjoseph:
This reverts commit 315eb551dbd18ecd424a2f32179d4c1f6f6edd26.

Over the past year, we've had several reports of "topology storms"
occurring where 2 external facing channels connected by one or more
local channels and bridges will get themselves in a state where
they continually send each other topology change requests.  This
usually manifests itself in no-audio calls and a flood of
"Exceptionally long queue length" messages.  It appears that this
commit is the cause so we're reverting it for now until we can
determine a more appropriate solution.

Resolves: #530

### DIFF
--- a/main/channel.c
+++ b/main/channel.c
@@ -10970,6 +10970,15 @@ int ast_channel_request_stream_topology_change(struct ast_channel *chan,
 		return -1;
 	}
 
+	if (ast_stream_topology_equal(ast_channel_get_stream_topology(chan), topology)) {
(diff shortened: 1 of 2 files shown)


gtjoseph:
cherry-pick-to: 18\ncherry-pick-to: 20\ncherry-pick-to: 21\n

jcolp (approved):

InterLinked1 (commented):
Small nitpick... *themseles* should be *themselves*

Issue #533, opened 2024-01-12
[improvement]: channel.c, func_frame_trace.c: Improve debuggability of channel frame queue


InterLinked1:
### Improvement Description

The `Exceptionally long queue length...` warning is a common source of fatal bridging problems in Asterisk, yet doesn't convey any specific information that can be useful for determining the problem. This adjusts the log message to make it more useful, and also adds a CLI command to `func_frame_trace` that can be used to easily inspect the frame queue and quickly narrow down the general direction of the problem. These enhancements were used to narrow down the issue described in #530

PR #534, opened 2024-01-12
func_frame_trace: Add CLI command to dump frame queue.

InterLinked1:
This adds a simple CLI command that can be used for analyzing all frames currently queued to a channel.

A couple log messages are also adjusted to be more useful in tracing bridging problems.

Resolves: #533

### DIFF
--- a/funcs/func_frame_trace.c
+++ b/funcs/func_frame_trace.c
@@ -35,6 +35,7 @@
 #include "asterisk/channel.h"
 #include "asterisk/pbx.h"
 #include "asterisk/framehook.h"
+#include "asterisk/cli.h"
 
 /*** DOCUMENTATION
 	<function name="FRAME_TRACE" language="en_US">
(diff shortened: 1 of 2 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

main/features.c
```diff
@@ -1011,7 +1011,7 @@ static int bridge_exec(struct ast_channel *chan, const char *data)
 			strlen(args.dest_chan));
 	}
 	if (!current_dest_chan) {
-		ast_verb(4, "Bridge failed because channel '%s' does not exist\n",
+		ast_log(LOG_WARNING, "Bridge failed because channel '%s' does not exist\n",
```
jcolp, line 1014: This seems like an unrelated logging change

jcolp (changes requested):

main/features.c
InterLinked1: It's related in that if bridging fails, it's now noticeable to the user. Before it was just a verbose, even though something went wrong.

But I'll put up another review at some point for log message related changes so if you'd rather it go with that, I can move it there.

jcolp: That's a separate thing and should be a separate review.

jcolp (approved):

PR #537, opened 2024-01-15
master issue 536

PeterHolik:
- chan_rtp.c: MulticastRTP missing refcount without codec option


jcolp:
This needs to be a single commit for the fix in question, and not include the prior one.

PR #547, opened 2024-01-21
app_confbridge: Don't emit warnings on valid configurations.

### LINKED ISSUE #546: [improvement]: confbridge.conf.sample has outdated information
## tl;dr
Seems like the sample configuration for  is outdated (tested on Asterisk 21): `auto` isn't accepted anymore.

## warning
```
 Loading app_confbridge.so.
[2024-01-21 05:27:24] WARNING[188400]: config_options.c:1495 uint_handler_fn: Attempted to set internal_sample_rate=auto, but set it to 0 instead due to default)
```

## location
https://github.com/asterisk/asterisk/blob/0668e5494a05b87b811033c139a58dfbc4a4c200/configs/samples/confbridge.conf.sample#L212-L218

## possible solution
```
;internal_sample_rate=0        ; Sets the internal native sample rate the
                               ; conference is mixed at.  This is set to automatically
                               ; adjust the sample rate to the best quality by default.
                               ; Other values can be anything from 8000-192000.  If a
                               ; sample rate is set that Asterisk does not support, the
                               ; closest sample rate Asterisk does support to the one requested
                               ; will be used.
```

seanbright:
The numeric bridge profile options `internal_sample_rate` and `maximum_sample_rate` are documented to accept the special values `auto` and `none`, respectively. While these values currently work, they also emit warnings when used which could be confusing for users.

In passing, also ensure that we only accept the documented range of sample rate values between 8000 and 192000.

Fixes #546

### DIFF
--- a/apps/confbridge/conf_config_parser.c
+++ b/apps/confbridge/conf_config_parser.c
@@ -2220,6 +2220,30 @@ static int user_template_handler(const struct aco_option *opt, struct ast_variab
 	return conf_find_user_profile(NULL, var->value, u_profile) ? 0 : -1;
 }
 
+static int sample_rate_handler(const struct aco_option *opt, struct ast_variable *var, void *obj)
+{
+	struct bridge_profile *b_profile = obj;
+	unsigned int *slot;
+
+	if (!strcasecmp(var->name, "internal_sample_rate")) {
+		slot = &b_profile->internal_sample_rate;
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #548, opened 2024-01-18
[improvement]: Get Record() audio duration/length


lukeescude:
### Improvement Description

Hello!

The Record() function has nice built-in features, but we lack the ability to retrieve the exact duration that elapsed during the record process.

It would be excellent to be able to access the length/duration from a dialplan variable set after Record() has completed (or set to 0 if an error occurred).

Reference doc: https://docs.asterisk.org/Asterisk_20_Documentation/API_Documentation/Dialplan_Applications/Record/

Thanks!

jcolp:
Request completed.

PR #549, opened 2024-01-22
app_record: Add RECORD_TIME output variable.

InterLinked1:
This adds the RECORD_TIME variable to Record(),
which is set to the recording duration before
the application returns.

Resolves: #548

### DIFF
--- a/apps/app_record.c
+++ b/apps/app_record.c
@@ -114,6 +114,10 @@
 					<value name="HANGUP">The channel was hung up.</value>
 					<value name="ERROR">An unrecoverable error occurred, which resulted in a WARNING to the logs.</value>
 				</variable>
+				<variable name="RECORD_TIME">
+					<para>Will be set to the duration of the recording, in milliseconds.</para>
+					<para>If the recording fails for whatever reason, this will still be set to 0.</para>
+				</variable>
 			</variablelist>
 		</description>
 	</application>
@@ -251,6 +255,9 @@ static int record_exec(struct ast_channel *chan, const char *data)
 	int ms;
 	struct timeval start;
 	const char *status_response = "ERROR";
+	char durationbuf[20];
+
+	pbx_builtin_setvar_helper(chan, "RECORD_TIME", "0"); /* Reset, in case already set */
 
 	/* The next few lines of code parse out the filename and header from the input string */
 	if (ast_strlen_zero(data)) { /* no data implies no filename or anything is present */
@@ -514,6 +521,9 @@ static int record_exec(struct ast_channel *chan, const char *data)
 		ast_channel_stop_silence_generator(chan, silgen);
 
 out:
+	snprintf(durationbuf, sizeof(durationbuf), "%ld", ast_tvdiff_ms(ast_tvnow(), start));
+	pbx_builtin_setvar_helper(chan, "RECORD_TIME", durationbuf);
+
 	if ((silence > 0) && rfmt) {
 		res = ast_set_read_format(chan, rfmt);
 		if (res) {


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

apps/app_record.c
```diff
@@ -514,6 +518,9 @@ static int record_exec(struct ast_channel *chan, const char *data)
 		ast_channel_stop_silence_generator(chan, silgen);
 
 out:
+	snprintf(durationbuf, sizeof(durationbuf), "%ld", ast_tvdiff_ms(ast_tvnow(), start));
+	pbx_builtin_setvar_helper(chan, "RECORD_TIME", durationbuf);
```
jcolp, line 522: I think in all cases RECORD_TIME should be set, in error cases it should be set to 0 so that there isn't state carryover between invocations, and it should be documented as such in the variable XML.

jcolp (changes requested):
A user note should be added to the commit message about the addition of this dialplan variable.

jcolp:
> This could clobber an existing variable. Maybe it should be master only?

Ooh, excellent point I hadn't considered. That or turning it into a dialplan function which can pull info about the recording just done.

jcolp (approved):

mbradeen (approved):

gtjoseph (changes requested):
Should be a UserNote (new feature) not UpgradeNote (possible breaking change).

InterLinked1:
> Should be a UserNote (new feature) not UpgradeNote (possible breaking change).

It was requested this be master only because it "could" be breaking, that's why I removed the cherry picks and made it an upgrade note. Should it still be a user note?

jcolp:
> > Should be a UserNote (new feature) not UpgradeNote (possible breaking change).
> 
> It was requested this be master only because it "could" be breaking, that's why I removed the cherry picks and made it an upgrade note. Should it still be a user note?

Since there could be a conflict in variable usage, I think UpgradeNote is appropriate.

seanbright:
Is there another way this information could be made available without creating a channel variable? If it was a dialplan function as @jcolp mentioned you could make other information available too (recording format, file size, etc.) and that could be easily extendable in the future.

asteriskteam:

InterLinked1:
> Is there another way this information could be made available without creating a channel variable? If it was a dialplan function as @jcolp mentioned you could make other information available too (recording format, file size, etc.) and that could be easily extendable in the future.
> 
> Edit: And then it would be a candidate for cherry picking which would be good,
> 
> Edit edit: Although I suppose a dialplan function have the same naming collision problem. I'm not sure if you can have a variable and a dialplan function with the same name.

Yeah, aside from that, a function to get the time feels kind of disjoint (at least to me), a variable is more consistent with the other things that are saved. Making it a function would probably also mean now a datastore is now needed to keep track of that, which seems like overkill for just exposing one piece of information. A function could be useful anyways but seems like it should be its own project to expose different information.

Personally I don't really feel like changing this if it's only a suggestion, but if you feel this should absolutely be a function, then I can do so.

seanbright:
> Edit edit: Although I suppose a dialplan function have the same naming collision problem. I'm not sure if you can have a variable and a dialplan function with the same name.

I just tested this and you can have a variable and a function with the same name - there is no conflict:

    same =     n,NoOp(${CURL})
    same =     n,Set(CURL=hello-${EXTEN})
    same =     n,NoOp(${CURL})
    same =     n,Hangup()

> Personally I don't really feel like changing this if it's only a suggestion, but if you feel this should absolutely be a function, then I can do so.

This PR is not waiting on my review as far as I can tell.

My opinion is that new functionality should not create channel variables that could trample those set by the user. It's a suggestion.

jcolp:
I agree regarding not using dialplan variables. Older things use dialplan variables because at the time there was no other way. For inclusion I will require it to be a dialplan function.

gtjoseph (dismissed):
~~UpgradeNotes should be for changes in existing behavior a user should be aware of.  This should be a UserNote.~~
Nevermind.  But, the upgrade note should explain the possible conflict.

asteriskteam:

seanbright:
Um. What?

jcolp:
When did I actually approve this... why... *looks back*

jcolp:
Oh, I never unapproved when we had our discussion. That explains it. I'm reverting this and reopening.

Issue #551, opened 2024-01-25
[bug]: manager: UpdateConfig triggers reload with "Reload: no"


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

manager.c

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

If the `Reload` header is provided, a reload will always occur, even if `Reload: no`. The documentation says that this should not happen, but the code is a little bit sloppy. This patch makes the header work as described in the documentation.

### Relevant log output

```shell
root@testdebian:/usr/local/src/cami# ./amicli -utest
*** Successfully logged in to AMI on 127.0.0.1 (test) ***
(Callback) Event Received: FullyBooted
(Callback) Event Received: SuccessfulAuth
Action: UpdateConfig
reload: no
srcfilename: /etc/asterisk/test.conf
dstfilename: /etc/asterisk/test.conf
PreserveEffectiveContext
...


******* RESPONSE *******
ActionID: 2 (Success) => # Events: 1
*** Event => # Fields: 2
Response        : Success
ActionID        : 2
(Callback) Event Received: Reload
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #552, opened 2024-01-25
manager.c: Fix erroneous reloads in UpdateConfig.

InterLinked1:
Currently, a reload will always occur if the
Reload header is provided for the UpdateConfig
action. However, we should not be doing a reload
if the header value has a falsy value, per the
documentation, so this makes the reload behavior
consistent with the existing documentation.

Resolves: #551


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

seanbright:
What should `Reload: yes` do? I feel like it should reload all modules but this change would prevent that. It’s possible I am reading the code incorrectly.

jcolp (approved):

Issue #565, opened 2024-01-31
[bug]: Application Read() returns immediately


btriller:
### Severity

Major

### Versions

20.6

### Components/Modules

app_read

### Operating Environment

```
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
```

### Frequency of Occurrence

Constant

### Issue Description

Application `Read(read_pin,,4)` returns immediately to dial plan without waiting for digits.
This worked in 20.5.2.
[Test](https://github.com/asterisk/testsuite/blob/master/tests/apps/read/test-config.yaml) also fails.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

InterLinked1:
I actually received a report of this as well, but hadn't had time to investigate yet.

jcolp:
It's the change from issue #172

jcolp:
I think it's a behavior difference between strsep and ast_strsep, in the ast_strsep case I think it is returning NULL which is skipping the logic, instead of returning a pointer to "".

jcolp:
Yeah, that seems to be the case and the logic is explicitly written as such, doesn't seem to be documented. I think ast_strsep should return NULL if passed NULL, but if it is passed an empty string it should return the empty string will match strsep behavior.

seanbright:
So you’re suggesting modifying `ast_strsep(…)` as a resolution?

jcolp:
Yes, I think it should behave the same as strsep in that case.

jcolp:
Maybe something as simple as having it return "st" instead of NULL, for the case where it is passed a NULL or empty string...

jcolp:
Pull request is up which makes the behavior match, and resolves this issue. Give it a whirl.

lukaszMaster:
I observed the same behavior on 18.21.0. The trick was to fill in the second parameter (name of sound file). After that 'Read' started to wait for digits.

PR #568, opened 2024-01-31
utils: Make behavior of ast_strsep* match strsep.

jcolp:
Given the scenario of passing an empty string to the ast_strsep functions the functions would return NULL instead of an empty string. This is counter to how strsep itself works.

This change alters the behavior of the functions to match that of strsep.

Fixes: #565

### DIFF
--- a/include/asterisk/strings.h
+++ b/include/asterisk/strings.h
@@ -271,7 +271,7 @@ enum ast_strsep_flags {
   AST_STRSEP_UNESCAPE unescapes '\' sequences.
   AST_STRSEP_ALL does all of the above processing.
   \return The next token or NULL if done or if there are more than 8 levels of
-  nested quotes.
+  nested quotes. If provided an empty string, will return the empty string.
 
   This function acts like strsep with three exceptions...
   The separator is a single character instead of a string.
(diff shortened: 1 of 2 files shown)


jcolp:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

InterLinked1 (approved):

alex2grad (commented):
Need to change function description in include/asterisk/strings.h
  \return The next token or NULL if done or if there are more than 8 levels of
  nested quotes.

Are you sure this change won't break something else?
There are a lot of while ((token = ast_strsep...

What's about ast_strsep_quoted?
Should ast_strsep_quoted match ast_strsep behavior?

jcolp:
The ast_strsep_quoted function was also updated, which is why I used ast_strsep* to mean both occurrences to keep the length acceptable.

I've reviewed the usage of this and didn't see any cases where a problem would occur as a result of changing this, but code review exists so other people can look so I urge others to do so.

seanbright (approved):

gtjoseph (approved):

gtjoseph:
@jcolp You may have a legit test failure....
rest_api/asterisk/config/dynamic/pjsip/modify_endpoint
fails in all branches.

gtjoseph (changes requested):
rest_api/asterisk/config/dynamic/pjsip/modify_endpoint keep failing.

gtjoseph (approved):

Issue #569, opened 2024-02-01
[improvement]: Add option to interleave input and output frames on spied channel


mbradeen:
### Improvement Description

CommUnity UCaaS submission:  Add option to app chanspy that allows the input and output to the spied channel to be interleaved rather than mixed.  This would allow the receiver of the spied audio to extract the input and output separately.

PR #570, opened 2024-02-01
app_chanspy: Add 'D' option for dual-channel audio

mbradeen:
Adds the 'D' option to app chanspy that causes the input and output
frames of the spied channel to be interleaved in the spy output frame.
This allows the input and output of the spied channel to be decoded
separately by the receiver.

If the 'o' option is also set, the 'D' option is ignored as the
audio being spied is inherently one direction.

Fixes: #569

UserNote: The ChanSpy application now accepts the 'D' option which
will interleave the spied audio within the outgoing frames. The
purpose of this is to allow the audio to be read as a Dual channel
stream with separate incoming and outgoing audio. Setting both the
'o' option and the 'D' option and results in the 'D' option being
ignored.

### DIFF
--- a/apps/app_chanspy.c
+++ b/apps/app_chanspy.c
@@ -245,6 +245,11 @@
 							</enum>
 						</enumlist>
 					</option>
+					<option name="D">
+						<para>Interleave the audio coming from the channel and the audio coming to the channel in
+						the output audio as a dual channel stream, rather than mix it. Does nothing if 'o'
+						is also set.</para>
+					</option>
 					<option name="e">
 						<argument name="ext" required="true" />
 						<para>Enable <emphasis>enforced</emphasis> mode, so the spying channel can
@@ -393,6 +398,7 @@ enum {
 	OPTION_EXITONHANGUP      = (1 << 18),   /* Hang up when the spied-on channel hangs up. */
 	OPTION_UNIQUEID          = (1 << 19),	/* The chanprefix is a channel uniqueid or fully specified channel name. */
 	OPTION_LONG_QUEUE        = (1 << 20),	/* Allow usage of a long queue to store audio frames. */
+	OPTION_INTERLEAVED       = (1 << 21),	/* Interleave the Read and Write frames in the output frame. */
 };
 
 enum {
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

apps/app_chanspy.c
```diff
@@ -471,6 +478,46 @@ static int spy_generate(struct ast_channel *chan, void *data, int len, int sampl
...
+		short write_buf[samples];
+		f = ast_audiohook_read_frame_all(&csth->spy_audiohook, samples, ast_format_slin, &fr_read, &fr_write);
+		if (f) {
+			ast_frame_free(f, 0);
+			f = NULL;
+		}
+		memset(read_buf, 0, sizeof(read_buf));
+		memset(write_buf, 0, sizeof(write_buf));
+		if (fr_read) {
+			memcpy(read_buf, fr_read->data.ptr, sizeof(read_buf));
```
InterLinked1, line 501: Why memset right before we memcpy?
Would make more sense to have:
if fr_read memcpy else memset

apps/app_chanspy.c
```diff
@@ -471,6 +478,46 @@ static int spy_generate(struct ast_channel *chan, void *data, int len, int sampl
...
+		}
+		if (fr_write) {
+			memcpy(write_buf, fr_write->data.ptr, sizeof(write_buf));
+		}
+
+		for (int i = 0; i < samples; i++) {
```
InterLinked1, line 507: You can't declare variables in the for loop pre-condition, move `i` to the top of the block.

InterLinked1 (changes requested):

apps/app_chanspy.c
```diff
@@ -245,6 +245,11 @@
 							</enum>
 						</enumlist>
 					</option>
+					<option name="D">
+						<para>Interleave the audio coming from the channel and the audio coming to the channel in
+						the output audio as a Dual channel stream, rather than mix it. Does nothing if 'o'
```
jcolp, line 250: dual channel

jcolp (changes requested):

jcolp (approved):

Issue #572, opened 2024-02-01
[improvement]: Copy partial speech results when Asterisk is ready to move on but the speech backend is not


chrsmj:
### Improvement Description

Some speech recognition engines, such as third party Vosk, produce partial results but the timeout imposed by Asterisk dial plan settings might be too short for the engine to confirm the results. Patch will address this and return whatever is available at the timeout.

PR #573, opened 2024-02-01
Partial speech results retrieval on timeout

chrsmj:
Resolves: #572


chrsmj:
Wrong branch, disregard please.

PR #574, opened 2024-02-01
Partial speech results collection even when backend is not complete done.

chrsmj:
Some speech backends might generate partial results before the final results. If Asterisk timeout happens before they are finalized, then this patch will at least return the partial results instead of empty strings.

Resolves: #572

### DIFF
--- a/apps/app_speech_utils.c
+++ b/apps/app_speech_utils.c
@@ -933,6 +933,11 @@ static int speech_background(struct ast_channel *chan, const char *data)
 		}
 	}
 
+	/* Copy to speech structure the results, even partial ones, if available */
+	if (speech->state == AST_SPEECH_STATE_READY) {
+		speech->results = ast_speech_results_get(speech);
+	}
+
 	if (!ast_strlen_zero(dtmf)) {
 		/* We sort of make a results entry */
 		speech->results = ast_calloc(1, sizeof(*speech->results));


jcolp:
This can not be looked at until a contributor license agreement is signed: https://oss-cla.sangoma.com/asterisk/asterisk?pullRequest=574

The commit message also does not follow the standard: https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/?h=commit+messages

chrsmj:
CLA bot is goofy. Clicking your link shows "You have agreed to the CLA for multiple repositories or organizations" at the top.

jcolp:
Ah, yes, I see you signed yesterday afternoon.

chrsmj:
Thank you. There were older agreements probably on file. Anyhow, commit message is updated now as well.

apps/app_speech_utils.c
```diff
@@ -933,6 +933,11 @@ static int speech_background(struct ast_channel *chan, const char *data)
 		}
 	}
 
+	/* Copy to speech structure the results, even partial ones, if available */
```
jcolp, line 936: This behavior should be documented in the documentation for the dialplan application.

apps/app_speech_utils.c
```diff
@@ -933,6 +933,11 @@ static int speech_background(struct ast_channel *chan, const char *data)
...
+	if (speech->state == AST_SPEECH_STATE_READY) {
+		speech->results = ast_speech_results_get(speech);
+	}
+
 	if (!ast_strlen_zero(dtmf)) {
 		/* We sort of make a results entry */
 		speech->results = ast_calloc(1, sizeof(*speech->results));
```
jcolp, line 943: If the speech engine returns results, but DTMF has been received, then this will leak memory.

jcolp (changes requested):

chrsmj:
Comments considered and abandoning this for another PR.

PR #576, opened 2024-02-02
app_speech_utils.c: Allow partial speech results.

chrsmj:
Adds 'p' option to SpeechBackground() application.
With this option, when the app timeout is reached,
whatever the backend speech engine collected will
be returned as if it were the final, full result.
(This works for engines that make partial results.)

Resolves: #572

UserNote: The words might be less than the best.

### DIFF
--- a/apps/app_speech_utils.c
+++ b/apps/app_speech_utils.c
@@ -105,6 +105,9 @@
 					<option name="n">
 						<para>Don't answer the channel if it has not already been answered.</para>
 					</option>
+					<option name="p">
+						<para>Return partial results when backend is terminated by timeout.</para>
+					</option>
 				</optionlist>
 			</parameter>
 		</syntax>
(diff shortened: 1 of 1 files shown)


gtjoseph:
You need to sign the CLA.  Click on the "Details" link in the license/cla line in the checks list above.

jcolp:
@gtjoseph It is signed, not sure why it is not showing up on here.

gtjoseph:
> @gtjoseph It is signed, not sure why it is not showing up on here.

Bah.

jcolp (changes requested):
The UserNote is what gets placed into release notes. As it is, it gives no indication as to what it is referring to or what it means. It needs to be specific, for example:

"The SpeechBackground dialplan application now supports a 'p' optional that will return partial results from speech engines that provide them when a timeout occurs."

chrsmj:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (approved):

PR #578, opened 2024-02-06
pseudo-MOS calculation formula change

romryz:
res_rtp_asterisk.c: Correct coefficient in MOS calculation.

Media Experience Score relies on incorrect pseudo_mos variable
calculation. According to forming an opinion section of the
documentation, calculation relies on ITU-T G.107 standard:

    https://docs.asterisk.org/Deployment/Media-Experience-Score/#forming-an-opinion

ITU-T G.107 Annex B suggests to calculate MOS with a coefficient
"seven times ten to the power of negative six", 7 * 10^(-6). which
would mean 6 digits after the decimal point. Current implementation
has 7 digits after the decimal point, which downrates the calls.

Fixes: #597

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -6240,7 +6240,7 @@ static double calc_media_experience_score(struct ast_rtp_instance *instance,
 	} else if (r_value > 100) {
 		pseudo_mos = 4.5;
 	} else {
-		pseudo_mos = 1 + (0.035 * r_value) + (r_value * (r_value - 60) * (100 - r_value) * 0.0000007);
+		pseudo_mos = 1 + (0.035 * r_value) + (r_value * (r_value - 60) * (100 - r_value) * 0.000007);
 	}
 
 	/*


jcolp:
Please see the code contribution instructions: https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/

The contributor license agreement needs to be signed, and the commit message updated to standard.

jcolp:
As well, https://docs.asterisk.org/Deployment/Media-Experience-Score/ has links to what was used for reference information and formula details.

romryz:
@jcolp thanks, the ITU-T G.107 suggests the coefficient 7*10^(-6), which is 0.000007 (6 digits after the decimal point) vs current 7 digits after decimal point

gtjoseph (changes requested):
A few things...   

* Add the info from the PR description to the commit message.
* Please open an issue and reference it in the commit message.
* Add a comment with the appropriate 'cherry-pick-to' lines.

See ...

seanbright:
The analysis and change are correct. @romryz you need to update the commit message (`git commit --amend`) and then force-push your changes (`git push --force origin patch-1`). I wouldn't worry about creating an issue.

seanbright:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

seanbright:
You can refer to issue #597

seanbright:
@romryz the commit message is still not correct. I’ll provide you with one.

seanbright:
@romryz Update to use [this commit message](https://gist.github.com/seanbright/ff1fef05fe3818f4579df9d868bfe3b3). Just replace your entire commit message with that one and force-push the change again.

seanbright:
That’s not correct still. Are you using the web UI?

romryz:
no, it's the gh cli. is it the top two lines that have to be removed?

romryz:
@seanbright could you check now please, updated (again)

seanbright:
Perfect. Thank you for doing all that.

seanbright (approved):

romryz:
@seanbright thank you for supporting me on it!

seanbright:
You should be able to re-request @gtjoseph’s review

gtjoseph:
@romryz Do you happen to have an example of a "before and after" so we can see what the actual effect of changing the decimal places is on the MES?    If not, that's fine.  I'm going to test it later today anyway.

jcolp (approved):

romryz:
@gtjoseph we made some comparison on formulas to implement so I reused that template to showcase the difference - the only variable here is the T time from ITU-T, but gives an idea of how glissade looks like:
MOS_comparison

gtjoseph:
> @gtjoseph we made some comparison on formulas to implement so I reused that template to showcase the difference - the only variable here is the T time from ITU-T, but gives an idea of how glissade looks like: 

Prefect, thanks!  I should be able to get this merged this afternoon or tomorrow morning.

gtjoseph:
I just tested a call between 2 local phones using ulaw both with and without the change and got an MES of 85 without the change and 88 with the change.  Those being a pseudo_mos of 4.25 and 4.4 respectively.  I thought I might have referenced "85" as the theoretical max in the documentation but I didn't so I'm good.

Issue #582, opened 2024-02-06
[improvement]: Reduce unneeded logging during startup and shutdown


gtjoseph:
### Improvement Description

When started with a verbose level of 3, asterisk can emit over 1500 verbose message that serve no real purpose other than to fill up logs.  When asterisk shuts down, it emits another 1100 that are of even less use.  Since the testsuite runs asterisk with a verbose level of 3, and asterisk starts and stops for every one of the 700+ tests, the number of log messages is staggering.

The PR associated with this issue changes the log level for those verbose messages to 5 instead of 3 which reduces the number of log messages to only a handful.  Of course, NOTICE, WARNING and ERROR message are unaffected.

PR #583, opened 2024-02-06
Reduce startup/shutdown verbose logging

gtjoseph:
When started with a verbose level of 3, asterisk can emit over 1500
verbose message that serve no real purpose other than to fill up
logs. When asterisk shuts down, it emits another 1100 that are of
even less use. Since the testsuite runs asterisk with a verbose
level of 3, and asterisk starts and stops for every one of the 700+
tests, the number of log messages is staggering.  Besides taking up
resources, it also makes it hard to debug failing tests.

This commit changes the log level for those verbose messages to 5
instead of 3 which reduces the number of log messages to only a
handful. Of course, NOTICE, WARNING and ERROR message are
unaffected.

There's also one other minor change...
ast_context_remove_extension_callerid2() logs a DEBUG message
instead of an ERROR if the extension you're deleting doesn't exist.
The pjsip_config_wizard calls that function to clean up the config
and has been triggering that annoying error message for years.

Resolves: #582

### DIFF
--- a/apps/app_amd.c
+++ b/apps/app_amd.c
@@ -567,7 +567,7 @@ static int load_config(int reload)
 
 	ast_config_destroy(cfg);
 
-	ast_verb(3, "AMD defaults: initialSilence [%d] greeting [%d] afterGreetingSilence [%d] "
+	ast_verb(5, "AMD defaults: initialSilence [%d] greeting [%d] afterGreetingSilence [%d] "
 		"totalAnalysisTime [%d] minimumWordLength [%d] betweenWordsSilence [%d] maximumNumberOfWords [%d] silenceThreshold [%d] maximumWordLength [%d]\n",
 		dfltInitialSilence, dfltGreeting, dfltAfterGreetingSilence, dfltTotalAnalysisTime,
 		dfltMinimumWordLength, dfltBetweenWordsSilence, dfltMaximumNumberOfWords, dfltSilenceThreshold, dfltMaximumWordLength);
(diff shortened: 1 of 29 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

InterLinked1 (commented):
Personally, I agree with most of your changes, though I don't think the messages from loader.c should go all the way down to 5. Loading/unloading a module is kind of a big deal. Maybe 2 or 3 instead? (There is a hierarchy of log messages, so making them the same I think is inappropriate, as they are different now).

gtjoseph:
> Personally, I agree with most of your changes, though I don't think the messages from loader.c should go all the way down to 5. Loading/unloading a module is kind of a big deal. Maybe 2 or 3 instead? (There is a hierarchy of log messages, so making them the same I think is inappropriate, as they are different now).

What info do you get from them though?  If there's an error, you'll still get it.
I can change the ones that were a 1 to 4 and the ones that were 2 to 5, or all of them to a 4  but I don't want any of them at 3 or lower.  If you really want them, set your verbose level appropriately.

InterLinked1:
> > Personally, I agree with most of your changes, though I don't think the messages from loader.c should go all the way down to 5. Loading/unloading a module is kind of a big deal. Maybe 2 or 3 instead? (There is a hierarchy of log messages, so making them the same I think is inappropriate, as they are different now).
> 
> What info do you get from them though? If there's an error, you'll still get it. I can change the ones that were a 1 to 4 and the ones that were 2 to 5, or all of them to a 4 but I don't want any of them at 3 or lower. If you really want them, set your verbose level appropriately.

1 -> 4 and 2-> 5 sounds good then, at least. It's more about preserving the logging hierarchy. Otherwise, there would be no way to see module loads/unloads without all the other individual "component" registration/unregistration, too.

Personally, I always run with a verbose of 10, but I can see why this could be relevant.

InterLinked1 (approved):

jcolp (approved):

Issue #586, opened 2024-02-07
[bug]: The "restrict" keyword used in chan_iax2.c isn't supported in older gcc versions


gtjoseph:
### Severity

Major

### Versions

18.21.0,20.6.0,21.1.0

### Components/Modules

chan_iax2

### Operating Environment

GCC versions <= 4.9

### Frequency of Occurrence

None

### Issue Description

Commit 59bc6ceb61cacd4f751b made use of the `restrict` keyword in the definition of the new `auth_method_names` function but that keyword isn't supported in gcc versions <= 4.9 unless the `-std=gnu99` flag is used (which it isn't by default).  This causes compilation to fail in distros like CentOS7 which have only gcc 4.8 available.

The work around is to add the flag when running ./configure:
```
$ ./configure --with-jansson-bundled CFLAGS=-std=gnu99
```

### Relevant log output

```shell
[CC] chan_iax2.c → chan_iax2.o
chan_iax2.c:418:64: error: expected ‘;’, ‘,’ or ‘)’ before ‘buf’
static char *auth_method_names(int authmethods, char *restrict buf)
^
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
@InterLinked1 Can you create a PR to remove the restrict keyword?

InterLinked1:
> @InterLinked1 Can you create a PR to remove the restrict keyword?

Yeah, I can... but would it be fine to detect the GCC version and keep restrict if it's >= 4.9?
I want to say I've seen restrict used a few other places in the codebase as well...

InterLinked1:
Actually, I'm leaning more towards having configure check for restrict availability, or at least some kind of macro that could be used automatically to either restrict or no op, like we do for some other things. What do you think?

jcolp:
That is generally the best approach, checking for availability instead of versioning and making it optional.

PR #587, opened 2024-02-08
configure: Rerun bootstrap on modern platform.

InterLinked1:
The last time configure was run, it was run on a system that
did not enable -std=gnu11 by default, which meant that the
restrict qualifier would not be recognized on certain platforms.
This regenerates the configure files from running bootstrap.sh,
so that these should be recognized on all supported platforms.

Resolves: #586


seanbright:
Out of curiosity, what does using `restrict` here buy you? There don't appear to be any optimization opportunities here for the compiler so I am wondering what the upside is?

InterLinked1:
> Out of curiosity, what does using `restrict` here buy you? There don't appear to be any optimization opportunities here for the compiler so I am wondering what the upside is?

It's not a huge deal in this particular case, but we should be able to use restrict in general if needed, so I'm just doing this now rather than punting it.

seanbright:
I guess what I really meant to ask was - what is the benefit of using `restrict` in this specific case?

InterLinked1:
> I guess what I really meant to ask was - what is the benefit of using `restrict` in this specific case?

Not that much, I suppose. You're right there's no performance benefit since it's not also reading. It's mostly semantic in that that variable isn't being used anywhere else. Do you think it should be removed?

seanbright:
I don't have an opinion, I was just trying to understand the rationale for it being there in the first place.

configure.ac
```diff
@@ -1279,6 +1279,17 @@ AC_RUN_IFELSE(
...
+AC_COMPILE_IFELSE(
+	[AC_LANG_PROGRAM([], [void modifybuf(char *restrict x) { }])],
+	AC_MSG_RESULT(yes)
+	[HAVE_RESTRICT_QUALIFIER=1],
+	[HAVE_RESTRICT_QUALIFIER=0]
+	AC_MSG_RESULT(no)
+)
+AC_SUBST(HAVE_RESTRICT_QUALIFIER)
+
+AST_GCC_ATTRIBUTE(__restrict__)
```
gtjoseph, line 1292: So, here's the deal.  You don't actually need any of this.
Apparently, the last time bootstrap.sh was run it was run on an older system which generated a configure script that didn't detect that the compiler supports -std=gnu11 but doesn't enable it by default.  If you just run bootstrap.sh again on your system and commit the results, when someone runs configure on an older system the next time, configure should automatically add the `-std=gnu11` flag to the CC variable in makeopts.  This will allow chan_iax2 to compile.

channels/chan_iax2.c
```diff
@@ -415,7 +415,11 @@ static const char *auth_method_labels[] = {
  * \param[out] buf Buffer into which to write the names. Must be of size AUTH_METHOD_NAMES_BUFSIZE.
  * \return Auth methods name
  */
+#ifdef HAVE_RESTRICT_QUALIFIER
 static char *auth_method_names(int authmethods, char *restrict buf)
+#else
+static char *auth_method_names(int authmethods, char *buf)
+#endif
```
gtjoseph, line 422: Not needed.  See below.

gtjoseph (changes requested):

configure.ac
InterLinked1: Hmm... so you're saying that flag should always be available on all supported platforms and thus restrict support can implicitly be a hard requirement? Wasn't sure if there might be other compatibility issues.

gtjoseph: As far as I can tell, as long as the configure script was generated on a recent system,it'll enable -std=gnu11 on gcc's as far back as we care to support.

InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

seanbright:
The title and description of this PR is now in conflict with what is actually being proposed.

InterLinked1:
> The title and description of this PR is now in conflict with what is actually being proposed.

The commit message is correct.

The PR doesn't update automatically, but I'll sync it now.

seanbright:
I know that the commit message is correct. I mentioned the title and description of the PR. Thank you for making the change.

gtjoseph (approved):

gtjoseph:
Tested on CentOS 7 with gcc (GCC) 4.8.5 20150623 and ./configure now correctly sets `CC=gcc -std=gnu11` in makeopts.

gtjoseph (changes requested):
The cherry-picks all failed so I think you're gonna have to submit a PR for each branch :(

gtjoseph (approved):

Issue #588, opened 2024-02-08
[new-feature]: app_dial: Allow Dial to be aborted if early media is not received


InterLinked1:
### Feature Description

This extends the timeout option for dial from answering to progress as well.

PR #589, opened 2024-02-08
app_dial: Add dial time for progress/ringing.

InterLinked1:
Add a timeout option to control the amount of time to wait if no early media is received before giving up. This allows aborting early if the destination
is not being responsive.

Resolves: #588

UserNote: The timeout argument to Dial now allows
specifying the maximum amount of time to dial if
early media is not received.

### DIFF
--- a/apps/app_dial.c
+++ b/apps/app_dial.c
@@ -88,9 +88,12 @@
 				</argument>
 				<xi:include xpointer="xpointer(/docs/info[@name='Dial_Resource'])" />
 			</parameter>
-			<parameter name="timeout" required="false">
+			<parameter name="timeout" required="false" argsep="^">
 				<para>Specifies the number of seconds we attempt to dial the specified devices.</para>
 				<para>If not specified, this defaults to 136 years.</para>
+				<para>If a second argument is specified, this controls the number of seconds we attempt to dial the specified devices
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

chrsmj:
The use of pipe as argument separator is discouraged, per conversation with @jcolp at AstriCon.

apps/app_dial.c
```diff
@@ -88,9 +88,12 @@
 				</argument>
 				<xi:include xpointer="xpointer(/docs/info[@name='Dial_Resource'])" />
 			</parameter>
-			<parameter name="timeout" required="false">
+			<parameter name="timeout" required="false" argsep="|">
```
jcolp, line 91: As was mentioned in a comment, | is discouraged - why the use of it here? Is there an alternative?

jcolp (changes requested):
This should have test coverage added as well.

apps/app_dial.c
InterLinked1: That was somewhat arbitrary, I've changed it to `^`, which I think is more common as the secondary arg delimiter.

InterLinked1:
Tests added here: https://github.com/asterisk/testsuite/pull/44

jcolp (approved):

Issue #593, opened 2024-02-09
[new-feature]: Add horse power in to speech engines


chrsmj:
### Feature Description

Current Speech...() apps only let you work with one engine at a time. This presents challenges in multi-lingual IVR environments, situations where you are testing different engines, cloud evaporation events, etc.

The forthcoming patch will allow you to configure the horses inside your engine and sequentially feed them audio frames. It is still a little leaky, eschews linked lists for arrays, and requires additional patches to the engines themselves to do anything useful (patches also forthcoming there eg. Vosk.)

But the dialplan will look like this -- note the carets used to separate the horses (because :horse: :heart: :carrot: :)

```
exten => 46773,1,NoOp(H-O-R-S-E) 
 same = n,Set(TIMEOUT(D)=1)
 same = n,Set(SPEECH_DTMF_MAXLEN=2)
 same = n,SpeechCreate(vosk^mage)
 same = n,SpeechCreate(vosk^secretariat)
 same = n,SpeechBackground(dial-here-often,5,,mage^secretariat)
 same = n,Set(speechtext0mage=${SPEECH_TEXT(0^mage)})
 same = n,Set(speechtext0secretariat=${SPEECH_TEXT(0^secretariat)})
 same = n,SpeechDestroy(vosk^mage)
 same = n,SpeechDestroy(vosk^secretariat)
 same = n,SpeechCreate(vosk^mage)
 same = n,SpeechCreate(vosk^secretariat)
 same = n,SpeechBackground(that-tickles,5,p,mage^secretariat)
 same = n,Set(speechtext0mage=${SPEECH_TEXT(0^mage)})
 same = n,Set(speechtext0secretariat=${SPEECH_TEXT(0^secretariat)})
 same = n,SpeechDestroy(vosk^mage)
 same = n,SpeechDestroy(vosk^secretariat)
```

And the corresponding res_speech_vosk.conf:
```
[general]

[mage]
type=horse
url = ws://localhost:2700

[secretariat]
type=horse
url = ws://localhost:2701
```

*You might replace **mage** and **secretariat** with **en** and **es**...*

jcolp:
This sounds like a major rearchitecture. Such things should really be discussed ahead of time, for example this would need to be backwards compatible to go into any current branches or even master.

jcolp:
With the place to discuss such things being https://groups.io/g/asterisk-dev

chrsmj:
[...]

PR #594, opened 2024-02-09
app_speech_utils.c: Add horses to speech engines.

chrsmj:
Speech Engines can now have multiple horses inside that race to get you results from different backends.

UserNote: The various Speech...() dialplan applications and the SPEECH_TEXT function now support the use of carets to control the horses in your engines.
SpeechCreate(engine^horse1)
SpeechCreate(engine^horse2)
SpeechBackground(audio,,,horse1^horse2)
SPEECH_TEXT(0^horse1)
SPEECH_TEXT(0^horse2)
SpeechDestroy(engine^horse1)
SpeechDestroy(engine^horse2)

Resolves: #593

### DIFF
--- a/apps/app_speech_utils.c
+++ b/apps/app_speech_utils.c
@@ -40,6 +40,8 @@
 #include "asterisk/app.h"
 #include "asterisk/speech.h"
 
+#define MAX_HORSES 10
+
 /*** DOCUMENTATION
 	<application name="SpeechCreate" language="en_US">
 		<synopsis>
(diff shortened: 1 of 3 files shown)


jcolp:
I don't agree with the use of "horses" in this. From a user facing perspective it doesn't really make sense and is not intuitive.

chrsmj:
cherry-pick-to: none

chrsmj:
The use of the word "engine" in software comes directly from cars. **And car engines have horse power.** But "power" alone is too generic. (Also considered was powertrain but that is too long.)

chrsmj:
Patch needs more work, will convert to draft for now.

Issue #595, opened 2024-02-09
[improvement]: dsp.c: Fix and improve confusing warning message.


InterLinked1:
### Improvement Description

Improve codec-related warnings that can currently be very misleading as to the problem.

PR #596, opened 2024-02-09
dsp.c: Fix and improve potentially inaccurate log message.

InterLinked1:
If ast_dsp_process is called with a codec besides slin, ulaw, or alaw, a warning is logged that in-band DTMF is not supported, but this message is not always appropriate or correct, because ast_dsp_process is much more generic than just DTMF detection.

This logs a more generic message in those cases, and also improves codec-mismatch logging throughout dsp.c by ensuring incompatible codecs are printed out.

Resolves: #595

### DIFF
--- a/main/dsp.c
+++ b/main/dsp.c
@@ -413,7 +413,7 @@ struct ast_dsp {
 	int totalnoise;
 	int features;
 	int ringtimeout;
-	int busymaybe;
+	int busymaybe; /* Boolean, could be a bitfield */
 	int busycount;
 	struct ast_dsp_busy_pattern busy_cadence;
 	int historicnoise[DSP_HISTORY];
@@ -428,8 +428,8 @@ struct ast_dsp {
 	int digitmode;
 	int faxmode;
 	int freqmode;
-	int dtmf_began;
-	int display_inband_dtmf_warning;
+	int dtmf_began; /* Boolean, could be a bitfield */
+	int display_inband_dtmf_warning; /* Boolean, could be a bitfield */
 	float genergy;
 	int mute_fragments;
 	unsigned int sample_rate;
@@ -1223,7 +1223,8 @@ int ast_dsp_call_progress(struct ast_dsp *dsp, struct ast_frame *inf)
 		return 0;
 	}
 	if (!ast_format_cache_is_slinear(inf->subclass.format)) {
-		ast_log(LOG_WARNING, "Can only check call progress in signed-linear frames\n");
+		ast_log(LOG_WARNING, "Can only check call progress in signed-linear frames, %s not supported\n",
+			ast_format_get_name(inf->subclass.format));
 		return 0;
 	}
 	return __ast_dsp_call_progress(dsp, inf->data.ptr, inf->datalen / 2);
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

main/dsp.c
```diff
@@ -429,7 +429,7 @@ struct ast_dsp {
 	int faxmode;
 	int freqmode;
 	int dtmf_began;
-	int display_inband_dtmf_warning;
+	unsigned int display_inband_dtmf_warning:1;
```
seanbright, line 432: Unrelated change that snuck in? If not can you explain the rationale for this change?

main/dsp.c
```diff
@@ -1223,7 +1223,7 @@ int ast_dsp_call_progress(struct ast_dsp *dsp, struct ast_frame *inf)
 		return 0;
 	}
 	if (!ast_format_cache_is_slinear(inf->subclass.format)) {
-		ast_log(LOG_WARNING, "Can only check call progress in signed-linear frames\n");
+		ast_log(LOG_WARNING, "Can only check call progress in signed-linear frames, %s not supported\n", ast_format_get_name(inf->subclass.format));
```
seanbright, line 1226: These lines are pretty long now. I might look nicer to add a line break after the format string. This applies to all of the logging statement changes you made in this file.

seanbright (changes requested):

main/dsp.c
InterLinked1: It is related; I reviewed all the code where this is used and set, and it only needs to be a bitfield, not an int.
I considered renaming the variable too for clarity but guess I didn't do that.

seanbright: But you aren't changing any code that reads or sets this structure member, so it's not clear to me what the benefit is. Does it change the generated code? Due to alignment concerns I don't think it would change the size of the struct either?

InterLinked1: It's more semantic. You can look at the struct and know that it's a boolean variable, which was not conveyed before. I think it helps with code clarity to do this, regardless of space savings, since it serves as documentation.

If more bitfields are added in the future, then it would actually reduce memory usage.

seanbright: > If more bitfields are added in the future, then it would actually reduce memory usage.

I don't think that is correct, but I will defer to your expertise on this.

jcolp: I don't like change for the sake of change. If that time ever comes, then it can be changed at that point. Right now it is premature to touch it or care about it. Just changing the log message is fine.

jcolp (changes requested):

main/dsp.c
InterLinked1: It wasn't really for a future hypothetical performance benefit, it was more for making the code more self-documenting, since that variable is already a misnomer. I've changed it back but added a documentation note about this.

jcolp (approved):

main/dsp.c
seanbright: Looks like `busymaybe` could be a bitfield too. I would add the same comment there so people aren't confused.

jcolp (approved):

PR #603, opened 2024-02-12
configure: Rerun bootstrap on modern platform.

InterLinked1:
The last time configure was run, it was run on a system that did not enable -std=gnu11 by default, which meant that the restrict qualifier would not be recognized on certain platforms. This regenerates the configure files from running bootstrap.sh, so that these should be recognized on all supported platforms.

Resolves: #586


gtjoseph:
cherry-pick-to: none

gtjoseph (approved):

Issue #592, opened 2024-02-09
[bug]: In certain circumstances, "pjsip show channelstats" can segfault when a fax session is active


gtjoseph:
### Severity

Major

### Versions

18,20,21,master

### Components/Modules

chan_pjsip/cli_commands

### Operating Environment

any

### Frequency of Occurrence

None

### Issue Description

Under certain circumstances, when a fax stream is added to a pjsip session, the audio stream isn't cleaned up fully.  If the "pjsip show channelstats" cli command is run in this state, it can think there's still an audio stream but because the stream has garbage in it, it can segfault when trying to retrieve the rtp stats.

### Relevant log output

```shell
#0  0x000055bfd453fa54 in ast_rtp_instance_get_stats (instance=0x39, stats=stats@entry=0x7f19da88df20, stat=stat@entry=AST_RTP_INSTANCE_STAT_ALL) at rtp_engine.c:2463
#1  0x00007f1b621159e2 in cli_channelstats_print_body (obj=0x7f1a10076588, arg=arg@entry=0x7f19da88e2b0, flags=flags@entry=2) at pjsip/cli_commands.c:383
#2  0x000055bfd445411d in internal_ao2_traverse (self=self@entry=0x7f1abc004818, flags=flags@entry=OBJ_NODATA, cb_fn=<optimized out>, arg=arg@entry=0x7f19da88e2b0, data=data@entry=0x0, type=type@entry=AO2_CALLBACK_DEFAULT, tag=0x0, file=0x7f1b622efb9f "res_pjsip/pjsip_cli.c", line=215, func=0x7f1b622effb0 <__PRETTY_FUNCTION__.29132> "ast_sip_cli_traverse_objects") at astobj2_container.c:328
#3  0x000055bfd445468c in __ao2_callback (c=c@entry=0x7f1abc004818, flags=flags@entry=OBJ_NODATA, cb_fn=<optimized out>, arg=arg@entry=0x7f19da88e2b0, tag=tag@entry=0x0, file=file@entry=0x7f1b622efb9f "res_pjsip/pjsip_cli.c", line=215, func=0x7f1b622effb0 <__PRETTY_FUNCTION__.29132> "ast_sip_cli_traverse_objects") at astobj2_container.c:414
#4  0x00007f1b622df098 in ast_sip_cli_traverse_objects (e=<optimized out>, cmd=<optimized out>, a=0x7f19da88e3e0) at res_pjsip/pjsip_cli.c:215
[...]

PR #606, opened 2024-02-12
pjsip show channelstats: Prevent possible segfault when faxing

gtjoseph:
Under rare circumstances, it's possible for the original audio
session in the active_media_state default_session to be corrupted
instead of removed when switching to the t38/image media session
during fax negotiation.  This can cause a segfault when a "pjsip
show channelstats" attempts to print that audio media session's
rtp statistics.  In these cases, the active_media_state
topology is correctly showing only a single t38/image stream
so we now check that there's an audio stream in the topology
before attempting to use the audio media session to get the rtp
statistics.

Resolves: #592


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

PR #614, opened 2024-02-17
Fix typos

omahs:
Fix typos


InterLinked1:
CLA is not signed, and commits need to be squashed into one (with a commit message that follows the commit message guidelines: https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/).

PR #615, opened 2024-02-17
strings.h: Ensure ast_str_buffer(…) returns a 0 terminated string.

seanbright:
If a dynamic string is created with an initial length of 0, `ast_str_buffer(…)` will return an invalid pointer.

This was a secondary discovery when fixing #65.

### DIFF
--- a/include/asterisk/strings.h
+++ b/include/asterisk/strings.h
@@ -753,7 +753,10 @@ char * attribute_pure ast_str_buffer(const struct ast_str *buf),
 	 * being returned; eventually, it should become truly const
 	 * and only be modified via accessor functions
 	 */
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

seanbright:
Are the cherry pick test failures ones you expect or are they related to this PR?

gtjoseph:
> Are the cherry pick test failures ones you expect or are they related to this PR?

Those look like new failures but if they were related to this PR I would have expected the same tests to fail in all branches and they didn't.

I'm going to retest just to make sure.

PR #628, opened 2024-02-28
alembic: Synchronize alembic heads between supported branches.

seanbright:
This adds a dummy migration to 18 and 20 so that our alembic heads are synchronized across all supported branches.

In this case the migration we are stubbing (24c12d8e9014) is:

https://github.com/asterisk/asterisk/commit/775352ee6c2a5bcd4f0e3df51aee5d1b0abf4cbe


seanbright:
cherry-pick-to: 20

jcolp (approved):
I like this.

Issue #629, opened 2024-02-29
[bug]: app_voicemail: Multiple executions of unit tests cause segfault


InterLinked1:
### Severity

Minor

### Versions

21.0.0

### Components/Modules

app_voicemail

### Operating Environment

Debian 12

### Frequency of Occurrence

Constant

### Issue Description

If you run `test execute category /apps/app_voicemail/` and then run `test execute category /apps/app_voicemail/` again, Asterisk will segfault.

The cause is consistent, an attempt to call `ast_copy_string` on NULL in the `test_voicemail_vm_info` unit test:

```
#1  0x00007f87680449d8 in acf_vm_info (chan=0x7f8734166600, cmd=0x7f8762a0928e "VM_INFO", args=0x7f8762a09080 "00000000@test,email", buf=0x7f8762a09180 "", len=256) at app_voicemail.c:13164
#2  0x00007f876804eb3b in test_voicemail_vm_info (info=0x56485a20b7f0, cmd=TEST_EXECUTE, test=0x56485a20b7f0) at app_voicemail.c:15683
```

Technically, this could happen since we do not check if allocation fails:

`vmu->email = ast_strdup("vm-info-test@example.net");`

However, that is not what's happening here, and this happens regardless of whether the executions are concurrent or sequential. It seems that the tests (as a whole, not this specific one per se) do not clean up properly after themselves, as running the test wipes my current voicemail users and leaves the system in this state:

```
debian*CLI> voicemail show users
Context    Mbox  User                      Zone       NewMsg
test       00000002 Mrs. Test                 european        0
test       00000000                                           0
2 voicemail users configured.
```

Therefore, the next time the tests are running, I'm guessing that `vmu->email` is NULL for something that was created and left behind by the test, causing `VM_INFO` to try to copy `vmu->email` when it's NULL. I narrowed this down to `test execute category /apps/app_voicemail/ name test_voicemail_msgcount` specifically as an offending test,.

Running `module refresh app_voicemail` between test executions prevents the SEGV, since the bad state is cleaned up when the module is discarded and loaded again (until the tests are run again).

The underlying root cause is the tests are not cleaning up properly and restoring the original state.

### Relevant log output

```shell
Thread 1 (Thread 0x7f8762a0a6c0 (LWP 2071219)):
#0  0x00007f87680204d9 in ast_copy_string (dst=0x7f8762a09180 "", src=0x0, size=256) at /usr/src/asterisk-21.0.0/include/asterisk/strings.h:412
[...]

PR #630, opened 2024-02-29
app_voicemail: Properly reinitialize config after unit tests.

InterLinked1:
Most app_voicemail unit tests were not properly cleaning up after themselves after running. This led to test mailboxes lingering around in the system. It also meant that if any unit tests in app_voicemail that create mailboxes were executed and the module was not unloaded/loaded again prior to running the test_voicemail_vm_info unit test, Asterisk would segfault due to an attempt to copy a NULL string.

The load_config test did actually have logic to reinitialize the config after the test. However, this did not work in practice since load_config() would not reload the config since voicemail.conf had not changed during the test; thus, additional logic has been added to ensure that voicemail.conf is truly reloaded, after any unit tests which modify the users list.

This prevents the SEGV due to invalid mailboxes lingering around, and also ensures that the system state is restored to what it was prior to the tests running.

Resolves: #629

### DIFF
--- a/apps/app_voicemail.c
+++ b/apps/app_voicemail.c
@@ -767,6 +767,18 @@ static const char * const mailbox_folders[] = {
 	"Urgent",
 };
 
+/*!
+ * \brief Reload voicemail.conf
+ * \param reload Whether this is a reload as opposed to module load
+ * \param force Forcefully reload the config, even it has not changed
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

PR #631, opened 2024-03-01
res_prometheus: Fix duplicate output of metric and help text

zecke:
The prometheus exposition format requires each line to be unique[1]. This is handled by struct prometheus_metric having a list of children that is managed when registering a metric. In case the scrape callback is used, it is the responsibility of the implementation to handle this correctly.

Originally the bridge callback didn't handle NULL snapshots, the crash fix lead to NULL metrics, and fixing that lead to duplicates.

The original code assumed that snapshots are not NULL and then relied on "if (i > 0)" to establish the parent/children relationship between metrics of the same class. This is not workerable as the first bridge might be invisible/lacks a snapshot.

Fix this by keeping a separate array of the first metric by class. Instead of relying on the index of the bridge, check whether the array has an entry. Use that array for the output.

Add a test case that verifies that the help text is not duplicated.

Resolves: #642

[1] https://prometheus.io/docs/instrumenting/exposition_formats/#grouping-and-sorting

### DIFF
--- a/res/prometheus/bridges.c
+++ b/res/prometheus/bridges.c
@@ -81,7 +81,8 @@ static void bridges_scrape_cb(struct ast_str **response)
 	struct ao2_container *bridges;
 	struct ao2_iterator it_bridges;
 	struct ast_bridge *bridge;
-	struct prometheus_metric *bridge_metrics;
+	struct prometheus_metric *metrics;
+	struct prometheus_metric **bridge_metrics;
 	char eid_str[32];
 	int i, j, num_bridges, num_outputs = 0;
 	struct prometheus_metric bridge_count = PROMETHEUS_METRIC_STATIC_INITIALIZATION(
@@ -116,8 +117,15 @@ static void bridges_scrape_cb(struct ast_str **response)
 		return;
 	}
 
-	bridge_metrics = ast_calloc(ARRAY_LEN(bridge_metric_defs) * num_bridges, sizeof(*bridge_metrics));
+	metrics = ast_calloc(ARRAY_LEN(bridge_metric_defs) * num_bridges, sizeof(*metrics));
+	if (!metrics) {
+		ao2_ref(bridges, -1);
+		return;
+	}
+
+	bridge_metrics = ast_calloc(ARRAY_LEN(bridge_metric_defs), sizeof(bridge_metrics));
 	if (!bridge_metrics) {
+		ast_free(metrics);
 		ao2_ref(bridges, -1);
 		return;
 	}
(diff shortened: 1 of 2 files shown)


res/prometheus/bridges.c
```diff
@@ -116,12 +117,18 @@ static void bridges_scrape_cb(struct ast_str **response)
...
+	if (!metrics) {
 		ao2_ref(bridges, -1);
 		return;
 	}
 
+	bridge_metrics = ast_calloc(ARRAY_LEN(bridge_metric_defs), sizeof(bridge_metrics));
+	if (!bridge_metrics) {
+		ao2_ref(bridges, -1);
+		ast_free(metrics);
+	}
```
seanbright, line 130: `return;` here?

tests/test_res_prometheus.c
```diff
@@ -710,10 +710,23 @@ static void safe_bridge_destroy(struct ast_bridge *bridge)
 	ast_bridge_destroy(bridge, 0);
 }
 
+static int word_count(const char *str, const char *needle)
```
seanbright, line 713: Minor: `word_count` sounds like it counts words. `match_count` instead?

res/prometheus/bridges.c
```diff
@@ -116,12 +117,18 @@ static void bridges_scrape_cb(struct ast_str **response)
...
+	metrics = ast_calloc(ARRAY_LEN(bridge_metric_defs) * num_bridges, sizeof(*metrics));
+	if (!metrics) {
 		ao2_ref(bridges, -1);
 		return;
 	}
 
+	bridge_metrics = ast_calloc(ARRAY_LEN(bridge_metric_defs), sizeof(bridge_metrics));
+	if (!bridge_metrics) {
+		ao2_ref(bridges, -1);
+		ast_free(metrics);
```
seanbright, line 129: Minor: Maybe swap these two lines to free in the reverse order that we allocated?

seanbright (changes requested):

res/prometheus/bridges.c
zecke: Done. I also applied this order at the end of the function.

gtjoseph (changes requested):
@zecke Can you open a "bug" issue for this and add the appropriate "Resolves:" trailer to the commit message and the PR description?

jcolp (approved):

zecke:
> @zecke Can you open a "bug" issue for this and add the appropriate "Resolves:" trailer to the commit message and the PR description?

Done. Mind having a look?

gtjoseph (approved):

gtjoseph:
I'm closing and re-opening just to trigger a retest.

mbradeen (approved):

Issue #634, opened 2024-03-04
[bug]: make install doesn't create the stir_shaken cache directory


gtjoseph:
### Severity

Minor

### Versions

18,20,21,master

### Components/Modules

build_system

### Operating Environment

all

### Frequency of Occurrence

Constant

### Issue Description

The default stir_shaken cache directory is "/var/lib/asterisk/keys/stir_shaken/cache" but on a `make install`, we're creating only "/var/lib/asterisk/keys/stir_shaken".  We need to create the "cache" sub-directory.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #635, opened 2024-03-04
Makefile: Add stir_shaken/cache to directories created on install

gtjoseph:
The default location for the stir_shaken cache is
/var/lib/asterisk/keys/stir_shaken/cache but we were only creating
/var/lib/asterisk/keys/stir_shaken on istall.  We now create
the cache sub-directory.

Resolves: #634


jcolp (approved):

gtjoseph:
FYI... I'm using this PR to test a few CI things.

Issue #636, opened 2024-03-05
[bug]: Possible SEGV in res_stir_shaken due to wrong free function


gtjoseph:
### Severity

Blocker

### Versions

18,20,21,master

### Components/Modules

res_stir_shaken

### Operating Environment

all

### Frequency of Occurrence

None

### Issue Description

In attestation_config:as_check_common_config is calling ast_std_free instead of ast_free on raw_key which can cause a SEGV if MALLOC_DEBUG is enabled.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #637, opened 2024-03-05
attestation_config.c: Use ast_free instead of ast_std_free

gtjoseph:
In as_check_common_config, we were calling ast_std_free on
raw_key but raw_key was allocated with ast_malloc so it
should be freed with ast_free.

Resolves: #636


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

PR #659, opened 2024-03-20
alembic: Fix compatibility with SQLAlchemy 2.0+.

seanbright:
SQLAlchemy 2.0 changed the way that commits/rollbacks are handled causing the final `UPDATE` to our `alembic_version_<whatever>` tables to be rolled back instead of committed.

We now use one connection to determine which `alembic_version_<whatever>` table to use and another to run the actual migrations. This prevents the erroneous rollback.

This change is compatible with both SQLAlchemy 1.4 and 2.0.

### DIFF
--- a/contrib/ast-db-manage/env.py
+++ b/contrib/ast-db-manage/env.py
@@ -67,6 +67,8 @@ def run_migrations_online():
     and associate a connection with the context.
 
     """
+    script_location = config.get_main_option('script_location')
+
     engine = engine_from_config(
                 config.get_section(config.config_ini_section),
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

mbradeen (approved):
Tested locally

jcolp (changes requested):
Minor error in commit message,

"We now one connection" should be "We now use one connection"

jcolp (approved):

Issue #480, opened 2023-12-04
[improvement]: pbx_variables.c: Prevent infinite recursion and stack overflow with variable expansion


InterLinked1:
### Improvement Description

It is possible to craft dialplan (though not likely intentionally) that will result in an infinite recursion of variable substitution. Currently, Asterisk will just segfault due to stack overflow. This patch detects this and instead safely aborts and logs an error for the user to fix the offending dialplan.

jcolp:
How does this recursion occur...

InterLinked1:
> How does this recursion occur...

Here's an example:

```
[lookup-context]
exten => _X,1,Return(1)
exten => _1X,1,Return(2)
exten => _N0,1,Return(3)
exten => _XX,1,Return(${EVAL_EXTEN(${CONTEXT},${EXTEN:-2:1}0,1)})

[overflow]
exten => s,1,NoOp(${EVAL_EXTEN(lookup-context,${ARG1:-2},1)})
	same => n,Assert(0) ; crashes before it gets here
```

Correct dialplan shouldn't recurse, but `exten => _XX,1,Return(${EVAL_EXTEN(${CONTEXT},${EXTEN:-2:1}0,1)})` is bad dialplan since it recurses on itself. This was a typo and should have actually been `_X5`, not `_XX`.

jcolp:
Ah, through the use of the EVAL_EXTEN dialplan function you added. Would it have been possible prior to the inclusion of that?

InterLinked1:
> Ah, through the use of the EVAL_EXTEN dialplan function you added. Would it have been possible prior to the inclusion of that?

I haven't thought of any cases where it would be, and I suspect maybe not, but I'm not positive.

Recursion with `EVAL_EXTEN` is not inherently problematic, and as in the (corrected) example above, may even be desired, but it can be easy to shoot yourself in the foot this way - great power but no guardrails.

PR #663, opened 2024-03-21
pbx_variables.c: Prevent SEGV due to stack overflow.

InterLinked1:
It is possible for dialplan to result in an infinite recursion of variable substitution, which eventually leads to stack overflow. If we detect this, abort
substitution and log an error for the user to fix
the broken dialplan.

Resolves: #480

UpgradeNote: The maximum amount of dialplan recursion using variable substitution (such as by using EVAL_EXTEN) is capped at 15.


InterLinked1:
cherry-pick-to: none

gtjoseph (approved):

mbradeen (approved):

Issue #676, opened 2024-03-30
[bug]: res_stir_shaken implicit declaration of function errors/warnings


lukeescude:
### Severity

Minor

### Versions

20.7.0

### Components/Modules

res_stir_shaken

### Operating Environment

CentOS 7.9 x64
openssl.x86_64 1:1.0.2k-26.el7_9

### Frequency of Occurrence

Constant

### Issue Description

We are unable to get Asterisk 20.7 compiled with stir_shaken support, and this may be why. During the build process on CentOS, the following errors/warnings occur:

res_stir_shaken/crypto_utils.c: In function 'crypto_get_raw_pubkey_from_cert':
res_stir_shaken/crypto_utils.c:285:2: warning: implicit declaration of function 'X509_get0_pubkey' [-Wimplicit-function-declaration]
  public_key = X509_get0_pubkey(cert);
  ^
res_stir_shaken/crypto_utils.c:285:13: warning: assignment makes pointer from integer without a cast [enabled by default]
  public_key = X509_get0_pubkey(cert);
             ^
res_stir_shaken/crypto_utils.c: In function 'crypto_lock_cert_store':
res_stir_shaken/crypto_utils.c:322:2: warning: implicit declaration of function 'X509_STORE_lock' [-Wimplicit-function-declaration]
  return X509_STORE_lock(store) == 1 ? 0 : -1;
  ^
res_stir_shaken/crypto_utils.c: In function 'crypto_unlock_cert_store':
res_stir_shaken/crypto_utils.c:331:2: warning: implicit declaration of function 'X509_STORE_unlock' [-Wimplicit-function-declaration]
  return X509_STORE_unlock(store) == 1 ? 0 : -1;
  ^
res_stir_shaken/crypto_utils.c: In function 'crypto_show_cli_store':
res_stir_shaken/crypto_utils.c:380:2: warning: implicit declaration of function 'X509_STORE_get0_objects' [-Wimplicit-function-declaration]
  certs = X509_STORE_get0_objects(store);
  ^
res_stir_shaken/crypto_utils.c:380:8: warning: assignment makes pointer from integer without a cast [enabled by default]
  certs = X509_STORE_get0_objects(store);
        ^
res_stir_shaken/crypto_utils.c:384:3: warning: implicit declaration of function 'X509_OBJECT_get0_X509' [-Wimplicit-function-declaration]
   X509 *c = X509_OBJECT_get0_X509(o);
   ^
[...]

PR #680, opened 2024-04-01
res_stir_shaken:  Fix compilation for CentOS7 (openssl 1.0.2)

gtjoseph:
* OpenSSL 1.0.2 doesn't support X509_get0_pubkey so we now use
  X509_get_pubkey.  The difference is that X509_get_pubkey requires
  the caller to free the EVP_PKEY themselves so we now let
  RAII_VAR do that.
* OpenSSL 1.0.2 doesn't support upreffing an X509_STORE so we now
  wrap it in an ao2 object.
* OpenSSL 1.0.2 doesn't support X509_STORE_get0_objects to get all
  the certs from an X509_STORE and there's no easy way to polyfill
  it so the CLI commands that list profiles will show a "not
  supported" message instead of listing the certs in a store.

Resolves: #676

### DIFF
--- a/res/res_stir_shaken/crypto_utils.c
+++ b/res/res_stir_shaken/crypto_utils.c
@@ -279,10 +279,8 @@ int crypto_extract_raw_pubkey(EVP_PKEY *key, unsigned char **buffer)
 int crypto_get_raw_pubkey_from_cert(X509 *cert,
 	unsigned char **buffer)
 {
-	RAII_VAR(BIO *, bio, NULL, BIO_free_all);
-	EVP_PKEY *public_key;
+	RAII_VAR(EVP_PKEY *, public_key, X509_get_pubkey(cert), EVP_PKEY_free);
 
-	public_key = X509_get0_pubkey(cert);
 	if (!public_key) {
 		crypto_log_openssl(LOG_ERROR, "Unable to retrieve pubkey from cert\n");
 		return -1;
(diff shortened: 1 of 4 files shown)


res/res_stir_shaken/crypto_utils.c
```diff
@@ -279,10 +279,12 @@ int crypto_extract_raw_pubkey(EVP_PKEY *key, unsigned char **buffer)
 int crypto_get_raw_pubkey_from_cert(X509 *cert,
 	unsigned char **buffer)
 {
-	RAII_VAR(BIO *, bio, NULL, BIO_free_all);
```
mbradeen, line 282: was this just copypasta before?

mbradeen (approved):

res/res_stir_shaken/crypto_utils.c
gtjoseph: Yeah, it wasn't needed or used.

gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

gtjoseph:
FYI...  I still need to run my leak check tests.

gtjoseph:
Local functional and leak testing passed.  PR should be good to go.

lukeescude:
Tested and works well for me

bkford (approved):

mbradeen (approved):

Issue #681, opened 2024-04-01
[new-feature]: callerid.c: Parse all received parameters


InterLinked1:
### Feature Description

Some received Caller ID parameters are currently ignored. Following previous changes that allowed all parameters to be set, this allows all received parameters to be provided up to the dialplan.

PR #682, opened 2024-04-01
callerid.c: Parse previously ignored Caller ID parameters.

InterLinked1:
Commit f2f397c1a8cc48913434ebb297f0ff50d96993db previously made it possible to send Caller ID parameters to FXS stations which, prior to that, could not be sent.

This change is complementary in that we now handle receiving all these parameters on FXO lines and provide these up to the dialplan, via chan_dahdi. In particular:

* If a redirecting reason is provided, the channel's redirecting reason is set. No redirecting number is set, since there is no parameter for this in the Caller ID protocol, but the reason can be checked to determine if and why a call was forwarded.
* If the Call Qualifier parameter is received, the Call Qualifier variable is set.
* Some comments have been added to explain why some of the code is the way it is, to assist other people looking at it.

With this change, Asterisk's Caller ID implementation is now reasonably complete for both FXS and FXO operation.

Resolves: #681

### DIFF
--- a/funcs/func_callerid.c
+++ b/funcs/func_callerid.c
@@ -192,6 +192,8 @@
 					You are responsible for setting it if/when needed.</para>
 					<para>Supporting Caller ID units will display the LDC
 					(Long Distance Call) indicator when they receive this parameter.</para>
+					<para>For incoming calls on FXO ports, if the Call Qualifier parameter is received,
+					this variable will also be set to 1.</para>
 					<para>This option must be used with a channel driver
 					that allows Asterisk to generate the Caller ID spill,
 					which currently only includes <literal>chan_dahdi</literal>.</para>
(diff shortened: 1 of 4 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

include/asterisk/callerid.h
```diff
@@ -199,6 +199,9 @@ int callerid_feed_jp(struct callerid_state *cid, unsigned char *ubuf, int sample
  */
 void callerid_get(struct callerid_state *cid, char **number, char **name, int *flags);
 
+/*! \brief Same as callerid_get, but also retrieve the redirecting reason, if available */
```
seanbright, line 202: Can you reproduce the full doxygen block from `callerid_get(...)` here and be sure to document `redirecting` obviously.

funcs/func_callerid.c
```diff
@@ -192,6 +192,8 @@
 					You are responsible for setting it if/when needed.</para>
 					<para>Supporting Caller ID units will display the LDC
 					(Long Distance Call) indicator when they receive this parameter.</para>
+					<para>On incoming calls on FXO port, if the Call Qualifier is detected,
```
seanbright, line 195: Minor: This would read better as:

> For incoming calls on an FXO port ...

or:

> For incoming calls on FXO ports ...

mbradeen (approved):

jcolp (approved):

Issue #683, opened 2024-04-01
[improvement]: func_callerid: Warn if invalid redirecting reason is set


InterLinked1:
### Improvement Description

Currently, setting REDIRECTING(reason) to an invalid reason will fail silently. If REDIRECTING(orig-reason) is set to an invalid value, a warning is emitted; however, if REDIRECTING(reason) is set to an invalid value, no warning is emitted (likely a copy/paste omission at some point).

This adds a warning if the redirecting reason provided is invalid.

PR #684, opened 2024-04-01
func_callerid: Emit warning if invalid redirecting reason set.

InterLinked1:
Emit a warning if REDIRECTING(reason) is set to an invalid reason, consistent with what happens when
REDIRECTING(orig-reason) is set to an invalid reason.

Resolves: #683


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #689, opened 2024-04-02
[bug] Document the `Events` argument of the `Login` AMI action


7-zete-7:

gtjoseph:
The XML documentation in manager.c for the "Login" action needs to be updated with the description for the "Events" parameter.

gtjoseph:
It looks like the AuthType and Key parameters are also missing.

7-zete-7:
Thank you @gtjoseph!

PR #690, opened 2024-04-03
manager.c: Add missing parameters to Login documentation

gtjoseph:
* Added the AuthType and Key parameters for MD5 authentication.

* Added the Events parameter.

Resolves: #689

### DIFF
--- a/main/manager.c
+++ b/main/manager.c
@@ -165,14 +165,41 @@
 			<parameter name="Username" required="true">
 				<para>Username to login with as specified in manager.conf.</para>
 			</parameter>
+			<parameter name="AuthType">
+				<para>Authorization type. Valid values are:</para>
+				<enumlist>
+					<enum name="plain"><para>Plain text secret. (default)</para></enum>
+					<enum name="MD5"><para>MD5 hashed secret.</para></enum>
+				</enumlist>
+			</parameter>
 			<parameter name="Secret">
-				<para>Secret to login with as specified in manager.conf.</para>
+				<para>Plain text secret to login with as specified in manager.conf.</para>
+			</parameter>
+			<parameter name="Key">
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 16
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

gtjoseph:
N.B.  We still produce the documentation for the 16 branch even though we don't accept changes to it.

main/manager.c
```diff
@@ -165,14 +165,32 @@
...
+				</enumlist>
+			</parameter>
 			<parameter name="Secret">
-				<para>Secret to login with as specified in manager.conf.</para>
+				<para>Plain text secret to login with as specified in manager.conf.</para>
+			</parameter>
+			<parameter name="Key">
+				<para>Key to use with MD5 authentication.  To create the key, you must
+				first call the <literal>Challenge</literal> action, create an MD5 hash
+				with the result, then update the hash with the secret as specified
```
seanbright, line 181: What needs to be hashed specifically? The challenge response concatenated with the plain text password?

gtjoseph: Just as I described.  You need to create the hash first with the challenge response, then update the hash with with the secret.  Exactly how you do it depends on the technology you're using to interface with AMI.

If you're using python for instance...
```
import hashlib
m = hashlib.md5()
m.update(response_from_challenge)
m.update(your_secret)
key = m.hexdigest()
## '031edd7d41651593c5fe5c006fa5752b'
```

Do I need to include the example?

gtjoseph: Added python example to the description.

seanbright (approved):

jcolp (changes requested):
The commit message has "parameteers" instead of "parameters".

jcolp (approved):

PR #706, opened 2024-04-11
Initial RTP inactivity check must consider the rtp_timeout setting.

### LINKED ISSUE #710: [bug]: RTP inactivity monitor hang up calls after 500ms.
### Severity

Major

### Versions

18, 20, 21

### Components/Modules

res_pjsip_sdp_rtp.c

### Operating Environment

N/A

### Frequency of Occurrence

Occasional

### Issue Description

RTP activity monitor is causing calls to be hung up under certain conditions.

Steps to reproduce:
Given rtp_timeout is 10 seconds

1. Receive a call
2. Let it ring for a duration longer than 10 seconds.
3. Answer the call. 
4. Rtp activity started after 500ms but before the 10 second rtp_timeout. 

Actual Result:
Call was hung up 500ms after answered due to RTP inactivity.

Expected result:
Call should not hang up due to RTP inactivity.

Example of the problem:
Given rtp_timeout is 10 seconds.

If the call is answered immediately we would have 10 seconds to receive the first rtp before "softhangup" due to rtp inactivity
If the call is answered after 5 seconds we would have 5 seconds to receive the first rtp before "softhangup" due to rtp inactivity
If the call is answered after 15 seconds we would have 0.5 seconds to receive the first rtp before "softhangup"due to rtp inactivity

Conclusion, it should not matter how long the call was in the ringing state for the rtp timeout logic to trigger the "softhangup".

### Relevant log output

_No response_

### Asterisk Issue Guidelines

henlil:
First rtp activity check was performed after 500ms regardless of the rtp_timeout setting. Having a call in ringing state for more than rtp_timeout and the first rtp package is received more than 500ms after sdp negotiation and before the rtp_timeout, erronously caused the call to be hungup. Changed to perform the first rtp inactivity check after the timeout setting preventing calls to be disconnected before the rtp_timeout has elapsed since sdp negotiation.

Fixes #710

### DIFF
--- a/res/res_pjsip_sdp_rtp.c
+++ b/res/res_pjsip_sdp_rtp.c
@@ -2101,6 +2101,7 @@ static int apply_negotiated_sdp_stream(struct ast_sip_session *session,
 	enum ast_media_type media_type = session_media->type;
 	char host[NI_MAXHOST];
 	int res;
+	int rtp_timeout;
 	struct ast_sip_session_media *session_media_transport;
 	SCOPE_ENTER(1, "%s Stream: %s\n", ast_sip_session_get_name(session),
 		ast_str_tmp(128, ast_stream_to_str(asterisk_stream, &STR_TMP)));
@@ -2246,8 +2247,10 @@ static int apply_negotiated_sdp_stream(struct ast_sip_session *session,
 		ast_rtp_instance_set_timeout(session_media->rtp, session->endpoint->media.rtp.timeout_hold);
 	}
 
-	if (ast_rtp_instance_get_timeout(session_media->rtp)) {
-		session_media->timeout_sched_id = ast_sched_add_variable(sched,	500, rtp_check_timeout,
+	rtp_timeout = ast_rtp_instance_get_timeout(session_media->rtp);
+
+	if (rtp_timeout) {
+		session_media->timeout_sched_id = ast_sched_add_variable(sched,	rtp_timeout*1000, rtp_check_timeout,
 			session_media, 1);
 	}
 


res/res_pjsip_sdp_rtp.c
```diff
@@ -2246,8 +2246,10 @@ static int apply_negotiated_sdp_stream(struct ast_sip_session *session,
 		ast_rtp_instance_set_timeout(session_media->rtp, session->endpoint->media.rtp.timeout_hold);
 	}
 
-	if (ast_rtp_instance_get_timeout(session_media->rtp)) {
-		session_media->timeout_sched_id = ast_sched_add_variable(sched,	500, rtp_check_timeout,
+	int rtp_timeout = ast_rtp_instance_get_timeout(session_media->rtp);
```
jcolp, line 2249: This will not compile under dev mode (when developing you should always build using "./configure --enable-dev-mode" which will turn warnings into errors and stop builds).  Variables need to be declared at the start of the function.

res/res_pjsip_sdp_rtp.c
```diff
@@ -2246,8 +2246,10 @@ static int apply_negotiated_sdp_stream(struct ast_sip_session *session,
...
+
+	if (rtp_timeout) {
+		session_media->timeout_sched_id = ast_sched_add_variable(sched,	rtp_timeout*1000, rtp_check_timeout,
```
jcolp, line 2252: Looking at the commit history, the code was originally this and was changed in:

https://github.com/asterisk/asterisk/commit/6624e34580f5f5c39d3dbbfb47ba29839397f3d7

Were the changes done in rtp_check_timeout insufficient? How so?

jcolp (changes requested):
The commit message does not conform to standard:

A comment also needs to be added for what this should be cherry picked into:

https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/#cherry-picking

res/res_pjsip_sdp_rtp.c
henlil: https://github.com/asterisk/asterisk/commit/6624e34580f5f5c39d3dbbfb47ba29839397f3d7 does not say the reason to why initial timer value changed to a fixed 500ms, but I believe it introduced a bug. 

1. Invite received. create_rtp is invoked setting rtp_rtp->last_rx = time(NULL)
2. Let it ring for a duration longer than the rtp_timeout setting.
3. Answer the call. sdp is negotiated and scheduler is setup to trigger rtp_check_timeout in 500ms.
4. 500ms elapses without any rtp or rtcp packages received.
5. rtp_check_timeout is triggered by the scheduler. The call is now in state "UP" and the elapsed check is done using the initial setup of last_rx, done when the invite received (in step 1 above). The elapsed time is calculated from the wrong starting point, making the elapsed time greater than rtp_timeout, causing a hang up half a second after it's been answered! 

When the call is answered, sdp negotiated, it should not hangup before rtp_timeout elapsed. Therefor it does not make sense to invoke rtp_check_timeout the first time before the rtp_timeout elapsed.

Example of the problem:
Given rtp_timeout is 10 seconds.
If the call is answered immediately we would have 10 seconds to receive the first rtp before "softhangup"
If the call is answered after 5 seconds we would have 5 seconds to receive the first rtp before "softhangup"
If the call is answered after 15 seconds we would have 0.5 seconds to receive the first rtp before "softhangup"

Conclusion it should not matter how long the call was in the ringing ("UP") state for the rtp timeout logic to trigger the hangup.

My solution would perform the first rtp check after 10 seconds from when the call was answered (=sdp negotiated)

henlil:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

henlil:
When reading this https://docs.asterisk.org/About-the-Project/Asterisk-Versions/ my understanding is that this fix, if approved, should not go into 18, even if the bug exists there since it's not a security fix?

jcolp (approved):

gtjoseph:
@henlil Just FYI for next time...  The PR description doesn't get automatically updated when you update the commit message and the "Fixes" line has to appear in the PR description or the related issue won't be automatically closed.  I fixed it for you.

PR #711, opened 2024-04-18
logger.h:  Add SCOPE_CALL and SCOPE_CALL_WITH_RESULT

gtjoseph:
If you're tracing a large function that may call another function
multiple times in different circumstances, it can be difficult to
see from the trace output exactly which location that function
was called from.  There's no good way to automatically determine
the calling location.  SCOPE_CALL and SCOPE_CALL_WITH_RESULT
simply print out a trace line before and after the call.
    
The difference between SCOPE_CALL and SCOPE_CALL_WITH_RESULT is
that SCOPE_CALL ignores the function's return value (if any) where
SCOPE_CALL_WITH_RESULT allows you to specify the type of the
function's return value so it can be assigned to a variable.
SCOPE_CALL_WITH_INT_RESULT is just a wrapper for SCOPE_CALL_WITH_RESULT
and the "int" return type.

### DIFF
--- a/include/asterisk/logger.h
+++ b/include/asterisk/logger.h
@@ -832,6 +832,76 @@ unsigned long _ast_trace_dec_indent(void);
 	} \
 
 /*!
+ * \def SCOPE_CALL
+ * \brief Wrap a function call in "--> Calling:" and "<-- Return from:" trace messages
+ *
+ * \param level The trace level (usually -1 to use the same level as the calling function)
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

mbradeen (approved):

PR #712, opened 2024-04-19
app_voicemail_odbc: Allow audio to be kept on disk

gtjoseph:
This commit adds a new voicemail.conf option 'odbc_audio_on_disk'
which when set causes the ODBC variant of app_voicemail to leave
the message and greeting audio files on disk and only store the
message metadata in the database.  This option came from a concern
that the database could grow to large and cause remote access
and/or replication to become slow.  In a clustering situation
with this option, all asterisk instances would share the same
database for the metadata and either use a shared filesystem
or other filesystem replication service much more suitable
for synchronizing files.

The changes to app_voicemail to implement this feature were actually
quite small but due to the complexity of the module, the actual
source code changes were greater.  They fall into the following
categories:

* Tracing.  The module is so complex that it was impossible to
figure out the path taken for various scenarios without the addition
of many SCOPE_ENTER, SCOPE_EXIT and ast_trace statements, even in
code that's not related to the functional change.  Making this worse
was the fact that many "if" statements in this module didn't use
braces.  Since the tracing macros add multiple statements, many "if"
statements had to be converted to use braces.

* Excessive use of PATH_MAX.  Previous maintainers of this module
used PATH_MAX to allocate character arrays for filesystem paths
and SQL statements as though they cost nothing.  In fact, PATH_MAX
is defined as 4096 bytes!  Some functions had (and still have)
multiples of these.  One function has 7.  Given that the vast
majority of installations use the default spool directory path
`/var/spool/asterisk/voicemail`, the actual path length is usually
less than 80 bytes.  That's over 4000 bytes wasted.  It was the
same for SQL statement buffers.  A 4K buffer for statement that
only needed 60 bytes.  All of these PATH_MAX allocations in the
ODBC related code were changed to dynamically allocated buffers.
The rest will have to be addressed separately.

* Bug fixes.  During the development of this feature, several
pre-existing ODBC related bugs were discovered and fixed.  They
had to do with leaving orphaned files on disk, not preserving
original message ids when moving messages between folders,
not honoring the "formats" config parameter in certain circumstances,
etc.

UserNote: This commit adds a new voicemail.conf option
'odbc_audio_on_disk' which when set causes the ODBC variant of
app_voicemail_odbc to leave the message and greeting audio files
on disk and only store the message metadata in the database.
Much more information can be found in the voicemail.conf.sample
file.

### DIFF
--- a/apps/app_voicemail.c
+++ b/apps/app_voicemail.c
@@ -689,9 +689,13 @@ static AST_LIST_HEAD_STATIC(vmstates, vmstate);
 #define VM_FWDURGAUTO    (1 << 18)  /*!< Autoset of Urgent flag on forwarded Urgent messages set globally */
 #define VM_EMAIL_EXT_RECS (1 << 19)  /*!< Send voicemail emails when an external recording is added to a mailbox */
 #define VM_MARK_URGENT   (1 << 20)	/*!< After recording, permit the caller to mark the message as urgent */
+#define VM_ODBC_AUDIO_ON_DISK  (1 << 21) /*!< When using ODBC, leave the message and greeting recordings on disk instead of moving them into the table */
+
 #define ERROR_LOCK_PATH  -100
 #define ERROR_MAX_MSGS   -101
 #define OPERATOR_EXIT     300
+
+#define MSGFILE_LEN (7) /*!< Length of the message file name: msgXXXX */
 
 enum vm_box {
 	NEW_FOLDER = 		0,
@@ -978,13 +982,16 @@ struct vm_state {
 #ifdef ODBC_STORAGE
 static char odbc_database[80] = "asterisk";
 static char odbc_table[80] = "voicemessages";
-#define RETRIEVE(a,b,c,d) retrieve_file(a,b)
-#define DISPOSE(a,b) remove_file(a,b)
-#define STORE(a,b,c,d,e,f,g,h,i,j,k) store_file(a,b,c,d)
-#define EXISTS(a,b,c,d) (message_exists(a,b))
-#define RENAME(a,b,c,d,e,f,g,h) (rename_file(a,b,c,d,e,f))
-#define COPY(a,b,c,d,e,f,g,h) (copy_file(a,b,c,d,e,f))
-#define DELETE(a,b,c,d) (delete_file(a,b))
+size_t odbc_table_len = sizeof(odbc_table);
+#define COUNT(a, b) odbc_count_messages(a,b)
+#define LAST_MSG_INDEX(a) odbc_last_message_index(a)
+#define RETRIEVE(a,b,c,d) odbc_retrieve_message(a,b)
+#define DISPOSE(a,b) odbc_remove_files(a,b)
+#define STORE(a,b,c,d,e,f,g,h,i,j,k) odbc_store_message(a,b,c,d)
+#define EXISTS(a,b,c,d) (odbc_message_exists(a,b))
+#define RENAME(a,b,c,d,e,f,g,h) (odbc_rename_message(a,b,c,d,e,f))
+#define COPY(a,b,c,d,e,f,g,h) (odbc_copy_message(a,b,c,d,e,f))
+#define DELETE(a,b,c,d) (odbc_delete_message(a,b))
 #define UPDATE_MSG_ID(a, b, c, d, e, f) (odbc_update_msg_id((a), (b), (c)))
 #else
 #ifdef IMAP_STORAGE
@@ -997,6 +1004,8 @@ static char odbc_table[80] = "voicemessages";
 #define DELETE(a,b,c,d) (vm_imap_delete(a,b,d))
 #define UPDATE_MSG_ID(a, b, c, d, e, f) (vm_imap_update_msg_id((a), (b), (c), (d), (e), (f)))
 #else
+#define COUNT(a, b) count_messages(a,b)
+#define LAST_MSG_INDEX(a) last_message_index(a)
 #define RETRIEVE(a,b,c,d)
 #define DISPOSE(a,b)
 #define STORE(a,b,c,d,e,f,g,h,i,j,k)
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 21

gtjoseph:
Branches 18 and 20 need a slightly tweaked version of this patch because they still have the macrocontext field.  When this PR is approved, I'll submit the PR for 18 and 20.

gtjoseph:
Oh, one thing to note... the PR mentions app_voicemail_odbc but you'll see the diff is for app_voicemail.c.  Remember, app_voicemail_odbc.c is just a symlink to app_voicemail.c.

apps/app_voicemail.c
```diff
@@ -3887,6 +4026,43 @@ static int msg_id_incrementor;
...
+ * \internal
+ * \brief Create a buffer containing the SQL statement with the table name inserted.
+ *
+ * \param __sql_fmt The SQL statement with a single '%s' where the table name should be inserted.
+ *
+ * \note The buffer is allocated on the stack and should not be freed.
+ *
+ * \return A pointer to a buffer containing the SQL statement with the table name inserted.
+ */
+#define MAKE_SQL_PTRA(__sql_fmt) \
```
jkroonza, line 4040: Why not pass the arguments here more generically using ...?

Then you can MAKE_SQL_PTRA("UPDATE %s SET msg_id=? WHERE dir=? AND msgnum=?", odbc_table); iinstead, but the advantage is that you can nuke MAKE_SQL_PTRA2 as well, and deal with it there, even better might to turn this into an inline function with attribute's so gcc can check formats?

apps/app_voicemail.c
```diff
@@ -3919,30 +4095,80 @@ static SQLHSTMT generic_prepare(struct odbc_obj *obj, void *data)
...
 	} else {
 		SQLFreeHandle(SQL_HANDLE_STMT, stmt);
 	}
 	ast_odbc_release_obj(obj);
-	return;
+	SCOPE_EXIT_RTN();
 }
 
+#define AUDIO_ON_DISK_MAGIC "AUDMAGIC"
+#define AUDIO_ON_DISK_MAGIC_LEN 8
```
jkroonza, line 4122: #define AUDIO_ON_DISK_MAGIG_LEN   (sizeof(AUDIO_ON_DISK_MAGIC) - 1) ??

jkroonza (commented):
No real input, the changes looks fine, and should not impact on IMAP which was the original request to check for.  Needs to be properly tested this is a huge diff.

apps/app_voicemail.c
```diff
@@ -1322,6 +1331,136 @@ static char *strip_control_and_high(const char *input, char *buf, size_t buflen)
...
+ * \returns A pointer to the buffer containing the path.
+ */
+#define MAKE_FILE_PTRA(dir, msgnum) \
+({ \
+    size_t __len = get_msg_path_len(dir); \
+    char *__var; \
+    if (msgnum < 0) { \
+        __var = ast_strdupa(dir); \
+    } else { \
+        __var = ast_alloca(__len); \
```
InterLinked1, line 1417: Would be nice if there was an asprintf version that used alloca instead of alloc... do you think it's worth adding a macro for that to simplify all these?

apps/app_voicemail.c
```diff
@@ -4053,46 +4270,77 @@ static int retrieve_file(char *dir, int msgnum)
...
-				lseek(fd, fdlen - 1, SEEK_SET);
-				if (write(fd, tmp, 1) != 1) {
-					close(fd);
-					fd = -1;
-					continue;
+			ast_trace(-1, "Audio size: %ld\n", colsize2);
+			if (colsize2 == AUDIO_ON_DISK_MAGIC_LEN) {
+				res = SQLGetData(stmt, x + 1, SQL_BINARY, rowdata, AUDIO_ON_DISK_MAGIC_LEN, NULL);
+				if (memcmp(rowdata, AUDIO_ON_DISK_MAGIC, AUDIO_ON_DISK_MAGIC_LEN) != 0) {
+					ast_log(AST_LOG_WARNING, "Invalid audio magic number for '%s'\n", full_fn);
```
InterLinked1, line 4284: Could we also print the invalid magic number here?

apps/app_voicemail.c
```diff
@@ -4339,7 +4603,13 @@ static void delete_file(const char *sdir, int smsg)
 	}
 	ast_odbc_release_obj(obj);
 
-	return;
+	if (ast_test_flag((&globalflags), VM_ODBC_AUDIO_ON_DISK)) {
+		char *src_fn = MAKE_FILE_PTRA(sdir, smsg);
+		ast_trace(-1, "Audio stored on disk.  Deleting '%s'\n", src_fn);
```
InterLinked1, line 4608: Two spaces

apps/app_voicemail.c
jkroonza: sprintfa?  That would get very confusing but I do like the idea.

apps/app_voicemail.c
jkroonza: Are we guaranteed that 8 bytes of non-magic isn't perhaps raw audio? (I find it HIGHLY unlikely to be valid, but can it be guaranteed?)

main/app.c
```diff
@@ -1724,7 +1724,11 @@ static int global_maxsilence = 0;
  * \retval 't' Recording ended from the message exceeding the maximum duration, or via DTMF in prepend mode
  * \retval dtmfchar Recording ended via the return value's DTMF character for either cancel or accept.
  */
-static int __ast_play_and_record(struct ast_channel *chan, const char *playfile, const char *recordfile, int maxtime, const char *fmt, int *duration, int *sound_duration, int beep, int silencethreshold, int maxsilence, const char *path, int prepend, const char *acceptdtmf, const char *canceldtmf, int skip_confirmation_sound, enum ast_record_if_exists if_exists)
+static int __ast_play_and_record(struct ast_channel *chan, const char *playfile,
+	const char *recordfile, int maxtime, const char *fmt, int *duration,
+	int *sound_duration, int beep, int silencethreshold, int maxsilence,
+	const char *path, int prepend, const char *acceptdtmf, const char *canceldtmf,
+	int skip_confirmation_sound, enum ast_record_if_exists if_exists)
```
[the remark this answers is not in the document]
gtjoseph: Ah, yeah.  I had some tracing in there but removed it.  Forgot to undo the long line wrap.

apps/app_voicemail.c
gtjoseph: We already know the length and it doesn't make sense to run sizeof every time the macro is referenced.

apps/app_voicemail.c
gtjoseph: I'll see if I can clean these up.

apps/app_voicemail.c
seanbright: The sizeof would be done at compile by the preprocessor

apps/app_voicemail.c
gtjoseph: Don't understand.

apps/app_voicemail.c
gtjoseph: @InterLinked1 Yeah, I can print it.

@jkroonza If it's 8 bytes then even if it's raw audio, it's useless.

apps/app_voicemail.c
gtjoseph: See below.

apps/app_voicemail.c
gtjoseph: Oh, between "disk." and "Deleting"?

apps/app_voicemail.c
InterLinked1: > So here's the deal... part of the convenience of the macros is using alloca so you don't have to call ast_free or use RAII_VAR. You can't wrap alloca in a function because of its very nature of allocating on the stack. You also can't use VA_ARGS because the only way to know how much memory to allocate would be to call snprintf once with a `0` as the buffer length to get the buffer size needed, then allocate the memory and call snprintf again. That's just a waste for these limited cases. I'm going to leave the macros as is.

I know a function definitely won't work, but won't a macro wrapper work?

apps/app_voicemail.c
jkroonza: same applies for strlen() nowadays if optimization is enabled, if the string is a compile-time constant strlen() reduces to a compile-time constant ...

apps/app_voicemail.c
jkroonza: #define sprintf_alloca(fmt, ...) {( int l = snprintf(NULL, 0, fmt, ##__VA_ARGS__); char *b = alloca(l+1); if (b) { snprintf(b, l+1, fmt, ##__VA_ARGS__); }; b; })

What I don't like about this is the double-invocation of snprintf, which if __VA_ARGS__ contains function calls which isn't marked as pure could cause problems.

jkroonza: alternatively asprintf(), strdupa(), free(), but that just feels like a waste of CPU cycles, not to mention abusing the memory manager even more (as if asterisk doesn't already cause malloc enough headaches).

InterLinked1: > #define sprintf_alloca(fmt, ...) {( int l = snprintf(NULL, 0, fmt, ##**VA_ARGS**); char *b = alloca(l+1); if (b) { snprintf(b, l+1, fmt, ##**VA_ARGS**); }; b; })

The second snprintf is not necessary, sprintf could be used safely in that case.

> What I don't like about this is the double-invocation of snprintf, which if **VA_ARGS** contains function calls which isn't marked as pure could cause problems.

But this is true in existing usage as well, no?

The macro could have a `note` that notes that.

> alternatively asprintf(), strdupa(), free(), but that just feels like a waste of CPU cycles, not to mention abusing the memory manager even more (as if asterisk doesn't already cause malloc enough headaches).

I agree, we should only add a macro if that will simplify without adding actual overhead, which I think is possible but I didn't think about it extensively.

Regarding length, I noticed this in a few places. The place I commented seems to already have the length so the benefit is less here, but in other places, the length is calculated for alloca in which case it seems like a no brainer to use a macro like this.

jkroonza: perhaps two variants?  One that takes a length snprintf_alloca(len, fmt, ...) and sprintf_alloca(fmt, ...) where the latter uses snprintf to get the length and then just reverts to snprintf_alloca, being sure to *document* the double evaluation of arguments.  Just be wary of the off by one, so snprintf_alloca(len, fmt, ...) should alloca(len+1) such that len is the length of the string without the NULL (from printf man page "Upon successful return, these functions return the number of characters printed (excluding the null byte used to end output to strings)." so IMHO it makes sense to stick with the "excluding null byte" assumptions).

gtjoseph: So, it this just a philosophical discussion now or are you guys requesting changes?

InterLinked1: I would like to see a change, at least in the alloca case where it simplifies with no downside

jkroonza: I believe @gtjoseph already eliminated the macro here, so we should probably move the specific discussion elsewhere.

gtjoseph: > I would like to see a change, at least in the alloca case where it simplifies with no downside

What change specifically?  I'm seeing neither complexity nor downside in the macros as is.

InterLinked1: Okay, I just reviewed again and now nothing rubs me the wrong way anymore so I think you addressed them all :)

mbradeen (commented):
just some spelling stuff.

configs/samples/voicemail.conf.sample
```diff
@@ -138,12 +138,93 @@ maxlogins=3
...
+; prefer to store the audio files on disk and only store the message
+; metadata in the database, set the following option to 'yes'.
+; This can be advantageous in some scenarios, such as when the
+; database is on a separate server from the Asterisk server and
+; network latency and size is a concern or when the database is
+; not well-suited for storing large binary objects.  It can also
+; be useful when you want to use the same voicemail storage
+; configuration for multiple Asterisk servers.  In this situation
+; you can use a shared network file system to store the audio files
+; and but each Asterisk server use the same database for fast
```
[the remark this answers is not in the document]
gtjoseph: done

configs/samples/voicemail.conf.sample
```diff
@@ -138,12 +138,93 @@ maxlogins=3
...
+
+; If you set this option to "yes", new messages and greetings will
+; have their audio kept on disk but the audio for existing messages
+; and greetings will remain in the database until the next time
+; they are played.  At that time, the audio will be moved to disk
+; and erased from the database.
+
+; If you set this option to "no" after you've already stored messages
+; or greetings with it set to "yes", new messages and greetings will
+; have their audio stored in the database but but the audio for
```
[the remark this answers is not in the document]
gtjoseph: done

mbradeen (approved):

jcolp (approved):
This change is large enough that I think it requires a third set of eyes.

gtjoseph:
@seanbright The reviews section at the top of the PR says you left review comments but I can't seem to find them.  Did I address them?

seanbright:
> @seanbright The reviews section at the top of the PR says you left review comments but I can't seem to find them. Did I address them?

You addressed all of the comments that I remember making (which may be zero...)

apps/app_voicemail.c
```diff
@@ -1322,6 +1331,136 @@ static char *strip_control_and_high(const char *input, char *buf, size_t buflen)
...
+ * in practice.
+ */
+#ifdef ODBC_STORAGE
+/*!
+ * \internal
+ * \brief Get the length needed for a message path base.
+ *
+ * \param dir The absolute path to a directory
+ *
+ * \note The directory provided should include all componenets
```
bkford, line 1347: s/componenets/components

apps/app_voicemail.c
```diff
@@ -1322,6 +1331,136 @@ static char *strip_control_and_high(const char *input, char *buf, size_t buflen)
...
+ * but we'll make this 12 to be extra safe.
+ */
+#define MAX_SOUND_EXTEN_LEN 12
+/*!
+ * \internal
+ * \brief Get the length needed for a message sound or metadata file path.
+ *
+ * \param dir The absolute path to a directory
+ *
+ * \note The directory provided should include all componenets
```
bkford, line 1374: s/componenets/components

bkford (approved):
Just a couple minor typos, otherwise lgtm

jcolp (approved):

Issue #713, opened 2024-04-23
[bug]: SNI isn't being set on websocket client connections


gtjoseph:
### Severity

Minor

### Versions

all

### Components/Modules

tcptls.c,iostream.c

### Operating Environment

all

### Frequency of Occurrence

Constant

### Issue Description

When creating secure websocket client connections (wss), the destination hostname isn't being set in the TLS SNI extension.  This causes TLS negotiation failures when the remote server hosts multiple virtual servers and needs SNI to distinguish between them.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #714, opened 2024-04-23
tcptls/iostream:  Add support for setting SNI on client TLS connections

gtjoseph:
If the hostname field of the ast_tcptls_session_args structure is
set (which it is for websocket client connections), that hostname
will now automatically be used in an SNI TLS extension in the client
hello.

Resolves: #713

UserNote: Secure websocket client connections now send SNI in
the TLS client hello.

### DIFF
--- a/main/iostream.c
+++ b/main/iostream.c
@@ -46,6 +46,7 @@ struct ast_iostream {
 	int rbuflen;
 	char *rbufhead;
 	char rbuf[2048];
+	char *sni_hostname;
 };
 
 #if defined(DO_SSL)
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (changes requested):
This should go into certified too.

main/iostream.c
```diff
@@ -597,6 +608,7 @@ static void iostream_dtor(void *cookie)
 	 * put this in a conditional to avoid compiler warnings. */
 	struct ast_iostream *stream = cookie;
 #endif
+	ast_free(stream->sni_hostname);
```
jcolp, line 611: This won't work in non-dev mode as stream is not present.

jcolp (changes requested):

main/iostream.c
gtjoseph: done

jcolp (approved):

Issue #721, opened 2024-04-27
[improvement]: logger: Add unique verbose prefixes for higher verbose levels


InterLinked1:
### Improvement Description

Commit 6871d1cdfc4aba983935126863a15265b9518717 pushed a lot of module-related logging to higher verbose levels, which now makes it impossible to visually differentiate the hierarchy of any of these messages at a quick glance.

To improve this, unique verbose prefixes are now added for all levels, rather than using prefix 4 for levels 4 through 10. This allows for the same visual differentiation as before, as well as in all other cases for messages of different levels.

PR #722, opened 2024-04-27
logger: Add unique verbose prefixes for levels 5-10.

InterLinked1:
Add unique verbose prefixes for levels higher than 4, so that these can be visually differentiated from each other.

Resolves: #721

### DIFF
--- a/include/asterisk/logger.h
+++ b/include/asterisk/logger.h
@@ -41,10 +41,16 @@ extern "C" {
 
 #define _A_ __FILE__, __LINE__, __FUNCTION__
 
-#define VERBOSE_PREFIX_1 " "
-#define VERBOSE_PREFIX_2 "  == "
-#define VERBOSE_PREFIX_3 "    -- "
-#define VERBOSE_PREFIX_4 "       > "
+#define VERBOSE_PREFIX_1  " "
+#define VERBOSE_PREFIX_2  "  == "
+#define VERBOSE_PREFIX_3  "    -- "
+#define VERBOSE_PREFIX_4  "       > "
+#define VERBOSE_PREFIX_5  "        > "
+#define VERBOSE_PREFIX_6  "         > "
+#define VERBOSE_PREFIX_7  "          > "
+#define VERBOSE_PREFIX_8  "           > "
+#define VERBOSE_PREFIX_9  "            > "
+#define VERBOSE_PREFIX_10 "             > "
 
 #define AST_CALLID_BUFFER_LENGTH 13
 
@@ -161,7 +167,7 @@ void __attribute__((format(printf, 5, 6))) ast_queue_log(const char *queuename, 
  * This will print the message to the console if the verbose level is set to a level >= 3
  *
  * Note the absence of a comma after the VERBOSE_PREFIX_3.  This is important.
- * VERBOSE_PREFIX_1 through VERBOSE_PREFIX_4 are defined.
+ * VERBOSE_PREFIX_1 through VERBOSE_PREFIX_10 are defined.
  *
  * \version 11 added level parameter
  */
--- a/main/asterisk.c
+++ b/main/asterisk.c
@@ -2164,7 +2164,19 @@ static void set_header(char *outbuf, int maxout, char level)
 		break;
 	case 3: cmp = VERBOSE_PREFIX_3;
 		break;
-	default: cmp = VERBOSE_PREFIX_4;
+	case 4: cmp = VERBOSE_PREFIX_4;
+		break;
+	case 5: cmp = VERBOSE_PREFIX_5;
+		break;
+	case 6: cmp = VERBOSE_PREFIX_6;
+		break;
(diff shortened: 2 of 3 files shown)


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

include/asterisk/logger.h
```diff
@@ -45,6 +45,12 @@ extern "C" {
 #define VERBOSE_PREFIX_2 "  == "
 #define VERBOSE_PREFIX_3 "    -- "
 #define VERBOSE_PREFIX_4 "       > "
+#define VERBOSE_PREFIX_5 "        >> "
+#define VERBOSE_PREFIX_6 "         > "
+#define VERBOSE_PREFIX_7 "         >> "
+#define VERBOSE_PREFIX_8 "          > "
+#define VERBOSE_PREFIX_9 "          >> "
+#define VERBOSE_PREFIX_10 "          > "
```
gtjoseph, line 53: Why `>>` on the odd levels and `>` on the even levels?

gtjoseph (changes requested):

include/asterisk/logger.h
InterLinked1: No particular reason, really, that was just one method of furthering the hierarchy that came to mind, you can see they progressively get more indented. Open to other suggestions, but I didn't want to add a whole bunch of different symbols either.

gtjoseph: > you can see they progressively get more indented

Ah, no they don't.  8 and 10 are the same. :)

I'd just make them all `>` with the progressive spaces.

gtjoseph (changes requested):

include/asterisk/logger.h
gtjoseph: > I might be missing something, but 10 is definitely indented one more space than 8 is, if you compare with the column alignment for 9.

It only looks that way because "10" is longer than "8". If you line up the opening quotes, you can see it...

```
#define VERBOSE_PREFIX_4  "       > "
#define VERBOSE_PREFIX_5  "        >> "
#define VERBOSE_PREFIX_6  "         > "
#define VERBOSE_PREFIX_7  "         >> "
#define VERBOSE_PREFIX_8  "          > "
#define VERBOSE_PREFIX_9  "          >> "
#define VERBOSE_PREFIX_10 "          > "
```
> The downside to making them all just a single '>` more progressively spaced is that doubles the indentation as you go down. On small consoles, that can result in a lot more wasted space. The way it is now saves a few columns.

Yeah but we only have 1 level 10 verbose message and none at all for levels 6-9 so I don't see that as an issue.

If you think it really is, how about...

```
#define VERBOSE_PREFIX_4  "       > "
#define VERBOSE_PREFIX_5  "      +> "
#define VERBOSE_PREFIX_6  "     ++> "
#define VERBOSE_PREFIX_7  "    +++> "
#define VERBOSE_PREFIX_8  "   ++++> "
#define VERBOSE_PREFIX_9  "  +++++> "
#define VERBOSE_PREFIX_10 " ++++++> "
```

InterLinked1: > > I might be missing something, but 10 is definitely indented one more space than 8 is, if you compare with the column alignment for 9.
> 
> It only looks that way because "10" is longer than "8". If you line up the opening quotes, you can see it...

Ah, I missed that!

> ```
> #define VERBOSE_PREFIX_4  "       > "
> #define VERBOSE_PREFIX_5  "        >> "
> #define VERBOSE_PREFIX_6  "         > "
> #define VERBOSE_PREFIX_7  "         >> "
> #define VERBOSE_PREFIX_8  "          > "
> #define VERBOSE_PREFIX_9  "          >> "
> #define VERBOSE_PREFIX_10 "          > "
> ```
> 
> > The downside to making them all just a single '>` more progressively spaced is that doubles the indentation as you go down. On small consoles, that can result in a lot more wasted space. The way it is now saves a few columns.
> 
> Yeah but we only have 1 level 10 verbose message and none at all for levels 6-9 so I don't see that as an issue.

Fair enough... I think I use 6 or 7 in some of my stuff but not extensively so minor issue.

> If you think it really is, how about...
> 
> ```
> #define VERBOSE_PREFIX_4  "       > "
> #define VERBOSE_PREFIX_5  "      +> "
> #define VERBOSE_PREFIX_6  "     ++> "
> #define VERBOSE_PREFIX_7  "    +++> "
> #define VERBOSE_PREFIX_8  "   ++++> "
> #define VERBOSE_PREFIX_9  "  +++++> "
> #define VERBOSE_PREFIX_10 " ++++++> "
> ```

This is the *opposite* of what we should do since it looks like 10 is more "important" than 4... the indentation needs to be consistent on each side.

I'll just go with your original approach here, it works well enough I think.

gtjoseph (approved):

jcolp (approved):

jcolp (approved):

Issue #719, opened 2024-04-26
[bug]: segfault on start if compiled with DETECT_DEADLOCKS


alex2grad:
### Severity

Trivial

### Versions

18.22.0

### Components/Modules

lock.c

### Operating Environment

Slackware 15.0 x86_64

### Frequency of Occurrence

Constant

### Issue Description

If asterisk is compiled with DETECT_DEADLOCKS and without DEBUG_THREADS it's immediately crashes on startup.

CFLAGS="-g -Og -fPIC" \
CXXFLAGS="-g -Og -fPIC" \
./configure \
  --prefix=/usr \
  --libdir=/usr/lib64 \
  --sysconfdir=/etc \
  --localstatedir=/var \
  --mandir=/usr/man \
  --with-pjproject-bundled \
  --with-download-cache=/tmp/downloads \
  --with-libedit=/usr \
  --enable-dev-mode=yes \
  --build=x86_64-slackware-linux

make menuselect.makeopts
menuselect/menuselect \
    --disable res_calendar_caldav \
    --disable CORE-SOUNDS-EN-GSM \
    --disable MOH-OPSOUND-WAV \
    --enable DETECT_DEADLOCKS \
    menuselect.makeopts

### Relevant log output

```shell
# asterisk -fdvvvvvvvv
Seeding global EID 'xx:xx:xx:xx:xx:xx' from 'br0' using 'siocgifhwaddr'
Unable to open specified master config file '/etc/asterisk/asterisk.conf', using built-in defaults
Asterisk 18.22.0, Copyright (C) 1999 - 2022, Sangoma Technologies Corporation and others.
Created by Mark Spencer <markster@digium.com>
Asterisk comes with ABSOLUTELY NO WARRANTY; type 'core show warranty' for details.
This is free software, with components licensed under the GNU General Public
License version 2 and other licenses; you are welcome to redistribute it under
certain conditions. Type 'core show license' for details.
=========================================================================
Manager registered action DBGet
Segmentation fault

Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x0000000000465329 in __ao2_lock (user_data=user_data@entry=0x1, lock_how=lock_how@entry=AO2_LOCK_REQ_MUTEX, file=file@entry=0x637fb1 "astobj2.c", func=func@entry=0x638338 <__PRETTY_FUNCTION__.11> "__ao2_ref", line=line@entry=526, 
    var=var@entry=0x637fc7 "weakproxy") at astobj2.c:224
[...]

PR #723, opened 2024-04-27
make_buildopts_h: Always include DETECT_DEADLOCKS

gtjoseph:
Since DETECT_DEADLOCKS is now split from DEBUG_THREADS, it must
always be included in buildopts.h instead of only when
ADD_CFLAGS_TO_BUILDOPTS_H is defined.  A SEGV will result otherwise.

Resolves: #719


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

PR #732, opened 2024-05-12
res_pjsip_path.c: Fix path when dialing using PJSIP_DIAL_CONTACTS()

IgorrG:
When using the PJSIP_DIAL_CONTACTS() function for use in the Dial()
command, the contacts are returned in text form, so the input to
the path_outgoing_request() function is a contact value of NULL.
The issue was reported in ASTERISK-28211, but was not actually fixed
in ASTERISK-30100. This fix brings back the code that was previously
removed and adds code to search for a contact to extract the path
value from it.

### DIFF
--- a/res/res_pjsip_path.c
+++ b/res/res_pjsip_path.c
@@ -48,6 +48,109 @@ static struct ast_sip_aor *find_aor(struct ast_sip_contact *contact)
 	return ast_sip_location_retrieve_aor(contact->aor);
 }
 
+static struct ast_sip_aor *find_aor2(struct ast_sip_endpoint *endpoint, pjsip_uri *uri)
+{
+	char *configured_aors, *aor_name;
+	const pj_str_t *uri_username;
+	const pj_str_t *uri_hostname;
+	char *domain_name;
+	char *username;
+	struct ast_str *id = NULL;
+
+	if (ast_strlen_zero(endpoint->aors)) {
+		return NULL;
+	}
+
+	uri_hostname = ast_sip_pjsip_uri_get_hostname(uri);
+	domain_name = ast_alloca(uri_hostname->slen + 1);
+	ast_copy_pj_str(domain_name, uri_hostname, uri_hostname->slen + 1);
+
+	uri_username = ast_sip_pjsip_uri_get_username(uri);
+	username = ast_alloca(uri_username->slen + 1);
+	ast_copy_pj_str(username, uri_username, uri_username->slen + 1);
+
+	/*
+	 * We may want to match without any user options getting
+	 * in the way.
+	 */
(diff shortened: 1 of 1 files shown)


IgorrG:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

gtjoseph:
Links to original issues if you're interested...
https://issues-archive.asterisk.org/ASTERISK-28211
https://issues-archive.asterisk.org/ASTERISK-30100

res/res_pjsip_path.c
```diff
@@ -48,6 +48,109 @@ static struct ast_sip_aor *find_aor(struct ast_sip_contact *contact)
...
+{
+	struct ao2_iterator it_contacts;
+	struct ast_sip_contact *contact;
+	const pj_str_t *uri_username;
+	const pj_str_t *uri_hostname;
+	char *domain_name;
+	char *username;
+	RAII_VAR(struct ao2_container *, contacts, NULL, ao2_cleanup);
+
+	uri_hostname = ast_sip_pjsip_uri_get_hostname(uri);
```
mbradeen, line 128: the uri is getting passed in and used to build these names, was that to compare against the contact when iterating?

mbradeen (changes requested):

res/res_pjsip_path.c
IgorrG: Thanks for pointing this, I have figured out yesterday that some old patch uploaded. I'll update patch, correct way is to compare exact URI, without host/username extraction.

res/res_pjsip_path.c
```diff
@@ -48,6 +48,105 @@ static struct ast_sip_aor *find_aor(struct ast_sip_contact *contact)
...
+	contact_buf[contact_buf_len] = '\0';
+
+	it_contacts = ao2_iterator_init(contacts, 0);
+	for (; (contact = ao2_iterator_next(&it_contacts)); ao2_ref(contact, -1)) {
+		if (!strcmp(contact_buf, contact->uri)) {
+			break;
+		}
+	}
+	ao2_iterator_destroy(&it_contacts);
+	return contact;
```
mbradeen, line 146: if no match is found, what should be returned?

mbradeen (changes requested):

res/res_pjsip_path.c
IgorrG: Got it, will fix

mbradeen (approved):

gtjoseph:
I had to rebase this PR on current master so it would pick up the alembic fix that just merged.  When all the checks finish, this PR should merge successfully.

Issue #740, opened 2024-05-21
[new-feature]: Add multi-tenant identifier to chan_pjsip


bkford:
### Feature Description

Add multi-tenant identifier to chan_pjsip for general purpose use.

alex2grad:
Why don't use '@' variable for PJSIP endpoint?
https://www.asterisk.org/did-you-know-you-can-use-variables-in-pjsip-conf/

For example

[myendpoint]
type=endpoint
@tenantid=MY_COMPANY

jcolp:
That doesn't put such information in events or other things.

alex2grad:
> That doesn't put such information in events or other things.

The Feature Description is too short to understand why it is needed and how to use it.
Nothing about "events or other things."

I do multi-tenant in my setup using PJSIP endpoint '@'variable and CALLERID(tag)/CONNECTEDLINE(tag)

BTW, I read the PR description and am confused.

PR #743, opened 2024-05-21
channel: Add multi-tenant identifier. cert-18.9

bkford:
This patch introduces a new identifier for channels: tenantid. It's
a stringfield on the channel that can be used for general purposes. It
will be inherited by other channels the same way that linkedid is.

You can set tenantid in a few ways. The first is to set it in the
dialplan with the Set and CHANNEL functions:

exten => example,1,Set(CHANNEL(tenantid)=My tenant ID)

It can also be accessed via CHANNEL:

exten => example,2,NoOp(CHANNEL(tenantid))

Another method is to use the new tenantid option for pjsip endpoints in
pjsip.conf:

[my_endpoint]
type=endpoint
tenantid=My tenant ID

This is considered the best approach since you will be able to see the
tenant ID as early as the Newchannel event.

It can also be set using set_var in pjsip.conf on the endpoint like
setting other channel variable:

set_var=CHANNEL(tenantid)=My tenant ID

Note that set_var will not show tenant ID on the Newchannel event,
however.

Tenant ID has also been added to CDR. It's read-only and can be accessed
via CDR(tenantid). You can also get the tenant ID of the last channel
communicated with via CDR(peertenantid).

Tenant ID will also show up in CEL records if it has been set, and the
version number has been bumped accordingly.

Fixes: https://github.com/asterisk/asterisk/issues/740

UserNote: tenantid has been added to channels. It can be read in
dialplan via CHANNEL(tenantid), and it can be set using
Set(CHANNEL(tenantid)=My tenant ID). In pjsip.conf, it is recommended to
use the new tenantid option for pjsip endpoints (e.g., tenantid=My
tenant ID) so that it will show up in Newchannel events. You can set it
like any other channel variable using set_var in pjsip.conf as well, but
note that this will NOT show up in Newchannel events. Tenant ID is also
available in CDR and can be accessed with CDR(tenantid). The peer tenant
ID can also be accessed with CDR(peertenantid). CEL includes tenant ID
as well if it has been set.

UpgradeNote: A new versioned struct (ast_channel_initializers) has been
added that gets passed to __ast_channel_alloc_ap. The new function
ast_channel_alloc_with_initializers should be used when creating
channels that require the use of this struct. Currently the only value
in the struct is for tenantid, but now more fields can be added to the
struct as necessary rather than the __ast_channel_alloc_ap function. A
new option (tenantid) has been added to endpoints in pjsip.conf as well.
CEL has had its version bumped to include tenant ID.

### DIFF
--- a/res/ari/ari_model_validators.c
+++ b/res/ari/ari_model_validators.c
@@ -1145,6 +1145,15 @@ int ast_ari_validate_channel(struct ast_json *json)
 				res = 0;
 			}
 		} else
+		if (strcmp("tenantid", ast_json_object_iter_key(iter)) == 0) {
+			int prop_is_valid;
+			prop_is_valid = ast_ari_validate_string(
+				ast_json_object_iter_value(iter));
+			if (!prop_is_valid) {
+				ast_log(LOG_ERROR, "ARI Channel field tenantid failed validation\n");
+				res = 0;
+			}
+		} else
 		{
 			ast_log(LOG_ERROR,
 				"ARI Channel has undocumented field %s\n",
--- a/cel/cel_manager.c
+++ b/cel/cel_manager.c
@@ -229,6 +229,7 @@ static void manager_log(struct ast_event *event)
 	struct ast_cel_event_record record = {
 		.version = AST_CEL_EVENT_RECORD_VERSION,
 	};
+	RAII_VAR(char *, tenant_id, NULL, ast_free);
 
 	if (!enablecel) {
 		return;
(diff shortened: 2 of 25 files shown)


bkford:
cherry-pick-to: none

jcolp (approved):

gtjoseph (changes requested):
Disapproved pending resolution of testsuite failures on the main PR

res/ari/ari_model_validators.c
```diff
@@ -1145,6 +1145,15 @@ int ast_ari_validate_channel(struct ast_json *json)
 				res = 0;
 			}
 		} else
+		if (strcmp("tenantid", ast_json_object_iter_key(iter)) == 0) {
```
gtjoseph, line 1148: ari_model_validators.c and .h are generated by running make ari_stubs after modifying files in /rest-api/api-docs.

gtjoseph (changes requested):

gtjoseph (approved):

jcolp (approved):

PR #745, opened 2024-05-23
res_pjsip_session: Channel variable SIPFROMUSER to setup From header URI user

wdoekes:
Ticket [ASTERISK-27309](https://issues-archive.asterisk.org/ASTERISK-27309) aimed to create feature parity between chan_sip and chan_pjsip so people can migrate and not lose features.

In [ASTERISK-28489](https://issues-archive.asterisk.org/ASTERISK-28489), the SIPFROMDOMAIN was added to pjsip (c7270dca8156194004e4cfcb7bbf5a9a09dbf717). This setting allows overriding the domain-part of the From header. However there was no way to override the user-part of the From header. This changeset adds a SIPFROMUSER to complement SIPFROMDOMAIN.

## Background:

chan_sip did provide for a way to dynamically set both the user and domain in the From and the user and domain in the To headers:
the SIP/ dialstring.

#### Syntax:

  `SIP/devicename/extension/IPorHost!to!from`

  *Where 'to' can be 'to-user' or 'to-user@to-domain' and 'from' can be 'from-user' or 'from-user@from-domain'.*

#### For PJSIP, we have this Dial string:

  `PJSIP/endpoint/sip:extension@IPorHost`

And one can statically set from_user and from_domain on the endpoint.

Setting the from_domain dynamically: `Set(_SIPFROMDOMAIN=from-domain)`.

Setting the from_user dynamically was only possible by altering `${CALLERID(num)}`, which has the unfortunate side-effect that it alters CDR/CEL records.

This changeset adds the possibility to dynamically set from_user using `Set(_SIPFROMUSER=from-user)`.

(As for feature parity with the SIP/ dial string, that leaves the missing to-user@to-domain setting. Luckily, that can be resolved by abusing the `maddr=` uri parameter. Instead of `PJSIP/${endpoint}/sip:${to_user}@${proxy}` one can do `PJSIP/${endpoint}/sip:${to_user}@${to_domain};maddr=${proxy}`.)

Change-Id: I4e48fe6341d49e9f90ea9d2479a2a58c6089da06

### DIFF
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -1657,6 +1657,7 @@ static void set_from_header(struct ast_sip_session *session)
 	pjsip_sip_uri *dlg_contact_uri;
 	int restricted;
 	const char *pjsip_from_domain;
+	const char *pjsip_from_user;
 
 	if (!session->channel || session->saved_from_hdr) {
 		return;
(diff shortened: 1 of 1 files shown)


jcolp (changes requested):
SIPFROMDOMAIN as a dialplan variable had precedent and prior usage from chan_sip. Its usage also predates the existence of dialplan functions. Going forward special dialplan variables should not be added unless absolutely necessary. Instead the manipulation should be done through a dialplan function, such as extending the CHANNEL dialplan function to allow setting the from user and updating it.

Additionally a UserNote should be added to the commit message so that the addition appears in the release notes and change log explicitly.

wdoekes:
Thanks! I'll look into that :+1:

asteriskteam:

asteriskteam:

Issue #765, opened 2024-06-12
[improvement]: Add option to not log Restricted Caller ID to queue_log


alex2grad:
### Improvement Description

Add an option to strip the Caller ID when storing the ENTERQUEUE event in the queue log if the Caller ID is restricted.

PR #766, opened 2024-06-12
app_queue:  Add option to not log Restricted Caller ID to queue_log

alex2grad:
Add a queue option log-restricted-caller-id to strip the Caller ID when storing the ENTERQUEUE event
in the queue log if the Caller ID is restricted.

Resolves: #765

UpgradeNote: Add a new column to the queues table:
queue_log_option_log_restricted ENUM('0','1','off','on','false','true','no','yes')
to control whether the Restricted Caller ID will be stored in the queue log.

UserNote: Add a Queue option log-restricted-caller-id to control whether the Restricted Caller ID
will be stored in the queue log.
If log-restricted-caller-id=no then the Caller ID will be stripped if the Caller ID is restricted.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -1882,6 +1882,8 @@ struct call_queue {
 	int memberdelay;                    /*!< Seconds to delay connecting member to caller */
 	int autofill;                       /*!< Ignore the head call status and ring an available agent */
 
+	int log_restricted_caller_id:1;     /*!< Whether log Restricted Caller ID */
+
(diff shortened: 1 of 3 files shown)


alex2grad:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

alex2grad:
strange things with workflow... 13 successful and 1 expected checks
but label still 'pr-submit-tests-failed'

jcolp (approved):

Issue #770, opened 2024-06-13
[improvement]: pbx.c: expand fields width of "core show hints"


alex2grad:
### Improvement Description

The current width for \<extension\> is 20 and \<device state id\> is 20, which is too small.
The \<extension\> field contains \<ext\>@\<context\>, so 20 characters is not enough.
The \<device state id\> field, for example for Queue pause state contains Queue:\<queue_name\>_pause_PSJIP/\<endpoint\>, so the 20 characters is not enough.

Increase the width of \<extension\> field to 30 characters and the width of the \<device state id\> field to 60 characters.

PR #771, opened 2024-06-13
pbx.c: expand fields width of "core show hints"

alex2grad:
The current width for "extension" is 20 and "device state id" is 20, which is too small.
The "extension" field contains "ext"@"context", so 20 characters is not enough.
The "device state id" field, for example for Queue pause state contains Queue:"queue_name"_pause_PSJIP/"endpoint", so the 20 characters is not enough.

Increase the width of "extension" field to 30 characters and the width of the "device state id" field to 60 characters.

Resolves: https://github.com/asterisk/asterisk/issues/770

UserNote: The fields width of "core show hints" were increased.
The width of "extension" field to 30 characters and
the width of the "device state id" field to 60 characters.

### DIFF
--- a/main/pbx.c
+++ b/main/pbx.c
@@ -5198,7 +5198,7 @@ static char *handle_show_hints(struct ast_cli_entry *e, int cmd, struct ast_cli_
 			ast_get_extension_name(hint->exten),
 			ast_get_context_name(ast_get_extension_context(hint->exten)));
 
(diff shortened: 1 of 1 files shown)


alex2grad:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

jcolp (approved):

PR #773, opened 2024-06-14
bridge_basic.c: Make sure that ast_bridge_channel is not destroyed while iterating over bridge->channels.

Tinet-mucw:
bridge_basic.c: Make sure that ast_bridge_channel is not destroyed while iterating over bridge->channels.

From the gdb information, we can see that while iterating over bridge->channels, the ast_bridge_channel reference count is 0, indicating it has already been destroyed.Additionally, when ast_bridge_channel is removed from bridge->channels, the bridge is first locked. Therefore, locking the bridge before iterating over bridge->channels can resolve the race condition.


Tinet-mucw:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #776, opened 2024-06-15
[bug] DTMF broken after rtp_engine: add support for multirate RFC2833 digits commit


iueoa:
### Severity

Blocker

### Versions

624f509ce4c74df852269d6eddd93de678ec4d1b

### Components/Modules

rtp_engine

### Operating Environment

Ubuntu 22.04, Asterisk self-compiled from 21 branch in git

### Frequency of Occurrence

None

### Issue Description

Following the 624f509ce4c74df852269d6eddd93de678ec4d1b commit, DTMF is broken. Reverting to the previous commit and everything works.

### Relevant log output

```shell
Sample "Good" DTMF-working 200 OK SDP:

v=0
o=- 1718461316199841 3 IN IP4 34.x.x.x
s=Asterisk         
c=IN IP4 34.x.x.x
t=0 0
m=audio 10000 RTP/SAVP 120 101
a=crypto:1 AEAD_AES_128_GCM inline:DUx1dJS7jQzAFfUlAxiadjqmXIul5oKQfkrglA==
a=rtpmap:120 opus/48000/2
a=fmtp:120 maxaveragebitrate=24000;useinbandfec=1;usedtx=1
a=rtpmap:101 telephone-event/8000
a=fmtp:101 0-16
a=ptime:20
a=maxptime:20
a=sendrecv         


Sample "BAD" DTMF broken 200 OK SDP:

v=0
o=- 1718460531774378 3 IN IP4 34.x.x.x
s=Asterisk
c=IN IP4 34.x.x.x
t=0 0              
m=audio 10018 RTP/SAVP 120
a=crypto:1 AEAD_AES_128_GCM inline:lExhOEA86fMbVq5XODV4wtocHc7DcgFQ/IiOAQ==
a=rtpmap:120 opus/48000/2
a=fmtp:120 maxaveragebitrate=24000;useinbandfec=1;usedtx=1
a=ptime:20
a=maxptime:20
a=sendrecv         


/etc/asterisk/codecs.conf
[opusmed]           
type=opus           
packet_loss=20
fec=yes
max_playback_rate=24000
bitrate=24000
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

seanbright:
Assuming DTMF is actually broken, how is this "Minor?"

jcolp:
@iueoa What is the incoming INVITE request?

@seanbright I've gone ahead and edited that to Blocker for now.

[...]

PR #779, opened 2024-06-21
res_pjsip_sdp_rtp: Add support for default/mismatched 8K RFC 4733/2833 digits

mbradeen:
After change made in 624f509 to add support for non 8K RFC 4733/2833 digits,
Asterisk would only accept RFC 4733/2833 offers that matched the sample rate of
the negotiated codec(s).

This change allows Asterisk to accept 8K RFC 4733/2833 offers if the UAC
offfers 8K RFC 4733/2833 but negotiates for a non 8K bitrate codec.

A number of corresponding tests in tests/channels/pjsip/dtmf_sdp also needed to
be re-written to allow for these scenarios.

Fixes: #776

### DIFF
--- a/include/asterisk/rtp_engine.h
+++ b/include/asterisk/rtp_engine.h
@@ -106,6 +106,9 @@ extern "C" {
  */
 #define MAX_CHANNEL_ID 152
 
+/*!< DTMF samples per second */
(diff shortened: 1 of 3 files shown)


mbradeen:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #783, opened 2024-06-26
[bug]: Under certain circumstances a channel snapshot can get orphaned in the cache


gtjoseph:
### Severity

Minor

### Versions

18,20,21,master

### Components/Modules

stasis_channels.c

### Operating Environment

Any

### Frequency of Occurrence

Occasional

### Issue Description

Whenever a channel snapshot is create or when a channel is being destroyed, any existing channel snapshot has to be deleted from the snapshot cache.  It appears there is some, as yet unknown, circumstance where the existing snapshot isn't being deleted.  This is mostly benign to the call actually being processed but since all listing of channels from the CLI, AMI and ARI use the snapshot cache, it can appear that a channel still exists when in fact, it's gone.  This can cause issues for AMI and ARI applications and looks odd from the CLI.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #784, opened 2024-06-27
stasis_channels: Use uniqueid and name to delete old snapshots

gtjoseph:
Whenver a new channel snapshot is created or when a channel is
destroyed, we need to delete any existing channel snapshot from
the snapshot cache.  Historically, we used the channel->snapshot
pointer to delete any existing snapshots but this has two issues.

First, if something (possibly ast_channel_internal_swap_snapshots)
sets channel->snapshot to NULL while there's still a snapshot in
the cache, we wouldn't be able to delete it and it would be orphaned
when the channel is destroyed.  Since we use the cache to list
channels from the CLI, AMI and ARI, it would appear as though the
channel was still there when it wasn't.

Second, since there are actually two caches, one indexed by the
channel's uniqueid, and another indexed by the channel's name,
deleting from the caches by pointer requires a sequential search of
all of the hash table buckets in BOTH caches to find the matching
snapshots.  Not very efficient.

So, we now delete from the caches using the channel's uniqueid
and name.  This solves both issues.

This doesn't address how channel->snapshot might have been set
to NULL in the first place because although we have concrete
evidence that it's happening, we haven't been able to reproduce it.

Resolves: #783

### DIFF
--- a/main/stasis_channels.c
+++ b/main/stasis_channels.c
@@ -942,8 +942,8 @@ void ast_channel_publish_final_snapshot(struct ast_channel *chan)
 		return;
 	}
 
-	ao2_unlink(channel_cache, update->old_snapshot);
-	ao2_unlink(channel_cache_by_name, update->old_snapshot);
+	ao2_find(channel_cache, ast_channel_uniqueid(chan), OBJ_SEARCH_KEY | OBJ_UNLINK | OBJ_NODATA);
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #789, opened 2024-07-01
[bug]: Mediasec headers aren't sent on outgoing INVITEs


gtjoseph:
### Severity

Minor

### Versions

18,20,21,master

### Components/Modules

res_pjsip_rfc3329

### Operating Environment

any

### Frequency of Occurrence

None

### Issue Description

Security-Verify and associated Requires and Proxy-Requires headers aren't being sent in outgoing INVITE requests although they are working on outgoing REGISTER requests.  It appears that commit cacd98bb29ae8ca17 changed the process to only send those headers if a 401 response is received with Security-Server headers but 401 responses are handled early and automatically and not sent to session handlers like res_pjsip_rfc3329.

@maximilianfridrich can you comment on this?  Can you also comment on whether DTAG still requires mediasec?

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

maximilianfridrich:
I will look into it. It seems to be working fine in our setups.

I don't have much information except what they make publicly available on their websites, but according to their technical documentation they still need for at least some of their trunks (CompanyFlex). And we use it for that.

gtjoseph:
OK, I figured it out.  We have to have received Security-Server headers on a 200-OK response to a previous REGISTER before we can send Security-Verify headers on an outgoing INVITE.  Since I was testing against another asterisk instance and since we don't implement the server side of the mediasec negotiation, I wasn't getting the Security-Server headers back.   When I mocked up getting those headers in res-pjsip_outbound_registration/handle_registration_response, I was seeing the Security-Verify headers in the outgoing INVITEs.   HOWEVER, there is a bug in res_pjsip_rfc3329/add_outgoing_request_headers where although we're adding the Security-Verify headers, we're not adding the "Require: mediasec" or "Proxy-Require: mediasec" headers as we do for outgoing REGISTERs.  I'll be fixing this shortly.

maximilianfridrich:
Thank you! @gtjoseph

PR #791, opened 2024-07-03
security_agreement.c: Always add the Require and Proxy-Require headers

gtjoseph:
The `Require: mediasec` and `Proxy-Require: mediasec` headers need
to be sent whenever we send `Security-Client` or `Security-Verify`
headers but the logic to do that was only in add_security_headers()
in res_pjsip_outbound_register.  So while we were sending them on
REGISTER requests, we weren't sending them on INVITE requests.

This commit moves the logic to send the two headers out of
res_pjsip_outbound_register:add_security_headers() and into
security_agreement:ast_sip_add_security_headers().  This way
they're always sent when we send `Security-Client` or
`Security-Verify`.

Resolves: #789

### DIFF
--- a/res/res_pjsip/security_agreements.c
+++ b/res/res_pjsip/security_agreements.c
@@ -290,6 +290,8 @@ int ast_sip_add_security_headers(struct ast_sip_security_mechanism_vector *secur
 	int mech_cnt;
 	int i;
 	int add_qvalue = 1;
+	static const pj_str_t proxy_require = { "Proxy-Require", 13 };
+	static const pj_str_t require = { "Require", 7 };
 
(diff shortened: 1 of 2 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

gtjoseph:
I still need to write a testsuite test that does a REGISTER and an INVITE to ensure all the headers are set correctly.

jcolp (approved):
While approving, leaving a note that this does require testsuite updating.

asteriskteam:
Successfully merged to branch master and cherry-picked to ["18","20","21","certified/20.7"]

Issue #780, opened 2024-06-25
[bug]: Infinite loop of "Indicated Video Update", max CPU usage


duncannz:
### Severity

Major

### Versions

20.6.0,20.8.1

### Components/Modules

Local,PJSIP

### Operating Environment

Debian 11.7

### Frequency of Occurrence

Constant

### Issue Description

I am experiencing:

* Extremely high / maxed out CPU usage
* Log messages such as `Exceptionally long queue length (0 voice / 150 total) queuing to Recorder/ARI-00000001;1` are logged repeatedly for as long as the bridge exists
* If I turn on debug logging, the messages
  ```
  [2024-06-25 01:14:04] DEBUG[3144][C-00000003]: chan_pjsip.c:1644 chan_pjsip_indicate:  PJSIP/webrtc_client-00000001: Indicated Video Update
  [2024-06-25 01:14:04] DEBUG[3144][C-00000003]: chan_pjsip.c:1875 chan_pjsip_indicate:  PJSIP/webrtc_client-00000001
  ```
  are logged repeatedly at an extremely high rate for as long as the bridges exist

Bridge 1:

```
Id: 01J16F9GK7X976AJQ6F596SKN6
Type: stasis
Technology: softmix
Subclass: stasis
Creator: Stasis
Name:
Video-Mode: talker
Video-Source-Id:
Num-Channels: 3
Num-Active: 3
Duration: 00:00:30
Channel: PJSIP/webrtc_client-00000000
Channel: Recorder/ARI-00000001;2
Channel: Local/+XXXXX@reflector-00000000;1
```

Bridge 2:

```
Id: 01J16F9GT51PEHR8F66WGVE7D6
Type: stasis
Technology: softmix
Subclass: stasis
Creator: Stasis
Name:
Video-Mode: talker
Video-Source-Id:
Num-Channels: 3
Num-Active: 3
Duration: 00:00:38
Channel: PJSIP/webrtc_client-00000001
Channel: Recorder/ARI-00000002;2
Channel: Local/+XXXXX@reflector-00000000;2
```

This issue happens on 100% of calls that have the above configuration (one PJSIP channel in each bridge, one local channel connecting the two bridges, and each bridge being recorded).

The only other thing I could find on Google for "Indicated Video Update" is [this post](https://asteriskfaqs.org/2021/05/05/asterisk-users/cpu-spike.html), which describes the exact same issue, but in Asterisk 16.17.0, so I suspect this issue has been around for some time.

[...]

PR #810, opened 2024-07-17
bridge_softmix: Fix queueing VIDUPDATE control frames

gtjoseph:
softmix_bridge_write_control() now calls ast_bridge_queue_everyone_else()
with the bridge_channel so the VIDUPDATE control frame isn't echoed back.

softmix_bridge_write_control() was setting bridge_channel to NULL
when calling ast_bridge_queue_everyone_else() for VIDUPDATE control
frames.  This was causing the frame to be echoed back to the
channel it came from.  In certain cases, like when two channels or
bridges are being recorded, this can cause a ping-pong effect that
floods the system with VIDUPDATE control frames.

Resolves: #780


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #811, opened 2024-07-17
[new-feature]: ARI channel tone detect events.


gibbz00:
### Feature Description

At work we need to listen to a 1400Hz special information tone for conference call detection as specified in [Swedish tones in public communications network](https://its.se/wp-content/uploads/its-apg-31-utgava-1-faststalld-2023-11-09.pdf).

We would ideally like to register tone detection event listening on a given channel, for it to then be ready when `/ari/bridges/<id>/addChannel` is requested.

I read some [AstriDevCon notes](https://docs.asterisk.org/Development/Roadmap/AstriDevCon-2015/#sean-mccord) dating all the way back to 2015 where such a feature is mentioned as a nice to have. Has anyone looked into it since then? Feels very much like a non-trivial task given that 9 years have now gone by. 

I would be more than happy to send in a PR for this, but I'm afraid all of this is a bit over my head given that both C and Asterisk are completely new to me. I've begun looking at the source code and here's what I've deduced so far:

* Sending events over WS from subscribed stasis messages seems pretty straight forward. 

* Register event listening by translating an HTTP request into a `res_tonetetect::detect_write` call might be doable.

* But how should events be created?  In `channel.c`'s `__ast_read` by somehow calling `res_tonedetect::detect_read`?

Am I on the right path, or is another way recommended? Feed-back is greatly appreciated.

jcolp:
The issue tracker is really for issues, not for discussions such as this. In the future please use the asterisk-dev mailing list: https://groups.io/g/asterisk-dev

The res_tonedetect module already shows how to passively listen for tones, as it does it in the TONE_DETECT dialplan function. This function uses an audiohook which allows the dialplan function implementation to receive the audio from the channel, it then passes it to the DSP to do analysis, and acts on it. The only difference is that in your case you would want to raise an event. This is done by publishing a stasis message on the channel, with a type that has a converter into JSON. An example of this that exists already would be TALK_DETECT. No ARI modification was needed to support that because it is a dialplan function that can be placed on the channel using the /variable route of the channel. It then passively detects when speaking is happening, and raises an event over ARI.

gibbz00:
Hii, thanks for the quick response.

> The issue tracker is really for issues, not for discussions such as this. In the future please use the asterisk-dev mailing list: https://groups.io/g/asterisk-dev

Apologies. Now I know for next time.

I was aware of TALK_DETECT but not the `/variable` endpoint. Helped tremendously, thank you!!

PR #814, opened 2024-07-18
feat: ARI "ChannelToneDetected" event

gibbz00:
Fixes: #811

Refrained from adding any JSON blob to the message. Perhaps it should be part of a separate PR if anyone wants to build on this later on.

### DIFF
--- a/include/asterisk/stasis_channels.h
+++ b/include/asterisk/stasis_channels.h
@@ -657,6 +657,14 @@ struct stasis_message_type *ast_channel_talking_start(void);
 struct stasis_message_type *ast_channel_talking_stop(void);
 
 /*!
+ * \since 22.0.0
+ * \brief Message type for a channel tone detection
+ *
+ * \return A stasis message type
+ */
+struct stasis_message_type *ast_channel_tone_detect(void);
+
+/*!
  * \since 12
  * \brief Publish in the \ref ast_channel_topic or \ref ast_channel_topic_all
  * topics a stasis message for the channels involved in a dial operation.
--- a/res/res_tonedetect.c
+++ b/res/res_tonedetect.c
@@ -39,6 +39,7 @@
 #include "asterisk/channel.h"
 #include "asterisk/dsp.h"
 #include "asterisk/pbx.h"
+#include "asterisk/stasis_channels.h"
 #include "asterisk/audiohook.h"
 #include "asterisk/app.h"
 #include "asterisk/indications.h"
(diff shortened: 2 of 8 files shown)


gibbz00:
cherry-pick-to: none

include/asterisk/stasis_channels.h
```diff
@@ -656,6 +656,14 @@ struct stasis_message_type *ast_channel_talking_start(void);
  */
 struct stasis_message_type *ast_channel_talking_stop(void);
 
+/*!
+ * \since 21.3.2
```
jcolp, line 660: This is no longer correct, and this change is not commented to be cherry picked against 21 anyway

jcolp (changes requested):
This new JSON event needs to be documented in rest-api/api-docs/events.json and "make ari-stubs" re-run.

The commit message also needs to e expanded and a UserNote added about this addition.

include/asterisk/stasis_channels.h
gibbz00: `\since 22` now, is that ok?

gibbz00:
> This new JSON event needs to be documented in rest-api/api-docs/events.json and "make ari-stubs" re-run.

Done, thanks.

> The commit message also needs to e expanded and a UserNote added about this addition.

Updated the commit message, not 100% how much you want to be included.

jcolp (changes requested):
The commit message is rather technical and internal. The following is high level enough and easy for users to understand in the case of UserNote:

feat: ARI "ChannelToneDetected" event

A stasis event is now produced when using the TONE_DETECT dialplan
function. This event is published over ARI using the ChannelToneDetected
event. This change does not make it available over AMI.

Fixes: asterisk#811

UserNote: Setting the TONE_DETECT dialplan function on a channel
in ARI will now cause a ChannelToneDetected ARI event to be raised
when the specified tone is detected.

gibbz00:
Updated, thank you

res/res_tonedetect.c
```diff
@@ -394,6 +396,16 @@ static int detect_callback(struct ast_audiohook *audiohook, struct ast_channel *
 			}
 			ast_debug(1, "TONE_DETECT just got a hit (#%d in this direction, waiting for %d total)\n", now, di->hitsrequired);
 			if (now >= di->hitsrequired) {
+				message = ast_channel_blob_create_from_cache(ast_channel_uniqueid(chan), ast_channel_tone_detect(), NULL);
+
+				if (!message) {
+					ast_log(LOG_ERROR, "unable to create ast_channel_tone_detect message");
```
jcolp, line 402: This should include the channel name.

"Unable to publish tone detected event for ARI on channel '%s'"

jcolp (changes requested):

res/res_tonedetect.c
gibbz00: Yeah, fixed :)

jcolp (approved):

gtjoseph:
testsuite-test-pr: 54

Issue #809, opened 2024-07-17
[bug]: CLI stir_shaken show verification kills asterisk


oza4h07:
### Severity

Major

### Versions

20.8.1

### Components/Modules

PJSIP

### Operating Environment

Debian Bullseye

### Frequency of Occurrence

Constant

### Issue Description

CLI "stir_shaken show verification" kills asterisk.
stir_shaken.conf file is set with specific data (that privately) be shared if necessary.

[backtrace_files.tar.gz](https://github.com/user-attachments/files/16270581/backtrace_files.tar.gz)

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
I see what's happening.  Thanks for the report.

oza4h07:
PR #815 seems to solve this issue

PR #815, opened 2024-07-21
stir_shaken: CRL fixes and a new CLI command

gtjoseph:
* Fixed a bug in crypto_show_cli_store that was causing asterisk
to crash if there were certificate revocation lists in the
verification certificate store.  We're also now prefixing
certificates with "Cert:" and CRLs with "CRL:" to distinguish them
in the list.

* Added 'untrusted_cert_file' and 'untrusted_cert_path' options
to both verification and profile objects.  If you have CRLs that
are signed by a different CA than the incoming X5U certificate
(indirect CRL), you'll need to provide the certificate of the
CRL signer here.  Thse will show up as 'Untrusted" when showing
the verification or profile objects.

* Fixed loading of crl_path.  The OpenSSL API we were using to
load CRLs won't actually load them from a directory, only a file.
We now scan the directory ourselves and load the files one-by-one.

* Fixed the verification flags being set on the certificate store.
  - Removed the CRL_CHECK_ALL flag as this was causing all certificates
    to be checked for CRL extensions and failing to verify the cert if
    there was none.  This basically caused all certs to fail when a CRL
    was provided via crl_file or crl_path.
  - Added the EXTENDED_CRL_SUPPORT flag as it is required to handle
    indirect CRLs.

* Added a new CLI command...
`stir_shaken verify certificate_file <certificate_file> [ <profile> ]`
which will assist troubleshooting certificate problems by allowing
the user to manually verify a certificate file against either the
global verification certificate store or the store for a specific
profile.

* Updated the XML documentation and the sample config file.

Resolves: #809


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

oza4h07:
I'm not familiar with either Githut or Git in general.

On a Bookworm host, I've got:

cd /usr/src
wget -O asterisk-pr-809.tar.gz https://api.github.com/repos/asterisk/asterisk/tarball/pull/815/head
tar xzf asterisk-pr-809.tar.gz
cd asterisk-asterisk-c31f1c6
./configure --with-libjwt-bundled
make menuselect
make
make install
rasterisk

In /etc/asterisk, there is no strir_shaken.conf file and autoload in set to no in modules.conf


> module load res_stir_shaken.so
Unable to load module res_stir_shaken.so
Command 'module load res_stir_shaken.so' failed.
[Jul 22 15:52:12] ERROR[11726]: loader.c:283 module_load_error: res_stir_shaken has one or more unknown dependencies.

IMHO, "res_stir_shaken has one or more unknown dependencies" is mis-leading  as make menuselect doesn't show any missing dependency.
Anyway, I'll leave it up to more informed people to decide if the current report is worth to be looked at.

gtjoseph:
> In /etc/asterisk, there is no strir_shaken.conf file and autoload in set to no in modules.conf

Well you won't get far that way.  You need the same environment and config you were using when you reported the issues.   If you have a good asterisk source tree, you can just download this PR as a patchfile from https://github.com/asterisk/asterisk/pull/815.patch and apply it.  Then rebuild and install as you normally would.

oza4h07:
As reported in Community Forum, we could check incoming calls didn't use a revoked cert, thanks to this PR.
"stir_shaken" CLI now works as expected.

jcolp (approved):

Issue #816, opened 2024-07-23
[bug]: res_pjsip_config_wizard doesn't load properly if res_pjsip is loaded first


gtjoseph:
### Severity

Trivial

### Versions

18,20,21,master

### Components/Modules

res_pjsip_config_wizard

### Operating Environment

all

### Frequency of Occurrence

None

### Issue Description

Normal file system sort order usually loads res_pjsip_config_wizard before res_pjsip but if the order is overridden in modules.conf and res_pjsip gets loaded first, res_pjsip_config_wizard will load fine but not register any of its objects.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #817, opened 2024-07-23
res_pjsip_config_wizard.c: Refactor load process

gtjoseph:
The way we have been initializing the config wizard prevented it
from registering its objects if res_pjsip happened to load
before it.

* We now use the object_type_registered sorcery observer to kick
things off instead of the wizard_mapped observer.

* The load_module function now checks if res_pjsip has been loaded
already and if it was it fires the proper observers so the objects
load correctly.

Resolves: #816

UserNote: The res_pjsip_config_wizard.so module can now be reloaded.


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

bkford (approved):

jcolp (approved):

Issue #822, opened 2024-07-25
[bug]: segfault in main/rtp_engine.c:1489 after updating 20.8.1 -> 20.9.0


sircubbi:
### Severity

Major

### Versions

20.9.0

### Components/Modules

rtp_engine

### Operating Environment

RockyLinux 9

### Frequency of Occurrence

Frequent

### Issue Description

After upgrading from 20.8.1 to 20.9.0 we encounter segfaults in rtp_engine.c and had to rollback to 20.8.1. I saw three segfaults so far within a 20 minute timeframe after a constant load of around 100-150 parallel calls.

It seems related to commit 3b4d2a65064eb7a49d1d765a80e8ffa602532d0a (Issue #699). Unfortunately I cannot provide a backtrace, since no core dump was captured. However I am able to pinpoint the segfault to line 1489 in `rtp_engine.c`, so hopefully that gives you a clue on where to look.

I am happy to provide more details if needed of course.

### Relevant log output

```shell
Kernel-Logfile:

Jul 25 10:13:10 sip kernel: asterisk[4004920]: segfault at 0 ip 0000000000535a91 sp 00007f857a188c40 error 4 in asterisk[438000+1d1000] likely on CPU 1 (core 0, socket 1)
Jul 25 10:13:32 sip kernel: asterisk[47139]: segfault at 0 ip 0000000000535a91 sp 00007f56243fcc40 error 4 in asterisk[438000+1d1000] likely on CPU 3 (core 0, socket 3)
Jul 25 10:31:17 sip kernel: asterisk[47329]: segfault at 0 ip 0000000000535a91 sp 00007f75e8c38c40 error 4 in asterisk[438000+1d1000] likely on CPU 3 (core 0, socket 3)
```

Determine the code in question:
```
# addr2line -e asterisk 0000000000535a91
/root/asterisk-src/asterisk-20.9.0/main/rtp_engine.c:1489
#
```

Diff between 20.8.1 and 20.9.0 for that code:
```
@@ -1457,6 +1485,10 @@
 
 	if (payload < AST_VECTOR_SIZE(&codecs->payload_mapping_tx)) {
 		type = AST_VECTOR_GET(&codecs->payload_mapping_tx, payload);
+		/* remove the preferred format if we are unsetting its container. */
+		if (ast_format_cmp(type->format, codecs->preferred_format) == AST_FORMAT_CMP_EQUAL) {
+			ao2_replace(codecs->preferred_format, NULL);
+		}
[...]

PR #823, opened 2024-07-25
rtp_engine.c: Prevent segfault in ast_rtp_codecs_payloads_unset()

gtjoseph:
There can be empty slots in payload_mapping_tx corresponding to
dynamic payload types that haven't been seen before so we now
check for NULL before attempting to use 'type' in the call to
ast_format_cmp.

Note: Currently only chan_sip calls ast_rtp_codecs_payloads_unset()

Resolves: #822

### DIFF
--- a/main/rtp_engine.c
+++ b/main/rtp_engine.c
@@ -1485,12 +1485,21 @@ void ast_rtp_codecs_payloads_unset(struct ast_rtp_codecs *codecs, struct ast_rtp
 
 	if (payload < AST_VECTOR_SIZE(&codecs->payload_mapping_tx)) {
 		type = AST_VECTOR_GET(&codecs->payload_mapping_tx, payload);
-		/* remove the preferred format if we are unsetting its container. */
-		if (ast_format_cmp(type->format, codecs->preferred_format) == AST_FORMAT_CMP_EQUAL) {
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

main/rtp_engine.c
```diff
@@ -1485,12 +1485,21 @@ void ast_rtp_codecs_payloads_unset(struct ast_rtp_codecs *codecs, struct ast_rtp
...
+		 * There can be empty slots in payload_mapping_tx corresponding to
+		 * dynamic payload types that haven't been seen before so we need
+		 * to check for NULL before attempting to use 'type' in the call to
+		 * ast_format_cmp.
+		 */
+		if (type) {
+			if (ast_format_cmp(type->format, codecs->preferred_format) == AST_FORMAT_CMP_EQUAL) {
+				ao2_replace(codecs->preferred_format, NULL);
+			}
+			ao2_cleanup(type);
```
jcolp, line 1500: No need to do another if (type) check in ao2_cleanup,

ao2_ref(type, -1);

jcolp (changes requested):

jcolp (approved):

PR #827, opened 2024-08-02
res_pjsip_sdp_rtp.c: Fix DTMF Handling in Re-INVITE with dtmf_mode Set to Auto

Tinet-mucw:
res_pjsip_sdp_rtp.c: Fix DTMF Handling in Re-INVITE with dtmf_mode set to Auto

When the endpoint dtmf_mode is set to auto, a SIP request is sent to the UAC, and the SIP SDP from the UAC does not include the telephone-event. Later, the UAC sends an INVITE, and the SIP SDP includes the telephone-event. In this case, DTMF should be sent by RFC2833 rather than using inband signaling.

Resolves: asterisk#826

### DIFF
--- a/res/res_pjsip_sdp_rtp.c
+++ b/res/res_pjsip_sdp_rtp.c
@@ -384,9 +384,14 @@ static void get_codecs(struct ast_sip_session *session, const struct pjmedia_sdp
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Tinet-mucw:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

PR #830, opened 2024-08-05
configure:  Use . file rather than source file.

jkroonza:
source is a bash concept, so when /bin/sh points to another shell the existing construct won't work.

Reference: https://bugs.gentoo.org/927055

### DIFF
--- a/configure
+++ b/configure
@@ -15774,7 +15774,7 @@ else
 	PBX_JANSSON=1
 fi
 
-source ./third-party/versions.mak
+. ./third-party/versions.mak
 # Find required JWT support if bundled is not enabled.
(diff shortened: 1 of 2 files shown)


jkroonza:
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

gtjoseph (approved):

jcolp (approved):

jkroonza:
rest_api/applications/subscribe-endpoint/nominal/resource        [  16 of   96][Failed][  15][   1]

I do not see how this change can impact on that test?

seanbright:
> I do not see how this change can impact on that test?

It is unrelated to your change. Don't worry about it.

PR #834, opened 2024-08-08
manager.c: Add entries to Originate blacklist

gtjoseph:
Added Reload and DBdeltree to the list of dialplan application that
can't be executed via the Originate manager action without also
having write SYSTEM permissions.

Added CURL, DB*, FILE, ODBC and REALTIME* to the list of dialplan
functions that can't be executed via the Originate manager action
without also having write SYSTEM permissions.

If the Queue application is attempted to be run by the Originate
manager action and an AGI parameter is specified in the app data,
it'll be rejected unless the manager user has either the AGI or
SYSTEM permissions.

Resolves: #GHSA-c4cg-9275-6w44

### DIFF
--- a/main/manager.c
+++ b/main/manager.c
@@ -6325,6 +6325,145 @@ aocmessage_cleanup:
 	return 0;
 }
 
+struct originate_permissions_entry {
+	const char *search;
+	int permission;
+	int (*searchfn)(const char *app, const char *data, const char *search);
+};
+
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21

jcolp (approved):

Issue #847, opened 2024-08-21
[bug]: Asterisk not using negotiated fall-back 8K digits


mbradeen:
### Severity

Trivial

### Versions

18, 20, 21, 22, master

### Components/Modules

pjsip

### Operating Environment

all versions

### Frequency of Occurrence

None

### Issue Description

Asterisk sends Bob an INVITE with opus and ulaw with matching 48K and 8K telephone-event lines

Bob chooses opus but with the 8K telephone-event for digits.  Asterisk still sends 48K digits even though this was not negotiated for.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #848, opened 2024-08-21
res_pjsip_sdp_rtp: Use negotiated DTMF Payload types on bitrate mismatch

mbradeen:
When Asterisk sends an offer to Bob that includes 48K and 8K codecs with
matching 4733 offers, Bob may want to use the 48K audio codec but can not
accept 48K digits and so negotiates for a mixed set.

Asterisk will now check Bob's offer to make sure Bob has indicated this is
acceptible and if not, will use Bob's preference.

Fixes: #847

### DIFF
--- a/include/asterisk/rtp_engine.h
+++ b/include/asterisk/rtp_engine.h
@@ -764,10 +764,15 @@ struct ast_rtp_codecs {
 	unsigned int framing;
 	/*! The preferred format, as the mappings are numerically sorted */
 	struct ast_format *preferred_format;
+	/*! The preferred dtmf sample rate */
+	int preferred_dtmf_rate;
+	/*! The preferred dtmf payload type */
+	int preferred_dtmf_pt;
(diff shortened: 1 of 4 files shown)


mbradeen:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

testsuite-test-pr: 66

gtjoseph (changes requested):
There was a coredump during the unit tests and the pjs4 gate test.  Both are asserts in ast_rtp_codecs_payload_code_tx_sample_rate().

You can download the artifacts from those two tests from https://github.com/asterisk/asterisk/actions/runs/10493519086?pr=848

The backtraces are in them.

gtjoseph (approved):
It's not crashing any more.  That's as much as I'll approve. :)

jcolp (approved):

Issue #849, opened 2024-08-21
[improvement]: tighter systemd asterisk.service file


chrsmj:
### Improvement Description

Forthcoming patch implements more things from the SANDBOXING section of the systemd.exec manual page in order to help improve security. Many distros offer their own versions, but keeping a modern upstream version is helpful for end users.

jacky-w-j-li:
Hi,

I'd like to work on this as a first-time contributor. The plan is to harden `contrib/systemd/asterisk.service` using additional options from the SANDBOXING section of systemd.exec (e.g. `ProtectSystem=`, `ProtectHome=`, `PrivateTmp=`, `NoNewPrivileges=`, `RestrictAddressFamilies=`) while keeping it functional for typical installations.

Before I open a PR, could you confirm:
1. Is this still wanted and not already being worked on?
2. Are there specific sandboxing options you'd prefer to include or avoid (e.g. anything that might break DAHDI/hardware access or module loading)?

I'll keep the change minimal and scoped, follow the coding guidelines, and verify the unit still starts Asterisk cleanly.

Thanks!

PR #850, opened 2024-08-21
asterisk.service: add fresh sand to the box

chrsmj:
Adds ProtectHome, ProtectSystem, and ReadWritePaths -- with reasonable defaults -- to prevent Asterisk process access to /home, /root, and /run/user directories; and mount the rest of the file system as read-only (with a few minor exceptions.)

Uncomments RuntimeDirectory to comport with new sandboxing.

Also adds Documentation, and some LimitNOFILE and Nice ideas.

Resolves: #849

UserNote: Please review the updated SystemD service file to consider some of the security improvements that restrict the ability of the Asterisk process to access the file system.

### DIFF
--- a/contrib/systemd/asterisk.service
+++ b/contrib/systemd/asterisk.service
@@ -1,5 +1,6 @@
 [Unit]
 Description=Asterisk PBX and telephony daemon.
+Documentation=man:asterisk(8)
 After=network.target
 #include these if asterisk need to bind to a specific IP (other than 0.0.0.0)
 #Wants=network-online.target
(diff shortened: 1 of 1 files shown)


chrsmj:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

contrib/systemd/asterisk.service
```diff
@@ -16,19 +17,53 @@ Group=asterisk
...
+# "...long-running services (in particular network-facing ones)..."
+# This restricts any Asterisk access to /home, /root and /run/user
+ProtectHome=true
+
+# Mount the entire file system read-only, except for /dev, /proc, /sys
+# and any exceptions (see ReadWritePaths below for some exception ideas)
+ProtectSystem=strict
+
+# Minimalist operation - PID and ctl files, AstDB, Record(), and Logs
+ReadWritePaths=/var/run /var/lib/asterisk /var/log/asterisk
```
gtjoseph, line 59: When RuntimeDirectory is specified and `/var/run` is a tmpfs, systemd will create the directory before applying the protection.  If `/var/run` isn't a tmpfs then someone will have to have created `/var/run/asterisk` manually.  Either way,  there should be no need for RW on `/var/run` itself so it can be further scoped down to `/var/run/asterisk`.

I also think you need to include `/var/spool/asterisk` in the default config.  Voicemail and call recording also require it and those are fairly common.

Have you checked that these all still work if runuser and rungroup are set to a non-privileged user in asterisk.conf?  I don't know why they wouldn't but I'm just playing safe.

gtjoseph (changes requested):

contrib/systemd/asterisk.service
chrsmj: Excellent points, thank you. Will add limit to `/var/run/asterisk`. Tested that it works with asterisk/asterisk user/group on recent local Debian 12 VM. (Not sure if testsuite addresses systemd issues.)

contrib/systemd/asterisk.service
```diff
@@ -16,19 +17,51 @@ Group=asterisk
 ExecStart=/usr/sbin/asterisk -mqf -C /etc/asterisk/asterisk.conf
```
gtjoseph, line 17: Oh you know what?  I didn't expand the whole diff and forgot about
```
User=asterisk
Group=asterisk
```
Two issues...  

There's nothing in an asterisk install that creates that user and group.  It's totally up to the packager.

and...

When asterisk.conf has runuser specified, asterisk knows it's not going to be run as a 
privileged user so, thinking it was actually started as root as it would have been in the old days, it tries to get the `cap_net_admin` capability before switching the user.  If the service specifies User then that will fail and you'll see a message like `Unable to install capabilities.` almost immediately after startup but startup will continue.  If the service specifies User and runuser isn't in asterisk.conf then we don't even try.  IIRC, at one time we needed that capability to set QOS on packets when not running as a privileged user.  I _think_ that may still be the case.

So, I'm wondering if we should leave User and Group commented out but add a comment explaining that setting runuser and rungroup in asterisk.conf is preferred for the reason above.

Having said all of that, the proper systemd way to do this would probably be adding AmbientCapabilities=CAP_NET_ADMIN to the file but I've never tried it so I don't know if it works or not.

gtjoseph (changes requested):

contrib/systemd/asterisk.service
```diff
@@ -16,19 +17,51 @@ Group=asterisk
...
+# "...long-running services (in particular network-facing ones)..."
+# This restricts any Asterisk access to /home, /root and /run/user
+ProtectHome=true
+
+# Mount the entire file system read-only, except for /dev, /proc, /sys
+# and any exceptions (see ReadWritePaths below for some exception ideas)
+ProtectSystem=strict
+
+# Standard operation - PID, CLI socket, AstDB, Record(), Voicemail(), and logs
+ReadWritePaths=/var/run/asterisk /var/lib/asterisk /var/log/asterisk /var/spool/asterisk
```
gtjoseph, line 61: Sorry, how about one more comment indicating that if they use AMI to modify config files in /etc/asterisk, they'll need to include that as well.

gtjoseph (changes requested):

asteriskteam:

asteriskteam:

Issue #861, opened 2024-08-30
[bug]: ChanSpy unable to read audiohook read direction frame when no packet lost on both side of the call


Tinet-mucw:
### Severity

Trivial

### Versions

18.10

### Components/Modules

main

### Operating Environment

centos

### Frequency of Occurrence

Constant

### Issue Description

ChanSpy("Local/s@audio_socket-00000003;1", "PJSIP/test-00000002,qEoS")
- o: Only listen to audio coming from this channel.

The chanspy is monitoring the read direction；while asterisk write both read direction and write direction to spy_audiohook. This often causes the write direction to exceed the 100ms limit, which  leading to frequent cleanup the audiohook factory. As a result, the packets in the read direction are being cleaned up before they  are read,  even in cases where there is no packet loss.

### Relevant log output

```shell
[Aug 30 11:10:36] DEBUG[13372][C-0000000d]: audiohook.c:179 ast_audiohook_write_frame: direction:0 our_factory_ms:20 other_factory_ms:60
[Aug 30 11:10:36] DEBUG[13363][C-0000000d]: audiohook.c:179 ast_audiohook_write_frame: direction:1 our_factory_ms:120 other_factory_ms:0
[Aug 30 11:10:36] DEBUG[13363][C-0000000d]: audiohook.c:187 ast_audiohook_write_frame: Audiohook 0x7f64e8f8f6e0 has stale audio in its factories. Flushing them both; our_factory_ms:120 other_factory_ms:0
[Aug 30 11:10:36] DEBUG[13363][C-0000000d]: audiohook.c:179 ast_audiohook_write_frame: direction:0 our_factory_ms:21 other_factory_ms:20
[Aug 30 11:10:36] DEBUG[13372][C-0000000d]: audiohook.c:179 ast_audiohook_write_frame: direction:1 our_factory_ms:81 other_factory_ms:20
[Aug 30 11:10:36] DEBUG[13372][C-0000000d]: audiohook.c:179 ast_audiohook_write_frame: direction:0 our_factory_ms:20 other_factory_ms:80
[Aug 30 11:10:36] DEBUG[13363][C-0000000d]: audiohook.c:179 ast_audiohook_write_frame: direction:1 our_factory_ms:40 other_factory_ms:0
[Aug 30 11:10:36] DEBUG[13363][C-0000000d]: audiohook.c:179 ast_audiohook_write_frame: direction:0 our_factory_ms:18 other_factory_ms:40
[...]

PR #862, opened 2024-08-30
app_chanspy.c: resolving the issue with audiohook direction read

Tinet-mucw:
ChanSpy(${channel}, qEoS): When chanspy spy the direction read, reading frame is often failed when reading direction read audiohook. because chanspy only read audiohook direction read; write_factory_ms will greater than 100ms soon, then ast_slinfactory_flush will being called, then direction read will fail.

Resolves: #861


Tinet-mucw:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20
cherry-pick-to: 18

Issue #831, opened 2024-08-07
[bug]: app_voicemail ODBC


sysnux:
### Severity

Critical

### Versions

20.9.1

### Components/Modules

app_voicemail

### Operating Environment

Linux

### Frequency of Occurrence

Constant

### Issue Description

Crash in odbc_insert_data_cb and messagecount

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
Even though you've submitted a pull request for this, we still need the backtraces...
https://docs.asterisk.org/Development/Debugging/Getting-a-Backtrace/

sysnux:
Well, I spent most of yesterday chasing these two bugs, and they are quite obvious, I won't spend more time on that, sorry.

bkervaski:
Same issue, it appears starting with 20.9.0.

gtjoseph:
It should be fixed before the next set of releases.

seanbright:
Seems this can be closed now?

sysnux:
Yes, thanks!

sysnux:
Resolved

PR #865, opened 2024-09-03
app_voicemail: Fix sql insert mismatch caused by cherry-pick

gtjoseph:
When commit e8c9cb80 was cherry-picked in from master, the
fact that the 20 and 18 branches still had the old "macrocontext"
column wasn't taken into account so the number of named parameters
didn't match the number of '?' placeholders.  They do now.

We also now use ast_asprintf to create the full mailbox query SQL
statement instead of trying to calculate the proper length ourselves.

Resolves: #831


gtjoseph:
cherry-pick-to: 18

seanbright (approved):

jcolp (approved):

Issue #763, opened 2024-06-10
[bug]: autoservice thread stuck in an endless sleep


alex2grad:
### Severity

Major

### Versions

18.22.0

### Components/Modules

autoservice.c

### Operating Environment

Linux x86_64 Slackware 15.0

### Frequency of Occurrence

Occasional

### Issue Description

The autoservice_run thread on start of loop increments a variable _as_chan_list_state_.
Then on serviced channel calls _ast_read_
The _ast_read_ gets AST_FRAME_READ_ACTION_CONNECTED_LINE_MACRO and calls ast_channel_connected_line_sub.
The global variable CONNECTED_LINE_SEND_SUB is set as _intercept_connectedline,s,1_
The subroutine intercept_connectedline calls ODBC function GET_CONNECTED_LINE_ID
The acf_odbc_read calls ast_autoservice_start/ast_autoservice_stop
The function ast_autoservice_stop _waits while autoservice thread rebuilds its list_,
        while (chan_list_state == _as_chan_list_state_) {
                usleep(1000);
        }
BUT ast_autoservice_stop is called in the autoservice thread!
So autoservice thread stuck in an endless sleep which causes a complete lockup of Asterisk.

### Relevant log output

```shell
Thread 100 (Thread 0x7f6a39177640 (LWP 18314)):
#0  0x00007f6aae6ef775 in clock_nanosleep@GLIBC_2.2.5 () from /lib64/libc.so.6
#1  0x00007f6aae6f45e3 in nanosleep () from /lib64/libc.so.6
#2  0x00007f6aae72445b in usleep () from /lib64/libc.so.6
#3  0x0000000000470baa in ast_autoservice_stop (chan=chan@entry=0x7f6a52189750) at autoservice.c:314
#4  0x00007f6aaa08701c in acf_odbc_read (chan=<optimized out>, cmd=0x7f6a3916a7b0 "GET_CONNECTED_LINE_ID", s=<optimized out>, buf=0x7f6a3916a308 "", len=4096) at func_odbc.c:1117
#5  0x000000000052daa6 in ast_func_read (chan=chan@entry=0x7f6a52189750, function=function@entry=0x7f6a3916a8e0 "GET_CONNECTED_LINE_ID(xxxxxxxxxxx,yyyyyyyyyyyy)", workspace=workspace@entry=0x7f6a3916c8e0 "", len=len@entry=4096) at pbx_functions.c:617
[...]

PR #868, opened 2024-09-04
autoservice: Do not sleep if autoservice_stop is called within autoservice thread

alex2grad:
It's possible that ast_autoservice_stop is called within the autoservice thread.
In this case the autoservice thread is stuck in an endless sleep.

To avoid endless sleep ast_autoservice_stop must check that it's not called
within the autoservice thread.

Fixes: #763

### DIFF
--- a/main/autoservice.c
+++ b/main/autoservice.c
@@ -309,9 +309,16 @@ int ast_autoservice_stop(struct ast_channel *chan)
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


alex2grad:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

mbradeen (approved):

jcolp (approved):

Issue #881, opened 2024-09-11
[bug]: Long URLs are being rejected by the media cache because of an astdb key length limit


gtjoseph:
### Severity

Minor

### Versions

18,20,21,22,master

### Components/Modules

db.c

### Operating Environment

any

### Frequency of Occurrence

Occasional

### Issue Description

Long URLs, like those needed by Amazon S3, are being rejected by the media cache if they exceed the astdb key limit of 253 characters.

### Relevant log output

```shell
[2023-09-27 09:57:01.259+0200] WARNING[12413][C-000001cb]: db.c:346 ast_db_put: Family and key length must be less than 253 bytes
    -- <Announcer/ARI-000002a6;1> Playing 'https://s3-storage.xxx.yyy.net:8999/zzzzz-audio-file-rrrrrrrrrr/zzzzz/1111111/666666666/11112222333344445555666677778888/qqqqqqqqqqq-xxxxxxxxxxxxxxx.wav?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Date=20240902T075701Z&X-Amz-SignedHeaders=host&X-Amz-Expires=3600&X-Amz-Credential=????????????????%2F20240902%2Feu-west-1%2Fs3%2Faws4_request&X-Amz-Signature=ccccccyyyyyyuuuuuuiiiiiiooooooppppppaaaaaassssssddddddffffffgggg.slin' (language 'en')
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #883, opened 2024-09-11
db.c: Remove limit on family/key length

gtjoseph:
Consumers like media_cache have been running into issues with
the previous astdb "/family/key" limit of 253 bytes when needing
to store things like long URIs.  An Amazon S3 URI is a good example
of this.  Now, instead of using a static 256 byte buffer for
"/family/key", we use ast_asprintf() to dynamically create it.

Both test_db.c and test_media_cache.c were also updated to use
keys/URIs over the old 253 character limit.

Resolves: #881

UserNote: The `ast_db_*()` APIs have had the 253 byte limit on
"/family/key" removed and will now accept families and keys with a
total length of up to SQLITE_MAX_LENGTH (currently 1e9!).  This
affects the `DB*` dialplan applications, dialplan functions,
manager actions and `databse` CLI commands.  Since the
media_cache also uses the `ast_db_*()` APIs, you can now store
resources with URIs longer than 253 bytes.

### DIFF
--- a/main/db.c
+++ b/main/db.c
@@ -116,7 +116,6 @@
 	</manager>
  ***/
 
-#define MAX_DB_FIELD 256
 AST_MUTEX_DEFINE_STATIC(dblock);
 static ast_cond_t dbcond;
 static sqlite3 *astdb;
@@ -138,7 +137,7 @@ DEFINE_SQL_STATEMENT(deltree_all_stmt, "DELETE FROM astdb")
 DEFINE_SQL_STATEMENT(gettree_stmt, "SELECT key, value FROM astdb WHERE key || '/' LIKE ? || '/' || '%' ORDER BY key")
 DEFINE_SQL_STATEMENT(gettree_all_stmt, "SELECT key, value FROM astdb ORDER BY key")
 DEFINE_SQL_STATEMENT(showkey_stmt, "SELECT key, value FROM astdb WHERE key LIKE '%' || '/' || ? ORDER BY key")
-DEFINE_SQL_STATEMENT(create_astdb_stmt, "CREATE TABLE IF NOT EXISTS astdb(key VARCHAR(256), value VARCHAR(256), PRIMARY KEY(key))")
+DEFINE_SQL_STATEMENT(create_astdb_stmt, "CREATE TABLE IF NOT EXISTS astdb(key TEXT, value TEXT, PRIMARY KEY(key))")
 
 /* This query begs an explanation:
  *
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

alex2grad:
There is also VARCHAR in utils/astdb2sqlite3.c

utils/astdb2sqlite3.c:DEFINE_SQL_STATEMENT(create_astdb_stmt, "CREATE TABLE IF NOT EXISTS astdb(key VARCHAR(256), value VARCHAR(256), PRIMARY KEY(key))")

gtjoseph:
> There is also VARCHAR in utils/astdb2sqlite3.c
> 
> utils/astdb2sqlite3.c:DEFINE_SQL_STATEMENT(create_astdb_stmt, "CREATE TABLE IF NOT EXISTS astdb(key VARCHAR(256), value VARCHAR(256), PRIMARY KEY(key))")

That program was created in 2011 when Asterisk switched from using Berkeley DB to Sqlite3 so no one should have used it in years.  Sqlite3 doesn't enforce those limits anyway so I really don't want to touch that code.

> Also there are a lot of VARCHAR in res/res_config_sqlite.c
> SQLite ignores the length specified in VARCHAR().

As you point out, the lengths are ignored and this PR was specifically meant to address the 253 byte limit hard coded in db.c.  I'd rather not mess with res_config_sqlite.c at this time.

main/db.c
```diff
@@ -341,17 +340,17 @@ static int ast_db_rollback_transaction(void)
...
 {
-	char fullkey[MAX_DB_FIELD];
+	char *fullkey;
 	size_t fullkey_len;
 	int res = 0;
 
-	if (strlen(family) + strlen(key) + 2 > sizeof(fullkey) - 1) {
-		ast_log(LOG_WARNING, "Family and key length must be less than %zu bytes\n", sizeof(fullkey) - 3);
+	fullkey_len = ast_asprintf(&fullkey, "/%s/%s", family, key);
+	if (fullkey_len < 0) {
```
seanbright, line 348: `fullkey_len` is a `size_t`

gtjoseph: I wonder why the compiler didn't catch that.

seanbright (approved):

main/media_cache.c
```diff
@@ -333,7 +333,7 @@ int ast_media_cache_create_or_update(const char *uri, const char *file_path,
 	}
 
 	if (created && ast_bucket_file_create(bucket_file)) {
-		ast_log(LOG_WARNING, "Failed to create media for %s\n", uri);
+		ast_log(LOG_WARNING, "Failed to create media %d for %s\n", created, uri);
```
seanbright, line 336: I assume this is leftover debugging code of some kind?

seanbright (changes requested):

main/media_cache.c
gtjoseph: oops, yeah.  removed.

seanbright (approved):

jcolp (approved):

Issue #884, opened 2024-09-13
[bug]: A ':' at the top of in stir_shaken.conf make Asterisk producing a core file when starting


oza4h07:
### Severity

Minor

### Versions

Asterisk 20 with PR-85

### Components/Modules

core

### Operating Environment

Debian 11

### Frequency of Occurrence

Constant

### Issue Description

By accident, I used a ":" instead of a ";" when commenting my /etc/asterisk/stir_shaken.conf file.
When starting Asterisk with new stir_shaken.conf file in place :
- Asterisk refuses to start
- new core file is present (in /var/lib/asterisk)

A simple line triggering this is adding a single ':' at the top of the file.
I have not check yet if this is specific or not to Stir Shaken config file.

I rated this as Minor. Feel free to re-rate this issue differently.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
This is res_stir_shaken specific, it is not handling the case of an invalid configuration file in a particular path.

oza4h07:
Adding in strir_shaken.conf file, an include statement to a non existing file keeps Asterisk from starting.
- stop Asterisk
- echo '#include non-existing-file.conf' >> /etc/asterisk/stir_shaken.conf
- start Asterisk (Asterisk refuses to start)

I don't know if this case deserves a new dedicated issue or mentioning this here is better.

jcolp:
That's not a bug, that's on purpose:

https://docs.asterisk.org/Fundamentals/Asterisk-Configuration/Asterisk-Configuration-Files/Using-The-include-tryinclude-and-exec-Constructs/?h=include

If it should not be fatal than tryinclude exists.

oza4h07:
I didn't know about this tryinclude statement.
For curiosity, I searched Asterisk-PR85 samples directory and found a couple of "#include" but no instance of "#tryinclude" in any sample file.
Maybe replacing one of those "#include" with "#tryinclude" would help to spread the knowledge.

Anyway, I'm sorry for this false alarm on #include files.

PR #885, opened 2024-09-13
res_stir_shaken.c: Fix crash when stir_shaken.conf is invalid

gtjoseph:
* If the call to ast_config_load() returns CONFIG_STATUS_FILEINVALID,
check_for_old_config() now returns LOAD_DECLINE instead of continuing
on with a bad pointer.

* If CONFIG_STATUS_FILEMISSING is returned, check_for_old_config()
assumes the config is being loaded from realtime and now returns
LOAD_SUCCESS.  If it's actually not being loaded from realtime,
sorcery will catch that later on.

* Also refactored the error handling in load_module() a bit.

Resolves: #884

### DIFF
--- a/res/res_stir_shaken.c
+++ b/res/res_stir_shaken.c
@@ -311,13 +311,22 @@ static int check_for_old_config(void)
 	char *cat = NULL;
 
 	cfg = ast_config_load("stir_shaken.conf", config_flags);
-	if (cfg == NULL) {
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

PR #886, opened 2024-09-13
Pjannesen/cel absolute filename

pjannesen:
In logger.conf it is possible to specify a absolute path for the log file.
Make this also possible for cel_custom.conf.


jcolp (changes requested):
This should be a single commit, and its commit message should follow the defined format[1].

[1] https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/?h=commit+messages

pjannesen:
I have create a new pull request.
Same code. To the master branch with 1 commit and commit text conform policies (hopefully)

Issue #889, opened 2024-09-17
[bug]: res_stir_shaken/verification.c has a stale include for jansson.h that can cause compilation to fail


gtjoseph:
### Severity

Minor

### Versions

all

### Components/Modules

res_stir_shaken

### Operating Environment

Environments without a jansson-devel or janson-dev packages installed.

### Frequency of Occurrence

Constant

### Issue Description

res_stir_shaken/verification.c has a stale `#include <jansson.h>` left over from previous versions and since res_stir_shaken itself correctly doesn't depend on jansson, if you don't have a jansson development package installed and are using the bundled jansson, `#include <jansson.h>` will cause compilation to fail.

### Relevant log output

```shell
res_stir_shaken/verification.c:22:10: fatal error: jansson.h: No such file or directory
   22 | #include <jansson.h>
      |          ^~~~~~~~~~~
compilation terminated.
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #890, opened 2024-09-17
res_stir_shaken: Remove stale include for jansson.h in verification.c

gtjoseph:
verification.c had an include for jansson.h left over from previous
versions of the module.  Since res_stir_shaken no longer has a
dependency on jansson, the bundled version wasn't added to GCC's
include path so if you didn't also have a jansson development package
installed, the compile would fail.  Removing the stale include
was the only thing needed.

Resolves: #889

### DIFF
--- a/res/res_stir_shaken/verification.c
+++ b/res/res_stir_shaken/verification.c
@@ -19,7 +19,6 @@
 #include <sys/stat.h>
 
 #include <jwt.h>
-#include <jansson.h>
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #882, opened 2024-09-11
[bug]: Value CHANNEL(userfield) is lost by BRIDGE_ENTER


pjannesen:
### Severity

Minor

### Versions

20.9.3

### Components/Modules

cel

### Operating Environment

debian container with Asterisk build from sources.

### Frequency of Occurrence

None

### Issue Description

CHANNEL(userfield) is lost by BRIDGE_ENTER. 

Following sample from cel_custom logging
```
"1726076868.487565","CHAN_START","dhg1tn01-1726076868.27","",""
"1726076869.488925","ANSWER","dhg1tn01-1726076868.27","11690910/1",""
"1726076873.676815","BRIDGE_ENTER","dhg1tn01-1726076868.27","","{""bridge_id"":""8f7477f5-e62a-4da2-88ca-fc03964095c5"",""bridge_technology"":""simple_bridge""}"
"1726076919.807386","BRIDGE_EXIT","dhg1tn01-1726076868.27","","{""bridge_id"":""8f7477f5-e62a-4da2-88ca-fc03964095c5"",""bridge_technology"":""simple_bridge""}"
"1726076919.807929","HANGUP","dhg1tn01-1726076868.27","","{""hangupcause"":16,""hangupsource"":"""",""dialstatus"":""""}"
"1726076919.807929","CHAN_END","dhg1tn01-1726076868.27",""""
```

The 4th field in CHANNEL(userfield). After the channel is bridge with a other channel the CHANNEL(userfield) value is lost.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
Please provide the complete console output as well with this.

pjannesen:
Your application controls asterisk using the AMI.
Channel "dhg1tn01-1726076868.27" was an incoming channel that was answered and put on a wait loop.
The CHANNEL(userfield) was assigned using AMI SetVar.
The seconds channel was create by a AMI Originate.
When then seconds channel was up (and after some prompts) both channels are bridge using AMI: Bridge.

jcolp:
Please provide a way to reproduce this issue, such as strictly dialplan or a dialplan+script that someone can run. Your scenario is complicated enough that this is needed.

pjannesen:
I have found the bug the 'userfield' is not copied in channel_do_masquerade

pjannesen:
I have create a pull request #887

PR #894, opened 2024-09-20
channel: Value CHANNEL(userfield) is lost by BRIDGE_ENTER/BRIDGE_LEAVE.

pjannesen:
When a channel masquerade occurs, the value of CHANNEL(userfield) is not copied to the new ast_channel. A channel masquerade occurs when 2 existing channels are bridged using a simple bridge. Or when a simple bridge is broken and both channels continue independently.

Fixes: #882

### DIFF
--- a/main/channel.c
+++ b/main/channel.c
@@ -7216,6 +7216,9 @@ static void channel_do_masquerade(struct ast_channel *original, struct ast_chann
 	/* copy over accuntcode and set peeraccount across the bridge */
 	ast_channel_accountcode_set(original, S_OR(ast_channel_accountcode(clonechan), ""));
 
+	/* copy over userfield */
+	ast_channel_userfield_set(original, ast_channel_userfield(clonechan));
+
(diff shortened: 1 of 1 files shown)


pjannesen:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (changes requested):
The commit message isn't accurate. A masquerade will not always occur in the circumstances given, it depends on different factors and the state of the channels. I don't think specifics like that should be in the commit message. I propose an alternative:

channel: Preserve CHANNEL(userfield) on masquerade.

In certain circumstances a channel may undergo an operation
referred to as a masquerade. If this occurs the CHANNEL(userfield)
value was not preserved causing it to get lost. This change makes
it so that this field is now preserved.

Fixes: #882

pjannesen:
I have reworded the commit and did a force push.
@jcolp thanks for de commit message!

jcolp (approved):

Issue #900, opened 2024-09-24
[bug]: astfd.c: NULL pointer passed to fclose with nonnull attribute causes compilation failure


InterLinked1:
### Severity

Major

### Versions

21.5.0-rc1

### Components/Modules

astfd.c

### Operating Environment

Ubuntu 24.04 / gcc 13.2.0

### Frequency of Occurrence

Constant

### Issue Description

As shown in log:

### Relevant log output

```shell
[CCi] astfd.i -> astfd.o
astfd.c: In function ‘__ast_fdleak_fclose’:
astfd.c:283:24: error: argument 1 null where non-null expected [-Werror=nonnull]
  283 |                 return fclose(ptr);
      |                        ^~~~~~~~~~~
In file included from /usr/src/asterisk-21.5.0-rc1/include/asterisk/compat.h:49,
                 from /usr/src/asterisk-21.5.0-rc1/include/asterisk.h:22,
                 from astfd.c:30:
/usr/include/stdio.h:184:12: note: in a call to function ‘fclose’ declared ‘nonnull’
  184 | extern int fclose (FILE *__stream) __nonnull ((1));
      |            ^~~~~~
cc1: all warnings being treated as errors
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #901, opened 2024-09-24
astfd.c: Avoid calling fclose with NULL argument.

InterLinked1:
Don't pass through a NULL argument to fclose, which is undefined behavior, and instead return -1 and set errno appropriately. This avoids a compiler warning with gcc 13.2.0.

Resolves: #900

### DIFF
--- a/main/astfd.c
+++ b/main/astfd.c
@@ -280,7 +280,8 @@ int __ast_fdleak_fclose(FILE *ptr)
 {
 	int fd, res;
 	if (!ptr) {
-		return fclose(ptr);
+		errno = EINVAL;
+		return -1;
 	}
 
 	fd = fileno(ptr);


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (dismissed):

gtjoseph:
@InterLinked1 I was trying to determine if we should apply this patch to the certified branches but I can't seem to reproduce the issue.  I'm using Ubuntu22 with gcc 13.2.0 and have tried with and without devmode and with and without COMPILE_DOUBLE.  Can you reproduce the issue with a `make NOISY_BUILD=yes` and paste the full commands for `[CCp]` and `[CCi]`?  I'd like to compare with what I'm seeing.

seanbright (dismissed):
Now that I look at both the [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fclose.html) docs and `fclose(3)` on Linux (_"The behaviour of fclose() is undefined if the stream parameter is an illegal pointer"_), setting `errno` here is probably not technically correct. I think returning `-1` is fine though.

InterLinked1:
> @InterLinked1 I was trying to determine if we should apply this patch to the certified branches but I can't seem to reproduce the issue. I'm using Ubuntu22 with gcc 13.2.0 and have tried with and without devmode and with and without COMPILE_DOUBLE. Can you reproduce the issue with a `make NOISY_BUILD=yes` and paste the full commands for `[CCp]` and `[CCi]`? I'd like to compare with what I'm seeing.

I believe I noticed this on an Ubuntu 24.04 build with gcc 13.2.0, but not Ubuntu 22.04 or any other distro. Builds were in devmode with DONT_OPTIMIZE as well. I think this *might* have happened only in a GitHub runner and not other Ubuntu images but I don't remember for sure now.

InterLinked1:
> Now that I look at both the [POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fclose.html) docs and `fclose(3)` on Linux (_"The behaviour of fclose() is undefined if the stream parameter is an illegal pointer"_), setting `errno` here is probably not technically correct. I think returning `-1` is fine though.

Yeah, it's undefined behavior, but is there a reason it's better not to set errno here anyways? I was thinking there should be some way to indicate what the issue is, but both an assertion and even a log message seemed too harsh. It's only if we compile with `DEBUG_FD_LEAKS`, which aims to provide us more info about bad management of file descriptors, so setting errno seems like it would help with that goal... in theory.
In practice, nobody ever checks if `fclose` returns nonzero, so I suppose bothering to set errno doesn't help much anyways.

seanbright:
> Yeah, it's undefined behavior, but is there a reason it's better not to set errno here anyways?

Sorry, I could have been more clear before. I don't necessarily object to `errno` being set, but `EBADF` has a specific meaning:

> _The file descriptor underlying stream is not valid._

I don't feel strongly either way, so if you'd prefer to keep it as-is that's fine.

InterLinked1:
> @InterLinked1 I was trying to determine if we should apply this patch to the certified branches but I can't seem to reproduce the issue. I'm using Ubuntu22 with gcc 13.2.0 and have tried with and without devmode and with and without COMPILE_DOUBLE. Can you reproduce the issue with a `make NOISY_BUILD=yes` and paste the full commands for `[CCp]` and `[CCi]`? I'd like to compare with what I'm seeing.

Actually, after thinking about this, I don't the version of gcc matters. It has to do with when the nonnull attribute would've been added, which looks like glibc 2.38: https://github.com/bminor/glibc/commit/71d9e0fe766a3c22a730995b9d024960970670af

I'm guessing your Ubuntu system has something older, even 24.04 seems to come with 2.36 by default. My guess is glibc got upgraded on that system and the newer header file then triggered the build failure.

InterLinked1:
> > Yeah, it's undefined behavior, but is there a reason it's better not to set errno here anyways?
> 
> Sorry, I could have been more clear before. I don't necessarily object to `errno` being set, but `EBADF` has a specific meaning:
> 
> > _The file descriptor underlying stream is not valid._
> 
> I don't feel strongly either way, so if you'd prefer to keep it as-is that's fine.

Okay, thanks for clarifying... I guess "file descriptor underlying stream is not valid" *is* different from "stream itself is not valid".

I'll see if there's a more appropriate value to use or remove it otherwise.

main/astfd.c
```diff
@@ -276,11 +276,11 @@ FILE *__ast_fdleak_fopen(const char *path, const char *mode, const char *file, i
 }
 
 #undef fclose
-int __ast_fdleak_fclose(FILE *ptr)
-{
+int __ast_fdleak_fclose(FILE *ptr) {
```
InterLinked1, line 279: Not sure why this happened, but I'm working on it

seanbright (approved):

gtjoseph (approved):

Issue #904, opened 2024-09-24
[bug]: stir_shaken: attest_level isn't being propagated correctly from attestation to profile to tn


gtjoseph:
### Severity

Major

### Versions

all

### Components/Modules

res_stir_shaken

### Operating Environment

any

### Frequency of Occurrence

Constant

### Issue Description

If attest_level isn't specified on a profile it's defaulting to "A" and not being taken from the attestation section.  Same thing for tns.  If not specified, it's defaulting to "A" and not being taken from the profile or attestation sections.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #905, opened 2024-09-24
stir_shaken: Fix propagation of attest_level and a few other values

gtjoseph:
attest_level, send_mky and check_tn_cert_public_url weren't
propagating correctly from the attestation object to the profile
and tn.

* In the case of attest_level, the enum needed to be changed
so the "0" value (the default) was "NOT_SET" instead of "A".  This
now allows the merging of the attestation object, profile and tn
to detect when a value isn't set and use the higher level value.

* For send_mky and check_tn_cert_public_url, the tn default was
forced to "NO" which always overrode the profile and attestation
objects.  Their defaults are now "NOT_SET" so the propagation
happens correctly.

* Just to remove some redundant code in tn_config.c, a bunch of calls to
generate_sorcery_enum_from_str() and generate_sorcery_enum_to_str() were
replaced with a single call to generate_acfg_common_sorcery_handlers().

Resolves: #904

### DIFF
--- a/res/res_stir_shaken/common_config.c
+++ b/res/res_stir_shaken/common_config.c
@@ -100,6 +100,7 @@ const char *param_name ## _to_str( \
 }
 
 generate_enum_string_functions(attest_level, UNKNOWN,
+	{attest_level_NOT_SET, "not_set"},
 	{attest_level_A, "A"},
 	{attest_level_B, "B"},
 	{attest_level_C, "C"},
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

mbradeen (approved):

jcolp (approved):

PR #910, opened 2024-09-25
res_pjsip_refer.c: crash that lead to deadlock during transfer

n2p-litvinov:
Deadlock is during the call transfer. pjsip subscribes to stasis “bridge:all“ structure with callback function “refer_progress_bridge” and applied allowed message types and filters. When applying process is in progress function “refer_progress_bridge“ executed in separate thread where start to unsubscribe and remove subscription that lead to unlocking of null pointer.

1. Create subscription to “bridge:all“ with callback (res/res_pjsip_refer.c:703) (thread 1)
2. Start updating this subscription. (res/res_pjsip_refer.c:723-725) (thread 1)
3. Lock sub->topic (“bridge:all“ structure) (main/stasis.c:1043 && main/stasis.c:1086) (thread 1)
4. Fire the callback function (res/res_pjsip_refer.c:181) (thread 2)
5. Unsubscribe, sub->topic = null (res/res_pjsip_refer.c:217) (thread 2)
6. Unlock sub->topic that is null (main/stasis.c:1050 && main/stasis.c:1090) (thread 1)
7. Thread 1 crashes
8. “bridge:all” is locked by crashed thread and will never be unlocked we have deadlock

Fixes issue:
https://github.com/asterisk/asterisk/issues/930

### DIFF
--- a/res/res_pjsip_refer.c
+++ b/res/res_pjsip_refer.c
@@ -726,9 +726,11 @@ static void refer_blind_callback(struct ast_channel *chan, struct transfer_chann
 
 			ao2_cleanup(refer->progress);
 		} else {
+			ao2_bump(refer->progress->bridge_sub);
 			stasis_subscription_accept_message_type(refer->progress->bridge_sub, ast_channel_entered_bridge_type());
 			stasis_subscription_accept_message_type(refer->progress->bridge_sub, stasis_subscription_change_type());
 			stasis_subscription_set_filter(refer->progress->bridge_sub, STASIS_SUBSCRIPTION_FILTER_SELECTIVE);
+			ao2_cleanup(refer->progress->bridge_sub);
 		}
 	}
 


gtjoseph:
@n2p-litvinov Is there an issue open for this?  If not, you may want to open one.  This allows the fix to be highlighted at the top of the changelog.

Please also see the links below for correct commit message formatting:
See:

gtjoseph (changes requested):
Please see the links below for correct commit message formatting:
See:

n2p-litvinov:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

res/res_pjsip_refer.c
```diff
@@ -726,9 +726,11 @@ static void refer_blind_callback(struct ast_channel *chan, struct transfer_chann
 
 			ao2_cleanup(refer->progress);
 		} else {
+			ao2_bump(refer->progress->bridge_sub);
```
jcolp, line 729: I don't think this is the correct fix. This appears to just reduce the chance of this happening and not fix the underlying issue, which based on the description (since there's no backtrace) would be a synchronization issue. Specifically the threads should not be executing at the same time OR working on the same thing at the same time.

jcolp (changes requested):

res/res_pjsip_refer.c
n2p-litvinov: Stacktrace can be found in related issue:
https://github.com/asterisk/asterisk/issues/930

Do you have any better idea to fix this?

jcolp: I don't have a fix to propose, no, but this itself is not a fix but merely a workaround/reduction of the chance of it happening. A full audit of the reference counting and threading for the specific situation has to occur, and then based on that determining the appropriate fix. It should be ensured that either the reference count is already correct/maintained when this is invoked, or it should not be invoked.

asteriskteam:

asteriskteam:

PR #918, opened 2024-09-29
config.c: #tryinclude should not create new files

Allan-N:
Correct an issue in ast_config_text_file_save2() when updating configuration files with "#tryinclude" statements. The API currently replaces "#tryinclude" with "#include". The API also creates empty template files if the referenced files do not exist. This change resolves these problems.


Allan-N:
This change resolves Issue #920

Allan-N:
Note : I was asked about the "precomment" changes included in this PR.  To set the stage, let's talk about a sample configuration file :
```
 1. ; comment block #1
 2. ; more of comment block #1
 3. 
 4. [category]
 5. 
 6. ; comment block #2
 7. ; more of comment block #2
 8. 
 9. variable1 = value1	; inline comment
10.			; more of inline comment
11.
12. ; comment block #3
13. ; more of comment block #3
14. 
15. #include somefile
16. 
17. ; comment block #4
18. ; more of comment block #4
19. 
20. variable2 = value2
```
The pre-PR code would have lines 1-3 associated with "[category]" as pre-comments, lines 5-8 as pre-comments to "variable1", and lines 11-14 AND lines 16-19 as pre-comments to "variable2".  As-is, the code would rewrite the above as :
```
 1. ; comment block #1
 2. ; more of comment block #1
 3. 
 4. [category]
 5. 
 6. ; comment block #2
 7. ; more of comment block #2
 8. 
 9. variable1 = value1	; inline comment
10.			; more of inline comment
11. #include somefile
12.
13. ; comment block #3
14. ; more of comment block #3
15. 
16. 
17. ; comment block #4
18. ; more of comment block #4
19. 
20. variable2 = value2
```
The "precomment" changes included in this PR are to keep the comments prior to the "#include" in place.  I think this fix is essential.  The bonus change I included (lines 2559-2570) relates to the handling of leading whitespace on standalone and inline comments.

InterLinked1 (changes requested):
All your changes should be squashed into a single commit (amend your original commit when making edits, rather than making a new commit)

Allan-N:
>All your changes should be squashed into a single commit

Done.

main/config.c
```diff
@@ -255,7 +255,16 @@ struct ast_config {
...
 	int include_level;
 	int max_include_level;
-	struct ast_config_include *includes;  /*!< a list of inclusions, which should describe the entire tree */
+	/*! First inclusion in the list. */
+	struct ast_config_include *includes_root;
+	/*! Last inclusion in the list. */
+	struct ast_config_include *includes_last;
+};
+
+struct ast_config_included_file {
```
seanbright, line 264: Avoid the `ast_` prefix when creating new things that will not be exported.

main/config.c
```diff
@@ -265,7 +274,7 @@ struct ast_config_include {
 	 */
 	char *include_location_file;
 	int  include_location_lineno;    /*!< lineno where include occurred */
-	int  exec;                       /*!< set to non-zero if its a #exec statement */
+	int  include_type;               /*!< set positive if #exec statement, negative if #tryinclude statement */
```
seanbright, line 277: This structure is opaque, so this type can be changed to an `enum` to be more expressive.

seanbright (changes requested):

main/config.c
Allan-N: Poof.  The `ast_` prefix has been removed

main/config.c
Allan-N: Not knowing for certain if there was anyone calling ast_include_new() with a value other than 0 or 1 made me a bit hesitant about using an enum.  But, I agree that using an enum makes the code more expressive.

Enum added, code updated.

Allan-N:
>The commit message should wrap at 72 characters.

Done.

> Also, the title could maybe be a bit more descriptive, like "config.c: Fix #tryinclude being converted to #include on rewrite" or something.

Updated.

> Finally, the commit message should contain a "Resolves" line closing the linked issue (number 920 in this case).

Added.

InterLinked1:
> > The commit message should wrap at 72 characters.
> 
> Done.
> 
> > Also, the title could maybe be a bit more descriptive, like "config.c: Fix #tryinclude being converted to #include on rewrite" or something.
> 
> Updated.
> 
> > Finally, the commit message should contain a "Resolves" line closing the linked issue (number 920 in this case).
> 
> Added.

Looks like you updated the PR description, not the commit itself (which is what matters).

Allan-N:
>Looks like you updated the PR description, not the commit itself (which is what matters).

I have now updated the commit too :-)

InterLinked1:
> > Looks like you updated the PR description, not the commit itself (which is what matters).
> 
> I have now updated the commit too :-)

`This change resolves Issue #920` should be `Resolves: #920` as GitHub looks for specific syntax to autoclose the issue

include/asterisk/config.h
```diff
@@ -59,6 +59,18 @@ enum config_save_flags {
...
+ * \brief Types used for ast_include_new() include_type
+ */
+typedef enum {
+	/*! #include statements */
+	CONFIG_STATEMENT_INCLUDE    = 0,
+	/*! #exec statements */
+	CONFIG_STATEMENT_EXEC       = 1,
+	/*! #tryinclude statements */
+	CONFIG_STATEMENT_TRYINCLUDE = 2,
+} include_statement_type;
```
seanbright, line 72: I think my previous comments were confusing.

You should add this `enum` only in `main/config.c` - `ast_include_new()` will still be forced to take an `int` for API compatibility.

Allan-N: Understood.  I changed the `ast_include_new()` argument type back to an `int` and moved the `enum` to `main/config.c`.

Allan-N:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

main/config.c
```diff
@@ -362,20 +390,65 @@ struct ast_config_include *ast_include_new(struct ast_config *conf, const char *
...
+	switch (inc->include_type) {
+	case CONFIG_STATEMENT_INCLUDE :
+		inc_file = ast_calloc(1, sizeof(*inc_file) + strlen(fn) + 1);
+		if (inc_file) {
+			strcpy(inc_file->path, fn);	/* safe */
+			AST_LIST_INSERT_TAIL(&inc->included_files, inc_file, next);
+		}
+		break;
+	case CONFIG_STATEMENT_TRYINCLUDE :
+	    {
```
InterLinked1, line 412: whitespace

main/config.c
```diff
@@ -362,20 +390,65 @@ struct ast_config_include *ast_include_new(struct ast_config *conf, const char *
...
+	make_fn(fn, sizeof(fn), inc->included_file, from_file);
+	switch (inc->include_type) {
+	case CONFIG_STATEMENT_INCLUDE :
+		inc_file = ast_calloc(1, sizeof(*inc_file) + strlen(fn) + 1);
+		if (inc_file) {
+			strcpy(inc_file->path, fn);	/* safe */
+			AST_LIST_INSERT_TAIL(&inc->included_files, inc_file, next);
+		}
+		break;
+	case CONFIG_STATEMENT_TRYINCLUDE :
```
InterLinked1, line 411: shouldn't be a space before the colons here and elsewhere

main/config.c
```diff
@@ -255,17 +255,38 @@ struct ast_config {
...
+struct config_included_file {
+	AST_LIST_ENTRY(config_included_file) next;
+	/*|< #include path or matching #tryinclude paths */
+	char path[];
+};
+
+/*!
+ * \brief Types used for ast_include_new() include_type
+ */
+typedef enum {
```
InterLinked1, line 273: Do not make a typedef for this, just use this as an enum directly: https://docs.asterisk.org/Development/Policies-and-Procedures/Coding-Guidelines/#dont-use-unnecessary-typedefs

InterLinked1 (changes requested):
> And, as a bonus, I also found/fixed the issue where standalone and inline comments in the .conf files were being stripped of all leading whitespace.

Personally, I think this should be broken out into a separate patch/PR, since it's not related to the primary purpose of the patch.

Allan-N:
>>And, as a bonus, I also found/fixed the issue where standalone and inline comments in the .conf files were being stripped of all leading whitespace.

>Personally, I think this should be broken out into a separate patch/PR, since it's not related to the primary purpose of the patch.

I removed my "bonus" fix and will create a new issue/PR to address the issue(s) with configuration file rewrites mangling most of the formatted comments.

main/config.c
Allan-N: This is a 1/2 indent that follows the coding style already present in the file (look for "case PARSE_INADDR")

main/config.c
Allan-N: Fixed

main/config.c
Allan-N: Fixed

asteriskteam:

Allan-N:
@seanbright, @InterLinked1 : I believe that all requested changes have been made.  Can you kindly review?

asteriskteam:

Allan-N:
[PR #1047](https://github.com/asterisk/asterisk/pull/1047) pulled in one of the fixes (a reference leak) that I had found/addressed with my change.  I have since rebased and updated this pull request.

I believe that all of the previously requested changes have been made. Can you kindly review / approve / ... ?

include/asterisk/config.h
```diff
@@ -914,7 +914,7 @@ void ast_category_rename(struct ast_category *cat, const char *name);
 struct ast_variable *_ast_variable_new(const char *name, const char *value, const char *filename, const char *file, const char *function, int lineno);
 #define ast_variable_new(name, value, filename) _ast_variable_new(name, value, filename, __FILE__, __PRETTY_FUNCTION__, __LINE__)
 
-struct ast_config_include *ast_include_new(struct ast_config *conf, const char *from_file, const char *included_file, int is_exec, const char *exec_file, int from_lineno, char *real_included_file_name, int real_included_file_name_size);
+struct ast_config_include *ast_include_new(struct ast_config *conf, const char *from_file, const char *included_file, int include_type, const char *exec_file, int from_lineno, char *real_included_file_name, int real_included_file_name_size);
```
InterLinked1, line 917: I think it would be nice for argument `include_type` to have the more precise type `enum include_statement_type`, not `int`.

I know this is modifying a public API, but I think it should still work out to be backwards-compatible since you assigned the same values to the enum.

main/config.c
```diff
@@ -3019,12 +3163,9 @@ int ast_config_text_file_save2(const char *configfile, const struct ast_config *
...
 
 			/* output the respective include */
-			if (incl->exec)
-				fprintf(f,"#exec \"%s\"\n", incl->exec_file);
-			else
-				fprintf(f,"#include \"%s\"\n", incl->included_file);
-			fclose(f);
+			print_include(f, incl);
 			incl->output = 1;
+			fclose(f);
```
InterLinked1, line 3168: Why switch the order here?

include/asterisk/config.h
InterLinked1: Actually, now that I think about it, I'm not sure how the compiler would like passing ints if the parameter type is enum... if it complains about that, then maybe ignore this, or make the change master only.

seanbright:
> I believe that all of the previously requested changes have been made. Can you kindly review / approve / ... ?

Yup, that is what I was doing when I saw the reference count leak you identified. Will provide feedback shortly.

include/asterisk/config.h
seanbright: I _think_ going from `int` to `enum` would be fine ABI-wise (I believe it is possible that the compiler could choose a smaller type), but best not to change it in release branches.

If we're going to change the parameter type, best to do do it in a separate PR and maybe add some doxygen while we're at it.

Allan-N: When I first started working on these changes I was concerned about changing the parameter type.  As part of the review process with @seanbright (on 10/03) I changed the parameter type to an `enum` (defined in `include/asterisk/config.h` and used in `main/config.c`).  A followup review comment (on 10/13) suggested that the `enum` should be only in `main/config.c` ... so I made that change.

Let me know if you want me to switch back to using an `enum` in both the .h and .c files now, create a new PR making the change later (after this change has been merged), or leave this PR as-is.

main/config.c
Allan-N: For consistency.  Every other place were I call `print_include(f, incl);` is followed with `incl->output = 1;`

main/config.c
```diff
@@ -2750,15 +2868,30 @@ int ast_config_text_file_save2(const char *configfile, const struct ast_config *
...
 		incl->output = 0;
 
-		if (!incl->exec) {
-			/* now make sure we have write access to the include file or its parent directory */
-			make_fn(fn, sizeof(fn), incl->included_file, configfile);
-			/* If the file itself doesn't exist, make sure we have write access to the directory */
-			if (!is_writable(fn)) {
+		if (incl->include_type != CONFIG_STATEMENT_EXEC) {
+			struct config_included_file *included_file;
+			int ok = 0;
```
seanbright, line 2877: `int ok = 1;`

main/config.c
```diff
@@ -2750,15 +2868,30 @@ int ast_config_text_file_save2(const char *configfile, const struct ast_config *
...
+			/* Not a #exec. */
+			/* For #include, make sure we have write access to the include file
+			   file or its parent directory.  If the file itself doesn't exist,
+			   make sure we have write access to the directory. */
+			/* For #tryinclude, we check the file(s) that match the pattern. */
+
+			AST_LIST_TRAVERSE(&incl->included_files, included_file, next) {
+				if (!is_writable(included_file->path)) {
+					/* the #tryinclude path exists but is not writable */
+					ok = -1;
```
seanbright, line 2888: `ok = 0;`

main/config.c
```diff
@@ -2750,15 +2868,30 @@ int ast_config_text_file_save2(const char *configfile, const struct ast_config *
...
+					break;
+				}
+			}
+
+			if (ok != 0) {
```
seanbright, line 2893: `if (!ok) {`

seanbright (changes requested):
Here's a handful of things I've seen so far. Still working through it/testing/etc.

seanbright (changes requested):
So before your patch, this file:

```
[general]
a = b
c = d

; I'm just a comment

; #include "issue920/inc.conf"
; #include "issue920/glob/inc-*.conf"
#tryinclude "issue920/tryinc.conf"
#tryinclude "issue920/glob/tryinc-*.conf"
```

Saves as (ignoring the 'generated' header):

```
[general]
a = b
c = d
; I'm just a comment

; #include "issue920/inc.conf"
; #include "issue920/glob/inc-*.conf"
#include "issue920/glob/tryinc-*.conf"
#include "issue920/tryinc.conf"
```

After your patch, it saves as:

```
[general]
a = b
c = d
#tryinclude "issue920/tryinc.conf"
#tryinclude "issue920/glob/tryinc-*.conf"
```

So the `#tryinclude`s render correctly, but we lose the comments.

Allan-N:
> Saves as (ignoring the 'generated' header):
> ...
> After your patch, it saves as:
> ...

Those results do not match what I've been seeing with my testing.  But, regardless, I will make the requested updates and re-check the behavior with the handling of comments as you should be able to write the config with the file unchanged, modulo the updated header.

Stay tuned.

Allan-N:
> So the #tryincludes render correctly, but we lose the comments.

Well, it turns out that the .conf I was using for testing worked but your test case presented a different (and very reasonable) scenario ... and one that I had not accounted for in my changes.  That has since been corrected!  This PR now has a bit more but with some added refactoring the changes are a bit cleaner.

main/config.c
```diff
@@ -363,20 +392,65 @@ struct ast_config_include *ast_include_new(struct ast_config *conf, const char *
...
 		inc->included_file = ast_strdup(included_file);
 
-	inc->exec = is_exec;
-	if (is_exec)
-		inc->exec_file = ast_strdup(exec_file);
-
 	if (!inc->include_location_file
 		|| !inc->included_file
-		|| (is_exec && !inc->exec_file)) {
+		|| ((include_type == CONFIG_STATEMENT_EXEC) && !inc->exec_file)) {
```
seanbright, line 397: You moved the `inc->exec_file = ast_strdup(exec_file);` so that it occurs after this check. I haven't tested but I assume this means we won't get the `#exec` rendered in to the saved file.

main/config.c
```diff
@@ -363,20 +392,65 @@ struct ast_config_include *ast_include_new(struct ast_config *conf, const char *
...
+					strcpy(inc_file->path, matched_path);	/* safe */
+					AST_LIST_INSERT_TAIL(&inc->included_files, inc_file, next);
+				}
+			}
+		}
+		globfree(&globbuf);
+		break;
+	    }
+	case CONFIG_STATEMENT_EXEC:
+		inc->exec_file = ast_strdup(exec_file);
```
seanbright, line 443: I think anywhere in this function where memory allocation fails we should immediately return `NULL` (calling the necessary cleanup function first).

I realize there is existing code in this function that does not check for allocation failure, but there is no reason to keeping kicking that can down the road.

main/config.c
```diff
@@ -363,20 +392,65 @@ struct ast_config_include *ast_include_new(struct ast_config *conf, const char *
...
+
 	/* attach this new struct to the conf struct */
-	inc->next = conf->includes;
-	conf->includes = inc;
+	if (conf->includes_root && conf->includes_last) {
+		conf->includes_last->next = inc;
+	} else {
+		conf->includes_root = inc;
+	}
+	conf->includes_last = inc;
```
seanbright, line 453: Would prepending the list change semantics? That would avoid the need for the extra pointer. (minor)

seanbright (changes requested):
FWIW - This is a difficult PR to review because it both fixes the `#include` / `#tryinclude` issue as well as the comment rendering. I'm not suggesting that you split it up, just justifying the number of rounds of review involved.

Allan-N:
>FWIW - This is a difficult PR to review because it both fixes the #include / #tryinclude issue as well as the comment rendering. I'm not suggesting that you split it up, just justifying the number of rounds of review involved.

Yes, I realize that this is not a simple change and I very much appreciate your willingness to review the change and fixing an issue.

The gotcha is that the comments had been associated with the "category" or "variable".  There was no way to track a #include nested within pre-category or pre-variable comments.  Now, we track the comments before and after the #include's.  Now, you can also have comments before the #include (describing what's being included) and comments after the #include (describing whatever category/variable/end-of-file that follows),

main/config.c
Allan-N: Fixed

main/config.c
Allan-N: Yes, we follow the list when the actual #include lines are written to the .conf file.

main/config.c
```diff
@@ -363,20 +392,70 @@ struct ast_config_include *ast_include_new(struct ast_config *conf, const char *
...
-	if (is_exec)
-		inc->exec_file = ast_strdup(exec_file);
-
 	if (!inc->include_location_file
 		|| !inc->included_file
-		|| (is_exec && !inc->exec_file)) {
+		|| ((include_type == CONFIG_STATEMENT_EXEC) && !exec_file)) {
 		ast_includes_destroy(inc);
 		return NULL;
 	}
```
seanbright, line 400: These checks were originally checking for allocation failures for the previous 3 potential calls to `ast_strdup`, so checking `exec_file` here doesn't make sense. Remove the entire exec expression from here and check for allocation failure in the `CONFIG_STATEMENT_EXEC` branch of the switch statement below instead.

Allan-N: You are correct.  I have updated the allocation failure checks here and in the `CONFIG_STATEMENT_EXEC` branch.

main/config.c
```diff
@@ -2125,17 +2232,31 @@ static int process_text_line(struct ast_config *cfg, struct ast_category **cat,
...
-		ast_include_new(cfg, cfg->include_level == 1 ? "" : configfile, cur, !do_include, cur2, lineno, real_inclusion_name, sizeof(real_inclusion_name));
+		newinclude = ast_include_new(cfg, cfg->include_level == 1 ? "" : configfile, cur, include_type, cur2, lineno, real_inclusion_name, sizeof(real_inclusion_name));
+		if (!newinclude) {
+			return -1;
+		}
 
-		do_include = ast_config_internal_load(cur, cfg, flags, real_inclusion_name, who_asked) ? 1 : 0;
+		result = ast_config_internal_load(cur, cfg, flags, real_inclusion_name, who_asked);
 		if (!ast_strlen_zero(exec_file))
 			unlink(exec_file);
```
seanbright, line 2242: This used to happen regardless of the result of `ast_include_new` but now if `ast_include_new` fails we return early and this doesn't happen.

main/config.c
```diff
@@ -2737,6 +2864,41 @@ static int is_writable(const char *fn)
...
+			line = nl + 1;
+		}
+	}
+}
+
+static void print_include(FILE *f, struct ast_config_include *incl)
+{
+	switch (incl->include_type) {
+	case CONFIG_STATEMENT_INCLUDE:
+		fprintf(f,"#include \"%s\"\n", incl->included_file);
```
seanbright, line 2891: Make one more pass through your changes and sprinkle in some spaces where appropriate for coding guidelines. So here after `f,`. I'll highlight some more but not all of them. Mostly I've seen them here and in `for` loops.

main/config.c
```diff
@@ -2832,31 +3014,24 @@ int ast_config_text_file_save2(const char *configfile, const struct ast_config *
 			}
 
 			/* dump any includes that happen before this category header */
-			for (incl=cfg->includes; incl; incl = incl->next) {
-				if (strcmp(incl->include_location_file, cat->file) == 0){
+			for (incl=cfg->includes_root; incl; incl = incl->next) {
```
seanbright, line 3017: Spaces here around the `=`

main/config.c
```diff
@@ -2937,24 +3113,23 @@ int ast_config_text_file_save2(const char *configfile, const struct ast_config *
 					return -1;
 				}
 
-				/* dump any includes that happen before this category header */
-				for (incl=cfg->includes; incl; incl = incl->next) {
-					if (strcmp(incl->include_location_file, var->file) == 0){
+				/* dump any includes that happen before this variable */
+				for (incl=cfg->includes_root; incl; incl = incl->next) {
```
seanbright, line 3117: Spaces here around the `=`

main/config.c
```diff
@@ -2990,6 +3163,42 @@ int ast_config_text_file_save2(const char *configfile, const struct ast_config *
 			}
 			cat = cat->next;
 		}
+
+
+		/* Now, for files with trailing #include/#exec statements,
+		   we have to make sure every entry is output */
+		for (incl=cfg->includes_root; incl; incl = incl->next) {
```
seanbright, line 3170: Spaces here around the `=`

seanbright (changes requested):
A couple more small things and then I'm happy.

main/config.c
Allan-N: Fixed!

main/config.c
Allan-N: I made a pass through the entire file.  Most certainly added the space after a "," in my changes ... and likely a few extra places too :-)

seanbright (approved):

jcolp (approved):

Issue #928, opened 2024-10-02
[bug]: chan_dahdi: MWI while off-hook when hung up on after recall ring


InterLinked1:
### Severity

Major

### Versions

Git

### Components/Modules

chan_dahdi

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Steps to reproduce (phone B must be a DAHDI channel, phone A can be anything):

Required configuration:

1. In chan_dahdi, threewaycalling must be enabled and a mailbox should be assigned to a line (to trigger MWI).

2. Set autoreoriginate=no (the default), to reproduce reliably on all systems

Reproduction steps:

1. Call from A to B
2. Answer on B, then flash to put A on hold
4. Hang up B, which will trigger recall ring for the call on hold
5. Re-answer on B. Time does not matter (e.g. could before first full ring has completed, could be multiple rings, etc.)
6. Hang up A, which will drop B to reorder.
7. After a couple seconds, an MWI FSK burst will be heard on B and the reorder will stop completely. This should not happen!

This issue is resolved by tightening up the relevant logic in `do_monitor`.

### Relevant log output

```shell
[2024-09-08 09:42:17.839] DEBUG[271694][C-00000008]: sig_analog.c:3050 analog_handle_event: Got event ANALOG_EVENT_ONHOOK(1) on channel 122 (index 0)
[2024-09-08 09:42:17.841]     -- Channel DAHDI/122-1 left 'native_dahdi' basic-bridge <64983730-6685-4bfe-820c-d554378f3b64>
[2024-09-08 09:42:17.840] DEBUG[271694][C-00000008]: chan_dahdi.c:4983 dahdi_ec_enable: Enabled echo cancellation on channel 122
[2024-09-08 09:42:17.840] DEBUG[271694][C-00000008]: chan_dahdi.c:4983 dahdi_ec_enable: Enabled echo cancellation on channel 121
[2024-09-08 09:42:17.841] DEBUG[271694][C-00000008]: chan_dahdi.c:7403 dahdi_master_slave_unlink: Unlinking slave 121 from 122
[2024-09-08 09:42:17.841] DEBUG[271694][C-00000008]: chan_dahdi.c:4816 conf_del: Removed 9 from conference 9/122
[2024-09-08 09:42:17.841] DEBUG[271694][C-00000008]: chan_dahdi.c:4816 conf_del: Removed 10 from conference 9/121
[...]

PR #929, opened 2024-10-02
chan_dahdi: Never send MWI while off-hook.

InterLinked1:
In some circumstances, it is possible for the do_monitor thread to erroneously think that a line is on-hook and send an MWI FSK spill to it when the line is really off-hook and no MWI should be sent. Commit 0a8b3d34673277b70be6b0e8ac50191b1f3c72c6 previously fixed this issue in a more readily encountered scenario, but it has still been possible for MWI to be sent when it shouldn't be. To robustly fix this issue, query DAHDI for the hook status to ensure we don't send MWI on a line that is actually still off hook.

Resolves: #928


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #821, opened 2024-07-24
[bug]: app_dial:  The progress timeout doesn't cause Dial to exit


gtjoseph:
### Severity

Trivial

### Versions

18,20,21,master

### Components/Modules

app_dial

### Operating Environment

all

### Frequency of Occurrence

None

### Issue Description

The loop in app_dial:wait_for_answer doesn't exit when to_progress times out so even if you specify a progress_timeout, Dial will continue until the answer timeout.
Discovered by failing testsuite test tests/apps/dial/dial_progress.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

InterLinked1:
@gtjoseph I see why the test is broken and will fix that, but did you find an actual issue with the functionality? I can't replicate your described issue:

```

[2024-10-01 20:31:27.766]     -- Executing [19@from-internal:1] InbandDial("DAHDI/16-1", "Local/s@test,^3") in new stack
[2024-10-01 20:31:27.766]     -- Executing [s@test:1] Wait("Local/s@test-0000010f;2", "6") in new stack
[2024-10-01 20:31:27.766]     -- Called Local/s@test
[2024-10-01 20:31:30.770]     -- No early media received in 3000 ms
[2024-10-01 20:31:30.771]   == Spawn extension (test, s, 1) exited non-zero on 'Local/s@test-0000010f;2'
[2024-10-01 20:31:30.772]        > Call ended, DIALSTATUS: CHANUNAVAIL

[2024-10-01 20:33:56.341]     -- Executing [19@from-internal:1] InbandDial("DAHDI/16-1", "Local/s@test,6^3") in new stack
[2024-10-01 20:33:56.347]     -- Executing [s@test:1] Wait("Local/s@test-00000110;2", "6") in new stack
[2024-10-01 20:33:56.347]     -- Called Local/s@test
[2024-10-01 20:33:59.354]     -- No early media received in 3000 ms
[2024-10-01 20:33:59.354]   == Spawn extension (test, s, 1) exited non-zero on 'Local/s@test-00000110;2'
[2024-10-01 20:33:59.356]        > Call ended, DIALSTATUS: CHANUNAVAIL
```

gtjoseph:
It's been a while but what is `InbandDial()`?  One of your apps?  Does it do something different than `Dial()`?

InterLinked1:
[...]

PR #934, opened 2024-10-03
app_dial: Fix progress timeout.

InterLinked1:
Under some circumstances, the progress timeout feature added in commit 320c98eec87c473bfa814f76188a37603ea65ddd does not work as expected, such as if there is no media flowing. Adjust the waitfor call to explicitly use the progress timeout if it would be reached sooner than the answer timeout to ensure we handle the timers properly.

Resolves: #821


InterLinked1:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
testsuite-test-pr: 68

PR #935, opened 2024-10-04
feat: add redirecting as a channel state

gibbz00:
We need to disallow certain redirects at work, and we are unfortunately not able to rely on SIP diversion-limit headers due to incomplete support between different operators.

From the commit message body:

"Redirecting" has been added as an allowable value in the ARI Channel state property. Users will now be notified of a redirecting channel through the ARI ChannelStateChange event.

I could try to add a test-case in asterisk/testsuite if wished for by the maintainers.

### DIFF
--- a/rest-api/api-docs/channels.json
+++ b/rest-api/api-docs/channels.json
@@ -398,7 +398,6 @@
 							"reason": "Channel with given unique ID already exists."
 						}
 					]
-
 				},
 				{
 					"httpMethod": "DELETE",
@@ -2152,6 +2151,7 @@
 							"Busy",
 							"Dialing Offhook",
 							"Pre-ring",
+							"Redirecting",
 							"Unknown"
 						]
 					}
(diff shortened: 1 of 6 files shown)


rest-api/api-docs/channels.json
```diff
@@ -2200,4 +2200,4 @@
 			}
 		}
 	}
-}
```
gibbz00, line 2203: I see now that my editor autoformatted these parts, do you want them reverted?

jcolp (changes requested):
I am not a fan of altering the channel state to add this, it is not backwards compatible and would break outside applications and potentially logic inside of Asterisk in interesting ways.

A change was made last year to place the RDNIS in the channel snapshot:

https://github.com/asterisk/asterisk/commit/69590ba33e91aa543c8737f9de2599bc63797ec5

This value is present if the channel is redirected.

Is this not sufficient? If so, what is actually needed?

gibbz00:
Hii, thank you for taking the time to quickly respond to this.

This PR is the result of trying to get rid of some custom changes to our own fork. I suspect that this was added internally when an older asterisk version was used.

So yes, you're right. Closing it in favor of using `caller_rdnis`, apologies.

Issue #938, opened 2024-10-08
[bug]: memory leak - CBAnn leaks a small amount format_cap related memory for every confbridge


JonathanRRose:
### Severity

Trivial

### Versions

master, 18.9-certified

### Components/Modules

app_confbridge

### Operating Environment

Running AlmaLinux 8.9 in WSL2, Asterisk is compiled with --enable-dev-mode, menuselect options DONT_OPTIMIZE and MALLOC_DEBUG, everything else is standard

Tested against both certifed/18.9 (d03864911eeba347b4378fbade5e1aa3bc697855) at the time of posting as well as the most recent commit to master (06444290d048d8136acf4a3eae11888acf3c42bf) from github.

### Frequency of Occurrence

Constant

### Issue Description

Every CBAnn channel (unreal announcer channels created for making announcements to confbridge bridges) leaks a couple allocations in format_cap.c

I'm using the following extensions to test this:

```
[teststuff]
exten => confbridge,1,Answer()
same => n,Confbridge(testconf,default_bridge,default_user)

exten => 1,1,Answer()
same => n,Playback(hello-world)
same => n,Wait(20)
same => n,Playback(hello-world)
```

From here I start Asterisk (asterisk -gcvvv) and issue "memory show summary format_cap.c"

```
*CLI> memory show summary format_cap.c
     21168 bytes in        441 allocations by __ast_format_cap_append() line   205 of format_cap.c
     15296 bytes in         12 allocations by format_cap_framed_init() line   156 of format_cap.c
      4320 bytes in         10 allocations by format_cap_framed_init() line   164 of format_cap.c
       120 bytes in          3 allocations by      format_cap_init() line   109 of format_cap.c
     40904 bytes allocated in 466 selected allocations

  34362101 bytes in all allocations
    320435 bytes in deferred free large allocations
     29753 bytes in deferred free small allocations
    350188 bytes in deferred free allocations
  34712289 bytes in all allocations and deferred free allocations
```

I then originate using the following:

```
*CLI> channel originate LOCAL/1@teststuff extension confbridge@teststuff
```

Wait for the conference, confirm that the channels are all gone with 'core show channels', then check the memory summary again.

```
*CLI> core show channels
Channel                                                          Location                         State   Application(Data)
0 active channels
0 active calls
2 calls processed
*CLI>
*CLI>
*CLI> memory show summary format_cap.c
     21216 bytes in        442 allocations by __ast_format_cap_append() line   205 of format_cap.c
[...]

PR #946, opened 2024-10-15
core_unreal.c: Fix memory leak in ast_unreal_new_channels()

gtjoseph:
When the channel tech is multistream capable, the reference to
chan_topology was passed to the new channel.  When the channel tech
isn't multistream capable, the reference to chan_topology was never
released.  "Local" channels are multistream capable so it didn't
affect them but the confbridge "CBAnn" and the bridge_media
"Recorder" channels are not so they caused a leak every time one
of them was created.

Also added tracing to ast_stream_topology_alloc() and
stream_topology_destroy() to assist with debugging.

Resolves: #938


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph:
@JonathanRRose Confirmed that the PR fixes the leak.

jcolp (approved):

PR #947, opened 2024-10-15
core_unreal.c: Fix memory leak in ast_unreal_new_channels() (cert-18.9 only)

gtjoseph:
When the channel tech is multistream capable, the reference to
chan_topology was passed to the new channel.  When the channel tech
isn't multistream capable, the reference to chan_topology was never
released.  "Local" channels are multistream capable so it didn't
affect them but the confbridge "CBAnn" and the bridge_media
"Recorder" channels are not so they caused a leak every time one
of them was created.

Also added tracing to ast_stream_topology_alloc() and
stream_topology_destroy() to assist with debugging.

Resolves: #938


gtjoseph:
@JonathanRRose Confirmed that the PR fixes the leak.

jcolp (dismissed):

jcolp (approved):

PR #953, opened 2024-10-17
pjproject_bundled:  Tweaks to support out-of-tree development

gtjoseph:
* pjproject is now configured with --disable-libsrtp so it will
  build correctly when doing "out-of-tree" development.  Asterisk
  doesn't use pjproject for handling media so pjproject doesn't
  need libsrtp itself.

* The pjsua app (which we used to use for the testsuite) no longer
  builds in pjproject's master branch so we just skip it.  The
  testsuite no longer needs it anyway.

See third-party/pjproject/README-hacking.md for more info on building
pjproject "out-of-tree".

### DIFF
--- a/third-party/pjproject/Makefile
+++ b/third-party/pjproject/Makefile
@@ -58,6 +58,7 @@ ifeq ($(SPECIAL_TARGETS),)
         -include source/build.mak
         CF := $(filter-out -W%,$(CC_CFLAGS))
         CF := $(filter-out -I%,$(CF))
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

mbradeen (approved):

jcolp (approved):

Issue #960, opened 2024-10-23
[bug]: config.c: Template inheritance is not respected by ast_variable_update


InterLinked1:
### Severity

Major

### Versions

Git

### Components/Modules

config.c

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

This issue is complementary to #244 

244 has to do with GET operations, while this issue is the complementary issue/patch for SET operations.

Supposing a configuration such as this:
```
[template]
setting=0

[child](template)
```

Using the config APIs, if attempting to set child -> setting to `1`, we correctly get the following:
```
[template]
setting=0

[child](template)
setting = 1
```

If we want to set child -> setting back to `0`, that's where things go haywire.

Problems occur when toggling the value again, such as changing child's setting back to 0, which is not obeyed. When toggling the value back and forth, multiple (duplicate) `setting = 1` lines appear in the config file.

The problem here is that in `ast_variable_update`, the traversal encounters the template category value first, and will detect that the new value is the same as the template's value. However, this is not the value we should be comparing against, since it's not the effective value of the setting. Instead, we should use the *last* matching setting, not the *first* one, since if there are multiple, the first one is a template and the last one is the actual setting to update. This is very similar to the fix for #244, although this patch is even more important as the config update API is simply broken at present and does not behave in any logical manner.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #961, opened 2024-10-23
config.c Make ast_variable_update update last match.

InterLinked1:
ast_variable_update currently sets the first match for a variable, as opposed to the last one. This issue is complementary to that raised in #244.

This is incorrect and results in the wrong (or no) action being taken in cases where a section inherits from a template section. When the traversal occurs to update the setting, the existing code erroneously would use the first of possibly multiple matches in its update logic, which is wrong. Now, explicitly use the last match in the traversal, which will ensure that the actual setting is updated properly, and not skipped or ignored because a template from which the setting's section inherits was used for comparison.

Resolves: #960

UpgradeNote: Config variables, when set/updated, such as via AMI, will now have the corresponding setting updated, even if their sections inherit from template sections.

### DIFF
--- a/main/config.c
+++ b/main/config.c
@@ -1534,13 +1534,19 @@ int ast_variable_delete(struct ast_category *category, const char *variable, con
 int ast_variable_update(struct ast_category *category, const char *variable,
 						const char *value, const char *match, unsigned int object)
 {
-	struct ast_variable *cur, *prev=NULL, *newer=NULL;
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: none

Allan-N (approved):
Thank you for this fix!

jcolp (approved):

Issue #923, opened 2024-09-30
[bug]: Transport monitor shutdown callback only works on the first disconnection


escatg:
### Severity

Major

### Versions

18.23.1

### Components/Modules

res_pjsip/pjsip_transport_events, res_pjsip_outbound_registration

### Operating Environment

Ubuntu 20.04.6 LTS

### Frequency of Occurrence

Constant

### Issue Description

When an outbound registration is set up it also registers a transport monitor with a callback to be triggered when the transport disconnects, allowing to re-create the registration. In this case the type of the transport is TLS.

The first time the transport disconnects due to a network error, the callback is called, and a new registration is restarted, however it fails to register a new monitor (ast_sip_transport_monitor_register_replace_key returns AST_TRANSPORT_MONITOR_REG_NOT_FOUND).

Digging a bit more I see that the transport_monitor object created upon receiving the PJSIP_TP_STATE_CONNECTED event is actually directly removed shortly after creation. The cause is a late arrival of a PJSIP_TP_STATE_DESTROY event for the previous transport, which triggers a lookup by _transport key_, finds the new transport monitor and removes it. The new transport has the same transport key as the previous one since the destination host:port is the same. In the attached log, you can see how the "state:DESTROY" line is for transport tlsc0x7fc9342c5728 while it triggers MONITOR_DESTROYED for transport tlsc0x7fc9343cd918 that is the newly connected one.

Steps to reproduce:
- Set debug level 3 on res_pjsip
- Configure an outbound registration over TLS
- Use tcpkill or similar to cut the TLS connection

### Relevant log output

```shell
[2024-09-30 13:59:33.061609] DEBUG[439110] res_pjsip/pjsip_transport_events.c: Transport destination.example.com:5061(tlsc0x7fc9342c5728,TLS): RefCnt: 4 state:DISCONNECTED
[2024-09-30 13:59:33.061637] DEBUG[439110] res_pjsip/pjsip_transport_events.c: Transport destination.example.com:5061(tlsc0x7fc9342c5728,TLS): RefCnt: 4 state:SHUTDOWN
[2024-09-30 13:59:33.061645] DEBUG[439110] res_pjsip/pjsip_transport_events.c: Transport destination.example.com:5061(tlsc0x7fc9342c5728,TLS) RefCnt: 4 : running callback 0x7fc9afb9d8ed(0x7fc93400ea18)
[...]

PR #964, opened 2024-10-28
pjsip_transport_events: Avoid monitor destruction

escatg:
When a transport is disconnected, several events can arrive following
each other. The first event will be PJSIP_TP_STATE_DISCONNECT and it
will trigger the destruction of the transport monitor object. The lookup
for the transport monitor to destroy is done using the transport key,
that contains the transport destination host:port.

A reconnect attempt by pjsip will be triggered as soon something needs to
send a packet using that transport. This can happen directly after a
disconnect since ca

Subsequent events can arrive later like PJSIP_TP_STATE_DESTROY and will
also try to trigger the destruction of the transport monitor if not
already done. Since the lookup for the transport monitor to destroy is
done using the transport key, it can match newly created transports
towards the same destination and destroy their monitor object.

Because of this, it was sometimes not possible to monitor a transport
after one or more disconnections.

This fix adds an additional check on the transport pointer to ensure
only a monitor for that specific transport is removed.

Fixes: #923


escatg:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

mbradeen (approved):

jcolp (approved):

Issue #932, opened 2024-10-03
[bug]: When connected to multiple IP addresses the transport monitor is only set on the first one


escatg:
### Severity

Major

### Versions

18.24.3

### Components/Modules

res_pjsip/pjsip_transport_events, res_pjsip_outbound_registration

### Operating Environment

Ubuntu 20.04.6 LTS

### Frequency of Occurrence

None

### Issue Description

If you have one domain returning several A records and you use it as outbound proxy when setting up outbound registrations, PJSIP will automatically connect to each of these IP addresses.

The outbound registration module will setup a transport monitor for each registration, each time using `ast_sip_transport_monitor_register_replace_key`. Since the key is based on the destination domain, it is the same for each transport, so it ends up monitoring several times the first transport.

Example:
```
$ host registrar.example.com
registrar.example.com has address 192.0.2.66
registrar.example.com has address 192.0.2.67
registrar.example.com has address 192.0.2.68
registrar.example.com has address 192.0.2.69
```

```
> pjsip show registrations
pjsip show registrations

 <Registration/ServerURI..............................>  <Auth....................>  <Status.......>
==========================================================================================
 03522a0f-df87-4f0a-b9da-5b1e7e012a56/sip:example.com  example-auth  Registered        (exp. 1183s)
 (... 8 times ..)

> pjsip show transport-monitors
<Remote Host...................................> <State.....> <Direction> <RefCnt> <Monitors> <ObjName............>
 registrar.example.com:5061                      ACTIVE       Outgoing         5          8   tlsc0x7f75ac85b7b8
 registrar.example.com:5061                      ACTIVE       Outgoing         3          0   tlsc0x7f75ac8088e8
 registrar.example.com:5061                      ACTIVE       Outgoing         4          0   tlsc0x7f75ac81c1d8
 registrar.example.com:5061                      ACTIVE       Outgoing         4          0   tlsc0x7f75ac8e1ef8
```

This means that if `tlsc0x7f75ac85b7b8` is cut, every 8 registrations will be restarted, while some of them are actually using other connections. On the other hand if `tlsc0x7f75ac81c1d8`, the 2 registrations that use it won't be restarted.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #967, opened 2024-10-29
pjsip_transport_events: handle multiple addresses for a domain

escatg:
The key used for transport monitors was the remote host name for the
transport and not the remote address resolved for this domain.

This was problematic for domains returning multiple addresses as several
transport monitors were created with the same key.

Whenever a subsystem wanted to register a callback it would always end
up attached to the first transport monitor with a matching key.

The key used for transport monitors is now the remote address and port
the transport actually connected to.

Fixes: #932


escatg:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

Issue #948, opened 2024-10-06
[improvement]: Support SHA-256 algorithm on REGISTER and INVITE challenges


FurretUber:
### Is your feature or improvement request related to a problem? Please describe.

With the increasing usage of SHA-256 algorithm for authentication challenges on `WWW-Authenticate`, a normal Asterisk fails to REGISTER and to INVITE with some providers. The problem happens when SHA-256 is the only option available (so no multiple `WWW-Authenticate` lines.

### Describe the solution you'd like

Add support to the SHA-256 algorithm for the REGISTER and INVITE `WWW-Authenticate` challenge. PJSIP supports this and this works with MicroSIP. With an Asterisk patched to have SHA-256 available, it also seems to work.

### Describe alternatives you've considered

I'm sorry for asking for this feature.

### Additional context

I asked about this first on the [forum](https://community.asterisk.org/t/register-with-sha-256-algorithm/104763)

PR #972, opened 2024-10-31
Add SHA-256 and SHA-512-256 as authentication digest algorithms

gtjoseph:
* Refactored pjproject code to support the new algorithms and
added a patch file to third-party/pjproject/patches

* Added new parameters to the pjsip auth object:
  * password_digest = `<algorithm>:<digest>`
  * supported_algorithms_uac = List of algorithms to support
    when acting as a UAC.
  * supported_algorithms_uas = List of algorithms to support
    when acting as a UAS.
  See the auth object in pjsip.conf.sample for detailed info.

* Updated both res_pjsip_authenticator_digest.c (for UAS) and
res_pjsip_outbound_authentocator_digest.c (UAC) to suport the
new algorithms.

The new algorithms are only available with the bundled version
of pjproject, or an external version > 2.14.1.  OpenSSL version
1.1.1 or greater is required to support SHA-512-256.

Resolves: #948

UserNote: The SHA-256 and SHA-512-256 algorithms are now available
for authentication as both a UAS and a UAC.

### DIFF
--- a/res/res_pjsip/config_auth.c
+++ b/res/res_pjsip/config_auth.c
@@ -24,13 +24,106 @@
 #include "asterisk/logger.h"
 #include "asterisk/sorcery.h"
 #include "asterisk/cli.h"
+#include "asterisk/vector.h"
 #include "include/res_pjsip_private.h"
 #include "asterisk/res_pjsip_cli.h"
+
+#ifndef HAVE_PJSIP_AUTH_NEW_DIGESTS
+/*
+ * These are needed if the version of pjproject in use
+ * does not have the new digests.
+ * NOTE: We don't support AKA but we need to specify
+ * it to be compatible with the pjproject definition.
+ */
+#ifdef HAVE_OPENSSL
+#include "openssl/md5.h"
+#include "openssl/sha.h"
+#else
+#define MD5_DIGEST_LENGTH     16
+#define SHA256_DIGEST_LENGTH  32
+#endif
+
+const pjsip_auth_algorithm pjsip_auth_algorithms[] = {
+/*    TYPE                             IANA name            OpenSSL name */
+/*      Raw digest byte length  Hex representation length                */
+    { PJSIP_AUTH_ALGORITHM_NOT_SET,    {"", 0},             "",
+        0,                      0},
+    { PJSIP_AUTH_ALGORITHM_MD5,        {"MD5", 3},          "MD5",
+        MD5_DIGEST_LENGTH,      MD5_DIGEST_LENGTH * 2},
+    { PJSIP_AUTH_ALGORITHM_SHA256,     {"SHA-256", 7},      "SHA256",
+        SHA256_DIGEST_LENGTH,   SHA256_DIGEST_LENGTH * 2},
+    { PJSIP_AUTH_ALGORITHM_SHA512_256, {"SHA-512-256", 11}, "SHA512-256",
+        SHA256_DIGEST_LENGTH,   SHA256_DIGEST_LENGTH * 2},
+    { PJSIP_AUTH_ALGORITHM_AKAV1_MD5,  {"AKAv1-MD5", 9},    "",
+        MD5_DIGEST_LENGTH,      MD5_DIGEST_LENGTH * 2},
+    { PJSIP_AUTH_ALGORITHM_AKAV1_MD5,  {"AKAv2-MD5", 9},    "",
+        MD5_DIGEST_LENGTH,      MD5_DIGEST_LENGTH * 2},
+    { PJSIP_AUTH_ALGORITHM_COUNT,      {"", 0},             "",
+        0,                      0},
+};
+#endif
+
+const pjsip_auth_algorithm *ast_sip_auth_get_algorithm_by_type(
+	pjsip_auth_algorithm_type algorithm_type)
+{
+#ifdef HAVE_PJSIP_AUTH_NEW_DIGESTS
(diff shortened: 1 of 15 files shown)


gtjoseph:
This PR is in draft status because the pjproject patch has not been merged upstream yet.

gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph:
Just pushed an update to res_pjsip_authenticator_digest:digest_check_auth that skips trying to verify artificial auths since they can only be used to send challenges and can never be verified.

gtjoseph:
pjproject has approved the pull request related to the new patch in third-party/pjproject/patches and marked it for release with pjproject-2.15 therefore I'm marking this PR as ready-to-review.

gtjoseph:
pjproject patch merged

third-party/pjproject/patches/0001-Add-full-support-for-SHA-256-and-SHA-512-256-digest-.patch
```diff
@@ -0,0 +1,2443 @@
...
+     int wildcard = -1;
+ 
+     PJ_UNUSED_ARG(auth_scheme);
+ 
+     for (i=0; i<sess->cred_cnt; ++i) {
++        switch(sess->cred_info[i].data_type) {
++        case PJSIP_CRED_DATA_PLAIN_PASSWD:
++            /* PLAIN_PASSWD creds can be used for any algorithm other than AKA */
++            if (algorithm_type != PJSIP_AUTH_ALGORITHM_AKAV1_MD5
++                    || algorithm_type != PJSIP_AUTH_ALGORITHM_AKAV2_MD5) {
```
silentindark, line 1599: Small change in this commit: https://github.com/pjsip/pjproject/commit/c28635a2d3d27317c9965f7b9182063f20ca5a28.patch
From:
if (algorithm_type != PJSIP_AUTH_ALGORITHM_AKAV1_MD5
+                    || algorithm_type != PJSIP_AUTH_ALGORITHM_AKAV2_MD5) {
To:
            if (algorithm_type != PJSIP_AUTH_ALGORITHM_AKAV1_MD5
                    && algorithm_type != PJSIP_AUTH_ALGORITHM_AKAV2_MD5) {

gtjoseph: @silentindark Thanks for catching this.  Teluu will be releasing pjproject 2.15 shortly so we'll wait until we update bundled-pjproject to 2.15.  That way we'll get the new stuff plus the fix without needing the patch in third-party/pjproject/patches.

Moving back to draft status.

gtjoseph: Unresolving so the comment stays visible.

res/res_pjsip/config_auth.c
```diff
@@ -24,13 +24,101 @@
...
 #include "asterisk/res_pjsip_cli.h"
 
+#ifndef HAVE_PJSIP_AUTH_NEW_DIGESTS
+/*
+ * These are needed if the version of pjproject in use
+ * does not have the new digests.
+ * NOTE: We don't support AKA but we need to specify
+ * it to be compatible with the pjproject definition.
+ */
+#include "openssl/md5.h"
```
jcolp, line 38: We can be built without OpenSSL, so in that case I think this would cause a build failure.

res/res_pjsip_authenticator_digest.c
```diff
@@ -131,58 +139,132 @@ static const struct ast_sip_auth *get_auth(void)
...
-		return PJSIP_SC_FORBIDDEN;
+		/* This can only happen if the auth object was not saved to thread-local storage */
+		SCOPE_EXIT_RTN_VALUE(PJSIP_SC_FORBIDDEN, "%s:%s: No auth object found\n",
+			auth_name, src_name);
 	}
 
 	if (auth->type == AST_SIP_AUTH_TYPE_ARTIFICIAL) {
-		return PJSIP_SC_FORBIDDEN;
+		/*
+		 * This shouldn't happen because this function can only be invoked
```
jcolp, line 214: Wouldn't this be the case if we challenge on artificial auth, and they resent with Authorization?

jcolp (dismissed):

res/res_pjsip/config_auth.c
gtjoseph: done

res/res_pjsip_authenticator_digest.c
gtjoseph: There's no scenario where we would actually verify an artificial auth.  We can challenge with an artificial auth if the endpoint only has auth_username for instance, but when the challenge comes back it has a userid so we look that up to find the endpoint and auth object.  Besides, the artificial auth has no credentials to verify against.  This is checked down in digest_check_auth().

jcolp (dismissed):

jcolp:
Now needs #1020 before merging.

gtjoseph:
NOTE:  The only changes to this PR since last approval have been...
* Rebased to pull in bundled pjproject 2.15.1
* Updated configs/samples/pjsip.conf.sample
* Updated res/res_pjsip/pjsip_config.xml

You only need to re-review the two updated files.

jcolp (approved):

bkford (changes requested):

Issue #979, opened 2024-11-06
[improvement]: Add ability to suppress MOH when a remote endpoint sends "sendonly" or "inactive"


gtjoseph:
### Improvement Description

Let's say party A calls party B and party B returns a 180 Ringing followed by 183 progress message with an SDP that contains a "sendonly" or "inactive" attribute.  Since Asterisk (and common industry practice) takes those attributes to mean "hold", Asterisk will start playing MOH to party A.  If party B is sending early media or if your dialplan is already playing audio back to party A while the call is being set up, it'll be replaced with MOH.  There needs to be a way to not interrupt what's already playing to party A in this situation.

PR #980, opened 2024-11-06
res_pjsip: Add new endpoint option "suppress_moh_on_sendonly"

gtjoseph:
Normally, when one party in a call sends Asterisk an SDP with
a "sendonly" or "inactive" attribute it means "hold" and causes
Asterisk to start playing MOH back to the other party. This can be
problematic if it happens at certain times, such as in a 183
Progress message, because the MOH will replace any early media you
may be playing to the calling party. If you set this option
to "yes" on an endpoint and the endpoint receives an SDP
with "sendonly" or "inactive", Asterisk will NOT play MOH back to
the other party.

Resolves: #979

UserNote: The new "suppress_moh_on_sendonly" endpoint option
can be used to prevent playing MOH back to a caller if the remote
end sends "sendonly" or "inactive" (hold) to Asterisk in an SDP.

### DIFF
--- a/configs/samples/pjsip.conf.sample
+++ b/configs/samples/pjsip.conf.sample
@@ -1015,7 +1015,21 @@
                 ; using the same CHANNEL function if needed. Setting tenant ID here
                 ; will cause it to show up on channel creation and the initial
                 ; channel snapshot.
-
+;
+; suppress_moh_on_sendonly = no
+                ; Normally, when one party in a call sends Asterisk an SDP with
+                ; a "sendonly" or "inactive" attribute it means "hold" and
+                ; causes Asterisk to start playing MOH back to the other party.
+                ; This can be problematic if it happens at certain times, such
+                ; as in a 183 Progress message, because the MOH will replace
+                ; any early media you may be playing to the calling party. If
+                ; you set this option to "yes" on an endpoint and the endpoint
+                ; receives an SDP with "sendonly" or "inactive", Asterisk will
+                ; NOT play MOH back to the other party.
+                ; NOTE: This doesn't just apply to 183 responses.  MOH will
+                ; be suppressed when the attribute appears in any SDP received
+                ; including INVITEs, re-INVITES, and other responses.
+                ; (default: no)
 
 ;==========================AUTH SECTION OPTIONS=========================
 ;[auth]
--- a/res/res_pjsip/pjsip_config.xml
+++ b/res/res_pjsip/pjsip_config.xml
@@ -1533,6 +1533,27 @@
 				<configOption name="send_aoc" default="no">
 					<synopsis>Send Advice-of-Charge messages</synopsis>
 				</configOption>
+				<configOption name="suppress_moh_on_sendonly" default="no">
+					<synopsis>Suppress playing MOH to party A if party B sends
+					"sendonly" or "inactive" in an SDP</synopsis>
+					<description><para>
+					Normally, when one party in a call sends Asterisk an SDP with a "sendonly"
+					or "inactive" attribute it means "hold" and causes Asterisk to start
+					playing MOH back to the other party. This can be problematic if it happens at
+					certain times, such as in a 183 Progress message, because the MOH will
+					replace any early media you may be playing to the calling party. If you set
+					this option to "yes" on an endpoint and the endpoint receives an SDP
+					with "sendonly" or "inactive", Asterisk will NOT play MOH back to the other
+					party.
+					</para>
+					<note><para>
+					This doesn't just apply to 183 responses.  MOH will be suppressed when
+					the attribute appears in any SDP received including INVITEs, re-INVITES,
+					and other responses.
+					</para>
+					</note>
+					</description>
+				</configOption>
 			</configObject>
 			<configObject name="auth">
 				<synopsis>Authentication type</synopsis>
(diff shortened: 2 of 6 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

mbradeen (approved):

configs/samples/pjsip.conf.sample
```diff
@@ -1015,7 +1015,18 @@
                 ; using the same CHANNEL function if needed. Setting tenant ID here
                 ; will cause it to show up on channel creation and the initial
                 ; channel snapshot.
-
+;
+; suppress_moh_on_remote_hold = no
+                ; Normally, when Alice calls Bob and Bob sends Asterisk a
```
jcolp, line 1020: This is not accurate, as this extends beyond just a 183. It is during any SDP negotiation including before the call is answered or when it is answered - whether the "m" option is used or not.

res/res_pjsip/pjsip_config.xml
```diff
@@ -1533,6 +1533,21 @@
 				<configOption name="send_aoc" default="no">
 					<synopsis>Send Advice-of-Charge messages</synopsis>
 				</configOption>
+				<configOption name="suppress_moh_on_remote_hold" default="no">
+					<synopsis>Suppress playing MOH to party A if party B sends
+					"sendonly" or "inactive" in an SDP</synopsis>
+					<description><para>
+						Normally, when Alice calls Bob and Bob sends Asterisk a 183 Progress
```
jcolp, line 1540: Previous comment applies here as well.

res/res_pjsip_sdp_rtp.c
```diff
@@ -1528,7 +1528,7 @@ static void set_session_media_remotely_held(struct ast_sip_session_media *sessio
 			((addrs != NULL) && ast_sockaddr_is_any(addrs)) ||
 			pjmedia_sdp_media_find_attr2(media, "sendonly", NULL) ||
 			pjmedia_sdp_media_find_attr2(media, "inactive", NULL)) {
-			if (!session_media->remotely_held) {
+			if (!session_media->remotely_held && !session->endpoint->suppress_moh_on_remote_hold) {
 				session_media->remotely_held = 1;
```
jcolp, line 1532: If we don't set remotely_held, what occurs if we send a re-INVITE? Do we try to set it back to sendrecv?

jcolp (dismissed):

configs/samples/pjsip.conf.sample
gtjoseph: Of course.  I only documented the case that brought this up to keep the description simple.  I'll update it though.

jcolp (approved):

gtjoseph:
just a rebase

res/res_pjsip_sdp_rtp.c
gtjoseph: @jcolp Oops, I didn't see this earlier.

Here's what I tested...
Alice has suppress_moh_on_sendonly set to no.
Bob has suppress_moh_on_sendonly set no yes.
Alice calls Bob.
Bob answers then places call on hold by sending re-INVITE with sendonly.
Asterisk responds with 200 OK with recvonly.
Nothing happens to Alice's media session and no MOH is played.
Alice places call on hold by sending re-INVITE with sendonly.
Asterisk responds with 200 OK with recvonly.
Nothing happens to Bob's media session. He's still in sendonly/recvonly.

Now both are in sendonly/recvonly.

If Alice resumes with sendrecv, she hears nothing.
If Alice stays in sendonly/recvonly and Bob resumes with sendrecv, he hears MOH because it's not suppressed on Alice's endpoint.
When both resume with sendrecv, two-way audio is resumed.

If you have a specific scenario in mind let me know and I'll test it.

jcolp: I was more thinking about Asterisk generated re-INVITE, for example if a connected line update is done.

gtjoseph: Just to be sure nothing else is affected, I've moved the check for suppress_moh_on_sendonly down to the apply_negotiated_sdp_stream() function so the only thing it does is suppress the MOH without disturbing the remotely_held flag...

```c
if (session_media->remotely_held) {
	/* The remote side has put us on hold */
	if (!session->endpoint->suppress_moh_on_sendonly) {
		ast_queue_hold(session->channel, session->endpoint->mohsuggest);
		ast_rtp_instance_stop(session_media->rtp);
		ast_queue_frame(session->channel, &ast_null_frame);
	}
	session_media->remotely_held_changed = 0;
} else {
	/* The remote side has taken us off hold */
	if (!session->endpoint->suppress_moh_on_sendonly) {
		ast_queue_unhold(session->channel);
		ast_queue_frame(session->channel, &ast_null_frame);
	}
	session_media->remotely_held_changed = 0;
}
```
I originally had the check here but moved it up to set_session_media_remotely_held() because I thought other things triggered by remotely_held needed to be suppressed.  When I realized that wasn't the case, I meant to move it back to apply_negotiated_sdp_stream() but forgot.

I'll push the change up when I'm done testing.

gtjoseph: > I was more thinking about Asterisk generated re-INVITE, for example if a connected line update is done.

Ah connected line.  I can test that easily.

gtjoseph: Yep, was an issue with the original PR.  The session got renegotiated back to sendrecv.
With the update above, the re-INVITE goes back to Bob as recvonly.
PR all better now.

PR #981, opened 2024-11-06
res_pjsip: Add new endpoint option "suppress_moh_on_sendonly" (cert-18.9 only)

gtjoseph:
Normally, when one party in a call sends Asterisk an SDP with
a "sendonly" or "inactive" attribute it means "hold" and causes
Asterisk to start playing MOH back to the other party. This can be
problematic if it happens at certain times, such as in a 183
Progress message, because the MOH will replace any early media you
may be playing to the calling party. If you set this option
to "yes" on an endpoint and the endpoint receives an SDP
with "sendonly" or "inactive", Asterisk will NOT play MOH back to
the other party.

Resolves: https://github.com/asterisk/asterisk/issues/979

UserNote: The new "suppress_moh_on_sendonly" endpoint option
can be used to prevent playing MOH back to a caller if the remote
end sends "sendonly" or "inactive" (hold) to Asterisk in an SDP.

### DIFF
--- a/configs/samples/pjsip.conf.sample
+++ b/configs/samples/pjsip.conf.sample
@@ -979,7 +979,21 @@
                 ; using the same CHANNEL function if needed. Setting tenant ID here
                 ; will cause it to show up on channel creation and the initial
                 ; channel snapshot.
-
+;
+; suppress_moh_on_sendonly = no
+                ; Normally, when one party in a call sends Asterisk an SDP with
+                ; a "sendonly" or "inactive" attribute it means "hold" and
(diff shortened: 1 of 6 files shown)


gtjoseph:
cherry-pick-to: none

mbradeen (approved):

jcolp (changes requested):
Marking changes requested per comments on other review.

jcolp (dismissed):

gtjoseph:
just a rebase

gtjoseph:
issue with connected line fixed.

jcolp (approved):

Issue #982, opened 2024-11-06
[bug]: The addition of tenantid to the ast_sip_endpoint structure broke ABI compatibility


gtjoseph:
### Severity

Major

### Versions

all

### Components/Modules

res_pjsip

### Operating Environment

All

### Frequency of Occurrence

None

### Issue Description

The tenantid field was added 

### Relevant log output

```shell
The tenantid field was originally added to the ast_sip_endpoint structure at the end of the AST_DECLARE_STRING_FIELDS block.  This caused everything after it in the structure to move down in memory and break ABI compatibility.  It needs to be moved to the end of the ast_sip_endpoint structure.
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

PR #984, opened 2024-11-06
res_pjsip: Move tenantid to end of ast_sip_endpoint (cert-18.9 only)

gtjoseph:
The tenantid field was originally added to the ast_sip_endpoint
structure at the end of the AST_DECLARE_STRING_FIELDS block.  This
caused everything after it in the structure to move down in memory
and break ABI compatibility.  It's now at the end of the structure
as an AST_STRING_FIELD_EXTENDED.  Given the number of string fields
in the structure now, the initial string field allocation was
also increased from 64 to 128 bytes.

Resolves: #982


gtjoseph:
cherry-pick-to: none

mbradeen (approved):

jcolp (approved):

Issue #921, opened 2024-09-30
[bug]: Stir-Shaken doesn’t allow B or C attestation for unknown callerid which is allowed by ATIS-1000074.v003, §5.2.4


oza4h07:
### Severity

Major

### Versions

20.10

### Components/Modules

stir_shaken

### Operating Environment

Debian 11

### Frequency of Occurrence

Constant

### Issue Description

When sending an outbound call from an unknown number, according ATIS-1000074.v003, §5.2.4, an ITSP should attest the call with a B level as it:

1.   he is responsible of the origination of the SIP call
2.   he has a direct authenticated relationship with the customer
3.   he has not established a verified relation with the callerid

As the number is unknown to him/her, the ITSP cannot configure in advance any corresponding TN entry.

Asterisk should allow the ITSP to define a default attestation B evel for those "unknown TNs".
Currently, as of 20.10.0-rc2, Asterisk doesn't define such default attestation level  and Asterisk sends these calls (from unrkown numbers) without any Identity header.
According some local regulations, a call without any Identity header may be dropped by any transit or termination provider.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
@oza4h07 Please test PR #987 and see if it does what you need.

PR #987, opened 2024-11-08
res_stir_shaken: Allow sending Identity headers for unknown TNs

gtjoseph:
Added a new option "unknown_tn_attest_level" to allow Identity
headers to be sent when a callerid TN isn't explicitly configured
in stir_shaken.conf.  Since there's no TN object, a private_key_file
and public_cert_url must be configured in the attestation or profile
objects.

Since "unknown_tn_attest_level" uses the same enum as attest_level,
some of the sorcery macros had to be refactored to allow sharing
the enum and to/from string conversion functions.

Also fixed a memory leak in crypto_utils:pem_file_cb().

Resolves: #921

UserNote: You can now set the "unknown_tn_attest_level" option
in the attestation and/or profile objects in stir_shaken.conf to
enable sending Identity headers for callerid TNs not explicitly
configured.

### DIFF
--- a/configs/samples/stir_shaken.conf.sample
+++ b/configs/samples/stir_shaken.conf.sample
@@ -99,6 +99,20 @@ One of "A", "B", "C"
 
 Default: none
 
+-- unknown_tn_attest_level --------------------------------------------
+Attestation level to use for unknown TNs.
+One of "A", "B", "C"
+
+Normally if a callerid TN isn't configured in stir_shaken.conf
+no Identity header will be created. If this option is set,
+however, an Identity header will be sent using this
+attestation level.  Since there's no TN object, you must
+ensure that a private_key_file and public_cert_url are
+configured in the attestation or profile objects for
+this to work.
+
+Default: none
+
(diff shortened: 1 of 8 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
testsuite_test_pr: 75

oza4h07:
I'm currently testing this PR. I'll report my findings, ASAP.

oza4h07:
With this PR applied, my testings that previously failed, are now passing with expected results !
At the moment, I only tested after setting `unknown_tn_attest_level` in profile section (*).

Thank you very, very much for this PR.

(*) I've got no reason to believe it wouldn't also work when configuring in attestation section

gtjoseph:
@oza4h07 Thanks for testing!

jcolp (approved):

oza4h07:
Am I correct thinking this PR is not included in Asterisk 20.11.0 but will be in 20.12.0 ?

jcolp:
Yes.

PR #1002, opened 2024-11-21
manager.c: Add Processed Call Count to CoreStatus output

mikepultz:
This update adds the processed call count to the CoreStatus AMI Action responsie. This output is similar to the values returned by "core show channels" or "core show calls" in the CLI.

UserNote: The current processed call count is now returned as CoreProcessedCalls from the CoreStatus AMI Action.


mikepultz:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1007, opened 2024-11-27
[improvement]: Cannot dynamically add queue member in paused state from dialplan or command line


viktike:
### Improvement Description

Adding a queue member dynamically in paused state is important because the following dialplan has issues:

```
1,AddQueueMember(markq,Local/${EXTEN}@${CONTEXT})
n,PauseQueueMember(markq,Local/${EXTEN}@${CONTEXT})
```

If _markq_ had waiting callers, the interface will ring.

The AMI Action: QueueAdd has a Paused: flag, but the Dialplan application AddQueueMember() has not and the `queue add member` CLI command also has not.

This fix also updates the sample configuration, fixing the depracated and removed SIP/ channel technology.

PR #1008, opened 2024-11-27
app_queue: allow dinamically adding a queue member in paused state.

viktike:
app_queue: allow dinamically adding a queue member in paused state.

Fixes: #1007

UserNote: use the p option of AddQueueMember() for paused member state. Optionally, use the r(reason) option to specify a custom reason for the pause.


jcolp (changes requested):
Commit message should have "dynamically" instead of "dinamically"

jcolp (approved):

Issue #1013, opened 2024-12-02
[improvement]: chan_pjsip: Send VIDUPDATE RTP frames for H.264 streams on endpoints without WebRTC


maximilianfridrich:
### Improvement Description

Currently, when a `chan_pjsip` channel receives a `VIDUPDATE` indication, an RTP `VIDUPDATE` frame is only queued on a H.264 stream if WebRTC is enabled on that endpoint. This restriction does not really make sense and a `VIDUPDATE` RTP frame should be written even if WebRTC is not enabled (as is the case for VP8, VP9, and H.265 streams).

I have a patch, feel free to assign this issue to me.

PR #1014, opened 2024-12-02
chan_pjsip: Send VIDUPDATE RTP frame for all H.264 streams

maximilianfridrich:
Currently, when a chan_pjsip channel receives a VIDUPDATE indication, an RTP VIDUPDATE frame is only queued on a H.264 stream if WebRTC is enabled on that endpoint. This restriction does not really make sense.

Now, a VIDUPDATE RTP frame is written even if WebRTC is not enabled (as is the case for VP8, VP9, and H.265 streams).

Resolves: #1013


maximilianfridrich:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

PR #1025, opened 2024-12-10
res_odbc: release threads from potential starvation.

jkroonza:
Whenever a slot is freed up due to a dead connection, or a connection failure, wake up another consumer.

Also fix the pointer log for ODBC handle %p dead which would always reflect NULL.

### DIFF
--- a/res/res_odbc.c
+++ b/res/res_odbc.c
@@ -945,6 +945,7 @@ struct odbc_obj *_ast_odbc_request_obj2(const char *name, struct ast_flags flags
 				if (odbc_obj_connect(obj) == ODBC_FAIL) {
 					ast_mutex_lock(&class->lock);
 					class->connection_cnt--;
+					ast_cond_signal(&class->cond);
 					ast_mutex_unlock(&class->lock);
 					ao2_ref(obj->parent, -1);
 					ao2_ref(obj, -1);
@@ -971,17 +972,18 @@ struct odbc_obj *_ast_odbc_request_obj2(const char *name, struct ast_flags flags
 			/* If the connection is dead try to grab another functional one from the
 			 * pool instead of trying to resurrect this one.
 			 */
-			ao2_ref(obj, -1);
-			obj = NULL;
-
 			ast_mutex_lock(&class->lock);
 
 			class->connection_cnt--;
+			/* this thread will re-acquire, and if that fails will signal,
+			 * thus no need to signal class->cond here */
 			ast_debug(2, "ODBC handle %p dead - removing from class '%s', new count is %zd\n",
 				obj, name, class->connection_cnt);
 
 			ast_mutex_unlock(&class->lock);
 
+			ao2_ref(obj, -1);
+			obj = NULL;
 		} else {
 			/* We successfully grabbed a connection from the pool and all is well!
 			 */


res/res_odbc.c
```diff
@@ -971,17 +972,17 @@ struct odbc_obj *_ast_odbc_request_obj2(const char *name, struct ast_flags flags
 			/* If the connection is dead try to grab another functional one from the
 			 * pool instead of trying to resurrect this one.
 			 */
-			ao2_ref(obj, -1);
-			obj = NULL;
-
 			ast_mutex_lock(&class->lock);
 
 			class->connection_cnt--;
+			ast_cond_signal(&class->cond);
```
jkroonza, line 978: Since obj = NULL, we will loop again, so I'm not 100% convinced of this one.

res/res_odbc.c
```diff
@@ -971,17 +972,17 @@ struct odbc_obj *_ast_odbc_request_obj2(const char *name, struct ast_flags flags
...
 			ast_debug(2, "ODBC handle %p dead - removing from class '%s', new count is %zd\n",
 				obj, name, class->connection_cnt);
 
 			ast_mutex_unlock(&class->lock);
 
+			ao2_ref(obj, -1);
+			obj = NULL;
```
jkroonza, line 985: Just moved down to the logging just above will have a non-NULL value for obj.  Of course we can ao2_ref(obj, -1) in original location, and just obj = NULL after logging, but personally I believe these two lines should always stick together.

jkroonza:
I'm somewhat debating the merit of this, but we're definitely seeing cases where func_odbc is deadlocking, whilst investigating a (potentially unrelated, in that case max_connections was just legitimately too low) issue I found this potential race when I was looking to better understand the implementation.  I *suspect*, but wondering how to prove, that users can get stuck on this piece of code here, and towards that, the next time we see this I will grab a bt all.

I don't see that this patch can hurt, but it certainly could potentially resolve issues in cases where we suddenly have a massive burst of connections going into the blocking state whilst other connections end up failing to establish DB connections.

Let's assume max_connections=10 and in a VERY SHORT burst we suddenly get 20 object requests, 10 of those enter the connection establishment (and end up failing), but 10 remain blocking on the cond_wait.  As the other 10 fail, I do think that the blocking 10 should be released to also fail, rather than blocking.

I don't see how this patch will completely stop our deadlock situation (we're simply seeing a bunch of channels getting stuck on func_odbc calls, and I've never properly investigated until now), but it will certainly prevent some potential thread starvation issues that might be happening.

jkroonza:
cherry-pick-to: 20, 21, 22

res/res_odbc.c
```diff
@@ -971,17 +972,17 @@ struct odbc_obj *_ast_odbc_request_obj2(const char *name, struct ast_flags flags
 			/* If the connection is dead try to grab another functional one from the
 			 * pool instead of trying to resurrect this one.
 			 */
-			ao2_ref(obj, -1);
-			obj = NULL;
-
 			ast_mutex_lock(&class->lock);
 
 			class->connection_cnt--;
+			/* ast_cond_signal(&class->cond); - this thread will re-acquire, and that fails will signal */
```
jcolp, line 978: Remove the ast_cond_signal in even the comment and just explain in a comment, I also think you missed an "if": "and if that fails"

jcolp (changes requested):

res/res_odbc.c
jkroonza: Should be better now?

jcolp (approved):

PR #1026, opened 2024-12-10
res_odbc: use a queue rather than a stack for cached connections.

jkroonza:
This hould result in lower risk of connection drops due to shorter idle times (the first connection to go onto the stack could in theory never be used again, ever, but sit there consuming resources, there could be multiple of these).

And with a queue rather than a stack, dead connections are guaranteed to be detected and purged eventually.

This should end up better balancing connection_cnt with actual load over time.


jkroonza:
cherry-pick: 20, 21, 22

jkroonza:
I don't see how this change could have caused the failure, unstable test?

jcolp (changes requested):
You missed a letter in the commit message "hould" instead of "should".

I'd also add a UserNote that the behavior will change so that if multiple connections are active we will go through them instead of preferring the same one like before.

jcolp:
Can you elaborate why it is better for you?

jkroonza:
> I think the strategy for releasing connections should be configured, because both strategies have merit. In my case releasing connections to the top is better.

@alex2grad I'm open to suggestions but I cannot imagine any use-case where the stack-based approach would be better, happy to put in the effort to make this configurable if you can justify a stack approach over a queue approach.

alex2grad:
> > I think the strategy for releasing connections should be configured, because both strategies have merit. In my case releasing connections to the top is better.
> 
> @alex2grad I'm open to suggestions but I cannot imagine any use-case where the stack-based approach would be better, happy to put in the effort to make this configurable if you can justify a stack approach over a queue approach.

When connections are released at the top of the list, connections are quickly reused, which maintains the performance of the database server, for example, by keeping cached query plans in the database active. Most database queries are authentication queries, which are the same SQL statement.

When connections are released at the top of the list, inactive connections can be dropped by the database server to reduce the total number of active connections on the server.
Example: the database server is overloaded for a short period, during this time asterisk opens more connections up to max_connections (for example, max_connections=100, usually 10 active connections).
When the server load returns to normal, asterisk continues to use only the top 10 connections, and the remaining 90 inactive connections can be dropped by the database server. If connections are released at the bottom of the list, asterisk continues to use all 100 connections.

jkroonza:
> > > I think the strategy for releasing connections should be configured, because both strategies have merit. In my case releasing connections to the top is better.
> > 
> > 
> > @alex2grad I'm open to suggestions but I cannot imagine any use-case where the stack-based approach would be better, happy to put in the effort to make this configurable if you can justify a stack approach over a queue approach.
> 
> When connections are released at the top of the list, connections are quickly reused, which maintains the performance of the database server, for example, by keeping cached query plans in the database active. Most database queries are authentication queries, which are the same SQL statement.

Most query caches that I've worked with are shared, so not per-connection.  And due to lock contention I generally switch them off, it's in my experience faster to just re-use the index compared to the query cache, especially when retrieving single-rows at a time.

> When connections are released at the top of the list, inactive connections can be dropped by the database server to reduce the total number of active connections on the server. Example: the database server is overloaded for a short period, during this time asterisk opens more connections up to max_connections (for example, max_connections=100, usually 10 active connections). When the server load returns to normal, asterisk continues to use only the top 10 connections, and the remaining 90 inactive connections can be dropped by the database server. If connections are released at the bottom of the list, asterisk continues to use all 100 connections.

Now you end up with 90  dead connections permanently cached on asterisk side which will never get cleared.  This is what I was hoping to address.  And should result in fewer reconnects overall.

Using the queue, and assuming the connections don't cycle insanely fast, the connections in the queue may (and should) reach that timeout still.  Which means that you may end up with say 20 connections eventually cycling the 10 needed.  So I get your point.  If they cycle fast enough you'll get stuck having maxconnections *actually* open, not just "pretend open" on asterisk side (which btw would cause other spike issues on asterisk side since you could then in a single thread loop to clean out say 90 connections on a thread just to re-establish a single new connection).

Two possible counter mechanisms I can think off:

1.  max_cache setting which you can set to say 5 or so to not cache more than 5 connections (I'd still use a queue).
2.  drop every nth connection returned to the pool rather than caching.

Number one would have somewhat high resource usage directly after the spike, or be a pain during a very jittery load spike (ie, nort particularly adaptive), compared to the latter which would adjust more slowly, but has the downside of forcing some re-connects.

One could always implement both of those, it's relatively simple:

1.  Keep a counter of connections in the cache (yes, one could count the linked list, easier to keep a counter) (cache_cnt and cycle_cnt).
2.  Keep the value of the limits (max_cache, drop_cycle); and
3.  Drop connections as appropriate.

No matter how I reason this, a stack here rather than a queue does not add up for me.  By default both of the cleanup mechanisms could be inactive, or with conservative defaults (max_cache defaults to half of maxconnections, and drop_cycle to say 10 x maxconnections).  I reckon we'd be setting drop_cycle to something closer to maxconnections itself, maybe 2x, but this is all gut-feeling-speculation.

Thoughts?

alex2grad:
> Now you end up with 90 dead connections permanently cached on asterisk side which will never get cleared. This is what I was hoping to address. And should result in fewer reconnects overall.

Why do you think dead connections will never get cleared?
The asterisk removes dead connection from the pool and disconnects it.
`                } else if (connection_dead(obj, class)) {
                        /* If the connection is dead try to grab another functional one from the
                         * pool instead of trying to resurrect this one.
                         */
                        ao2_ref(obj, -1);
                        obj = NULL;
        
                        ast_mutex_lock(&class->lock);
        
                        class->connection_cnt--;
                        ast_debug(2, "ODBC handle %p dead - removing from class '%s', new count is %zd\n",
                                obj, name, class->connection_cnt);
        
                        ast_mutex_unlock(&class->lock);

static void odbc_obj_destructor(void *data)
{
        struct odbc_obj *obj = data;

        odbc_obj_disconnect(obj);
}

static odbc_status odbc_obj_disconnect(struct odbc_obj *obj)
{
        int res;
        SQLINTEGER err;
        short int mlen;
        unsigned char msg[200], state[10];
        SQLHDBC con;
        
        /* Nothing to disconnect */
        if (!obj->con) {
                return ODBC_SUCCESS;
        }
        
        con = obj->con;
        obj->con = NULL;
        res = SQLDisconnect(con);
        
        if ((res = SQLFreeHandle(SQL_HANDLE_DBC, con)) == SQL_SUCCESS) {
                ast_debug(3, "Database handle %p (connection %p) deallocated\n", obj, con);
        } else {
                SQLGetDiagRec(SQL_HANDLE_DBC, con, 1, state, &err, msg, 100, &mlen);
                ast_log(LOG_WARNING, "Unable to deallocate database handle %p? %d errno=%d %s\n", con, res, (int)err, msg);
        }
 
        return ODBC_SUCCESS;
}

`

jkroonza:
@alex2grad could you please check if the attached would be more to your liking?  Unless I screwed up the defaults should retain historic behaviour.

Just to be clear on why this is a bit of a sticky point.  We have spikes on SOME systems where we need high connection counts (upwards of 100), and a number of these systems connect to the same database host, so we need to be able to adjust dynamically fairly quickly as not all systems are equally active.  We do use fairly short network timeouts on the database side, but these also can't be too low, we'd prefer if asterisk could control it's cache growth (and shrinking) again more dynamically.  We will probably use a combination of cache_size ~ half of max_connections and cache_drop ~ twice max_connections (which in other environments seems to work well, whether that'll be the case for asterisk type environments remains to be seen).

alex2grad:
> @alex2grad could you please check if the updated would be more to your liking? Unless I screwed up the defaults should retain historic behaviour.

I'll look at it.

> Just to be clear on why this is a bit of a sticky point. We have spikes on SOME systems where we need high connection counts (upwards of 100), and a number of these systems connect to the same database host, so we need to be able to adjust dynamically fairly quickly as not all systems are equally active. We do use fairly short network timeouts on the database side, but these also can't be too low, we'd prefer if asterisk could control it's cache growth (and shrinking) again more dynamically. We will probably use a combination of cache_size ~ half of max_connections and cache_drop ~ twice max_connections (which in other environments seems to work well, whether that'll be the case for asterisk type environments remains to be seen).

I see that you are going to implement a more advanced connection pooling. Have you checked the unixODBC connection pooling?

jkroonza:
To be honest, no I have not.  Haven't looked at unixODBC at all.  Don't know how it works, and not in the mood for a complete overhaul.  The changes that were made are fairly simple and (experimentally to me at least) known to work well.

jcolp:
We used UnixODBC connection pooling in the beginning and it did not work. In fact, it had a bug that was subsequently fixed (based on our report) which caused rampant memory usage. I had to drop what I was doing at the time and fix things up as it was so broken, which is why this was implemented in res_odbc. Going back to UnixODBC connection pooling is not an option.

alex2grad:
> We used UnixODBC connection pooling in the beginning and it did not work. In fact, it had a bug that was subsequently fixed (based on our report) which caused rampant memory usage. I had to drop what I was doing at the time and fix things up as it was so broken, which is why this was implemented in res_odbc. Going back to UnixODBC connection pooling is not an option. It would require a full audit of the UnixODBC version across common distros to see how much the minimum version requirement would change, full testing, a statement of why and the benefits, and more testing. It's not a "oh, we'll just use UnixODBC" thing. We did that before. It was a disaster that I don't want to repeat.

Why do you categorically deny the use of UnixODBC connection pooling by the user?
I think it's been a while since there were problems with UnixODBC connection pooling. A lot has been fixed and improved since then. I recently tested asterisk with UnixODBC connection pooling and found no problems. That's why I suggested to @jkroonza to test UnixODBC solution with connection pooling, and not to invent it again. I did not suggest that the asterisk project use or recommend UnixODBC connection pooling.

jcolp:
I was referring to any changes within Asterisk itself required to do that, if any, and fundamentally changing the recommendations and approach. Your response made it seem as though you were sending @jkroonza down the road of that and changing the Asterisk recommendation to it.

What people do on their own is up to them.

alex2grad:
> I was referring to any changes within Asterisk itself required to do that, if any, and fundamentally changing the recommendations and approach. Your response made it seem as though you were sending @jkroonza down the road of that and changing the Asterisk recommendation to it.
> 
> What people do on their own is up to them.

There may have been a misunderstanding. I wanted to suggest @jkroonza to test the UnixODBC connection pooling solution by setting up connection pooling in UnixODBC without any changes to Asterisk.

alex2grad:
I always thought that stack is the best way for DВ connection pools.
Here I found a discussion "When should connection pools be FIFO v. LIFO?"
https://www.reddit.com/r/PostgreSQL/comments/y2y2vl/when_should_connection_pools_be_fifo_v_lifo/

jkroonza:
> > @alex2grad could you please check if the updated would be more to your liking? Unless I screwed up the defaults should retain historic behaviour.
> > Just to be clear on why this is a bit of a sticky point. We have spikes on SOME systems where we need high connection counts (upwards of 100), and a number of these systems connect to the same database host, so we need to be able to adjust dynamically fairly quickly as not all systems are equally active. We do use fairly short network timeouts on the database side, but these also can't be too low, we'd prefer if asterisk could control it's cache growth (and shrinking) again more dynamically. We will probably use a combination of cache_size ~ half of max_connections and cache_drop ~ twice max_connections (which in other environments seems to work well, whether that'll be the case for asterisk type environments remains to be seen).
> 
> What is your goal? Free up resources used by asterisk for "dead" connections? Improve performance? I looked at your implementation and I think it will hurt performance because during heavy load asterisk will not cache some connections, which means asterisk will have to open a new connection more often, which is not good for performance.
> 
> For example:
> 
>     1. max_cache
>        max_connection=100, max_cache=50
>        During normal load, when there are less then 50 active connections, nothing changes.
>        During have load, when more than 50 connections are required, asterisk allows to establish up to 100 connections, but caches only 50, which means that all DB requests over 50 active connection must open/close a connection. This results in a significant performance drop.
> 
>     2. drop_nth
>        max_connection=100, drop_nth=10
>        During normal load, there may be a small performance drop, as a connection may need to be opened/closed for every 10th DB request.
>        During have load, every 10th DB request must open/close a connection, which results in at least a 10% performance drop.
> 
> 
> I think the best solution for cleaning up "dead" connections is a thread that will periodically check for "dead" or inactive connections and destroy them.

My goal is both to keep connections open and active, as well as to get dead connections out of the system faster (or completely avoid them), and manage the size of the cache.  The other options was added as a mechanism to manage the cache size overall to hopefully avoid too many connections going into the dead state in the first place.  The defaults should retain current behaviour (ie, max_cache >= max_connections will always cache, and drop_nth = 0 won't ever drop connections.  And by default now the cache will remain a stack.

I want to be able to set the max connections high enough to never be the blocking point, but also not keep an insane amount of database connections open.  We've seen systems where normally we only need a few tens of connections to never block, but then suddenly require a few hundred, thus max_connections is 1000 there, now we end up in a situation where that ends up never getting cleared.  Probably not the biggest issue as far as problems go, thus why we haven't investigated previously, but it's still a waste, and a source for possible jitter ... imagine you've got 900 dead connections on that stack that is then suddenly getting cleaned up by a single blocking thread.

After this I figured another mechanism may be a form of RED ... so up to max_cache we always cache connections, once we exceed max_cache we can either drop every nth, or use a random drop that's proportional to how far we are down the line to max connections, eg, max_cache = 50, max_connections = 100.  This is essentially the way SSH handles incoming connection management and IMHO it's a pain in that specific usage (accepting connections), but for releasing connections should be fine.

If we're to return connection number 75 to the cache, then we calculate (75-50)/(100-50) and get 0.5 (this number will always be negative if we ignore the lower bound for the check, with a max value of 1), with a random number in the range [0,1] being < calculated value then we drop the connection.

This is all about managing the cache *size*.  They're just options.

We've found in other environments that it's often better to tear down in a "distributed" manner if the cache is going to get too big rather than having to in another thread deal (during I need to connect) with too many dead connections.

Another idea may be to set a mechanism like max_cache_time, which will track for each connection the last time it was returned to the pool, and then a task worker that periodically looks at the head (or tail) of each cache and closes that connection if this time has been exceeded.

Or to base the decision whether or not to cache the returned connection based on the age of the head/tail of the list.

Happy to implement all of these mechanisms here, that would be three more then:

1. cache_size_is_low_level_mark = no (if yes, will start "random dropping" of connections where the probably is proportional to how far above cache_size we are such that cache_size represents a 0% probability of dropping, and max_connections reflects a 100% probability).
2. max_cache_time = 0 (if > 0 a task processor will periodically check the head/tail of each cache and drop those connections where the time in cache exceeds this).
3. max_cache_cycle = 0 (if >0 will when connections gets returned check the age of the oldest connection in the cache and if that connection has been in the cache for longer than max_cache_cycle, drop that connection from the cache).

Both 2 and 3 has a similar effect looking at that more simply put, it's really just the mechanism / time of check that differs, and obviously, which thread does the work.

Given that we likely want to drop the connection that's been in the cache the longest (most likely to be dropped by the server) I'll rework the existing code as well to always cache the just returned connection, and drop the head/tail of the cache rather (the connection that's been in the cache the longest).

alex2grad:
For most databases, a long-lived connection is usually more productive than a newly established one.
Therefore, I do not recommend dropping long-lived connections.
I recommend dropping  an "inactive" connection if it has not been used for a certain number of seconds.

jkroonza:
> For most databases, a long-lived connection is usually more productive than a newly established one. Therefore, I do not recommend dropping long-lived connections. I recommend dropping an "inactive" connection if it has not been used for a certain number of seconds.

In other words, the ones that have been waiting longest in the cache and has been idle?

alex2grad:
> > For most databases, a long-lived connection is usually more productive than a newly established one. Therefore, I do not recommend dropping long-lived connections. I recommend dropping an "inactive" connection if it has not been used for a certain number of seconds.
> 
> In other words, the ones that have been waiting longest in the cache and has been idle?

Not quite.
Keep the stack-based connection cache.
Need a new option cache_inactive_timeout - time for unused connections to remain in the cache in seconds.
When starting to use a connection, set is_used=1.
When stopping to use a connection, set expire_time = current time + cache_inactive_timeout and is_used=0.
Need a res_odbc thread that will check for unused inactive connections from the tail of the list.
If the connection is_used==0 and expire_time < current time, then drop/destroy this connection.
Or maybe to avoid creating a new thread we can check the last connection in the list every time we stop using the current connection.
I prefer the first option with the check thread.

jkroonza:
That's exactly the additional options 2 and 3 in my long post yesterday.

gtjoseph:
@jkroonza You need to fix the cherry-pick comment.  Each branch needs to be on its own line.

jkroonza:
> @jkroonza You need to fix the cherry-pick comment. Each branch needs to be on its own line.

Done.

There are however the further improvements that @alex2grad is pushing for, which I don't object to.  So the first commit I reckon is ok, and the latter two will do what they were intended to do, but I think they can be improved.  Will rework tomorrow.  Marking as WIP in the meantime.

alex2grad:
> That's exactly the additional options 2 and 3 in my long post yesterday.

I understood your options 2 and 3 differently, as if long-lived connections would be dropped regardless of whether they were used recently or not, since your description says nothing about inactivity. And even the name of the options speaks about the lifetime of the connections, and not about the inactivity of the connection in the cache.

jkroonza:
Sorry Alex, with max_cache_time I meant the timer restarts every time it gets returned to the cache, not from when the connection was made.  That's another option, but frankly, that would be pointless.  Keeping connections open is good *as long as they get used*, otherwise we're just wasting resources (both on asterisk and database side).

I'm thinking max_cache_size still has use, and probably so does drop-nth (as mentioned earlier, we use it elsewhere to successfully shrink caches, but I do see your concern and it's valid), but in both cases I do agree with drop the connection that has been *idle* the longest, not the connection that is being returned to the cache (current implemention), in other words, when these limits (drop-nth or max_cache_size) gets hit, still cache the returned connection, but then immediately dequeue (or pop) the connection that *would have been returned for the next request* and drop that connection instead, such that the connection that has been *idle* for the longest time gets dropped rather than the one that was *just busy* or *idle for the shortest time*.

I'm just wondering if max_cache_size should get a companion that affects the drop-nth mechanism to only do that if the cache size is larger than that watermark?  In other words, something like min_cache_drop_level where connections will never be dropped if the cache_size is below that (possibly not even long-idle connections, except if they're dead, more below).

Then max_cache_time is possibly a better control mechanism overall anyway, but we can use this same worker thread to periodically perform an aliveness check on the longest idle connection irrespective of if the max_cache_time has been exceeded or not.  But since this check would *probably* happen under lock (or we need to dequeue and then undo if the connection is alive), and as such I'm in two minds, as it may block other threads from de-queueing, so I think max_cache_time should be enforced irrespective of min_cache_drop_level (if the cache cycle time is that high, then probably the low level is too high anyway).

Given min_cache_drop_level we can then also do a RED mechanisms where if cache size exceeds min_cache_drop_level it does a proportional random drop.

Is this more clear?  And what's your thoughts?

alex2grad:
Honestly, I read it 3 times and got even more confused.
My opinion:
Keep the stack-based connection cache.
Drop inactive connections that remain in cache for more then N seconds, from the tail of the list.
As I described in https://github.com/asterisk/asterisk/pull/1026#issuecomment-2549148404

jkroonza:
I improved the max_cache_size patch:

1. Renamed the variable, so that a max_cache_time variable can be added too.
2. Drop longest-idle connection (fast in case of queue, slow in case of stack, due to having to traverse the full list - as long as max_cache_size is relatively low this shouldn't really matter).

For the time being:

Admin can choose FIFO (queue) or LIFO (stack).

Admin can choose how big to grow the cache.

If database closes the connections, behaviour in stack will remain unchanged.  In case of queue it becomes less likely that server will close connection, but if it does it will get detected more quickly.

I think some stats should be added here too to the class status:

* cache underflows (number of times new connections had to be opened).
* cache overflows (number of times we dropped connections due to cache growing too big).

I *think* this can be merged as is, and additional follow-ups can be made using separate PRs.

jkroonza:
> @jkroonza
> 
>     1. can you squash these 2 commits?

I can.  I do think logically they're two separate changes though.

>     2. The descriptions "the longest idle connection in the cache shall be dropped" is incorrect, because your patch does not incorporate a timestamp or tracking mechanism to determine the longest idle connection.

on it's not.  In case of queue the next entry in the queue will be the longest idle, irrespective of when it was requeued, in the case of a stack it'll be the entry at the bottom, same argument.

>     3. Still need max_cache_time to drop only inactive connections which have been waiting for a long time, otherwise it is very likely that the connection just used will be dropped.

See above.  This can now be implemented separately.

>     4. By default max_cache_size should be 0.
>        to check "if (max_cache_size && class->cur_cache >= class->max_cache_size)"
>        also do not print "Cache Usage" if max_cache_size is 0.

I don't think it matters much whether 0 or MAX_UINT is used.  I think a better change may be to use -1 as "infinite" and 0 as "cache disabled" perhaps rather?  Allowing use of unixODBC connection pooling as you suggested?

configs/samples/res_odbc.conf.sample
```diff
@@ -79,6 +79,22 @@ pre-connect => yes
...
+; Set the type of connection cache.  Historically res_odbc used a stack, resulting in
+; a single connection being re-used frequently, whilst other connections could end up
+; dying.  Using a queue or round-robin strategy results in all connections being re-used
+; and kept alive.  Each strategy has pro's and con's as debated on
+; https://github.com/asterisk/asterisk/pull/1026.  The default remains a stack to be backwards
+; compatible, setting this value to any of rr, roundrobin or queue will result
+; in a round-robin queue being used.
+;cache_type => roundrobin
+;
+; Limit the number of connections to cache.  On extremely spiky loads this could be
```
jcolp, line 92: What's the difference between max_connections and cache_size? I think that interaction needs to be described here clearly.

configs/samples/res_odbc.conf.sample
```diff
@@ -79,6 +79,22 @@ pre-connect => yes
 ; default this is set to 5000 milliseconds (or 5 seconds). If you would like to
 ; disable the WARNING message it can be set to "0".
 ;slow_query_limit => 5000
+;
+; Set the type of connection cache.  Historically res_odbc used a stack, resulting in
+; a single connection being re-used frequently, whilst other connections could end up
+; dying.  Using a queue or round-robin strategy results in all connections being re-used
+; and kept alive.  Each strategy has pro's and con's as debated on
```
jcolp, line 86: pros and cons

res/res_odbc.c
```diff
@@ -823,7 +845,33 @@ void ast_odbc_release_obj(struct odbc_obj *obj)
...
+		--class->connection_cnt;
+		ast_mutex_unlock(&class->lock);
+
+		ast_debug(2, "ODBC Pool '%s' exceeded cache size, dropping '%p', connection count is %zd (%u cached)\n",
+			class->name, obj, class->connection_cnt, class->cur_cache);
+
+		ao2_ref(obj, -1);
+
+		ast_mutex_lock(&class->lock);
+	} else
```
jcolp, line 873: } else {

res/res_odbc.c
```diff
@@ -919,6 +967,8 @@ struct odbc_obj *_ast_odbc_request_obj2(const char *name, struct ast_flags flags
 		ast_mutex_lock(&class->lock);
 
 		obj = AST_LIST_REMOVE_HEAD(&class->connections, list);
+		if (obj)
```
jcolp, line 970: if (obj) {

jcolp (changes requested):

jkroonza:
Notes:

1. connection cache can be disabled by setting max_cache_size = 0, which will effectively disable the cache.  unixodbc connection pooling can then (I think) be configured in unixodbc itself.
2. hopefully the config descriptions aren't too verbose, and are coherent (sorry, not native english, I tend to screw things up, so please do double check me there).


Open questions:

1. To squash or not to squash.
2. To first implement max_cache_time prior to merge, or just do this as is for now.

max_cache_time will require:

1. Adding a return to pool timer for each connection.
2. Having a separate task processor that runs periodically, or at calculated intervals to prune connections that have been in the queue too long.

Other tuning statistics that might be helpful:

1. Number of connections requested
2. Number of such requests served from the cache
3. Number of cache_size overflows (ie, connections dropped due to max_cache_size exceeded).
4. Number of dead connections dropped from the pool due to not functioning any more.
5. Max number of connections used ever.
6. Maximum time a connection was idle in the cache before re-use.
7. Minimum idle time for in-cache pooling.

Of course, all of these statistics takes up additional resources (ram to store, cpu to calculate).  And the benefit is probably marginal.  I'm quite happy with the current implementation and feel that these can be added later, but should not be overly much effort to do now if it's desired.

asteriskteam:

jcolp:
> Notes:
> 
>     1. connection cache can be disabled by setting max_cache_size = 0, which will effectively disable the cache.  unixodbc connection pooling can then (I think) be configured in unixodbc itself.
> 
>     2. hopefully the config descriptions aren't too verbose, and are coherent (sorry, not native english, I tend to screw things up, so please do double check me there).
> 
> 
> Open questions:
> 
>     1. To squash or not to squash.
> 
>     2. To first implement max_cache_time prior to merge, or just do this as is for now.
> 
> 
> max_cache_time will require:
> 
>     1. Adding a return to pool timer for each connection.
> 
>     2. Having a separate task processor that runs periodically, or at calculated intervals to prune connections that have been in the queue too long.
> 
> 
> Other tuning statistics that might be helpful:
> 
>     1. Number of connections requested
> 
>     2. Number of such requests served from the cache
> 
>     3. Number of cache_size overflows (ie, connections dropped due to max_cache_size exceeded).
> 
>     4. Number of dead connections dropped from the pool due to not functioning any more.
> 
>     5. Max number of connections used ever.
> 
>     6. Maximum time a connection was idle in the cache before re-use.
> 
>     7. Minimum idle time for in-cache pooling.
> 
>     8. Number of requests not serviced due to max_connections exceeded.
> 
>     9. Number of requests not serviced due to connection failures.
> 
> 
> Of course, all of these statistics takes up additional resources (ram to store, cpu to calculate). And the benefit is probably marginal. I'm quite happy with the current implementation and feel that these can be added later, but should not be overly much effort to do now if it's desired.

I don't personally care either way for merging, and as for max_cache_time due to the complexity that should be done separate. At this point noone else is really clamoring for all the tuning statistics you've mentioned, so I don't think they should be done/added unless circumstances require them.

gtjoseph:
@jkroonza Can you remove the "signed-off-by" lines from the commit messages or move them above the UserNotes?  Where they are now will make them part of the UserNotes.

jkroonza:
Hmm, looking at the failures I can't imagine any of those are because of anything I've done ... suggestions?

gtjoseph:
> Hmm, looking at the failures I can't imagine any of those are because of anything I've done ... suggestions?

Not your fault.  There's a new test that assumes a more recent asterisk pull request.  Don't worry about it.

Issue #1029, opened 2024-12-16
[bug]: chan_dahdi: Wrong channel state set when RINGING received


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

devicestate.c, chan_dahdi

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

When chan_dahdi receives RINGING, it sets the state to `AST_STATE_RINGING`, which is wrong, since this refers to lines that are actually themselves ringing, not ones listening to audible ringback tone from the peer channel.

The documentation for `AST_STATE_RINGING` and `AST_STATE_RING` is also flipped.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

InterLinked1:
Okay, I did some tests and I think I see what's going on here.

In this example, 25 calls 27, so you'd expect one channel's state would be Ring and one would be Ringing. In fact, *both* are Ringing, which makes absolutely no sense. Consequently, both also have a device state of `RINGING`, which is wrong: only 27 is RINGING, and 25's device state should just be `INUSE` since it's off-hook.

```
[2024-12-17 17:08:04.921]     -- DAHDI/27-1 is ringing
[2024-12-17 17:08:04.922]     -- DAHDI/27-1 is making progress passing it to DAHDI/25-1
proliant*CLI> core show channels
Channel                                                          Location                         State   Application(Data)
DAHDI/27-1                                                       27@from-internal:1               Ringing AppDial((Outgoing Line))
DAHDI/25-1                                                       27@fxs:1                         Ringing Dial(DAHDI/27)
2 active channels
1 active call
3 calls processed
proliant*CLI> dialplan eval function DEVICE_STATE(DAHDI/25)
Return Value: Success (0)
Result: RINGING
[2024-12-17 17:08:12.919]     -- DAHDI/27-1 is ringing
proliant*CLI> dialplan eval function DEVICE_STATE(DAHDI/27)
Return Value: Success (0)
Result: RINGING
[2024-12-17 17:08:15.094]     -- Hanging up on 'DAHDI/27-1'
```

So, the above suggest's the `chan_dahdi` is setting the channel state wrong somewhere, though which one is wrong isn't super clear just from that.

Another example with an IAX2 to PJSIP call, IAX2 has state Ring while PJSIP has state Ringing. Good that they're different, and in this case, the device states are logical.

```
polypbx*CLI> core show channels
[...]

PR #1030, opened 2024-12-16
chan_dahdi: Fix wrong channel state when RINGING recieved.

InterLinked1:
Previously, when AST_CONTROL_RINGING was received by
a DAHDI device, it would set its channel state to
AST_STATE_RINGING. However, an analysis of the codebase
and other channel drivers reveals RINGING corresponds to
physical power ringing, whereas AST_STATE_RING should be
used for audible ringback on the channel. This also ensures
the correct device state is returned by the channel state
to device state conversion.

Since there seems to be confusion in various places regarding
AST_STATE_RING vs. AST_STATE_RINGING, some documentation has
been added or corrected to clarify the actual purposes of these
two channel states, and the associated device state mapping.

An edge case that prompted this fix, but isn't explicitly
addressed here, is that of an incoming call to an FXO port.
The channel state will be "Ring", which maps to a device state
of "In Use", not "Ringing" as would be more intuitive. However,
this is semantic, since technically, Asterisk is treating this
the same as any other incoming call, and so "Ring" is the
semantic state (put another way, Asterisk isn't ringing anything,
like in the cases where channels are in the "Ringing" state).

Since FXO ports don't currently support Call Waiting, a suitable
workaround for the above would be to ignore the device state and
instead check the channel state (e.g. IMPORT(DAHDI/1-1,CHANNEL(state)))
since it will be Ring if the FXO port is idle (but a call is ringing
on it) and Up if the FXO port is actually in use. (In both cases,
the device state would misleadingly be "In Use".)

Resolves: #1029

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -9395,8 +9395,11 @@ static int dahdi_indicate(struct ast_channel *chan, int condition, const void *d
 				if ((ast_channel_state(chan) != AST_STATE_RING) ||
 					((p->sig != SIG_FXSKS) &&
 				 (p->sig != SIG_FXSLS) &&
-				 (p->sig != SIG_FXSGS)))
-				ast_setstate(chan, AST_STATE_RINGING);
+				 (p->sig != SIG_FXSGS))) {
+					/* We're playing audible ringback tone on the channel,
+					 * so set state to AST_STATE_RING, not AST_STATE_RINGING. */
+					ast_setstate(chan, AST_STATE_RING);
+				}
 			}
 			break;
 		case AST_CONTROL_INCOMPLETE:
--- a/include/asterisk/channelstate.h
+++ b/include/asterisk/channelstate.h
@@ -37,8 +37,8 @@ enum ast_channel_state {
 	AST_STATE_RESERVED,		/*!< Channel is down, but reserved */
 	AST_STATE_OFFHOOK,		/*!< Channel is off hook */
 	AST_STATE_DIALING,		/*!< Digits (or equivalent) have been dialed */
-	AST_STATE_RING,			/*!< Line is ringing */
-	AST_STATE_RINGING,		/*!< Remote end is ringing */
+	AST_STATE_RING,			/*!< Remote end is ringing (e.g. listening to audible ringback tone). Also often used for initial state for a new channel. */
+	AST_STATE_RINGING,		/*!< Line is ringing */
 	AST_STATE_UP,			/*!< Line is up */
 	AST_STATE_BUSY,			/*!< Line is busy */
 	AST_STATE_DIALING_OFFHOOK,	/*!< Digits (or equivalent) have been dialed while offhook */
--- a/main/devicestate.c
+++ b/main/devicestate.c
@@ -179,8 +179,8 @@ static const struct chan2dev {
 	{ AST_STATE_RESERVED,        AST_DEVICE_INUSE },
 	{ AST_STATE_OFFHOOK,         AST_DEVICE_INUSE },
 	{ AST_STATE_DIALING,         AST_DEVICE_INUSE },
-	{ AST_STATE_RING,            AST_DEVICE_INUSE },
-	{ AST_STATE_RINGING,         AST_DEVICE_RINGING },
+	{ AST_STATE_RING,            AST_DEVICE_INUSE }, /* Audible ringback tone */
+	{ AST_STATE_RINGING,         AST_DEVICE_RINGING }, /* Actual ringing */
 	{ AST_STATE_UP,              AST_DEVICE_INUSE },
 	{ AST_STATE_BUSY,            AST_DEVICE_BUSY },
 	{ AST_STATE_DIALING_OFFHOOK, AST_DEVICE_INUSE },


InterLinked1:
cherry-pick-to: none

alex2grad:
Maybe I'm wrong, but that's how I understand it, this is for PJSIP:
R1. PJSIP endpoint makes outgoing call, while remote party is ringing, the PJSIP channel state is AST_STATE_RING.
R2. PJSIP endpoint receives incoming calls, while PJSIP endpoint is ringing, the PJSIP channel state is AST_STATE_RINGING.
Which means in case R1 the PJSIP endpoint is INUSE, in case R2 the PJSIP endpoint is RINGING.
BLF status for PJSIP endpoint in case R1 is "In use", in case of R2 is "Ringing" and can be picked up.
Everything works correctly.

InterLinked1:
> Maybe I'm wrong, but that's how I understand it, this is for PJSIP: R1. PJSIP endpoint makes outgoing call, while remote party is ringing, the PJSIP channel state is AST_STATE_RING. R2. PJSIP endpoint receives incoming calls, while PJSIP endpoint is ringing, the PJSIP channel state is AST_STATE_RINGING. Which means in case R1 the PJSIP endpoint is INUSE, in case R2 the PJSIP endpoint is RINGING. BLF status for PJSIP endpoint in case R1 is "In use", in case of R2 is "Ringing" and can be picked up. Everything works correctly.

Yeah, it's certainly possible that `chan_dahdi` is simply using the wrong channel state (`Ring` vs `Ringing` here) and that is what should be fixed. That didn't seem super consistent when I was looking around so wanted to float this to get other perspectives. What you're saying makes sense to me.

Intuitively I'd think it would be the other way around from how it is, especially given the existence of `prering`.

In that case, it seems like usage of `AST_STATE_RING` should be audited as there are several places where the wrong one is used.

InterLinked1:
> Maybe I'm wrong, but that's how I understand it, this is for PJSIP: R1. PJSIP endpoint makes outgoing call, while remote party is ringing, the PJSIP channel state is AST_STATE_RING. R2. PJSIP endpoint receives incoming calls, while PJSIP endpoint is ringing, the PJSIP channel state is AST_STATE_RINGING. Which means in case R1 the PJSIP endpoint is INUSE, in case R2 the PJSIP endpoint is RINGING. BLF status for PJSIP endpoint in case R1 is "In use", in case of R2 is "Ringing" and can be picked up. Everything works correctly.

So I double-checked this and it seems like it is in fact the more intuitive way, at least that's how they're defined: https://github.com/asterisk/asterisk/blob/0bf31781c302f5bc6612c539eec7a19c056599b7/include/asterisk/channelstate.h#L40-L41

So this suggests `AST_STATE_RING` should be used when a phone line is physically ringing, and `AST_STATE_RINGING` is the state when listening to audible ringback tone.

However, the mapping is defined exactly the opposite: https://github.com/asterisk/asterisk/blob/0bf31781c302f5bc6612c539eec7a19c056599b7/main/devicestate.c#L182-L183

Going by the previous definitions and my understanding, it would make sense for `AST_STATE_RINGING` to map to `AST_DEVICE_INUSE` and `AST_STATE_RING` to map to `AST_STATE_RINGING`... in which case I *think* the original change here seems correct, but incomplete (the other one should be swapped).

If anyone's seeing something here that I'm not, please correct me.

alex2grad:
https://docs.asterisk.org/Configuration/Interfaces/Asterisk-REST-Interface-ARI/Introduction-to-ARI-and-Channels/ARI-and-Channels-Manipulating-Channel-State/
"A channel's state reflects the current state of the path of communication between Asterisk and a device."

Party A calls Party B via Asterisk.
There are 2 channels: Channel A (between Asterisk and device Party A) and Channel B (between Asterisk and device Party B).
There are 2 device states: state of device Party A and state of device Party B.
Party A picked up the handset and dialed a number and Party B is RINGING (for SIP, the Party B sent 180 RINGING).
The channel A state is RING (from the asterisk's point of view the 'Line is ringing') and device Party A state is IN USE.
The channel B state is RINGING (from the asterisk's point of view the 'Remote end is ringing') and device Party B state is RINGING.

alex2grad:
May be the name and comment of state "AST_STATE_RING: Line is ringing" not quite correct.
May be it should be AST_STATE_ALERTING or AST_STATE_PROGRESS and comment "It indicates that called party alerting/ringing has been initiated".
So I think the mapping for channel states to device states is correct.
It may be necessary to change or add more detailed comments for ast_channel_state and for mapping.

InterLinked1:
> There are 2 channels: Channel A (between Asterisk and device Party A) and Channel B (between Asterisk and device Party B).
There are 2 device states: state of device Party A and state of device Party B.
Party A picked up the handset and dialed a number and Party B is RINGING (for SIP, the Party B sent 180 RINGING).
The channel A state is RING (from the asterisk's point of view the 'Line is ringing') and device Party A state is IN USE.
The channel B state is RINGING (from the asterisk's point of view the 'Remote end is ringing') and device Party B state is RINGING.

But based on the way the comments are written, they seem to imply A would have state RINGING (and device state should be INUSE) and B would be RING (and device state would be RINGING).

I haven't tested this with non-DAHDI channels yet to see what the channel states are, I will see if there is a similar pattern there.

> May be the name and comment of state "AST_STATE_RING: Line is ringing" not quite correct. May be it should be AST_STATE_ALERTING or AST_STATE_PROGRESS and comment "It indicates that called party alerting/ringing has been initiated". So I think the mapping for channel states to device states is correct. It may be necessary to change or add more detailed comments for ast_channel_state and for mapping.

It's definitely confusing. It may be right, though something is definitely wrong somewhere. I see similar usage for `AST_STATE_RING` in other places, so it's not super clear. Hopefully someone from the project can shed some additional insight on this.

jcolp:
You're unlikely to find any answer from anyone with any certainty, as this stuff predates all of us. The only history is the mailing list archive and the commits.

alex2grad:
The current mapping is correct for PJSIP channels. 
If you change the mapping, then a lot of things in PJSIP will break.
I think that at the present time most asterisk installations use SIP, not DANDI.
So I think that there is no need to change the mapping, but rather fix the usage of the AST_STATE_RING/AST_STATE_RINGING in DANDI.

InterLinked1:
> The current mapping is correct for PJSIP channels. If you change the mapping, then a lot of things in PJSIP will break. I think that at the present time most asterisk installations use SIP, not DANDI. So I think that there is no need to change the mapping, but rather fix the usage of the AST_STATE_RING/AST_STATE_RINGING in DANDI.

Except it isn't DAHDI logic, it's core Asterisk logic in `devicestate.c` (the fallback logic used for every channel driver). So I think this is likely broken for a bunch of stuff.

InterLinked1:
Okay, I've hashed everything out and narrowed this down to a minor fix within chan_dahdi, along with some documentation fixes. This shouldn't be any compatibility concerns either.

alex2grad:
> Okay, I've hashed everything out and narrowed this down to a minor fix within chan_dahdi, along with some documentation fixes. There shouldn't be any compatibility concerns either.

I recommend splitting this into 2 PRs:
 fix for chan_dahdi
 and comments for channel states and mapping for channel states to device states .
I don't agree with new comments.
 1. There is no physical/power ring for SIP. 
 2. Your comments are probably correct for DANDI, but completely incorrect for SIP. And since this is channel_state and mapping is used for all types of channels, there should not be a description only for DANDI.

InterLinked1:
> > Okay, I've hashed everything out and narrowed this down to a minor fix within chan_dahdi, along with some documentation fixes. There shouldn't be any compatibility concerns either.
> 
> I recommend splitting this into 2 PRs: fix for chan_dahdi and comments for channel states and mapping for channel states to device states .
> 
> I don't agree with new comments.
> 
> 1. There is no physical/power ring for SIP.
> 2. Your comments are probably correct for DANDI, but completely incorrect for SIP. And since this is channel_state and mapping is used for all types of channels, there should not be a description only for DANDI.

Sure, I can make the comments a little bit more generic, but the documentation comments *are* backwards and do still need to be fixed.

alex2grad:
My opinion
AST_STATE_RING - the caller is listening to audible ringback tone
AST_STATE_RINGING - the called is ringing

alex2grad:
BTW, the title and descriptions of the issue https://github.com/asterisk/asterisk/issues/1029 should be updated

jcolp (approved):

bkford (approved):

PR #1035, opened 2024-12-31
Header fixes for compiling C++ source files

gtjoseph:
A few tweaks needed to be done to some existing header files to allow them to
be compiled when included from C++ source files.

logger.h had declarations for ast_register_verbose() and
ast_unregister_verbose() which caused C++ issues but those functions were
actually removed from logger.c many years ago so the declarations were just
removed from logger.h.


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

PR #1041, opened 2025-01-04
Add C++ Standard detection to configure and fix a new C++20 compile issue

gtjoseph:
* The autoconf-archive package contains macros useful for detecting C++
  standard and testing other C++ capabilities but that package was never
  included in the install_prereq script so many existing build environments
  won't have it.  Even if it is installed, older versions won't newer C++
  standards and will actually cause an error if you try to test for that
  version. To make it available for those environments, the
  ax_cxx_compile_stdcxx.m4 macro has copied from the latest release of
  autoconf-archive into the autoconf directory.

* A convenience wrapper(ast_cxx_check_std) around ax_cxx_compile_stdcxx was
  also added so checking the standard version and setting the
  asterisk-specific PBX_ variables becomes a one-liner:
  `AST_CXX_CHECK_STD([std], [force_latest_std])`.
  Calling that with a version of `17` for instance, will set PBX_CXX17
  to 0 or 1 depending on whether the current c++ compiler supports stdc++17.
  HAVE_CXX17 will also be 'defined" or not depending on the result.

* C++ compilers hardly ever default to the latest standard they support.  g++
  version 14 for instance supports up to C++23 but only uses C++17 by default.
  If you want to use C++23, you have to add `-std=gnu++=23` to the g++
  command line.  If you set the second argument of AST_CXX_CHECK_STD to "yes",
  the macro will automatically keep the highest `-std=gnu++` value that
  worked and pass that to the Makefiles.

* The autoconf-archive package was added to install_prereq for future use.

* Updated configure.ac to use AST_CXX_CHECK_STD() to check for C++
  versions 11, 14, 17, 20 and 23.

* Updated configure.ac to accept the `--enable-latest-cxx-std` option which
  will set the second option to AST_CXX_CHECK_STD() to "yes".  The default
  is "no".

* ast_copy_string() in strings.h declares the 'sz' variable as volatile and
  does an `sz--` on it later.  C++20 no longer allows the `++` and `--`
  increment and decrement operators to be used on variables declared as
  volatile however so that was changed to `sz -= 1`.


gtjoseph:
cherry-pick-to: none

gtjoseph:
git won't correctly merge the changes to the generated ./configure and ./include/asterisk/autoconf.h.in files to other branches so this will be done manually.

PR #1045, opened 2025-01-07
POC: Alternate Channel Storage Backends

gtjoseph:
Full details: http://s.asterisk.net/f4c7d153

This commit is a Proof of Concept for Alternate Channel Storage Backends
and not meant to be merged.

It contains the following alternate backends in addition to the
original "ao2_legacy" container backend (default)...

* ao2_optimized
* cpp_map_name
* cpp_map_name_id
* boost_map_name
* boost_multi_index
* sqlite3

Testing the Boost alternatives requires the boost-devel(rh) or
libboost-dev(deb) packages to be installed.

To select which backends are compiled, use the new Alternate Channel
Storage Backends category in menuselect.  Compiling them all increases the
size of the "asterisk" binary by about 20% (wth the boost alternatives
accounting for 16 of that 20%) but doesn't affect performance otherwise.

To select which one is used at runtime, set the "channel_storage_backend"
option in asterisk.conf to one of the values described in
asterisk.conf.sample.

YOU ARE ENCOURAGED TO TEST but please read the full documentation linked
above before doing so.


gtjoseph:
Final PR #1227 is up for review.

PR #1048, opened 2025-01-07
samples: Use "asterisk" instead of "postgres" for username

seuros:
These files were using postgres instead of asterisk.

### DIFF
--- a/configs/samples/cel_pgsql.conf.sample
+++ b/configs/samples/cel_pgsql.conf.sample
@@ -3,71 +3,70 @@
 ;
 
 ; Sample Asterisk config file for CEL logging to PostgreSQL
-;
+
 ; CEL field names:
-;
-;	eventtype
-;	  CHANNEL_START = 1
-;	  CHANNEL_END = 2
-;	  HANGUP = 3
-;	  ANSWER = 4
-;	  APP_START = 5
-;	  APP_END = 6
-;	  BRIDGE_START = 7
-;	  BRIDGE_END = 8
-;	  CONF_START = 9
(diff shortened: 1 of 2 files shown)


configs/samples/cel_pgsql.conf.sample
```diff
@@ -3,71 +3,70 @@
...
-;	appdata
-;	accountcode
-;	peeraccount
-;	uniqueid
-;	linkedid
-;	amaflags  (an int)
-;	userfield
-;	peer
-;	extra
+;   eventtype
```
gtjoseph, line 8: Was the change in formatting intentional or an artifact of your editor saving the file?

gtjoseph (changes requested):
"chore(psql)" isn't going to mean much in a a changelog.  How about...

samples: Use "asterisk" instead of "postgres" for username

Also, the cherry-pick comments need to be in a separate comment and formatted correctly.

configs/samples/cel_pgsql.conf.sample
seuros: It was intentional to follow the spacing of the other files.

I can revert back.

gtjoseph: Nope that's fine.

seanbright (dismissed):
Sorry to bikeshed but we don't use the 'chore' thing on this project, so rather than:

`chore(samples): ...`

Just:

`samples: ...`

seuros:
@seanbright Thanks for pointing that out! I've made the change to follow the project's convention.

seanbright (approved):

PR #1050, opened 2025-01-08
logger.c fix: malformed JSON template

DooMMasteR:
this typo was mentioned before, but never got fixed.  https://community.asterisk.org/t/logger-cannot-log-long-json-lines-properly/87618/6

(I found it elsewhere too but wasn't able to dig it up)

We are hunting down an issue where the logger fails to log our Error (probably caused by invalid chars in the caller-ID).

### DIFF
--- a/main/logger.c
+++ b/main/logger.c
@@ -287,7 +287,7 @@ static int format_log_json(struct logchannel *channel, struct logmsg *msg, char 
 	}
 
 	json = ast_json_pack("{s: s, s: s, "
-		"s: {s: i, s: s} "
+		"s: {s: i, s: s}, "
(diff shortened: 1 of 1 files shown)


DooMMasteR:
cherry-pick: 22
cherry-pick: 21
cherry-pick: 20

seanbright (approved):
This is fine.

The reality is that [this bug has existed for over 9 years](https://github.com/asterisk/asterisk/commit/2d7a4a3357cbaea7acd8fb202c28512980508c88), so the correct PR is probably to remove it entirely rather than to fix it.

jcolp (approved):

PR #1051, opened 2025-01-08
res_config_pgsql: normalize database connection options

seuros:
The syntax will match the other modules cel and cdr.

Let me know if this is the correct way to correct syntaxes.

Also i noticed that this module is the only one that use socket as option.

### DIFF
--- a/res/res_config_pgsql.c
+++ b/res/res_config_pgsql.c
@@ -1441,7 +1441,7 @@ static int parse_config(int is_reload)
 	config = ast_config_load(RES_CONFIG_PGSQL_CONF, config_flags);
 	if (config == CONFIG_STATUS_FILEUNCHANGED) {
 		if (is_reload && pgsqlConn && PQstatus(pgsqlConn) != CONNECTION_OK) {
-			ast_log(LOG_WARNING,  "PostgreSQL RealTime: Not connected\n");
+			ast_log(LOG_WARNING, "PostgreSQL RealTime: Not connected\n");
 		}
 		return 0;
 	}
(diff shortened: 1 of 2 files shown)


seuros:
cherry-pick-to: none

res/res_config_pgsql.c
```diff
@@ -1459,61 +1459,66 @@ static int parse_config(int is_reload)
...
-	if (!ast_strlen_zero(dbhost)) {
-		/* No socket needed */
-	} else if (!(s = ast_variable_retrieve(config, "general", "dbsock"))) {
-		ast_log(LOG_WARNING,
-				"PostgreSQL RealTime: No database socket found, using '/tmp/.s.PGSQL.%d' as default.\n", dbport);
-		strcpy(dbsock, "/tmp");
-	} else {
-		ast_copy_string(dbsock, s, sizeof(dbsock));
-	}
+    if ((s = ast_variable_retrieve(config, "general", "dbuser"))) {
```
jcolp, line 1462: The indenting on this appears to be incorrect and not to coding guidelines.

res/res_config_pgsql.c
```diff
@@ -1459,61 +1459,66 @@ static int parse_config(int is_reload)
...
+        ast_log(AST_LOG_WARNING, "Option 'dbuser' has been deprecated in favor of 'user'. Please make that change in your res_pgsql.conf.\n");
```
jcolp, line 1463: I don't see any benefit to deprecating the old configuration option names. Supporting both is fine for consistency, but deprecating the old ones for the sake of deprecating them in the name of consistency doesn't sit well with me.

jcolp (changes requested):

res/res_config_pgsql.c
seuros: Should we support both in the other modules ?

jcolp: I don't think so - one should just be picked and stuck with.

seuros: got it.

res/res_config_pgsql.c
```diff
@@ -1551,7 +1570,6 @@ static int parse_config(int is_reload)
 
 	ast_verb(2, "PostgreSQL RealTime reloaded.\n");
 
-	/* Done reloading. Release lock so others can now use driver. */
```
jcolp, line 1554: Why this removal?

seuros: vim :D .

I fixed the indentation with vim.

Restored.

jcolp (approved):

Issue #1054, opened 2025-01-10
[bug]: chan_iax2: Frames unnecessarily backlogged with jitterbuffer if no voice frames have been received yet


InterLinked1:
### Severity

Major

### Versions

Git

### Components/Modules

chan_iax2

### Operating Environment

Jitterbuffer enabled for the channel

### Frequency of Occurrence

Constant

### Issue Description

If the jitter buffer is enabled, on channels where no voice frames have yet been received, non-voice frames can get backlogged unnecessarily by code that was added to address a crash by using a NULL format. In this case, the normal variables containing the format have not yet been set. We do have a separate variable, `chosenformat`, specifically used for saving the negotiated format for channel creation. This fix leverages that to use as a fallback to ensure that we can process frames  in the jitterbuffer prior to a voice format being received.

### Relevant log output

```shell
[2025-01-09 19:34:36.960] WARNING[2062818]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[2025-01-09 19:34:36.961] WARNING[2062822]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[2025-01-09 19:34:36.961] WARNING[2062824]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[2025-01-09 19:34:36.961] WARNING[2062819]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[2025-01-09 19:34:36.962] WARNING[2062821]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[2025-01-09 19:34:36.962] WARNING[2062825]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[2025-01-09 19:34:36.963] WARNING[2062816]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[2025-01-09 19:34:36.963] WARNING[2062817]: chan_iax2.c:4213 __get_from_jb: No voice format and no peer format available on IAX2/REDACTED, backlogging frame
[...]

PR #1055, opened 2025-01-10
chan_iax2: Always set chosenformat upon call connection.

InterLinked1:
Currently, when receiving an unauthenticated call, we keep track of the negotiated format in the chosenformat, which allows us to later create the channel using the right format. However, this was not done for authenticated calls. This meant that in certain circumstances, if we had not yet received a voice frame from the peer, only certain other types of frames (e.g. text), there were no variables containing the appropriate frame. This led to problems in the jitterbuffer callback where we unnecessarily bailed out of retrieving a frame from the jitterbuffer. This was logic intentionally added in commit 73103bdcd5b342ce5dfa32039333ffadad551151 in response to an earlier regression, and while this prevents crashes, it also backlogs legitimate frames unnecessarily.

The abort logic was initially added because at this point in the code, we did not have the negotiated format available to us. However, it should always be available to us as a last resort in chosenformat, so we now pull it from there if needed. This allows us to process frames the jitterbuffer even if voicefmt and peerfmt aren't set and still avoid the crash. The failsafe logic is retained, but now it shouldn't be triggered anymore.

Resolves: #1054

### DIFF
--- a/channels/chan_iax2.c
+++ b/channels/chan_iax2.c
@@ -786,7 +786,8 @@ struct chan_iax2_pvt {
 	unsigned short peercallno;
 	/*! Negotiated format, this is only used to remember what format was
 	    chosen for an unauthenticated call so that the channel can get
-	    created later using the right format */
+	    created later using the right format. We also use it for
+		authenticated calls to check the format from __get_from_jb. */
 	iax2_format chosenformat;
 	/*! Peer selected format */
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

PR #1059, opened 2025-01-13
fixed semicolons

michaelnugent:
This fixes missing semicolons typoed in the latest release.  

I do this all the time when I'm writing python all day and then start writing C.


michaelnugent:
It's 2 characters.  Not even, it's just 1 character twice.  I dont think we need to go through legal.

PR #1060, opened 2025-01-13
docs:  Split manager XML docs into separate file and Add <since> tags for all AMI actions (cert-20.7).

gtjoseph:
To keep the source tree somewhat compatible with the base 20 branch
the XML documentation from manager.c has been extracted into manager_doc.xml.
This will give future cherry-picks a better channce of succeeding without
manual intervention.

Also cherry-picked @seanbright's "Add `<since>` tags for all AMI actions." PR.


gtjoseph:
cherry-pick-to: none

PR #1061, opened 2025-01-13
Fix build with no AST_DEVMODE in res_pjsip_authenticator_digest.c

umerov1999:
Fix build logger and res_pjsip_authenticator_digest when no AST_DEVMODE


seanbright (changes requested):
Looks fine. Can you squash these down to a single commit with a commit message that follows [the Commit Message policy](https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/)?

Something like:

`logger.h: Fix build when AST_DEVMODE is not defined.`

seanbright:
Um, what happened?

Issue #1064, opened 2025-01-14
[improvement]: ast_tls_script: Add option to skip passphrase for CA private key


InterLinked1:
### Improvement Description

Currently, the script is hardcoded to use the -des3 option, which requires a passphrase. This adds an option to bypass this.

PR #1065, opened 2025-01-14
ast_tls_cert: Add option to skip passphrase for CA private key.

InterLinked1:
Currently, ast_tls_cert file is hardcoded to use the -des3 option for 3DES encryption, and the script needs to be manually modified to not require a passphrase. Add an option (-e) that disables encryption of the CA private key so no passphrase is required.

Resolves: #1064


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1028, opened 2024-12-11
[bug]: "pjsip show endpoints" shows some identifies on endpoints that shouldn't be there


rasstegar:
### Severity

Minor

### Versions

Asterisk 22.1.0

### Components/Modules

res_pjsip , res_pjsip_endpoint_identifier_ip.c

### Operating Environment

Debian 12 (bookworm)

### Frequency of Occurrence

None

### Issue Description

[123456789c1]
type=endpoint
context=from-trunk-c1
disallow=all
allow=ulaw
from_domain=a.main.com
from_user=123456789c1
allow = !all,g722,ulaw,alaw

[my-identify1]
type=identify
endpoint = 123456789c1
match = 192.168.5.24

------------------------------------

[123456789c2]
type=endpoint
context=from-trunk-c1
disallow=all
allow=ulaw
from_domain=a.main.com
from_user=123456789c1
allow = !all,g722,ulaw,alaw

[my-identify2]
type=identify
endpoint = 123456789c2
match = 192.168.5.25

I create two endpoints and two identify with different names. 
But when I look in Cli I see that two identify are set for each endpoint. As you can see below:
Which causes problems with incoming calls.
I think, the identify does not consider letters after the number => "123456789c1" .

### Relevant log output

```shell
Endpoint:  123456789c1                                          Unavailable   0 of inf
   Identify:  my-identify1/123456789c1
        Match: 192.168.5.24/32
   Identify:  my-identify2/123456789c2
        Match: 192.168.5.25/32

 Endpoint:  123456789c2                                          Unavailable   0 of inf
   Identify:  my-identify1/123456789c1
        Match: 192.168.5.24/32
   Identify:  my-identify2/123456789c2
        Match: 192.168.5.25/32
```

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
Do you have an actual log showing that this is an underlying problem with inbound matching, aside from the cosmetic output of "pjsip show endpoints"?

jcolp:
Or to be more specific, after examining the code I believe the shown CLI output while it should be fixed is just a result of the way the CLI output works. It does not work the same way in inbound matching, and would not have the same problem.

rasstegar:
In a real example I have the same match value and when I make a invite request with each user the first identify is picked and sent to that context. When I work without identify and it involves autenticate it is correctly redirected to the desired context for each user.
[123456789c1]
type=endpoint

[...]

PR #1066, opened 2025-01-15
strings.c: Improve numeric detection in `ast_strings_match()`.

seanbright:
Essentially, we were treating 1234x1234 and 1234x5678 as 'equal' because we were able to convert the prefix of each of these strings to the same number.

Resolves: #1028


seanbright:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

jcolp (approved):

PR #1070, opened 2025-01-16
docs: Add version information to configObject and configOption XML elements (20 only)

gtjoseph:
Most of the configObjects and configOptions that are implemented with
ACO or Sorcery now have `<since>/<version>` elements added.  There are
probably some that the script I used didn't catch.  The version tags were
determined by the following...
 * Do a git blame on the API call that created the object or option.
 * Do a `git log --grep <summary>` to find the cherry-pick commits in all
   branches that match.
 * Do a `git patch-id` to ensure the commits are all related and didn't get
   a false match on the summary.
 * Do a `git tag --contains <commit>` to find the tags that contain each
   commit.
 * Weed out all tags not <major>.<minor>.0.
 * Sort and discard any <major>.0.0 and following tags where the commit
 * The result is a single tag for each branch where the API was last touched.

configObjects and configOptions elements implemented with the base
ast_config APIs were just not possible to find due to the non-deterministic
way they are accessed.

Also note that if the API call was on modified after it was added, the
version will be the one it was last modified in.

Final note:  The configObject and configOption elements were introduced in
12.0.0 so options created before then may not have any XML documentation.

### DIFF
--- a/apps/app_agent_pool.c
+++ b/apps/app_agent_pool.c
@@ -268,11 +268,13 @@
 				<synopsis>Unused, but reserved.</synopsis>
 			</configObject>
 			<configObject name="agent-id">
+				<since><version>12.0.0</version></since>
 				<synopsis>Configure an agent for the pool.</synopsis>
 				<description>
 					<xi:include xpointer="xpointer(/docs/configInfo[@name='app_agent_pool']/description/note)" />
 				</description>
 				<configOption name="ackcall">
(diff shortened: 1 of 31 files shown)


gtjoseph:
cherry-pick-to: none

gtjoseph:
The pjsip.conf:system object and options were initially missing but are now included.

jcolp (approved):

PR #1073, opened 2025-01-17
docs: Add version information to application and function XML elements

gtjoseph:
* Do a git blame on the embedded XML application or function element.

* Do a git log --grep <summary> to find the cherry-pick commits in all
  branches that match.

* Do a git patch-id to ensure the commits are all related and didn't get
  a false match on the summary.

* Do a git tag --contains <commit> to find the tags that contain each
  commit.

* Weed out all tags not ..0.

* Sort and discard any .0.0 and following tags where the commit

* The result is a single tag for each branch where the application or function
  was defined.

The applications and functions defined in the following files were done by
hand because the XML was extracted from the C source file relatively recently.
* channels/pjsip/dialplan_functions_doc.xml
* main/logger_doc.xml
* main/manager_doc.xml
* res/res_geolocation/geoloc_doc.xml
* res/res_stir_shaken/stir_shaken_doc.xml

### DIFF
--- a/apps/app_adsiprog.c
+++ b/apps/app_adsiprog.c
@@ -57,6 +57,7 @@ static const char app[] = "ADSIProg";
 
 /*** DOCUMENTATION
 	<application name="ADSIProg" language="en_US">
+		<since><version>1.8.0</version></since>
 		<synopsis>
 			Load Asterisk ADSI Scripts into phone
 		</synopsis>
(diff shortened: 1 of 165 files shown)


gtjoseph:
cherry-pick-to: none

seanbright (approved):

PR #1074, opened 2025-01-19
Update dialplan_functions.c

fgadot:
**Fix parse_uri_cb to handle display names in SIP URIs**

This pull request fixes an issue in the parse_uri_cb function in res/pjsip/dialplan_functions.c, where it fails to correctly parse SIP URIs that include a display name in the From or To headers. The issue affects scenarios where SIP headers use the name-addr format as defined in RFC 3261.

**Problem**
SIP headers often include display names in the name-addr format, e.g.:
`From: Alice <sip:alice@example.com>;tag=12345
`
The current implementation of parse_uri_cb directly passes the full header value (e.g., Alice <sip:alice@example.com>) to pjsip_parse_uri, which fails to parse the URI because it doesn't handle the display name.

This results in the following warning in Asterisk logs:
`WARNING[1234]: pjsip/dialplan_functions.c:668 parse_uri_cb: Failed to parse URI 'Alice <sip:alice@example.com>;tag=12345'`

The pjsip_parse_uri function expects only the addr-spec portion (e.g., <sip:alice@example.com>) to be passed for parsing.

**Solution**
Modified parse_uri_cb to:
Detect if < and > delimiters are present in the URI.
Extract the portion between < and > (i.e., the addr-spec).
If < and > are not found, fallback to using the entire string as the URI.

This approach ensures compliance with the name-addr syntax from RFC 3261, Section 25.1.
**Updated logic:**
```
const char *start = strchr(args->uri, '<');
const char *end = strchr(args->uri, '>');

if (start && end && start < end) {
    tmp.ptr = (char *)(start + 1);  // Skip '<'
    tmp.slen = end - start - 1;     // Length between '<' and '>'
} else {
    // If no '<' and '>' are found, fallback to the original URI
    pj_strdup2_with_null(pool, &tmp, args->uri);
}

uri = (pjsip_name_addr *)pjsip_parse_uri(pool, tmp.ptr, tmp.slen, PJSIP_PARSE_URI_AS_NAMEADDR);
```

**Testing**
**Test Cases**

- Valid Name-Addr with Display Name:
Input:
`From: Alice <sip:alice@example.com>;tag=12345`
Expected Behavior:
No warnings in Asterisk logs.
URI sip:alice@example.com is correctly parsed.

- Valid Name-Addr without Display Name:
Input:
`From: <sip:alice@example.com>;tag=12345`
Expected Behavior:
URI sip:alice@example.com is correctly parsed.

- URI Without < and >:
Input:
`From: sip:alice@example.com;tag=12345`
Expected Behavior:
URI sip:alice@example.com is correctly parsed.

- Malformed URI:
Input:
`From: Alice sip:alice@example.com;tag=12345`
Expected Behavior:
Warning in logs indicating parsing failure:
`WARNING[1234]: pjsip/dialplan_functions.c:668 parse_uri_cb: Failed to parse URI`

**Testing Environment**
Asterisk version: 22.1.1
Test scenarios executed using SIPp and manual SIP INVITE messages.

**References**
[RFC 3261 Section 25.1: Name-Addr Syntax](https://www.rfc-editor.org/rfc/rfc3261.html#page-219)

**Impact**
This fix ensures that SIP headers with name-addr format are correctly parsed, improving compatibility with SIP endpoints and preventing unnecessary warnings in Asterisk logs.

### DIFF
--- a/channels/pjsip/dialplan_functions.c
+++ b/channels/pjsip/dialplan_functions.c
@@ -662,7 +662,18 @@ static int parse_uri_cb(void *data)
 		return 0;
 	}
 
-	pj_strdup2_with_null(pool, &tmp, args->uri);
+	/* Let's make sure that the parser only treat that's in between < and > */
+	const char *start = strchr(args->uri, '<');
+	const char *end = strchr(args->uri, '>');
+
+	if (start && end && start < end) {
+		tmp.ptr = (char *)(start + 1);  // Skip '<'
+		tmp.slen = end - start - 1;     // Length between '<' and '>'
+	} else {
+		// If no '<' and '>' are found, fallback to the original URI
+		pj_strdup2_with_null(pool, &tmp, args->uri);
+	}
+
 	uri = (pjsip_name_addr *)pjsip_parse_uri(pool, tmp.ptr, tmp.slen, PJSIP_PARSE_URI_AS_NAMEADDR);
 	if (!uri || (!PJSIP_URI_SCHEME_IS_SIP(uri) && !PJSIP_URI_SCHEME_IS_SIPS(uri))) {
 		ast_log(LOG_WARNING, "Failed to parse URI '%s'\n", args->uri);
@@ -670,6 +681,7 @@ static int parse_uri_cb(void *data)
 		args->ret = -1;
 		return 0;
 	}
+
 
 	if (!strcmp(args->type, "scheme")) {
 		ast_copy_pj_str(args->buf, pjsip_uri_get_scheme(uri), args->buflen);


fgadot:
Erratum: It looks like the current code can't handle whatever is after the > actually. 
Regardless, this fix will strip anything before the < and anything after the > keeping the URI Clean.

alex2grad:
Could you please provide your test dialplan which fails to parse.

Attached [parse_uri_test_dialplan.txt](https://github.com/user-attachments/files/18482079/parse_uri_test_dialplan.txt) my test dialplan to check the function PJSIP_PARSE_URI on Asterisk 20.11.1.
All tests passed, see [parse_uri_test_result.txt](https://github.com/user-attachments/files/18482083/parse_uri_test_result.txt).

alex2grad:
I understood why your example failed, it's because of ";tag=".
If URI is "Alice <sip:alice@example.com>", it doesn't fail.

alex2grad:
You should use `${CHANNEL(pjsip,local_uri)}` or `${CHANNEL(pjsip,remote_uri)}` to get URI without tag

fgadot:
hello @alex2grad 

The INVITE my legacy PBX is sending is this:

INVITE sip:1998@192.168.100.98:5060 SIP/2.0
Accept-Language: en
Allow: INVITE, ACK, CANCEL, BYE, PRACK, NOTIFY, REFER, SUBSCRIBE, OPTIONS, UPDATE, INFO
Supported: replaces,timer,path
User-Agent: OmniPCX Enterprise R100.1 n2.534.7.b
Session-Expires: 1800;refresher=uac
Min-SE: 900
P-Alcatel-CSBU: categparty=internal;rd=noanswer;call_condition=forwardeduser
History-Info: <sip:1000@192.168.100.100:5060?reason=SIP%3bcause%3d408%3btext%3d%22Request%20Timeout%22>;index=1,<sip:1998@node000000:5060>;index=1.1
Content-Type: application/sdp
To: <sip:1000@192.168.100.100:5060;user=phone>
From: "First Name last" <sip:1234@192.168.100.100:5060;user=phone>;tag=9301d7dc28c38f07a7bd3ecf8b3b9427
Contact: <sip:1234@192.168.100.100:5060;transport=UDP>
Call-ID: 9c983b11f9146863054a5d2dbad1c076@192.168.100.100
CSeq: 1422312506 INVITE
Via: SIP/2.0/UDP 192.168.100.100;branch=z9hG4bK51a6c2bc6c4c9759ade49672ed087aa7
Max-Forwards: 70
Content-Length: 398
(...)


The line in the dialplan causing issue was the ${PJSIP_PARSE_URI(...)} in the Verbose section:
```
	; Extract the extension from the "From:" header (caller)
	same => n,Verbose(1, Extracting the From/user from the PJSIP Header ${PJSIP_PARSE_URI(${PJSIP_HEADER(read,From)},user)})
	same => n,Set(extracted_from=${PJSIP_PARSE_URI(${PJSIP_HEADER(read,From)},user)})
	same => n,Verbose(1, Extracted origin: ${extracted_from})
```


I need the From user part only of the URI. 
The documentation [here](https://docs.asterisk.org/Latest_API/API_Documentation/Dialplan_Functions/PJSIP_PARSE_URI/) specifically stipulates that it is capable of returning part of the URI which I am correctly passing. 

Using your work around works. But since the tag is a normal parameter in the TO / FROM , I'd request to make this parser more fail proof and accept the fix which keeps the parser doing what it's doing, but as expected in its name only reads the URI without head and tail parameters outside the < and > 

Your thoughts?

alex2grad:
replace 
`	same => n,Set(extracted_from=${PJSIP_PARSE_URI(${PJSIP_HEADER(read,From)},user)})
`
with
`	same => n,Set(extracted_from=${PJSIP_PARSE_URI(${CHANNEL(pjsip,remote_uri)},user)})
`

fgadot:
Hi, 
THat's what I did after you mentioned it, and the work around works. 
But you should either update the documentation to explain that the only safe parameter that PJSIP_PARSE_URI will accept is from CHANNEL(pjsip,local/remote_uri) because PJSIP_HEADER(read,From) will work unless it has a tag, tag which is OK from the RFC ...

alex2grad:
You are passing the whole header, not URI.
The header  contains URI and Header Parameters.
In case of "From" header, the Header Parameter is "tag".

alex2grad:
The function PJSIP_PARSE_URI parses the string which must contain ONLY URI.
If you need to parse the whole SIP header you can add  a new function (for example PJSIP_PARSE_HEADER) to parse the whole header.

fgadot:
ok - You can close this PR. If possible, it would be great to update the PJSIP_PARSE_URI documentation and mention 
${PJSIP_PARSE_URI(${CHANNEL(pjsip,local_uri)},user)}
${PJSIP_PARSE_URI(${CHANNEL(pjsip,remote_uri)},user)}
To extract the From / To fields

Thank you.

jcolp:
The documentation is kept in XML in the source code, so another PR can be submitted to update the documentation if you wish.

PR #1076, opened 2025-01-20
docs: Add version information to manager event instance XML elements

gtjoseph:
* Do a git blame on the embedded XML managerEvent elements.

* Do a git log --grep <summary> to find the cherry-pick commits in all
  branches that match.

* Do a git patch-id to ensure the commits are all related and didn't get
  a false match on the summary.

* Do a git tag --contains <commit> to find the tags that contain each
  commit.

* Weed out all tags not ..0.

* Sort and discard any .0.0 and following tags where the commit

* The result is a single tag for each branch where the application or function
  was defined.

The events defined in res/res_pjsip/pjsip_manager.xml were done by hand
because the XML was extracted from the C source file relatively recently.

Two bugs were fixed along the way...

* The get_documentation awk script was exiting after it processed the first
  DOCUMENTATION block it found in a file.  We have at least 1 source file
  with multiple DOCUMENTATION blocks so only the first one in them was being
  processed.  The awk script was changed to continue searching rather
  than exiting after the first block.

* Fixing the awk script revealed an issue in logger.c where the third
  DOCUMENTATION block contained a XML fragment that consisted only of
  a managerEventInstance element that wasn't wrapped in a managerEvent
  element.  Since logger_doc.xml already existed, the remaining fragments
  in logger.c were moved to it and properly organized.

### DIFF
--- a/apps/app_agent_pool.c
+++ b/apps/app_agent_pool.c
@@ -181,6 +181,7 @@
 	</manager>
 	<managerEvent language="en_US" name="Agents">
 		<managerEventInstance class="EVENT_FLAG_AGENT">
+			<since><version>12.0.0</version></since>
 			<synopsis>
 				Response event in a series to the Agents AMI action containing
 				information about a defined agent.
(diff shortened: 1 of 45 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

seanbright (approved):

Issue #1085, opened 2025-01-24
[bug]: utils: Compilation failure with DEVMODE due to old-style definitions


InterLinked1:
### Severity

Critical

### Versions

Git

### Components/Modules

sqlite3 (libdb)

### Operating Environment

Fedora 42

### Frequency of Occurrence

Constant

### Issue Description

Builds now fail in dev mode with old style definition gcc warnings.

### Relevant log output

```shell
[CCi] hash/hash.i -> hash/hash.o
   [CCi] hash/hash_bigkey.i -> hash/hash_bigkey.o
hash/hash.c: In function '__hash_open':
hash/hash.c:96:1: error: old-style function definition [-Werror=old-style-definition]
   96 | __hash_open(file, flags, mode, info, dflags)
      | ^~~~~~~~~~~
hash/hash.c: In function 'hash_close':
hash/hash.c:246:1: error: old-style function definition [-Werror=old-style-definition]
  246 | hash_close(dbp)
      | ^~~~~~~~~~
hash/hash.c: In function 'hash_fd':
hash/hash.c:262:1: error: old-style function definition [-Werror=old-style-definition]
  262 | hash_fd(dbp)
      | ^~~~~~~
hash/hash.c: In function 'init_hash':
hash/hash.c:280:1: error: old-style function definition [-Werror=old-style-definition]
  280 | init_hash(hashp, file, info)
      | ^~~~~~~~~
hash/hash.c: In function 'init_htab':
hash/hash.c:351:1: error: old-style function definition [-Werror=old-style-definition]
  351 | init_htab(hashp, nelem)
      | ^~~~~~~~~
hash/hash.c: In function 'hdestroy':
hash/hash.c:397:1: error: old-style function definition [-Werror=old-style-definition]
  397 | hdestroy(hashp)
      | ^~~~~~~~
hash/hash.c: In function 'hash_sync':
hash/hash.c:457:1: error: old-style function definition [-Werror=old-style-definition]
  457 | hash_sync(dbp, flags)
      | ^~~~~~~~~
hash/hash.c: In function 'flush_meta':
hash/hash.c:486:1: error: old-style function definition [-Werror=old-style-definition]
  486 | flush_meta(hashp)
      | ^~~~~~~~~~
hash/hash.c: In function 'hash_get':
hash/hash.c:534:1: error: old-style function definition [-Werror=old-style-definition]
  534 | hash_get(dbp, key, data, flag)
      | ^~~~~~~~
[...]

PR #1086, opened 2025-01-24
utils: Disable old style declaration/definition warnings for libdb.

InterLinked1:
Newer versions of gcc now warn about old style declarations and definitions, such as those in libdb, which causes compilation failure with DEVMODE enabled. Ignore these warnings for libdb.

Resolves: #1085


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

seanbright (approved):

jcolp (approved):

Issue #1088, opened 2025-01-24
[bug]: app_sms: Compilation failure in DEVMODE due to stringop-overflow error in GCC 15 pre-release


InterLinked1:
### Severity

Major

### Versions

22.2.0-rc1

### Components/Modules

app_sms

### Operating Environment

Fedora 42

### Frequency of Occurrence

Constant

### Issue Description

gcc 15 warning about possible writing beyond bounds of a buffer.

Looks like a false positive to me since there is a check for each byte if the end of the buffer has been reached, and the bounds are correct, but it's possible I've missed something...

### Relevant log output

```shell
[CCi] app_stack.i -> app_stack.o
In function 'packsms8',
    inlined from 'sms_exec' at app_sms.c:1997:76:
app_sms.c:470:24: error: writing 32 bytes into a region of size 6 [-Werror=stringop-overflow=]
  470 |                 o[p++] = u;
      |                 ~~~~~~~^~~
app_sms.c: In function 'sms_exec':
app_sms.c:450:23: note: at offset [134, 140] into destination object 'dummy' of size 140
  450 |         unsigned char dummy[SMSLEN_8];
      |                       ^~~~~
cc1: all warnings being treated as errors
make[1]: *** [/usr/src/asterisk-22.2.0-rc1/Makefile.rules:151: app_sms.o] Error 1
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

seanbright:
Does putting an explicit cast there work? I don't know where it is getting 32 bytes from, `u` is a `long`.

    o[p++] = (unsigned char) u;

InterLinked1:
> Does putting an explicit cast there work? I don't know where it is getting 32 bytes from, `u` is a `long`.
> 
> ```
> o[p++] = (unsigned char) u;
> ```

Nope.

```
[...]

PR #1089, opened 2025-01-24
app_sms: Ignore false positive stringop-overflow warning.

InterLinked1:
Ignore gcc warning about writing 32 bytes into a region of size 6, since we check that we don't go out of bounds for each byte.

Resolves: #1088

### DIFF
--- a/apps/Makefile
+++ b/apps/Makefile
@@ -52,6 +52,8 @@ $(call MOD_ADD_C,app_confbridge,$(wildcard confbridge/*.c))
 app_confbridge.o: _ASTCFLAGS+=$(AST_NO_FORMAT_TRUNCATION)
 app_meetme.o: _ASTCFLAGS+=$(AST_NO_FORMAT_TRUNCATION)
 app_minivm.o: _ASTCFLAGS+=$(AST_NO_FORMAT_TRUNCATION)
+# Needed to work around gcc 15 bug: https://github.com/asterisk/asterisk/issues/1088
+app_sms.o: _ASTCFLAGS+=-fno-tree-vectorize
 app_voicemail.o: _ASTCFLAGS+=$(AST_NO_FORMAT_TRUNCATION) -DFILE_STORAGE
 app_voicemail_odbc.o: _ASTCFLAGS+=$(AST_NO_FORMAT_TRUNCATION) -DODBC_STORAGE
 app_voicemail_imap.o: _ASTCFLAGS+=$(AST_NO_FORMAT_TRUNCATION) -DIMAP_STORAGE


seanbright:
Just my 2 cents but I’d wait for gcc 15 to be released. This seems like a compiler bug.

InterLinked1:
> Just my 2 cents but I’d wait for gcc 15 to be released. This seems like a compiler bug.
> 
> Update: this is a GCC vectorization bug which explains why the error is complaining about "32 bytes" while storing a long. You can confirm this by turning off vectorization in apps/Makefile:
> 
> ```diff
> +app_sms.o: _ASTCFLAGS+=-fno-tree-vectorize
> ```
> 
> If we end up needing to silence this error after GCC 15 is released this is the approach I would prefer to take.

Thanks, this sounds like the best approach. I tested it and it works.

You haven't already reported this to gcc by chance, have you? I didn't see anything similar in their bugtracker already, so considering reporting it.

seanbright:
> You haven't already reported this to gcc by chance, have you? I didn't see anything similar in their bugtracker already, so considering reporting it.

I have not. I spent some time trying to come up with a small repro I could submit but didn’t have any luck.

InterLinked1:
> Workflow PRCheck failed [master-ari1](https://github.com/asterisk/asterisk/actions/runs/13606505345/job/38038715317): FAILED TEST: rest_api/bridges/error [master-ari2](https://github.com/asterisk/asterisk/actions/runs/13606505345/job/38038715017): FAILED TEST: rest_api/channels/create_dial_bridge/dial/off_nominal [master-pjs5](https://github.com/asterisk/asterisk/actions/runs/13606505345/job/38038715893): FAILED TEST: channels/pjsip/stir_shaken/stir_shaken_anon_callerid

recheck

InterLinked1:
> > You haven't already reported this to gcc by chance, have you? I didn't see anything similar in their bugtracker already, so considering reporting it.
> 
> I have not. I spent some time trying to come up with a small repro I could submit but didn’t have any luck.

I might take a stab at it, I was thinking of just starting with `app_sms` and whittling it down, then submitting that + the underlying gcc command used to compile, unless you already tried that and it didn't work. Were you able to reproduce the original issue as in see the compilation failure yourself? Or was it just the repro attempt that wouldn't fail?

I think I only observed this in the Fedora 42 runner in GitHub, even when testing on Fedora 42 locally with gcc 15, I couldn't reproduce the issue for some reason.

seanbright:
> I might take a stab at it, I was thinking of just starting with `app_sms` and whittling it down, then submitting that + the underlying gcc command used to compile, unless you already tried that and it didn't work.

I didn’t try that. I just extracted a chunk of code and tried to compile it. I didn’t spend a lot of time on it. 

> Were you able to reproduce the original issue as in see the compilation failure yourself?

I was. I built GCC from source and then tried to build Asterisk and got the error you reported.

seanbright (changes requested):
Can you add a comment here that references the associated bug?

InterLinked1:
> Can you add a comment here that references the associated bug?

I assume you mean the gcc bug, and a comment in the Makefile?

I'm compiling gcc 15 from source now so I can try to put together an example program to submit to the bugtracker. Once I've opened that, I'll add a link.

seanbright:
> > Can you add a comment here that references the associated bug?
> 
> I assume you mean the gcc bug, and a comment in the Makefile?

No, I mean the Asterisk issue. It can be backed in to from a `git blame` but I think it would be nicer to understand why the change was made at a glance.

If there _is_ an associated `gcc` bug than that would also be a good addition.

InterLinked1:
> > > Can you add a comment here that references the associated bug?
> > 
> > 
> > I assume you mean the gcc bug, and a comment in the Makefile?
> 
> No, I mean the Asterisk issue. It can be backed in to from a `git blame` but I think it would be nicer to understand why the change was made at a glance.
> 
> If there _is_ an associated `gcc` bug than that would also be a good addition.

I haven't filed any report yet, but I've done some more testing. I think I'm running into what you described earlier. As soon as I try to strip away basically anything in `app_sms.c`, then it compiles fine. It's rather bizarre. It does make me question if there is some other issue in the code.

Following some of the gcc bug reporting tips, I tried adding a few flags, and when I add Address Sanitizer in, it also compiles fine:

```
/usr/src/sms# gcc -o /dev/null -c app_sms.c -pthread -pipe -Wall -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declarations -g3 -Werror -Wunused -Wdeclaration-after-statement -Wtrampolines -Wundef -Wmissing-format-attribute -Wformat=2 -O0  -march=native -fPIC -DAST_MODULE=\"app_sms\" -DAST_MODULE_SELF_SYM=__internal_app_sms_self    -O3 -fno-partial-inlining -fno-strict-aliasing -fwrapv -fno-aggressive-loop-optimizations -Wno-sign-compare -Wno-unused-parameter -Wextra
In function ‘packsms8’,
    inlined from ‘sms_exec’ at app_sms.c:1999:24:
app_sms.c:472:24: error: writing 32 bytes into a region of size 6 [-Werror=stringop-overflow=]
  472 |                 o[p++] = u;
      |                 ~~~~~~~^~~
app_sms.c: In function ‘sms_exec’:
app_sms.c:452:23: note: at offset [134, 140] into destination object ‘dummy’ of size 140
  452 |         unsigned char dummy[SMSLEN_8];
      |                       ^~~~~
/usr/src/sms# gcc -o /dev/null -c app_sms.c -pthread -pipe -Wall -Wstrict-prototypes -Wmissing-prototypes -Wmissing-declarations -g3 -Werror -Wunused -Wdeclaration-after-statement -Wtrampolines -Wundef -Wmissing-format-attribute -Wformat=2 -O0  -march=native -fPIC -DAST_MODULE=\"app_sms\" -DAST_MODULE_SELF_SYM=__internal_app_sms_self    -O3 -fno-partial-inlining -fno-strict-aliasing -fwrapv -fno-aggressive-loop-optimizations -Wno-sign-compare -Wno-unused-parameter -Wextra -fsanitize=address,undefined
/usr/src/sms#
```

(either address and/or undefined, doesn't matter)

seanbright:
Using `git bisect` on the GCC repo I was able to back in to [the problematic change](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=68326d5d1a593dc0bf098c03aac25916168bc5a9). A brief search of the GCC bugzilla didn't really turn up anything useful but this subject is well outside my sphere of knowledge.

It seems to be related to inlining `packsms8` into `sms_exec` (if you add a `__attribute__((noinline))` to `packsms8` the error also goes away). I tried rearranging the loops a bit but didn't have any luck. Having to turn off individual compiler options for specific files is annoying, but whatever.

gtjoseph:
Just FYI...  I installed Fedora 42 over the weekend which ships with GCC 15.0.1  and I have no issues compiling app_sms.  I tried with and without DEVMODE, with and without DONT_OPTIMIZE and with and without COMPILE_DOUBLE.

```
gcc (GCC) 15.0.1 20250329 (Red Hat 15.0.1-0)
Copyright (C) 2025 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

seanbright:
Red Hat may have rolled out the problematic code. I promise it’s a real GCC bug.

gtjoseph:
Interesting, I just updated and got 15.1.1. Still compiles fine.  Interestingly enough though, I went through RedHat's gcc patches and didn't see anything obvious that would have fixed the issue. :shrug:

seanbright:
Strange 🤷🏻

seanbright:
Some additional info - this only manifests when `-march=native` which @gtjoseph explicitly disables in his builds.

jcolp (approved):

gtjoseph:
@seanbright Are you OK with the current solution?

seanbright (approved):
Sure

PR #1090, opened 2025-01-24
docs: Add version information to AGI command XML elements.

gtjoseph:
This process was a bit different than the others because everything
is in the same file, there's an array that contains the command
names and their handler functions, and the last command was created
over 15 years ago.

* Dump a `git blame` of res/res_agi.c from BEFORE the handle_* prototypes
  were changed.
* Create a command <> handler function xref by parsing the the agi_command
  array.
* For each entry, grep the function definition line "static int handle_*"
  from the git blame output and capture the commit.  This will be the
  commit the command was created in.
* Do a `git tag --contains <commit> | sort -V | head -1` to get the
  tag the function was created in.
* Add a single since/version element to the command XML.  Multiple versions
  aren't supported here because the branching and tagging scheme changed
  several times in the 2000's.

### DIFF
--- a/res/res_agi.c
+++ b/res/res_agi.c
@@ -73,6 +73,9 @@
 
 /*** DOCUMENTATION
 	<agi name="answer" language="en_US">
+		<since>
+			<version>0.4.0</version>
+		</since>
 		<synopsis>
 			Answer channel
 		</synopsis>
@@ -87,6 +90,9 @@
 		</see-also>
 	</agi>
 	<agi name="asyncagi break" language="en_US">
+		<since>
+			<version>1.8.0</version>
+		</since>
 		<synopsis>
 			Interrupts Async AGI
 		</synopsis>
@@ -101,6 +107,9 @@
 		</see-also>
 	</agi>
 	<agi name="channel status" language="en_US">
+		<since>
+			<version>0.4.0</version>
+		</since>
 		<synopsis>
 			Returns status of the connected channel.
 		</synopsis>
@@ -143,6 +152,9 @@
 		</see-also>
 	</agi>
 	<agi name="control stream file" language="en_US">
+		<since>
+			<version>1.2.0</version>
+		</since>
 		<synopsis>
 			Sends audio file on channel and allows the listener to control the stream.
 		</synopsis>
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

res/res_agi.c
```diff
@@ -73,6 +73,11 @@
 
 /*** DOCUMENTATION
 	<agi name="answer" language="en_US">
+		<since>
+			<version>0.1.0</version>
+			<version>1.6.1.0</version>
+			<version>1.8.0</version>
+		</since>
```
seanbright, line 80: Shouldn't this just be 0.1.0? This applies to most of this file.

seanbright (dismissed):

InterLinked1 (dismissed):
Also, typo in the commit message:

> teh commit

res/res_agi.c
gtjoseph: I'm not sure what's right or wrong any more. :)

gtjoseph: At least I indented the versions though.

gtjoseph:
> Also, typo in the commit message:
> 
> > teh commit

fixed

seanbright (approved):

mbradeen (dismissed):

res/res_agi.c
```diff
@@ -945,6 +1059,9 @@
 		</see-also>
 	</agi>
 	<agi name="speech create" language="en_US">
+		<since>
```
jcolp, line 1062: The speech stuff came in 1.6.0 I believe

jcolp (dismissed):

res/res_agi.c
gtjoseph: I might have to re-think the approach.  This file has had several major refactors so it's very hard to tell when something appeared.

gtjoseph:
Back in 2009, the prototypes for all of the handle_* functions were changed but there hasn't been a command addition since then so I changed the way the commits were determined.  This resulted in a much more accurate determination of the earliest tag a command appeared in.

jcolp (approved):

Issue #211, opened 2023-07-14
[bug]: stasis: Off-nominal channel leave causes bridge to be destroyed


11qeq11:
### Severity

Major

### Versions

18.x.x, 20.x.x

### Components/Modules

Resources/res_stasis

### Operating Environment

CentOS 7, kernel 3.10.0-1160.83.1.el7.x86_64.

### Frequency of Occurrence

Occasional

### Issue Description

This was already reported in https://issues-archive.asterisk.org/ASTERISK-30445, I will summarize everything important.

After upgrade to version `18.6.0` from `16.29.0`, sometimes (once a day or so) Asterisk started to crash with spamming "FRACK!..." log entries. If left unnoticed, kernel would kill Asterisk process, and it will be simply restarted:
```
Feb 24 09:32:48 localhost kernel: asterisk[11361]: segfault at 1 ip 0000000000552827 sp 00007f7ed3e0a6b0 error 4 in asterisk[400000+2a2000]
```

During the incident, everything is growing indefinitely (calls/channels/bridges count), until Asterisk is restarted/OpenSIPS in front of it will recognize that it's faulty.
Previously, I thought that it's tied to the overall number of calls, because the bug occurs only during working hours, however, appeared that it's caused by just one particular client of ours (it's, however, unknown what is wrong with him).

After providing DEBUG logs and core dumps, @jcolp has recognized issue with bridges, quote:

> There is an off-nominal scenario where channel departure fails, and the bridge is destroyed while it is subsequently still used:
```
[2023-04-17 08:59:43.075] DEBUG[2971] stasis/control.c: 1681721248.708: Channel is no longer in departable state
[2023-04-17 08:59:43.075] DEBUG[2971] bridge.c: Bridge 1681721248.708: actually destroying stasis bridge, nobody wants it anymore
[2023-04-17 08:59:43.075] DEBUG[2971] stasis_bridges.c: Update: 0x7f35400b72f8  Old: 1681721248.708  New: <none>
[2023-04-17 08:59:43.075] DEBUG[2971] stasis_bridges.c: Update: 0x7f35400b72f8  Old: 1681721248.708  New: <none>
[2023-04-17 08:59:43.075] DEBUG[2971] stasis.c: Destroying topic. name: bridge:all/bridge:1681721248.708, detail: 
[2023-04-17 08:59:43.076] DEBUG[2971] stasis.c: Topic 'bridge:all/bridge:1681721248.708': 0x7f354c04a3d0 destroyed
```
[...]

PR #1094, opened 2025-01-27
bridging: Fix multiple bridging issues causing SEGVs and FRACKs.

gtjoseph:
Issues:

* The bridging core allowed multiple bridges to be created with the same
  unique bridgeId at the same time.  Only the last bridge created with the
  duplicate name was actually saved to the core bridges container.

* The bridging core was creating a stasis topic for the bridge and saving it
  in the bridge->topic field but not increasing its reference count.  In the
  case where two bridges were created with the same uniqueid (which is also
  the topic name), the second bridge would get the _existing_ topic the first
  bridge created.  When the first bridge was destroyed, it would take the
  topic with it so when the second bridge attempted to publish a message to
  it it either FRACKed or SEGVd.

* The bridge destructor, which also destroys the bridge topic, is run from the
  bridge manager thread not the caller's thread.  This makes it possible for
  an ARI developer to create a new one with the same uniqueid believing the
  old one was destroyed when, in fact, the old one's destructor hadn't
  completed. This could cause the new bridge to get the old one's topic just
  before the topic was destroyed.  When the new bridge attempted to publish
  a message on that topic, asterisk could either FRACK or SEGV.

* The ARI bridges resource also allowed multiple bridges to be created with
  the same uniqueid but it kept the duplicate bridges in its app_bridges
  container.  This created a situation where if you added two bridges with
  the same "bridge1" uniqueid, all operations on "bridge1" were performed on
  the first bridge created and the second was basically orphaned.  If you
  attempted to delete what you thought was the second bridge, you actually
  deleted the first one created.

Changes:

* A new API `ast_bridge_topic_exists(uniqueid)` was created to determine if
  a topic already exists for a bridge.

* `bridge_base_init()` in bridge.c and `ast_ari_bridges_create()` in
  resource_bridges.c now call `ast_bridge_topic_exists(uniqueid)` to check
  if a bridge with the requested uniqueid already exists and will fail if it
  does.

* `bridge_register()` in bridges.c now checks the core bridges container to
  make sure a bridge doesn't already exist with the requested uniqueid.
  Although most callers of `bridge_register()` will have already called
  `bridge_base_init()`, which will now fail on duplicate bridges, there
  is no guarantee of this so we must check again.

* The core bridges container allocation was changed to reject duplicate
  uniqueids instead of silently replacing an existing one. This is a "belt
  and suspenders" check.

* A global mutex was added to bridge.c to prevent concurrent calls to
  `bridge_base_init()` and `bridge_register()`.

* Even though you can no longer create multiple bridges with the same uniqueid
  at the same time, it's still possible that the bridge topic might be
  destroyed while a second bridge with the same uniqueid was trying to use
  it. To address this, the bridging core now increments the reference count
  on bridge->topic when a bridge is created and decrements it when the
  bridge is destroyed.

* `bridge_create_common()` in res_stasis.c now checks the stasis app_bridges
  container to make sure a bridge with the requested uniqueid doesn't already
  exist.  This may seem like overkill but there are so many entrypoints to
  bridge creation that we need to be safe and catch issues as soon in the
  process as possible.

* The stasis app_bridges container allocation was changed to reject duplicate
  uniqueids instead of adding them. This is a "belt and suspenders" check.

* The `bridge show all` CLI command now shows the bridge name as well as the
  bridge id.

* Response code 409 "Conflict" was added as a possible response from the ARI
  bridge create resources to signal that a bridge with the requested uniqueid
  already exists.

* Additional debugging was added to multiple bridging and stasis files.

Resolves: #211

### DIFF
--- a/include/asterisk/bridge.h
+++ b/include/asterisk/bridge.h
@@ -75,6 +75,10 @@ extern "C" {
 #include "asterisk/dsp.h"
 #include "asterisk/uuid.h"
 
+/* Macros to assist debugging */
+#define BRIDGE_PRINTF_SPEC "%s(%s)(%p)"
+#define BRIDGE_PRINTF_VARS(bridge) S_OR((bridge)->uniqueid, "<unknown>"), S_OR((bridge)->name, "<unknown>"), (bridge)
+
 struct a02_container;
 struct ast_bridge_technology;
 struct ast_bridge;
--- a/include/asterisk/stasis_bridges.h
+++ b/include/asterisk/stasis_bridges.h
@@ -79,6 +79,15 @@ struct stasis_topic *ast_bridge_topic(struct ast_bridge *bridge);
 struct stasis_topic *ast_bridge_topic_all(void);
 
 /*!
+ * \brief Check if a stasis topic exists for a bridge uniqueid
+ *
+ * \param uniqueid The uniqueid to test
+ * \retval 1 if the topic exists
+ * \retval 0 if the topic does not exist
+ */
+int ast_bridge_topic_exists(const char *uniqueid);
+
+/*!
  * \since 12
  * \brief Publish the state of a bridge
  *
--- a/main/bridge.c
+++ b/main/bridge.c
@@ -133,6 +133,8 @@ static struct ao2_container *bridges;
 
 static AST_RWLIST_HEAD_STATIC(bridge_technologies, ast_bridge_technology);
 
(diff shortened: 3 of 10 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

mbradeen (approved):

jcolp (approved):

Issue #1095, opened 2025-01-27
[bug]: res_pjsip missing "Failed to authenticate" log entry for unknown endpoint


alex2grad:
### Severity

Major

### Versions

20.12.0-rc1

### Components/Modules

res/res_pjsip_authenticator_digest

### Operating Environment

Linux, Slackware 15.0 x86_64

### Frequency of Occurrence

Constant

### Issue Description

While testing the new Asterisk 20.12.0-rc1, I noticed that the log was missing an entry for failed authentication for an unknown endpoint.

Version 20.11.1 log 
++
NOTICE[28107]` res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:902@192.168.168.168>' failed for '192.168.168.169:54734' (callid: 1583355179-981804311-1748446211) - No matching endpoint found
NOTICE[21110] res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:902@192.168.168.168>' failed for '192.168.168.169:54734' (callid: 1583355179-981804311-1748446211) - No matching endpoint found
NOTICE[21110] res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:902@192.168.168.168>' failed for '192.168.168.169:54734' (callid: 1583355179-981804311-1748446211) - Failed to authenticate
++

Version 20.12.0-rc1 log
++
NOTICE[5480] res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:902@192.168.168.168>' failed for '192.168.168.169:54960' (callid: 735306937-1538200074-923933159) - No matching endpoint found
NOTICE[5480] res_pjsip/pjsip_distributor.c: Request 'INVITE' from '<sip:902@192.168.168.168>' failed for '192.168.168.169:54960' (callid: 735306937-1538200074-923933159) - No matching endpoint found
++

Our fail2ban asterisk filter was set to catch "Failed to authenticate" only. 
Since there are no "Failed to authenticate" entries, fail2ban doesn't catch anything.

Suspected commits:
https://github.com/asterisk/asterisk/commit/a0987672f0487a9caf92d91c095115799076a713
https://github.com/asterisk/asterisk/commit/0c272429e60d825481300dc25e232d3d9bafc85a

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
[...]

PR #1098, opened 2025-01-28
res_pjsip_authenticator_digest: Make correct error messages appear again.

gtjoseph:
When an incoming request can't be matched to an endpoint, the "artificial"
auth object is used to create a challenge to return in a 401 response and we
emit a "No matching endpoint found" log message. If the client then responds
with an Authorization header but the request still can't be matched to an
endpoint, the verification will fail and, as before, we'll create a challenge
to return in a 401 response and we emit a "No matching endpoint found" log
message.  HOWEVER, because there WAS an Authorization header and it failed
verification, we should have also been emitting a "Failed to authenticate"
log message but weren't because there was a check that short-circuited that
it if the artificial auth was used.  Since many admins use the "Failed to
authenticate" message with log parsers like fail2ban, those attempts were not
being recognized as suspicious.

Changes:

* digest_check_auth() now always emits the "Failed to authenticate" log
  message if verification of an Authorization header failed even if the
  artificial auth was used.

* The verification logic was refactored to be clearer about the handling
  of the return codes from verify().

* Comments were added clarify what return codes digest_check_auth() should
  return to the distributor and the implications of changing them.

Resolves: #1095

### DIFF
--- a/res/res_pjsip_authenticator_digest.c
+++ b/res/res_pjsip_authenticator_digest.c
@@ -558,6 +558,26 @@ static char *check_auth_result_str[] = {
  * options. If \b any of the incoming Authorization headers result in successful
  * authentication, then authentication is considered successful.
  *
+ * \warning The return code from the function is used by the distributor to
+ * determine which log messages (if any) are emitted.  Many admins will be
+ * using log parsers like fail2ban to block IPs that are repeatedly failing
+ * to authenticate so changing the return code could have unintended
+ * consequences.
+ *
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph:
@alex2grad confirmed that this PR fixed the issue with the missing "Failed to authenticate" log message.

jcolp (approved):

PR #1102, opened 2025-01-30
ARI/pjsip: Make it possible to control transfers through ARI (21,20 only)

zecke:
Introduce a ChannelTransfer event and the ability to notify progress to ARI. Implement emitting this event from the PJSIP channel instead of handling the transfer in Asterisk when configured.

Introduce a dialplan function to the PJSIP channel to switch between the "core" and "ari-only" behavior.

UserNote: Call transfers on the PJSIP channel can now be controlled by ARI. This can be enabled by using the PJSIP_TRANSFER_HANDLING(ari-only) dialplan function.

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -1851,6 +1851,8 @@ static int chan_pjsip_indicate(struct ast_channel *ast, int condition, const voi
 		break;
 	case AST_CONTROL_STREAM_TOPOLOGY_SOURCE_CHANGED:
 		break;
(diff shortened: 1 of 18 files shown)


zecke:
cherry-pick-to: 20
testsuite-test-pr: 77

jcolp (dismissed):

gtjoseph (changes requested):
Documentation versions need to be updated.

jcolp (approved):

Issue #1101, opened 2025-01-29
[bug]: when setting a  var with a double quotes and using Set(HASH)


phoneben:
### Severity

Minor

### Versions

20.11.1

### Components/Modules

hash_read()

### Operating Environment

centos7 ,9

### Frequency of Occurrence

Constant

### Issue Description

CUSTOMPARAM1 =test abu"se
Set(__USR-${EVAL(${CUSTOMPARAM1})}=${EVAL(${CUSTOMPARAM2})});
Set(HASH(USRVARS,${EVAL(${CUSTOMPARAM1})})=${STRREPLACE(var,",","§")});

### Relevant log output

```shell
have backtraces 
Core was generated by `/usr/sbin/asterisk -f -vvvg -c'.
Program terminated with signal 11, Segmentation fault.
#0  0x00007fe1acdfa442 in __strncat_sse2_unaligned () from /lib64/libc.so.6
[...]

PR #1103, opened 2025-01-30
func_strings.c: Prevent SEGV in HASH single-argument mode.

gtjoseph:
When in single-argument mode (very rarely used), a malformation of a column
name (also very rare) could cause a NULL to be returned when retrieving the
channel variable for that column.  Passing that to strncat causes a SEGV.  We
now check for the NULL and print a warning message.

Resolves: #1101


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph:
> That comment has a small typo: "problably" should be "probably."

typos are my trademark :)

jcolp (approved):

Issue #829, opened 2024-08-05
[bug]: Segmentation fault - Failed assertion bad magic number 0x0 for object


duncannz:
### Severity

Major

### Versions

20.8.1

### Components/Modules

core

### Operating Environment

Debian 11.7

### Frequency of Occurrence

Occasional

### Issue Description

We have experienced this segfault twice so far with Asterisk version 20.8.1.

Since we have experienced it twice with version 20.8.1 in ~1 month, but didn't experience it with version 20.6.0 in ~5 months, we think it may be a regression since 20.6.0, however obviously can't say this for certain based on this alone.

The "Failed assertion bad magic number 0x0 for object" error message appears to be logged for each channel that exists at the time of the crash, because it is logged more times in the second occurrence (with more active calls at the time) than the first time (with only 2 active calls at the time). The logs for both occurrences are provided.

I extracted the core dump for the first occurrence and have attached the backtrace text files:

[core-asterisk-2024-07-27T08-19-00Z-brief.txt](https://github.com/user-attachments/files/16491172/core-asterisk-2024-07-27T08-19-00Z-brief.txt)
[core-asterisk-2024-07-27T08-19-00Z-full.txt](https://github.com/user-attachments/files/16491173/core-asterisk-2024-07-27T08-19-00Z-full.txt)
[core-asterisk-2024-07-27T08-19-00Z-info.txt](https://github.com/user-attachments/files/16491175/core-asterisk-2024-07-27T08-19-00Z-info.txt)
[core-asterisk-2024-07-27T08-19-00Z-locks.txt](https://github.com/user-attachments/files/16491177/core-asterisk-2024-07-27T08-19-00Z-locks.txt)
[core-asterisk-2024-07-27T08-19-00Z-thread1.txt](https://github.com/user-attachments/files/16491178/core-asterisk-2024-07-27T08-19-00Z-thread1.txt)

I don't really know how to read them, but a search for "Cannot access memory at address" in full.txt seems interesting.

### Relevant log output

**Occurrence 1 (only 1 active call at the time):**

```
FRACK!, Failed assertion bad magic number 0x0 for object 0x7cd84822f370 (0) at line 3225 in ast_settimeout_full of channel.c
[2024-07-27 07:09:48] ERROR[2418942][C-0000b6c0]: channel.c:3225 ast_settimeout_full: FRACK!, Failed assertion bad magic number 0x0 for object 0x7cd84822f370 (0)
[...]

PR #1104, opened 2025-01-31
res_musiconhold: Appropriately lock channel during start.

jkroonza:
This relates to #829

This doesn't fully solve the Ops issue, but it solves the specific crash there.  Further PRs to follow.

In the specific crash the generator was still under construction when moh was being stopped, which then proceeded to close the stream whilst it was still in use.

### DIFF
--- a/res/res_musiconhold.c
+++ b/res/res_musiconhold.c
@@ -1544,8 +1544,10 @@ static int _moh_unregister(struct mohclass *moh, const char *file, int line, con
  */
 static void local_ast_moh_cleanup(struct ast_channel *chan)
 {
-	struct moh_files_state *state = ast_channel_music_state(chan);
+	struct moh_files_state *state;
 
+	ast_channel_lock(chan);
+	state = ast_channel_music_state(chan);
 	if (state) {
 		ast_channel_music_state_set(chan, NULL);
 		if (state->class) {
@@ -1560,6 +1562,7 @@ static void local_ast_moh_cleanup(struct ast_channel *chan)
 		/* Only held a module reference if we had a music state */
 		ast_module_unref(ast_module_info->self);
 	}
+	ast_channel_unlock(chan);
 }
 
 /*! \brief Support routing for 'moh unregister class' CLI
--- a/res/res_musiconhold.c
+++ b/res/res_musiconhold.c
@@ -291,6 +291,9 @@ static void moh_post_stop(struct ast_channel *chan)
 	ao2_cleanup(message);
 }
 
+/*
+ * Called with `chan` locked.
+ */
 static void moh_files_release(struct ast_channel *chan, void *data)
 {
 	struct moh_files_state *state;
@@ -323,7 +326,10 @@ static void moh_files_release(struct ast_channel *chan, void *data)
 	state->class = mohclass_unref(state->class, "Unreffing channel's music class upon deactivation of generator");
 }
 
-static int ast_moh_files_next(struct ast_channel *chan)
+/*
+ * Called with `chan` locked.
+ */
+static int moh_files_next(struct ast_channel *chan)
 {
 	struct moh_files_state *state = ast_channel_music_state(chan);
 	struct ast_vector_string *files;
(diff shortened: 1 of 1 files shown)


jkroonza:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

res/res_musiconhold.c
```diff
@@ -1762,10 +1762,13 @@ static int local_ast_moh_start(struct ast_channel *chan, const char *mclass, con
 
 			if (ast_test_flag(global_flags, MOH_CACHERTCLASSES)) {
 				/* CACHERTCLASSES enabled, let's add this class to default tree */
+				ast_channel_lock(chan);
+				state = ast_channel_music_state(chan);
```
seanbright, line 1766: Can we just do this (and the unlock) one time at the top of the function? Or do we need to serialize all access to `state`?

jkroonza: I suspect there's potentially some long-running sub-tasks which is why I went a bit more granular.

jcolp: Long-running sub tasks like what?

jkroonza: Good question.  You know better than me.  I didn't dig into this much further beyond this specific PR, life happened and I deviated from this.  My gut was that the construction of some of these generators could be time-consuming, and that that could have adverse impact on the channel structure.

But yea, if any of those do take overly long, it's going to be the case that the channel gets no audio anyway, and it would most likely be signalling stuff that just waits for a "long" duration, and that's probably OK if those gets blocked for a few hundred ms.

Based on that, sure, I think I can resubmit with "lock at top, release at bottom" and we see what happens?

seanbright: If memory serves, when this originally came up I did an audit of this file and I _think_ that the channel lock is serializing access to members of the `moh_files_state` structure, so the way it is here in this PR is probably the safest approach.

It would be helpful to go through and note which functions are being called with the channel lock already held, but that may be out of scope for this PR.

seanbright: > It would be helpful to go through and note which functions are being called with the channel lock already held, but that may be out of scope for this PR.

I went ahead and did this and pushed an additional commit that you can throw away or merge - up to you.

After your change, the only place in the module where the music state is manipulated without the channel lock held is in `local_ast_moh_cleanup` which is only called during channel destruction (but `ast_moh_cleanup()` is currently API so anyone _could_ call it).

I think locking around state manipulation in `local_ast_moh_cleanup` is technically the right thing, and the channel lock is still available at the time it is called, so it should be safe, and I've added another change to this PR that does that.

ashtonmoomoo:
Hi everyone. We appreciate the work that is going into resolving #829.

I wanted to share some very coarse numbers with everybody. We have a script (which we unfortunately can't share, since it interfaces with our application, not Asterisk) that can reproduce #829, and we have been running it against various revisions to see how long it takes before reproduction.

**20.9.3 (no patches applied)**
| Run | # Calls before crash |
| :-: | :-- |
| 1 | 54 |
| 2 | 292 |
| 3 | 208 |
| 4 | 300 |

**20.9.3 (with just [4f5ecca](https://github.com/asterisk/asterisk/pull/1104/commits/4f5ecca43b08878fc75f508e1bb8a746ccf21b82))**
| Run | # Calls before crash |
| :-: | :-- |
| 1 | 77 |
| 2 | 486 |
| 3 | 233 |
| 4 | 257 |

**20.9.3 (with all commits on this branch up to [36f842b](https://github.com/asterisk/asterisk/pull/1104/commits/36f842b88d805d9e348a2ab1aabb237b6d9afae7))**
| Run | # Calls before crash |
| :-: | :-- |
| 1 | 349 |
| 2 | 29 |
| 3 | 567 |
| 4 | 133 |

Obviously more runs would be desirable in all cases but I am not sure if there's value in letting the script run for longer.

So it seems like there is still something going on. If there's any additional diagnostic information that we can share (other than what we shared in #829), please let us know.

jcolp:
What is the current state of this, reviewable despite not resolving the original issue?

jkroonza:
> What is the current state of this, reviewable despite not resolving the original issue?

It solves the original/first lock scenario, there's a second issue.  I've lost interest in this to be honest so won't be spending more time on this.  Someone else is welcome to use this as base, but I suspect the channel was basically being manipulated in such a way that it was being added to a bridge, as well as having moh started on it after it was added to a bridge, resulting in this.  Don't recall the exact details, but the logic for the flow didn't make much sense to me.

Issue #1109, opened 2025-02-04
[bug]: Off nominal memory leak in res/ari/resource_channels.c


gtjoseph:
### Severity

Minor

### Versions

all

### Components/Modules

res/ari/resource_channels.c

### Operating Environment

All

### Frequency of Occurrence

Occasional

### Issue Description

When there's a failure in res/ari/resource_channels.c:ast_ari_channels_external_media(), the `variables` structure isn't being freed.

Thanks @alex2grad for finding.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1110, opened 2025-02-04
resource_channels.c: Fix memory leak in ast_ari_channels_external_media.

gtjoseph:
Between ast_ari_channels_external_media(), external_media_rtp_udp(),
and external_media_audiosocket_tcp(), the `variables` structure being passed
around wasn't being cleaned up properly when there was a failure.

* In ast_ari_channels_external_media(), the `variables` structure is now
  defined with RAII_VAR to ensure it always gets cleaned up.

* The ast_variables_destroy() call was removed from external_media_rtp_udp().

* The ast_variables_destroy() call was removed from
  external_media_audiosocket_tcp(), its `endpoint` allocation was changed to
  to use ast_asprintf() as external_media_rtp_udp() does, and it now
  returns an error on failure.

* ast_ari_channels_external_media() now checks the new return code from
  external_media_audiosocket_tcp() and sets the appropriate error response.

Resolves: #1109

### DIFF
--- a/res/ari/resource_channels.c
+++ b/res/ari/resource_channels.c
@@ -2111,7 +2111,6 @@ static int external_media_rtp_udp(struct ast_ari_channels_external_media_args *a
 		NULL,
 		args->format,
 		response);
-	ast_variables_destroy(variables);
 
 	ast_free(endpoint);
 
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1112, opened 2025-02-05
[bug]: STIR/SHAKEN verification doesn't allow anonymous callerid to be passed to the dialplan.


gtjoseph:
### Severity

Minor

### Versions

all

### Components/Modules

res_stir_shaken.so

### Operating Environment

All

### Frequency of Occurrence

Occasional

### Issue Description

Originally reported on the community forums... [STIR/SHAKEN doesn’t continue for Anonymous calls](https://community.asterisk.org/t/stir-shaken-doesnt-continue-for-anonymous-calls/106788)

When verification is enabled for an endpoint and the failure_action is continue or continue_return_reason and a call is received with no or Anonymous caller_id, the call is terminated rather than allowed to continue into the dialplan.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1113, opened 2025-02-05
res_stir_shaken: Allow missing or anonymous CID to continue to the dialplan.

gtjoseph:
The verification check for missing or anonymous callerid was happening before
the endpoint's profile was retrieved which meant that the failure_action
parameter wasn't available.  Therefore, if verification was enabled and there
was no callerid or it was "anonymous", the call was immediately terminated
instead of giving the dialplan the ability to decide what to do with the call.

* The callerid check now happens after the verification context is created and
  the endpoint's stir_shaken_profile is available.

* The check now processes the callerid failure just as it does for other
  verification failures and respects the failure_action parameter.  If set
  to "continue" or "continue_return_reason", `STIR_SHAKEN(0,verify_result)`
  in the dialplan will return "invalid_or_no_callerid".

* If the endpoint's failure_action is "reject_request", the call will be
  rejected with `433 "Anonymity Disallowed"`.

* If the endpoint's failure_action is "continue_return_reason", the call will
  continue but a `Reason: STIR; cause=433; text="Anonymity Disallowed"`
  header will be added to the next provisional or final response.

Resolves: #1112

### DIFF
--- a/res/res_pjsip_stir_shaken.c
+++ b/res/res_pjsip_stir_shaken.c
@@ -44,10 +44,10 @@ enum sip_response_code {
 	SIP_RESPONSE_CODE_OK = 200,
 	SIP_RESPONSE_CODE_STALE_DATE = 403,
 	SIP_RESPONSE_CODE_USE_IDENTITY_HEADER = 428,
+	SIP_RESPONSE_CODE_ANONYMITY_DISALLOWED = 433,
 	SIP_RESPONSE_CODE_BAD_IDENTITY_INFO = 436,
 	SIP_RESPONSE_CODE_UNSUPPORTED_CREDENTIAL = 437,
 	SIP_RESPONSE_CODE_INVALID_IDENTITY_HEADER = 438,
-	SIP_RESPONSE_CODE_USE_SUPPORTED_PASSPORT_FORMAT = 428,
 	SIP_RESPONSE_CODE_INTERNAL_ERROR = 500,
 };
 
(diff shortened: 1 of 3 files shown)


gtjoseph:
The reporter @marrold confirmed that this PR fixed his issue.

res/res_pjsip_stir_shaken.c
```diff
@@ -44,6 +44,7 @@ enum sip_response_code {
 	SIP_RESPONSE_CODE_OK = 200,
 	SIP_RESPONSE_CODE_STALE_DATE = 403,
 	SIP_RESPONSE_CODE_USE_IDENTITY_HEADER = 428,
+	SIP_RESPONSE_CODE_ANONIMITY_DISALLOWED = 433,
```
seanbright, line 47: ANONIMITY → ANONYMITY

gtjoseph: Heh, I copied that from pjproject. :)

oza4h07:
Can this bug happen, if you receive an incoming call from an carrier with which anonymous calls come in with P-Asserted-Identity and Privacy headers being both set ?

gtjoseph:
@oza4h07 When asterisk parses the P-Asserted-Identity header it also checks for a Privacy header and sets a flag on the internal ast_party_id structure if the Privacy header is found and it's value isn't "none".  The stir/shaken module does NOT use that flag however so if there's a valid sip uri in the PAI header, then stir/shaken will use it.

PR #1115, opened 2025-02-06
channel.c: Remove dead AST_GENERATOR_FD code.

seanbright:
Nothing ever sets the `AST_GENERATOR_FD`, so this block of code will never execute. It also is the only place where the `generate` callback is called with the channel lock held which made it difficult to reason about the thread safety of `ast_generator`s.

In passing, also note that `AST_AGENT_FD` isn't used either.


seanbright:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1124, opened 2025-02-18
[bug]: Race condition between bridge and channel delete can over-write cause code set in hangup.


mbradeen:
### Severity

Trivial

### Versions

All

### Components/Modules

ARI

### Operating Environment

All

### Frequency of Occurrence

None

### Issue Description

If an ARI application sends an bridge delete and channel hangup with cause code simultaneously, this can cause a race condition where the channel exiting the bridge can over-write the cause set in the hangup.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1125, opened 2025-02-18
bridge_channel: don't set cause code on channel during bridge delete if already set

mbradeen:
Due to a potential race condition via ARI when hanging up a channel hangup with cause
while also deleting a bridge containing that channel, the bridge delete can over-write
the hangup cause code resulting in Normal Call Clearing instead of the set value.

With this change, bridge deletion will only set the hangup code if it hasn't been
previously set.

Resolves: #1124


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

PR #1128, opened 2025-02-24
Introduces support for real-time text conversation over RTP (RFC 4103).

xkaraman:
* Co-authored-by: Henning Westerholt <hw@gilawa.com>

Based on a PR from Bharat Ramaswamy Nandakumar <bharat@bramsoft.com>
* Co-authored-by: Kevin Basov <kevin.basov@cgi.com>
* Co-authored-by: Sean Bright <sean@seanbright.com>

Previous changes from PR https://github.com/asterisk/asterisk/pull/998
* chan_pjsip - added write_text function to the pjsip_tech driver
* Added case for AST_FRAME_TEXT in chan_pjsip_write_stream
* Added case to handle text in chan_pjsip_write
* pjsip/dialplan_functions.c: added logic to handle media for TEXT
* res_pjsip.h: added variables to deal with endpoint media
* channel.c: added logic to add a stream_num for a frame
* frame.c: added code to handle and fix issues with red.
* res_pjsip_sdp_rtp.c: added logic to setup SDP.
* Added logic to handle tos and cos text
* Added logic to RED for red enabled
* res_pjsip_session.c: aded logic for max_text_streams
* res_rtp_asterisk.c: Added logic to handle Red and dealing with a repeating RTP
  issue
* pjsip_configuration.c: Added logic to handle text for PJSIP library

New additions:
* Maintain ABI compatibility
* Add alembic
* Add docs
* Adjust length based on stream
* fix whitespace formatting on some cases
* refactoring existing SDP attribute logic to use the existing framework
* fix redundant generation SDP payload mismatch
* remove not needed t140 debug fix, the t140 lengths are reset in the red case
* extend error logging and add some debug messages
* add more comment documentation regarding edge cases and workarounds
* add missing reference counting decrement for scheduler deletion
* add native bridging support for real-time RTP support
* add streaming support for real-time text
* update pjsip_config to new versions

UserNote: This introduces support for real-time text conversation over RTP (RFC 4103).
To enable it you need to add t140 and preferable also red as supported codecs.
You can configure this functionality by three new parameters in the endpoint
section of pjsip.conf: tos_text, cos_text and max_text_streams. If you are
using a database, please have a look to the upgrade notes.

UpgradeNote: This change introduce a schema modification for the ps_endpoint table by
alembic. If you don't use the scripts to update your database, please
adapt the table manually to the new structure. New fields were tos_text,
cos_text and max_text_streams.

Resolves: https://github.com/asterisk/asterisk/issues/997
[...]

PR #1129, opened 2025-02-24
res_rtp_asterisk.c: Use correct timeout value for T.140 RED timer.

seanbright:
Found while reviewing #1128

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -623,7 +623,6 @@ struct rtp_red {
 	unsigned char len[AST_RED_MAX_GENERATION]; /*!< length of each generation */
 	int num_gen; /*!< Number of generations */
 	int schedid; /*!< Timer id */
-	int ti; /*!< How long to buffer data before send */
 	unsigned char t140red_data[64000];
 	unsigned char buf_data[64000]; /*!< buffered primary data */
 	int hdrlen;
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

xkaraman:
Hey @seanbright,

Also as noted by  [docs](https://github.com/asterisk/asterisk/blob/2d7948fa483c8703cf9948811fd67147f4d560d4/include/asterisk/sched.h#L207-L224), the last argument is: `data`.

Previously, the whole `rtp` of type `ast_rtp` was passed. 

In this new patch I see `struct ast_rtp *rtp = ast_rtp_instance_get_data(instance);` but in the 	`rtp->red->schedid = ast_sched_add(rtp->sched, rtp->red->ti, red_write, instance);` `instance` of type `ast_rtp_instance` is passed.

Can you verify or confirm what should be the type expected by `ast_rtp_instance_get_data` since to protoypy is just a `void*`?

seanbright:
I’m sorry I don’t follow. This PR is not changing the `data` argument at all.

Based on how`red_write` is implemented, `data` should be a `struct ast_rtp_instance *`.

seanbright:
The `data` argument is meaningless to `ast_sched_add`. It just blindly passes it to the callback function (in this case `red_write`).

xkaraman:
> I’m sorry I don’t follow. This PR is not changing the `data` argument at all.

Yeah I am asking if it should be changed to `rtp` variable or not due to the modularization you mentioned earlier in https://github.com/asterisk/asterisk/pull/1128.

> Based on how `red_write` is implemented, `data` should be a `struct ast_rtp_instance *`.

Ok, then seems to be correct as is! Thanks for the quick reply!

jcolp (approved):

PR #1136, opened 2025-02-28
Revert "res_rtp_asterisk.c: Set Mark on rtp when timestamp skew is too big"

### LINKED ISSUE #1135: [bug]: Problems with video decoding due to RTP marker bit set
### Severity

Blocker

### Versions

22.2.0

### Components/Modules

res_rtp_asterisk

### Operating Environment

Debian 11 "bullseye" x86_64

### Frequency of Occurrence

Frequent

### Issue Description

The change #912 causes video issues when a large RTP timestamp skew is detected in Asterisk which now results in the marker bit being set in the RTP header. This is not how the marker bit is intended to be used, especially not for video data and it can break the video decoder of the UA receiving the video stream.

In [RFC 3550 section 5.1](https://datatracker.ietf.org/doc/html/rfc3550#section-5.1):

```
   marker (M): 1 bit
      The interpretation of the marker is defined by a profile.  It is
      intended to allow significant events such as frame boundaries to
      be marked in the packet stream.
```

So the usage of the marker bit is defined by the RTP profile, which is a mandatory companion document to RFC 3550. One of those is for example [RFC 3551](https://datatracker.ietf.org/doc/html/rfc3551). For audio, in [section 4.1](https://datatracker.ietf.org/doc/html/rfc3551#section-4.1):

```
   For applications which send either no packets or occasional comfort-
   noise packets during silence, the first packet of a talkspurt, that
   is, the first packet after a silence period during which packets have
   not been transmitted contiguously, SHOULD be distinguished by setting
   the marker bit in the RTP data header to one.  The marker bit in all
   other packets is zero.  The beginning of a talkspurt MAY be used to
   adjust the playout delay to reflect changing network delays.
   Applications without silence suppression MUST set the marker bit to
   zero.
```

For video in [section 5](https://datatracker.ietf.org/doc/html/rfc3551#section-5):

```
   Most of these video encodings also specify that the marker bit of the
   RTP header SHOULD be set to one in the last packet of a video frame
   and otherwise set to zero.
```

And specifically for the H.264 video codec in [RFC 6184 section 5.1](https://datatracker.ietf.org/doc/html/rfc6184#section-5.1):

```
   Marker bit (M): 1 bit
      Set for the very last packet of the access unit indicated by the
      RTP timestamp, in line with the normal use of the M bit in video
      formats, to allow an efficient playout buffer handling.  For
      aggregation packets (STAP and MTAP), the marker bit in the RTP
      header MUST be set to the value that the marker bit of the last
      NAL unit of the aggregation packet would have been if it were
      transported in its own RTP packet.  Decoders MAY use this bit as
      an early indication of the last packet of an access unit but MUST
      NOT rely on this property.
```

So the change in #912 of unconditionally setting the marker bit when a large timestamp skew is detected does not make much sense and breaks things. This change must be reverted and the reporter's issue #927 must be fixed differently.

### Asterisk Issue Guidelines

maximilianfridrich:
This reverts commit f30ad96b3f467739c38ff415e80bffc4afff1da7.

The original change was not RFC compliant and caused issues because it set the RTP marker bit in cases when it shouldn't be set. See the linked issue #1135 for a detailed explanation.

Fixes: #1135.

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -5267,11 +5267,6 @@ static int rtp_raw_write(struct ast_rtp_instance *instance, struct ast_frame *fr
 	}
 
 	if (ast_test_flag(frame, AST_FRFLAG_HAS_TIMING_INFO)) {
-		if (abs(frame->ts * rate - (int)rtp->lastts) > MAX_TIMESTAMP_SKEW) {
-			ast_verbose("(%p) RTP audio difference is %d set mark\n",
-				instance, abs(frame->ts * rate - (int)rtp->lastts));
-			mark = 1;
-		}
 		rtp->lastts = frame->ts * rate;
 	}
 


maximilianfridrich:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

gtjoseph:
**Attention!** This pull request may contain issues that could prevent it from being accepted.  Please review the checklist below and take the recommended action.  If you believe any of these are not applicable, just add a comment and let us know.

- [ ] The PR description and/or commit message has a malformed `Fixes` or `Resolves` trailer. The `Fixes` and `Resolves` keywords MUST be preceeded by a blank line and followed immediately by a colon, a space, a hash sign(`#`), and the issue number. If you have multiple issues to reference, you can add additional `Fixes` and `Resolves` trailers on consecutive lines as long as the first one has the preceeding blank line. A malformed trailer will prevent the issue from being automatically closed when the PR merges and from being listed in the release change logs.<br>   Regular expression: `^(Fixes|Resolves): #[0-9]+$`.<br>   Example: `Fixes: #9999`. 

Documentation:<br>
* [Asterisk Developer Documentation](https://docs.asterisk.org/Development/)<br>
  * [Code Contribution](https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/)<br>
  * [Commit Messages](https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages)<br>
  * [Alembic Scripts](https://docs.asterisk.org/Development/Reference-Information/Other-Reference-Information/Alembic-Scripts/)<br>

gtjoseph:
I love it when a process works. :)  Thanks for being a PR Checklist test subject @maximilianfridrich !

maximilianfridrich:
Anytime @gtjoseph :smile:

PR #1154, opened 2025-03-07
res_rtp_asterisk.c: Don't truncate spec-compliant `ice-ufrag` or `ice-pwd`.

seanbright:
RFC 8839[1] indicates that the `ice-ufrag` and `ice-pwd` attributes can be up to 256 bytes long. While we don't generate values of that size, we should be able to accomodate them without truncating.

1. https://www.rfc-editor.org/rfc/rfc8839#name-ice-ufrag-and-ice-pwd-attri


seanbright:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1155, opened 2025-03-08
[new-feature]: pbx.c: Allow PBX arguments to be modified while PBX is running


InterLinked1:
### Feature Description

Allow PBX arguments to be modified while PBX is running

PR #1156, opened 2025-03-08
pbx.c: Allow PBX arguments to be modified while PBX is running.

InterLinked1:
Currently, an ast_pbx_args structure can only be passed in prior to PBX creation, and the PBX arguments cannot be changed while the PBX is running. This allows the arguments to be changed while the PBX is running (currently, this only allows setting the no_hangup_chan flag as that is the only argument). To enable this, we also ensure that __ast_pbx_run always receives an args structure even if one wasn't passed in, and add an API to update the args later on.

Resolves: #1155

### DIFF
--- a/include/asterisk/pbx.h
+++ b/include/asterisk/pbx.h
@@ -215,8 +215,8 @@ int ast_destroy_timing(struct ast_timing *i);
 struct ast_pbx {
 	int dtimeoutms;				/*!< Timeout between digits (milliseconds) */
 	int rtimeoutms;				/*!< Timeout for response (milliseconds) */
+	struct ast_pbx_args *pbx_args;	/*!< PBX args */
 };
-
 
 /*!
  * \brief Register an alternative dialplan switch
@@ -416,6 +416,26 @@ struct ast_pbx_args {
 		};
 	};
 };
+
+/*!
+ * \brief Update a channel's existing PBX with new arguments
+ *
+ * \param c channel on which the PBX is running
+ * \param args Options for the PBX
+ *
+ * This updates the PBX arguments on an existing PBX with new arguments
+ * that replace any existing arguments. This is useful if PBX arguments
+ * need to be set or modified after the PBX has already been started.
+ * Arguments can be provided even if none were provided to start the PBX initially.
+ *
+ * If possible, avoid this function and use ast_pbx_run_args instead.
+ *
+ * No channel locks need to be held when calling this function.
+ *
+ * \retval 0 on success
+ * \retval -1 on failure (no PBX on channel, or PBX doesn't have an arguments structure)
+ */
+int ast_pbx_set_args(struct ast_channel *c, struct ast_pbx_args *args);
 
 /*!
  * \brief Execute the PBX in the current thread
--- a/main/pbx.c
+++ b/main/pbx.c
@@ -4338,6 +4338,8 @@ static enum ast_pbx_result __ast_pbx_run(struct ast_channel *c,
 		return AST_PBX_FAILED;
 	}
 
+	pbx->pbx_args = args; /* Keep a reference to our arguments, so that we can modify the arguments from the PBX later if desired */
+
 	callid = ast_read_threadstorage_callid();
 	/* If the thread isn't already associated with a callid, we should create that association. */
 	if (!callid) {
@@ -4700,6 +4702,21 @@ static void destroy_exten(struct ast_exten *e)
 	ast_free(e);
 }
 
+/*! \brief Wrapper for __ast_pbx_run. Provides an ast_pbx_args for the PBX if one wasn't passed by the caller. */
+static enum ast_pbx_result _ast_pbx_run(struct ast_channel *c, struct ast_pbx_args *args)
+{
+	struct ast_pbx_args stack_args;
+
+	if (!args) {
+		/* So that any PBX can be updated to add flags while the channel is running,
+		 * give every PBX an arguments structure if it doesn't already one. */
+		memset(&stack_args, 0, sizeof(stack_args));
+		args = &stack_args;
+	}
+
+	return __ast_pbx_run(c, args);
+}
+
 static void *pbx_thread(void *data)
 {
 	/* Oh joyous kernel, we're a new thread, with nothing to do but
@@ -4712,7 +4729,7 @@ static void *pbx_thread(void *data)
 	 */
 	struct ast_channel *c = data;
 
-	__ast_pbx_run(c, NULL);
+	_ast_pbx_run(c, NULL);
 	decrease_call_count();
 
 	pthread_exit(NULL);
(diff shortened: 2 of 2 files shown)


InterLinked1:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

jcolp (changes requested):

include/asterisk/pbx.h
```diff
@@ -417,6 +417,26 @@ struct ast_pbx_args {
...
+ * \param args Options for the PBX
+ *
+ * This updates the PBX arguments on an existing PBX with new arguments
+ * that replace any existing arguments. This is useful if PBX arguments
+ * need to be set or modified after the PBX has already been started.
+ * Arguments can be provided even if none were provided to start the PBX initially.
+ *
+ * If possible, avoid this function and use ast_pbx_run_args instead.
+ *
+ * No channel locks need to be held when calling this function.
```
jcolp, line 433: This doesn't really clarify what it is doing. It should state "This function will lock the channel."

jcolp:
What's the use case on this as well?

InterLinked1:
> What's the use case on this as well?

At a high level, it's to allow transferring ownership of a channel from a PBX thread to another thread. This used to be possible with `AST_PBX_KEEPALIVE`, but that was removed some while ago. The existing arg option is essentially that, but right now it can only be set when the PBX is started - which works for many cases but not all. In the case of dialplan, there are no args, so if we want to set the flag later, we need to be able to modify that argument. That's what this enables.

jcolp:
Can you elaborate further with actual real world examples requiring it?

InterLinked1:
> Can you elaborate further with actual real world examples requiring it?

This is one example: https://github.com/AllStarLink/app_rpt/blob/master/apps/app_rpt.c#L6836
Currently, the only way to make it work correctly is to terminate the PBX thread within the application, so the PBX thread doesn't destroy the channel after it returns, but that's not something applications should be doing.

And in general, if you wanted to accept a call via dialplan somehow and then impart it into some other worker thread (perhaps it services multiple channels in the same thread), this ability would be required. I wouldn't say it's a common use case, but it's something that could make sense to do for certain things.

jcolp:
I'm not going to look at outside code. For the purposes of moving a channel from one thread to another, the ast_channel_yank API call exists for this. Why is that not used?

InterLinked1:
> I'm not going to look at outside code.

Not sure what you are looking for then... you asked for an example - obviously, none would exist in this repo currently.

>For the purposes of moving a channel from one thread to another, the ast_channel_yank API call exists for this. Why is that not used?

That involves making a new channel (masquerading). The PBX thread would have to let the other thread know about the original channel, and then wait for that other thread to do the masquerade by yanking it, since if returned too easily that would still hang up the channel. It doesn't seem like the most elegant approach. At least, if that's the preferred approach for this sort of thing, it's not very efficient.

jcolp:
For real world examples I meant a written scenario/use case. I do not like to put things into the tree without knowing why or understanding the scenario/usage.

The ast_channel_yank function is the current defined mechanism for doing so. Another thread doesn't have to do the operation, the current thread can do it and then pass the new channel elsewhere wherever it may want. Using masquerade guarantees that no matter where the channel was, nothing there can/will interact with it anymore. Your change doesn't guarantee that. It provides the ability to tell the PBX core not to hang up later - but it doesn't guarantee that a caller of ast_pbx_run / ast_pbx_run_args won't still have a channel pointer, won't still interact with the channel, won't still think it has ownership.

InterLinked1:
> For real world examples I meant a written scenario/use case. I do not like to put things into the tree without knowing why or understanding the scenario/usage.

Okay, at a high level, picture an application accepts a call and does some processing. It then inserts it into a list of channels that the module is using for something, and there are other threads in the module that service these channels. The application does some more processing with the channel, and eventually it exits and wants the PBX not to hang it up since the other threads are still using it.

If were were to yank instead in this scenario, the yank would need to occur before anybody else learns about the channel, since if it occurred later, other threads would now be dealing with a channel that masqueraded. And if something happened prior to the application being called that we want to continue to run, a masquerade could be problematic. For example, I seem to recall there were some audiohook caveats where certain stuff would change on a masquerade (might be wrong about that though).

> The ast_channel_yank function is the current defined mechanism for doing so. Another thread doesn't have to do the operation, the current thread can do it and then pass the new channel elsewhere wherever it may want. Using masquerade guarantees that no matter where the channel was, nothing there can/will interact with it anymore. Your change doesn't guarantee that. It provides the ability to tell the PBX core not to hang up later - but it doesn't guarantee that a caller of ast_pbx_run / ast_pbx_run_args won't still have a channel pointer, won't still interact with the channel, won't still think it has ownership.

Yes, actually that's part of the benefit, the simplicity allows for more custom behavior. For example, both threads can still interact with the channel for a time being (locking appropriately, etc. obviously), whereas if the channel is yanked, then suddenly it is invalid everywhere else so this somewhat constricts the possible flows. I believe that was part of the intent with the flag as it exists now, to tell the PBX "not to worry about this channel anymore, we are in charge now and please forget about it".

jcolp:
That sounds complicated and clever channel usage and ownership.

Prior to this change there was a clear intention for the arguments and behavior. The caller of the API indicates what behavior it wants from the PBX core. It could indicate that the channel shouldn't be hung up by the PBX core, so that it could continue to use the channel or handle it, whatever it wishes.

Your change, allowing the arguments to be changed outside, fundamentally alters this behavior and the underlying contract. Instead of the intention of the API caller being observed it can now be ANYONE. It could be changed to go against what the API caller wanted, whether the API caller would tolerate it or not. You're using this as an easy way to stop the channel from being hung up so you can take it over. That's not the intention of this, and I disagree with making it be such a thing.

InterLinked1:
> That sounds complicated and clever channel usage and ownership.

Yes, I don't disagree with that.

> Prior to this change there was a clear intention for the arguments and behavior. The caller of the API indicates what behavior it wants from the PBX core. It could indicate that the channel shouldn't be hung up by the PBX core, so that it could continue to use the channel or handle it, whatever it wishes.
> 
> Your change, allowing the arguments to be changed outside, fundamentally alters this behavior and the underlying contract. Instead of the intention of the API caller being observed it can now be ANYONE. It could be changed to go against what the API caller wanted, whether the API caller would tolerate it or not.

Would you feel any differently if it was one way only, i.e. being able to set the flag but not clear it? That would preserve the existing behavior.

(Put another way, is it worth modifying the PR in some way or should we just close it now?)

> You're using this as an easy way to stop the channel from being hung up so you can take it over. That's not the intention of this, and I disagree with making it be such a thing.

Yes, ultimately, the API could be abused, just like many other APIs if used improperly.
I can understand though if you would prefer not providing that option though.

jcolp:
Changing it to be one way is insufficient. This will not be included.

PR #1162, opened 2025-03-14
res_pjsip_outbound_registration.c: Remove xpointer reference to user_agent.

gtjoseph:
Commit e86f937e added AMI documentation to the module but the cherry-pick
from master to the 20 branch pulled in an xpointer reference to the
`user_agent` config option which doesn't actually exist in 20.  This causes
asterisk to fail to start because it can't load the core-en_US.xml
documentation file.  The CI cherry-pick tests should have caught this but
there was bug in the cherry-picker that was causing the commits to be
cherry-picked to the wrong branch.  Just removing the xpointer reference
resolves the issue.


gtjoseph:
cherry-pick-to: none

jcolp (approved):

gtjoseph:
> My bad!

Mine too.  The cherry-pick tests should have caught it.

PR #1163, opened 2025-03-14
documentation: Update Gosub, Goto, and add new documentationtype.

bkford:
Gosub and Goto were not displaying their syntax correctly on the docs
site. This change adds a new way to specify an optional context, an
optional extension, and a required priority that the xml stylesheet can
parse without having to know which optional parameters come in which
order. In Asterisk, it looks like this:

  parameter name="context" documentationtype="dialplan_context"
  parameter name="extension" documentationtype="dialplan_extension"
  parameter name="priority" documentationtype="dialplan_priority" required="true"

The stylesheet will ignore the context and extension parameters, but for
priority, it will automatically inject the following:

  [[context,]extension,]priority

This is the correct oder for applications such as Gosub and Goto.

### DIFF
--- a/apps/app_stack.c
+++ b/apps/app_stack.c
@@ -49,9 +49,9 @@
 			Jump to label, saving return address.
 		</synopsis>
 		<syntax>
-			<parameter name="context" />
-			<parameter name="exten" />
-			<parameter name="priority" required="true" hasparams="optional">
+			<parameter name="context" documentationtype="dialplan_context" />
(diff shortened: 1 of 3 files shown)


bkford:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

InterLinked1 (changes requested):
"correct order" (typo)

gtjoseph (approved):
I built the documentation manually and it look good!

PR #1168, opened 2025-03-18
audiohook.c: Add ability to adjust volume with float

mkmer:
Add the capability to audiohook for float type volume adjustments.  This allows for adjustments to volume smaller than 6dB.  With INT adjustments, the first step is 2 which converts to ~6dB (or 1/2 volume / double volume depending on adjustment sign). 3dB is a typical adjustment level which can now be accommodated with an adjustment value of 1.41.

This is accomplished by the following:
  Convert internal variables to type float.
  Always use ast_frame_adjust_volume_float() for adjustments.
  Cast int to float in original functions ast_audiohook_volume_set(), and ast_volume_adjust().
  Cast float to int in ast_audiohook_volume_get()
  Add functions ast_audiohook_volume_get_float, ast_audiohook_volume_set_float, and ast_audiohook_volume_adjust_float.

This update maintains 100% backward comparability.


mkmer:
cherry-pick-to: none

mkmer:
Addresses issue https://github.com/asterisk/asterisk/issues/1171

InterLinked1 (changes requested):
The commit message itself needs to end with:

`Resolves: #1171`

jcolp (approved):

PR #1174, opened 2025-03-21
ARI: REST over Websocket

gtjoseph:
This commit adds the ability to make ARI REST requests over the same
websocket used to receive events.

For full details on how to use the new capability, visit...

https://docs.asterisk.org/Configuration/Interfaces/Asterisk-REST-Interface-ARI/ARI-REST-over-WebSocket/

Changes:

* Added utilities to http.c:
  * ast_get_http_method_from_string().
  * ast_http_parse_post_form().
* Added utilities to json.c:
  * ast_json_nvp_array_to_ast_variables().
  * ast_variables_to_json_nvp_array().
* Added definitions for new events to carry REST responses.
* Created res/ari/ari_websocket_requests.c to house the new request handlers.
* Moved non-event specific code out of res/ari/resource_events.c into
  res/ari/ari_websockets.c
* Refactored res/res_ari.c to move non-http code out of ast_ari_callback()
  (which is http specific) and into ast_ari_invoke() so it can be shared
  between both the http and websocket transports.

UpgradeNote: This commit adds the ability to make ARI REST requests over the same
websocket used to receive events.
See https://docs.asterisk.org/Configuration/Interfaces/Asterisk-REST-Interface-ARI/ARI-REST-over-WebSocket/


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
testsuite-test-pr: 83

gtjoseph:
WARNING: The protocol has changed with the latest push.  See the documentation for updated info...
https://docs.asterisk.org/Configuration/Interfaces/Asterisk-REST-Interface-ARI/ARI-REST-over-WebSocket/

mbradeen (approved):

main/json.c
```diff
@@ -861,6 +861,83 @@ enum ast_json_to_ast_vars_code ast_json_to_ast_variables(struct ast_json *json_v
 	return AST_JSON_TO_AST_VARS_CODE_SUCCESS;
 }
 
+enum ast_json_nvp_ast_vars_code ast_json_nvp_array_to_ast_variables(
+	struct ast_json *json_variables, struct ast_variable **variables)
+{
+	struct ast_variable *tail = NULL;
+	int i = 0;
+	size_t len = json_variables ? ast_json_array_size(json_variables) : 0;
+
```
jcolp, line 870: Should *variables be set to NULL at the start?

main/json.c
```diff
@@ -861,6 +861,83 @@ enum ast_json_to_ast_vars_code ast_json_to_ast_variables(struct ast_json *json_v
...
+		}
+		value = ast_json_string_get(json_key);
+
+		new_var = ast_variable_new(key, value, "");
+		if (!new_var) {
+			/* Error: OOM */
+			return AST_JSON_NVP_AST_VARS_CODE_OOM;
+		}
+
+		tail = ast_variable_list_append_hint(variables, tail, new_var);
```
jcolp, line 908: It's possible for an error to be returned but variables to be set. This behavior should be documented for the function.

res/ari/ari_websocket_requests.c
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	request = ast_calloc(1, sizeof(*request));
+	if (!request) {
+		SET_RESPONSE_AND_EXIT(500,
+			"Server error","Out of memory",
+			remote_addr, request, NULL);
+	}
+
+	/* transaction_id is optional */
```
jcolp, line 99: If transaction_id exists in request_msg, but the ast_strdup fails, does this break an understanding/contract between the server and the client as to the existence/usage of it? (Same applies to others)

jcolp (changes requested):

main/json.c
gtjoseph: No because there may already be variables you want to keep. For this case, they get accumulated from the query string, the query_strings property and the the message body.  I'll document that.  Actually, thinking about it, a new variable should replace an existing one with the same name.  I'll fix that as well.

gtjoseph: done

jcolp (approved):

Issue #1181, opened 2025-03-25
[bug]: Incorrect PJSIP Endpoint Device States on Multiple Channels


alex2grad:
### Severity

Trivial

### Versions

20.12.0,  21.7.0,  22.2.0,  20.13.0-rc1,  21.8.0-rc1,  22.3.0-rc1

### Components/Modules

channels/chan_pjsip

### Operating Environment

Linux, Slackware 15.0 x86_64

### Frequency of Occurrence

Constant

### Issue Description

The Asterisk PJSIP module incorrectly determines the aggregate device state when handling multiple channels.

1. Incorrect ONHOLD state determination:
When one channel is placed on hold, the device state is set to ONHOLD without checking other channels states.
The device state should be result of aggregate channels states.
If device_state_busy_at is equal to the number of channels in use, the correct state should be BUSY.

2. Incorrect channels usage determination:
A channel should be considered "in use" if its state is anything other than DOWN or RESERVED.
The current implementation does not correctly classify channels in use (only UP, RING and BUSY), potentially leading to inaccurate device states.

3. Incorrect BUSY state determination:
The endpoint device state should be BUSY if the number of channels "in use" is greater than or equal to device_state_busy_at.
Currently, if the number of channels in use is greater than device_state_busy_at, the system does not set the state to BUSY. Instead, it incorrectly assigns an aggregate device state.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

alex2grad:
I'll provide PR

PR #1182, opened 2025-03-25
chan_pjsip: set correct Endpoint Device State on multiple channels

alex2grad:
1. When one channel is placed on hold, the device state is set to ONHOLD
without checking other channels states.
In case of AST_CONTROL_HOLD set the device state as AST_DEVICE_UNKNOWN
to calculate aggregate device state of all active channels.

2. The current implementation incorrectly classifies channels in use.
The only channels that has the states: UP, RING and BUSY are considered as "in use".
A channel should be considered "in use" if its state is anything other than
DOWN or RESERVED.

3. Currently, if the number of channels "in use" is greater than device_state_busy_at,
the system does not set the state to BUSY. Instead, it incorrectly assigns an aggregate
device state.
The endpoint device state should be BUSY if the number of channels "in use" is greater
than or equal to device_state_busy_at.

Fixes: #1181

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -1221,15 +1221,14 @@ static int chan_pjsip_devicestate(const char *data)
 			ast_devstate_aggregate_add(&aggregate, ast_state_chan2dev(snapshot->state));
 		}
 
-		if ((snapshot->state == AST_STATE_UP) || (snapshot->state == AST_STATE_RING) ||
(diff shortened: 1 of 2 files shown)


alex2grad:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

PR #1187, opened 2025-04-02
ari_websockets: Fix frack if ARI config fails to load.

gtjoseph:
ari_ws_session_registry_dtor() wasn't checking that the container was valid
before running ao2_callback on it to shutdown registered sessions.


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

PR #1194, opened 2025-04-10
asterisk.c: Add "pre-init" and "pre-module" capability to cli.conf.

gtjoseph:
Commands in the "[startup_commands]" section of cli.conf have historically run
after all core and module initialization has been completed and just before
"Asterisk Ready" is printed on the console. This meant that if you
wanted to debug initialization of a specific module, your only option
was to turn on debug for everything by setting "debug" in asterisk.conf.

This commit introduces options to allow you to run CLI commands earlier in
the asterisk startup process.

A command with a value of "pre-init" will run just after logger initialization
but before most core, and all module, initialization.

A command with a value of "pre-module" will run just after all core
initialization but before all module initialization.

A command with a value of "fully-booted" (or "yes" for backwards
compatibility) will run as they always have been...after all
initialization and just before "Asterisk Ready" is printed on the console.

This means you could do this...

```
[startup_commands]
core set debug 3 res_pjsip.so = pre-module
core set debug 0 res_pjsip.so = fully-booted
```

This would turn debugging on for res_pjsip.so to catch any module
initialization debug messages then turn it off again after the module is
loaded.

UserNote: In cli.conf, you can now define startup commands that run before
core initialization and before module initialization.

### DIFF
--- a/configs/samples/cli.conf.sample
+++ b/configs/samples/cli.conf.sample
@@ -7,5 +7,21 @@
 ; Any commands listed in this section will get automatically executed
 ; when Asterisk starts as a daemon or foreground process (-c).
 ;
+; Commands with a value of "pre-init" will run just after the logger
+; is initialized but before all other core and module initialization.
+;
+; Commands with a value of "pre-module" will run just after core
+; initialization is done but before module initialization.
+;
+; These two values can be used for things like enabling debugging on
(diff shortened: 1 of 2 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

PR #1195, opened 2025-04-10
res_pjproject: Fix DTLS client check failing on some platforms

A1bi:
Certain platforms (mainly BSD derivatives) have an additional length field in `sockaddr_in6` and `sockaddr_in`.
`ast_sockaddr_from_pj_sockaddr()` does not take this field into account when copying over values from the `pj_sockaddr` into the `ast_sockaddr`. The resulting `ast_sockaddr` will have an uninitialized value for `sin6_len`/`sin_len` while the other `ast_sockaddr` (not converted from a `pj_sockaddr`) to check against in `ast_sockaddr_pj_sockaddr_cmp()` has the correct length value set.

This has the effect that `ast_sockaddr_cmp()` will always indicate an address mismatch, because it does a bitwise comparison, and all DTLS packets are dropped even if addresses and ports match.

`ast_sockaddr_from_pj_sockaddr()` now checks whether the length fields are available on the current platform and sets the values accordingly.

Resolves: #505

### DIFF
--- a/configure
+++ b/configure
@@ -28233,6 +28233,25 @@ printf "%s\n" "#define HAVE_STRUCT_STAT_ST_MTIMESPEC 1" >>confdefs.h
 stat_nsec_found=yes
 fi
 
+ac_fn_c_check_member "$LINENO" "struct sockaddr_in6" "sin6_len" "ac_cv_member_struct_sockaddr_in6_sin6_len" "#include <netinet/in.h>
+"
+if test "x$ac_cv_member_struct_sockaddr_in6_sin6_len" = xyes
+then :
+
+printf "%s\n" "#define HAVE_STRUCT_SOCKADDR_IN6_SIN6_LEN 1" >>confdefs.h
+
+
+fi
+ac_fn_c_check_member "$LINENO" "struct sockaddr_in" "sin_len" "ac_cv_member_struct_sockaddr_in_sin_len" "#include <netinet/in.h>
+"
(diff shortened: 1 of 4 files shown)


A1bi:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

seanbright (changes requested):
Thanks for the PR.

If the values of `sin_len` and `sin6_len` don't actually matter (save for the comparison) I would prefer to clear the destination `ast_sockaddr` on function entry and then we don't need to check for/set the non-standard fields. Essentially the `memset(...)` from the `else` branch below.

A1bi:
> If the values of `sin_len` and `sin6_len` don't actually matter (save for the comparison) I would prefer to clear the destination `ast_sockaddr` on function entry and then we don't need to check for/set the non-standard fields.

I’m not sure if I got it right. You want to first clear the target `ast_sockaddr` in `ast_sockaddr_from_pj_sockaddr()` so the length field will always be zero? This wouldn’t fix the issue, because the non-converted `ast_sockaddr` to compare against in `ast_sockaddr_pj_sockaddr_cmp()` will have the correct non-zero value for `sin_len` and `sin6_len` that was originally set by the platform’s socket function, so the lengths still don’t match.

seanbright:
> This wouldn’t fix the issue

Yes, of course. Sorry for the noise.

jcolp (approved):

Issue #1197, opened 2025-04-16
[bug]: ChannelHangupRequest does not show cause code in all cases


jcolp:
### Severity

Minor

### Versions

20, 21, 22

### Components/Modules

channel

### Operating Environment

All

### Frequency of Occurrence

Constant

### Issue Description

In some cases the ChannelHangupRequest event would not contain the cause code.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1198, opened 2025-04-16
channel: Always provide cause code in ChannelHangupRequest.

jcolp:
When queueing a channel to be hung up a cause code can be specified in one of two ways:

1. ast_queue_hangup_with_cause This function takes in a cause code and queues it as part of the hangup request, which ultimately results in it being set on the channel.

2. ast_channel_hangupcause_set + ast_queue_hangup This combination sets the hangup cause on the channel before queueing the hangup instead of as part of that process.

In the #2 case the ChannelHangupRequest event would not contain the cause code. For consistency if a cause code has been set on the channel it will now be added to the event.

Resolves: #1197

### DIFF
--- a/main/channel.c
+++ b/main/channel.c
@@ -1168,13 +1168,21 @@ int ast_queue_frame_head(struct ast_channel *chan, struct ast_frame *fin)
 /*! \brief Queue a hangup frame for channel */
 int ast_queue_hangup(struct ast_channel *chan)
 {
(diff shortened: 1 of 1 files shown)


jcolp:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph (approved):

PR #1199, opened 2025-04-16
contrib: Add systemd service and timer files for malloc trim.

bkford:
Adds two files to the contrib/systemd/ directory that can be installed
to periodically run "malloc trim" on Asterisk. These files do nothing
unless they are explicitly moved to the correct location on the system.
Users who are experiencing Asterisk memory issues can use this service
to potentially help combat the problem. These files can also be
configured to change the start time and interval. See systemd.timer(5)
and systemd.time(7) for more information.

UserNote: Service and timer files for systemd have been added to the
contrib/systemd/ directory. If you are experiencing memory issues,
install these files to have "malloc trim" periodically run on the
system.

### DIFF
--- a/contrib/systemd/asterisk-malloc-trim.service
+++ b/contrib/systemd/asterisk-malloc-trim.service
@@ -1,0 +1,9 @@
+# Run the `malloc trim` CLI command to return unused memory
+# to the OS.  Triggered automatically by asterisk-malloc-trim.timer.
+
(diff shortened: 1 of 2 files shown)


bkford:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph (approved):

PR #1202, opened 2025-04-17
pjsip: Allow for a transport-specific suffix onto received usernames.

jkroonza:


jkroonza:
Need testing.

Need documentation updates.

jkroonza:
I'm fairly certain I'm screwing something up here (applied to asterisk 21.7.0):

``` Endpoint:  299-plain/299                                        Not in use    0 of inf
     InAuth:  299/299
        Aor:  299                                               10
      Contact:  299/sip:299@165.16.203.251:5060;x-ast-orig 1f8112fffc Avail        10.777

 Endpoint:  299-rtc/299                                          Not in use    0 of inf
     InAuth:  299/299
        Aor:  299                                               10
      Contact:  299/sip:299@165.16.203.251:5060;x-ast-orig 1f8112fffc Avail        10.777

 Endpoint:  299-tls/299                                          Not in use    0 of inf
     InAuth:  299/299
        Aor:  299                                               10
      Contact:  299/sip:299@165.16.203.251:5060;x-ast-orig 1f8112fffc Avail        10.777
```

Client is registering over udp to the server on port 5059.

```
arthur*CLI> pjsip show aor 299 

      Aor:  <Aor..............................................>  <MaxContact>
    Contact:  <Aor/ContactUri............................> <Hash....> <Status> <RTT(ms)..>
==========================================================================================

      Aor:  299                                                 10
    Contact:  299/sip:299@165.16.203.251:5060;x-ast-orig-h 1f8112fffc Avail         8.901


 ParameterName        : ParameterValue
 =====================================================================================
 authenticate_qualify : false
 contact              : sip:299@165.16.203.251:5060;x-ast-orig-host=192.168.42.86:5060
 default_expiration   : 3600
 mailboxes            : 
 max_contacts         : 10
 maximum_expiration   : 7200
 minimum_expiration   : 60
 outbound_proxy       : 
 qualify_frequency    : 60
 qualify_timeout      : 2.000000
 remove_existing      : false
 remove_unavailable   : false
 support_path         : false
 voicemail_extension  : *97
```

Not in a position to be testing calls from the specific registering device right now, will only be with the physical device on Friday, but this in my opinion looks positive.  Using WebRTC does break though, with some extra debug in the find_endpoint and find_endpoint2 functions:

```
[Apr 29 22:41:50] DEBUG[31336]: res_pjsip_endpoint_identifier_user.c:184 username_identify: Attempting identify by From username '299' domain 'arthur.iewc.co.za'
[Apr 29 22:41:50] DEBUG[31336]: res_pjsip_endpoint_identifier_user.c:152 find_endpoint: Found transport: No
[Apr 29 22:41:50] DEBUG[31336]: res_pjsip_endpoint_identifier_user.c:106 find_endpoint2: Attempting lookup for 299@arthur.iewc.co.za
[Apr 29 22:41:50] DEBUG[31336]: res_pjsip_endpoint_identifier_user.c:188 username_identify: Endpoint not found for From username '299' domain 'arthur.iewc.co.za'
```

which means that find_transport doesn't find the transport in the case of WebRTC.

I can probably work around this in my use-case by just dealing with no-suffix as the RTC case and using suffix for -plain and -tls only, but that just seems inconcistent.  And way to get around this?

The three endpoints now share an aor, and I'm wondering if that's the reason for all three endpoints showing the Contact record?

jcolp:
Technically speaking a transport is not required to be defined in .conf for WebRTC, because the HTTP server is separately configured/setup/implemented. The logic for finding the transport may not work in that case. You're in uncharted territory with that so I don't know anything further.

An AOR contains contacts, if an endpoint references an AOR then it will have access to all the contacts on the AOR. That's how it works.

jkroonza:
modified to match the type_name from pjsip against the id of the transport_state on asterisk side such that it becomes possible to setup (pjsip.conf):

```
[WSS]
type=transport
protocol=ws
bind=[::1]:5062
endpoint_suffix=-rtc
```

For example.  There are a few things about this I don't like myself:

1. It's a massive special case, and can cause problems if people define transports named TCP, UDP or TLS as well.
2. Extra strcmp() for every object check match (if you have 8 transports, and a couple of templates, then that's potentially a fair number of strcmp()s).

I'm not seeing a better way to achieve this.  I'm not sure how other people deal with being able to use the same auth user over multiple transports and have endpoint settings that match the transport in use.

jcolp (changes requested):
I'm not a fan of this implementation either. It's an approach that can meet specific needs, but I think it's a band-aid approach. I'm opting to not allow inclusion of this in the tree.

jkroonza:
Can you think of a non-band-aid approach that I can look into?

jcolp:
Not off the top of my head, no. Requirements would have to be defined, then the inherent limitations defined of the existing, and out of that possibilities determined. This would be a discussion for the -dev list.

asteriskteam:

jkroonza:
@jcolp In the simplest way I can phrase the requirement:

The same authenticating user by way of username must be usable over any transport (udp, tcp, tls or websocket) and have parameters set according to the transport, not the username.

jcolp:
> @jcolp In the simplest way I can phrase the requirement:
> 
> The same authenticating user by way of username must be usable over any transport (udp, tcp, tls or websocket) and have parameters set according to the transport, not the username.

That's an implementation level detail requirement. Requirements would need to be higher than that, without implementation details.

jkroonza:
How about:

No matter which transport (udp, tcp, tls and rtc being the known options currently) the client chooses to use, credentials should remain the same whilst still selecting media transport parameters that's appropriate for the signalling protocol (for example, enable MUX when using webrtc, whether or not to enforce media encryption, as well as the type of encryption SDES vs DTLS).

Otherwise I don't know, transport here refers to the actual wire protocol (not the specific configured transport object in pjsip).  Of course the two are closely correlated, but the one refers to what's actually happening whereas the other refers to the way pjsip configures it.

jcolp:
Sure, though that's still technical level.

"I want to be able to use a single endpoint for both WebRTC and non-WebRTC devices."

That's a high level requirement. Drilling down into what that means and the scenarios would be the next step. An assumption in your case is that the signalling protocol dictates what to use, but that doesn't work if someone is terminating WSS outside using a proxy and it is UDP inside. Then WebRTC can flow over UDP.

jcolp:
Which then becomes fundamentally SDP negotiation and configuration aspects. Have you looked at RFCs or anything to see if there are defined methods for handling this?

Without doing that off the top of my head:

On incoming calls Asterisk can know what the remote side wants and adjust accordingly. That's the easy part. Outgoing is where it gets difficult, so could we instead have a fallback. That is: Offer non-WebRTC to the endpoint, if that gets a 488 enable WebRTC and send again. The consequence is 2 INVITEs of course.

jkroonza:
Your second last message makes a whole lot of sense and is not a use-case I considered.

Of course in such a case you could define two separate UDP transports, one for general use, and one for use by the proxy (which then uses RTC parameters).

But it does raise a point .. what if the proxy will "translate" from TCP, TLS and WebRTC to UDP?  Can we handle that on a single UDP transport on pjsip (asterisk) side?  For example by using some extra header from the REGISTER as part of the endpoint selection process ... (again, this is probably getting technical).

As a rule (and not detailing) what I've found is that parameters for devices using UDP + TCP as SIP transport is generally the same (no encryption), whereas TLS seems to use the SDES based encryption stuff, and RTC more on the DTLS side, and of course, requires RTP+RTCP muxing (single UDP flow).

There may be significantly more to it than that, but so far I've found that that covers it for us at least.

jcolp:
I can 100% guarantee that there are deployments that DO terminate the signaling via TCP, TLS, WebRTC on the proxy and then use a single UDP endpoint in Asterisk and single transport.

jkroonza:
Just the fact that you mentioned it already convinced me, and for that the current methodology is probably the simplest best possible option given the current constraints.

Taking that into consideration, and not knowing that that implementation looks like, but sounds like you do:

Is there something available in the REGISTER that asterisk can use to determine which protocol is used on the other side?

By default backwards compatibility should be maintained.  In this I firmly believe, but if we can incorporate that information then I believe it should be possible to make more intelligent decisions for endpoint selection.  Ie, if the proxy adds some header or tag that can be used to deviate the endpoint, then possibly that could improve things for everyone, not just us.

My patch is thus good enough for us, and I'm actually already starting to build on top of it, but nothing production yet, just playing with it currently.  So far it works.  But I agree that not currently in any shape ready for merging into the project here.  However, I do like the conceptual idea.

jcolp:
> Is there something available in the REGISTER that asterisk can use to determine which protocol is used on the other side?

Maybe, maybe not.

jkroonza:
> > Is there something available in the REGISTER that asterisk can use to determine which protocol is used on the other side?
> 
> Maybe, maybe not.

Can't say (as in don't know) or won't say as in NDA?

Again, my suggestion is that we should maintain backwards compatibility.

As such I reckon an option on transport to say to use some tag or header value (possibly transformed by a sed style transformation) to determine an endpoint suffix (similar to the patch) *could* potentially make things more flexible in such scenarios, or possibly a fixed string to be appended.  Thus three options [...] here is optional:

endpoint_suffix = header:X-Custom[:/^.*something:(.*)$/\1/]
endpoint_suffix = tag:transport[:/../../]
endpoint_suffix = -tls

This gets technical.  For our purposes the last should suffice.

By default there is no endpoint_suffix (current status quo is retained).

And if endpoint_suffix begins with header: or tag: then the appropriate tag or header is used to suffix.

jcolp:
> Can't say (as in don't know) or won't say as in NDA?

Neither.  You can purposely obfuscate the topology in the proxy side to hide such things so they aren't visible.

And yes, this is technical and still focused on how you want to fix the issue for your usage and the work you've done. That still doesn't mean this is the right solution in the grand scheme of things. I'm not ready to accept this as the "solution" even with changes.

jkroonza:
That's fine.  I can deal with an honest answer.

I believe you now better understand our requirement?  At least, I hope.

Given that, would you mind letting your mind think on it a bit and let me know your thoughts and what avenue I could pursue that would satisfy muster and eventually at least have potential of getting merged rather than having to maintain a separate patch?

Regarding "purposefully obfuscate", in such a case would not the proxy also potentially become responsible for translating/bridging potentially different RTP requirements?  After all, to work with that in the current situation asterisk would need to be configured per endpoint already, and none of what I propose dictate that to change, nor do I think the default behaviour should change.  I'm aiming to give another possible alternative.

So if a sysadmin wants to obfuscate those details, he can keep doing so by using the current configuration, but if a sysadmin wants to utilise template-driven setups then either the proxy can be configured to add a header or tag to indicate to asterisk which profile should be used, or a "profile" (endpoint suffix) can be selected/enforced based on the asterisk transport being used by the peer.

Either way, our base requirement is that media transport parameters be configurable based on the transport protocol in use, not based on the username that's being used to authenticate, such that the same set of credentials can be used irrespective of the transport chosen to connect to asterisk.

I think what you added expands on that quite a bit, and is a sensible addition.

Either way - this has to be conditional functionality such that the current mechanisms does not change *by default*.  This should be a hard requirement of any implementation here.

jcolp:
> I believe you now better understand our requirement? At least, I hope.

Sure, but the reason I asked for discussion to occur on the asterisk-dev mailing list originally is so that other people can provide input/experience/requirements as well. You have your usage you're trying to solve, but as I've demonstrated there are avenues that aren't applicable to you/that you didn't think of. The asterisk-dev list opens up discussion to others if they want to participate. On Github unless someone sees this, it's just you and I talking and we don't represent everyone or don't have other cases in our minds - and any solution should not just cover your usage. To the best of our ability we should ensure it covers others that are applicable to this.

> Regarding "purposefully obfuscate", in such a case would not the proxy also potentially become responsible for translating/bridging potentially different RTP requirements?

Maybe, maybe not. That's up to the deployer.

> Either way, our base requirement is that media transport parameters be configurable based on the transport protocol in use, not based on the username that's being used to authenticate, such that the same set of credentials can be used irrespective of the transport chosen to connect to asterisk.

I disagree that this is a base requirement. This reads as an approach to solving your issue. "If media transport parameters are configurable based on the transport protocol in use then it solves my problem".

Would it help that if instead of base requirements/requirements I said use cases?

1. Here are my use cases and what my existing configuration is like
2. Here are where they fail
3. Here is why they fail
4. Here is what I would like to happen

With this discussion can take place (and others can fill in gaps) and the best way to solve it can be determined. Maybe it is your approach, maybe it isn't. I want to ensure we make the best choice possible because ultimately we need to ensure it works for the needs of as many people as possible in the most straightforward way possible, and that it is something that the project is comfortable with and can maintain.

asteriskteam:

asteriskteam:

PR #1205, opened 2025-04-18
pbx_ael: unregister AELSub application and CLI commands on module load failure

andreas-wehrmann:
This fixes crashes/hangs I noticed with Asterisk 20.3.0 and 20.13.0 and quickly found out,
that the AEL module doesn't do proper cleanup when it fails to load.
This happens for example when there are syntax errors and AEL fails to compile in which case pbx_load_module()
returns an error but load_module() doesn't then unregister CLI cmds and the application.

### DIFF
--- a/pbx/pbx_ael.c
+++ b/pbx/pbx_ael.c
@@ -270,11 +270,21 @@ static int unload_module(void)
 
 static int load_module(void)
 {
+	int pbx_load_res = AST_MODULE_LOAD_SUCCESS;
+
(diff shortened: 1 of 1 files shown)


andreas-wehrmann:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

gtjoseph:
> I think the failed pipeline stages are unrelated?

Yep, nothing to worry about.

PR #1208, opened 2025-04-21
docs: Remove `module` attribute from XML docs.

seanbright:
It is only present in 3 modules and only appears to break some documentation links.

### DIFF
--- a/res/res_fax.c
+++ b/res/res_fax.c
@@ -90,7 +90,7 @@
 #include "asterisk/format_cache.h"
 
 /*** DOCUMENTATION
-	<application name="ReceiveFAX" language="en_US" module="res_fax">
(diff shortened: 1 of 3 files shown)


seanbright:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

InterLinked1 (commented):
This wouldn't be backwards compatible, but should it also be removed from the DTD then? If there is an issue then I suppose we don't want that attribute to be used.

Also, the 2nd 'only' in the commit message reads funny, was that intentional?

seanbright:
> Also, the 2nd 'only' in the commit message reads funny, was that intentional?

It was. As in “it’s only function seems to be to mess up the docs.”

PR #1211, opened 2025-04-24
file.c: Increment the channel reference count when it is added as owner to a ast_filestream

Josh-Hogan:
SEGFAULTS and memory corruption has been reported at locations where the ast_filestream->owner is assumed to point to a vaild channel, sometimes it does not (and appears to have been freed).

Fixes: #829

UserNote: Ensure we increment the channel reference count when it is added as the owner of a ast_filestream and decrement it when the ast_filestream is destroyed


Josh-Hogan:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

Josh-Hogan:
Closing in favor of #1213 as this is a bit of a mess now

PR #1212, opened 2025-04-24
ARI Outbound Websockets

gtjoseph:
Asterisk can now establish websocket sessions _to_ your ARI applications
as well as accepting websocket sessions _from_ from them.
Full details:
[ARI Outbound Websockets](https://docs.asterisk.org/Configuration/Interfaces/Asterisk-REST-Interface-ARI/ARI-Outbound-Websockets)

Code change summary:
* Added an ast_vector_string_join() function,
* Added ApplicationRegistered and ApplicationUnregistered ARI events.
* Converted res/ari/config.c to use sorcery to process ari.conf.
* Added the "outbound-websocket" ARI config object.
* Refactored res/ari/ari_websockets.c to handle outbound websockets.
* Refactored res/ari/cli.c for the sorcery changeover.
* Updated res/res_stasis.c for the sorcery changeover.
* Updated apps/app_stasis.c to allow initiating per-call outbound websockets.
* Added CLI commands to manage ARI websockets.
* Added the new "outbound-websocket" object to ari.conf.sample.

UserNote: Asterisk can now establish websocket sessions _to_ your ARI applications
as well as accepting websocket sessions _from_ from them.
Full details:
[ARI Outbound Websockets](https://docs.asterisk.org/Configuration/Interfaces/Asterisk-REST-Interface-ARI/ARI-Outbound-Websockets)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
testsuite-test-pr: 85

res/ari/config.c
```diff
@@ -22,239 +22,623 @@
...
+	}
+
+	if (res != 0) {
+		ast_log(LOG_WARNING, "%s: Outbound websocket configuration failed\n", id);
+	} else {
+		ast_debug(3, "%s: Outbound websocket configuration succeeded\n", id);
+
+		if (owc->reconnect_attempts == DEFAULT_RECONNECT_ATTEMPTS) {
+			if (owc->connection_type == ARI_WS_TYPE_OUTBOUND_PERSISTENT) {
+				owc->reconnect_attempts = INT_MAX;
```
bkford, line 167: It may not matter, but would it make sense to have this be a check in ari_websockets.c for persistent instead of counting down to make it truly infinite? The number is pretty high so it probably doesn't matter, but figured I'd just throw it out there in case. This isn't that big of a deal to me.

gtjoseph: Actually, it's a bug to set it to INT_MAX.  I changed the meaning for reconnect_attempts for persistent connections to be "reconnect forever but print a log message every reconnect_attempts tries" but forgot to change it here. :)  The default should be 4 for both types of connections.

jcolp (approved):

PR #1213, opened 2025-04-25
file.c: Increment the channel reference count when it is added as owner to a ast_filestream

Josh-Hogan:
SEGFAULTS and memory corruption has been reported at locations where the ast_filestream->owner is assumed to point to a vaild channel, sometimes it does not (and appears to have been freed).

Fixes: #829

UserNote: Ensure we increment the channel reference count when it is added as the owner of a ast_filestream and decrement it when the ast_filestream is destroyed

### DIFF
--- a/include/asterisk/file.h
+++ b/include/asterisk/file.h
@@ -356,7 +356,7 @@ struct ast_filestream *ast_openstream_full(struct ast_channel *chan, const char 
 struct ast_filestream *ast_openvstream(struct ast_channel *chan, const char *filename, const char *preflang);
 
 /*!
- * \brief Applies a open stream to a channel.
+ * \brief Applies a open stream to a channel. And bumps the reference count of the channel.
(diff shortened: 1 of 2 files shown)


Josh-Hogan:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

Josh-Hogan:
@duncannz Let me know how you get on testing this. You might need to cherry pick the changes onto the Asterisk version you are using - example I have done for 20.9.3 https://github.com/asterisk/asterisk/compare/20.9.3...Josh-Hogan:20.9.3-with-josh-patch.diff

If it works I can undraft this PR.  
If it doesn't work - happy to make more changes if you have more core-dumps.

Josh-Hogan:
Closing this for now as there is likely more complicated race conditions at play and no need for CICD to keep running

PR #1219, opened 2025-04-25
build_system: Process MAKEOPTS blocks in main source files.

gtjoseph:
The source files in the module directories are scanned for blocks
beginning with `/*** MAKEOPTS` which contain XML fragments that control
menuselect.  The source files in the main directory however weren't scanned.
This commit adds rules to the top-level Makefile and main/Makefile to
do the scanning.  The rules are a bit different for main because the directory
list that controls the MAKEOPTS scanning also controls compiler options for
modules and they're not appropriate for source files in main that will be
linked into the `asterisk` executable.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 *.exports
 *.moduleinfo
 *.makeopts
+*.main-makeopts
 *.makedeps
 .lastclean
 /.pc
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph:
Will submit as part of the final Channelstorage PR.

Issue #1222, opened 2025-04-26
[bug]: func_callerid: ANI2 is not always formatted as two digits


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

func_callerid

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

ANI2 is always supposed to be formatted as two digits. Since we store it as an integer and print it out directly, it is sometimes printed out as a single digit, which is wrong.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1223, opened 2025-04-26
func_callerid: Always format ANI2 as two digits.

InterLinked1:
ANI II is always supposed to be formatted as two digits, so zero pad when formatting it if necessary.

Resolves: #1222


InterLinked1:
cherry-pick-to: none

seanbright (approved):

jcolp (approved):

Issue #1224, opened 2025-04-26
[improvement]: app_meetme: Removal version is incorrect


InterLinked1:
### Improvement Description

`app_meetme` was not removed in version 21 as planned, so this documentation is incorrect.

PR #1225, opened 2025-04-26
app_meetme: Remove inaccurate removal version from xmldocs.

InterLinked1:
app_meetme is deprecated but wasn't removed as planned in 21, so remove the inaccurate removal version.

Resolves: #1224


InterLinked1:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

PR #1228, opened 2025-04-29
20 channel storage backends

gtjoseph:
- **build_system: Process MAKEOPTS blocks in main source files.**
- **Alternate Channel Storage Backends**


gtjoseph:
cherry-pick-to: none
multiple-commits: standalone

include/asterisk/channel.h
```diff
@@ -3119,48 +3121,65 @@ struct ast_channel *ast_channel_iterator_next(struct ast_channel_iterator *i);
 
 /*! @} End channel iterator definitions. */
 
+/*! @{ Channel search functions */
+
+/*!
+* \warning If the legacy ao2_container channel storage backend is in use,
```
jcolp, line 3127: This should just state that no channel locks be held. Since this is a developer comment, and the selection of the backend is a user runtime selection you can't really reconcile the two and take it into account.

include/asterisk/channel.h
```diff
@@ -5084,5 +5113,20 @@ void *ast_channel_get_stream_topology_change_source(struct ast_channel *chan);
...
 
+/*!
+ * \brief Get the name of the current channel storage driver
+ *
+ * \return The name of the current channel storage driver
+ */
+const char *ast_channel_get_current_storage_driver_name(void);
+
+/*!
+ * \brief Set the current channel storage driver
```
jcolp, line 5124: Document what happens if this is called multiple times

jcolp (changes requested):

include/asterisk/channel.h
gtjoseph: Done

include/asterisk/channel.h
gtjoseph: Now marked as internal and will return an error if called after ast_fdully_booted.  Documented appropriately.

jcolp (approved):

mbradeen (approved):

PR #1231, opened 2025-05-02
lock.h: Add include for string.h when DEBUG_THREADS is defined.

gtjoseph:
When DEBUG_THREADS is defined, lock.h uses strerror(), which is defined
in the libc string.h file, to print warning messages. If the including
source file doesn't include string.h then strerror() won't be found and
and compile errors will be thrown. Since lock.h depends on this, string.h
is now included from there if DEBUG_THREADS is defined.  This way, including
source files don't have to worry about it.

### DIFF
--- a/include/asterisk/lock.h
+++ b/include/asterisk/lock.h
@@ -54,6 +54,9 @@
 #ifdef HAVE_BKTR
 #include <execinfo.h>
 #endif
+#ifdef DEBUG_THREADS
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1234, opened 2025-05-05
[bug]: Set CalllerID lost on DTMF attended transfer


mbradeen:
### Severity

Minor

### Versions

All

### Components/Modules

chan_pjsip

### Operating Environment

All 18+

### Frequency of Occurrence

Frequent

### Issue Description

When performing an attended transfer, a race condition exists on the outbound Dial that may over-write the dialplan set Caller ID with the default information from the Transferer.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

psidlo:
https://github.com/asterisk/asterisk/issues/519 - is this the same issue?

jcolp:
It seems like it, yes.

igorolhovskiy:
Issue reproduces on Asterisk 20.18.2 with a 50/50 probability. Means sometimes CallerID is set from original caller, sometimes - according to script.

jcolp:
This is a closed issue. If you believe there is still a scenario that exhibits this, then you'll need to file a new issue with complete scenario and reproduction step by step instructions as well as an Asterisk debug log.

PR #1235, opened 2025-05-05
chan_pjsip: Serialize INVITE creation on DTMF attended transfer

mbradeen:
When a call is transfered via DTMF feature code, the Transfer Target and
Transferer are bridged immediately.  This opens the possibilty of a race
condition between the creation of an INVITE and the bridge induced colp
update that can result in the set caller ID being over-written with the
transferer's default info.

Fixes: #1234

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -2394,13 +2394,12 @@ static void update_initial_connected_line(struct ast_sip_session *session)
 
 static int call(void *data)
 {
-	struct ast_sip_channel_pvt *channel = data;
-	struct ast_sip_session *session = channel->session;
+	struct ast_sip_session *session = data;
 	pjsip_tx_data *tdata;
 	int res = 0;
 	SCOPE_ENTER(1, "%s Topology: %s\n",
 		ast_sip_session_get_name(session),
-		ast_str_tmp(256, ast_stream_topology_to_str(channel->session->pending_media_state->topology, &STR_TMP))
+		ast_str_tmp(256, ast_stream_topology_to_str(session->pending_media_state->topology, &STR_TMP))
 		);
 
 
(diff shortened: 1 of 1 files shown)


channels/chan_pjsip.c
```diff
@@ -2422,16 +2422,28 @@ static int call(void *data)
...
 
 	ao2_ref(channel, +1);
-	if (ast_sip_push_task(channel->session->serializer, call, channel)) {
+
+	session = ao2_bump(channel->session);
+	ast_channel_unlock(ast);
+
+	/* The creation of the INVITE needs to be pushed synchronously to prevent a race condition
+	   with bridging on attended transfers that can result in a loss of set Caller ID. */
+	if (ast_sip_push_task_wait_serializer(session->serializer, call, channel)) {
```
jcolp, line 2437: This assumes channel will remain valid for the lifetime of this, instead "session" should be passed and the "call" function be updated to use session .

jcolp (changes requested):
You misspelt update in the commit message.

mbradeen:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

channels/chan_pjsip.c
```diff
@@ -2414,24 +2413,36 @@ static int call(void *data)
...
 /*! \brief Function called by core to actually start calling a remote party */
 static int chan_pjsip_call(struct ast_channel *ast, const char *dest, int timeout)
 {
 	struct ast_sip_channel_pvt *channel = ast_channel_tech_pvt(ast);
+	struct ast_sip_session *session;
+
 	SCOPE_ENTER(1, "%s Topology: %s\n", ast_sip_session_get_name(channel->session),
 		ast_str_tmp(256, ast_stream_topology_to_str(channel->session->pending_media_state->topology, &STR_TMP)));
 
 	ao2_ref(channel, +1);
```
jcolp, line 2428: Manipulation of channel refcount can be removed

jcolp (changes requested):

jcolp (approved):

Issue #1141, opened 2025-03-02
[bug]: res_pjsip: Contact header set incorrectly for call redirect (302 Moved temp.) when external_* set


tw-mloschwitz:
### Severity

Minor

### Versions

22.1.0

### Components/Modules

chan_pjsip

### Operating Environment

Debian Bookworm, FreePBX 17

### Frequency of Occurrence

None

### Issue Description

https://issues-archive.asterisk.org/ASTERISK-30451 remains unfixed to this day. There is even a patch attached to it, please apply it to the Asterisk sources to allow for proper SIP 302 functionality.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

tw-mloschwitz:
I just rebuilt a version of res_pjsip_nat with the above patch applied and I can confirm that SIP 302 works as expected afterwards.

PR #1237, opened 2025-05-08
res_pjsip_nat.c: Do not overwrite transfer host

mbradeen:
When a call is transfered via dialplan behind a NAT, the
host portion of the Contact header in the 302 will no longer
be over-written with the external NAT IP and will retain the
hostname.

Fixes: #1141


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

PR #1241, opened 2025-05-12
frame.c: validate frame data length.

mkmer:
frame.c: validate frame data length is less than samples when adjusting volume

Resolves: asterisk#1230


mkmer:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

PR #1247, opened 2025-05-13
res_audiosocket.c: Add retry mechanism for reading data from AudioSocket

SvenKube:
The added retry mechanism addresses an issue that arises when fragmented TCP packets are received, each containing only a portion of an AudioSocket packet. This situation can occur if the external service sending the AudioSocket data has Nagle's algorithm enabled.

### DIFF
--- a/res/res_audiosocket.c
+++ b/res/res_audiosocket.c
@@ -307,6 +307,18 @@ struct ast_frame *ast_audiosocket_receive_frame_with_hangup(const int svc,
 	while (i < *length) {
 		n = read(svc, data + i, *length - i);
 		if (n == -1) {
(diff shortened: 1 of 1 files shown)


res/res_audiosocket.c
```diff
@@ -307,6 +307,18 @@ struct ast_frame *ast_audiosocket_receive_frame_with_hangup(const int svc,
 	while (i < *length) {
 		n = read(svc, data + i, *length - i);
 		if (n == -1) {
+			if (errno == EAGAIN || errno == EWOULDBLOCK) {
+				int poll_result = ast_wait_for_input(svc, 5);
+
+				if(poll_result == 1) {
```
silentindark, line 313: space

jcolp (approved):

PR #1248, opened 2025-05-14
res_websocket_client: Create common utilities for websocket clients.

gtjoseph:
Since multiple Asterisk capabilities now need to create websocket clients
it makes sense to create a common set of utilities rather than making
each of those capabilities implement their own.

* A new configuration file "websocket_client.conf" is used to store common
client parameters in named configuration sections.
* APIs are provided to list and retrieve ast_websocket_client objects created
from the named configurations.
* An API is provided that accepts an ast_websocket_client object, connects
to the remote server with retries and returns an ast_websocket object. TLS is
supported as is basic authentication.
* An observer can be registered to receive notification of loaded or reloaded
client objects.
* An API is provided to compare an existing client object to one just
reloaded and return the fields that were changed. The caller can then decide
what action to take based on which fields changed.

Also as part of thie commit, several sorcery convenience macros were created
to make registering common object fields easier.


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

res/res_websocket_client.c
```diff
@@ -0,0 +1,574 @@
...
+	if (ast_string_field_init(wc, 1024) != 0) {
+		ao2_cleanup(wc);
+		return NULL;
+	}
+
+	ast_debug(2, "%s: Allocated websocket client config\n", id);
+	return wc;
+}
+
+static int websocket_client_connection_type_from_str(const struct aco_option *opt,
```
mbradeen, line 317: what is the opt for?

res/res_websocket_client.c
```diff
@@ -0,0 +1,574 @@
...
+	} else if (strcasecmp(var->value, "per_call_config") == 0) {
+		ws->connection_type = AST_WS_TYPE_CLIENT_PER_CALL_CONFIG;
+	} else {
+		return -1;
+	}
+
+	return 0;
+}
+
+static int websocket_client_connection_type_to_str(const void *obj, const intptr_t *args, char **buf)
```
mbradeen, line 333: same as above, is args for standardization or future-proofing?

res/res_websocket_client.c
```diff
@@ -0,0 +1,574 @@
...
+		if (wc->reconnect_attempts == DEFAULT_RECONNECT_ATTEMPTS) {
+			if (wc->connection_type == AST_WS_TYPE_CLIENT_PERSISTENT) {
+				wc->reconnect_attempts = INT_MAX;
+			} else {
+				wc->reconnect_attempts = 4;
+			}
+		}
+	}
+
+	if (res != 0) {
```
mbradeen, line 382: Is this needed?

res/res_websocket_client.c
```diff
@@ -0,0 +1,574 @@
...
+	return res;
+}
+
+struct ao2_container *ast_websocket_client_retrieve_all(void)
+{
+	if (!sorcery) {
+		return NULL;
+	}
+
+	return ast_sorcery_retrieve_by_fields(sorcery, "websocket_client",
```
mbradeen, line 395: Should "websocket_client" be a #define or static string?

mbradeen (changes requested):

res/res_websocket_client.c
gtjoseph: That's the standard signature for sorcery/aco functions that convert a string into the internal value.  We rarely use "opt" but it has to be there.

res/res_websocket_client.c
gtjoseph: That's the standard signature for sorcery/aco functions that convert a the internal value to a string.  We rarely use "args" but it has to be there.

res/res_websocket_client.c
gtjoseph: It's only used in a few places so it's not really necessary.

mbradeen (commented):
Just the double return res on wc_apply

gtjoseph:
> Just the double return res on wc_apply

Oh, *that's* what you were talking about in the review. The review snippet didn't show the extra return so I didn't know what you were talking about. :)

res/res_websocket_client.c
gtjoseph: Now I get it.  Fixed.

mbradeen (approved):

jcolp (approved):

Issue #1240, opened 2025-05-09
[bug]: WebRTC invites failing on Chrome 136


jjsearle:
### Severity

Major

### Versions

18.26.1

### Components/Modules

res_pjsip

### Operating Environment

Problem happening on Asterisk 18.26.1 not tested on later versions yet

### Frequency of Occurrence

Constant

### Issue Description

After updating Chrome to version 136, invites over webrtc to Asterisk are failing (see log output and attached invite)

Working fine with Chrome 135.

I looked at the Chrome 136 release notes and saw they added support for the HVEC codec. This seems to have increased the content-length from `7823` to `8189`.

Some discussion on IRC suggested that I may be hitting the max number of attributes of 196

[invite.log](https://github.com/user-attachments/files/20122614/invite.log)

### Relevant log output

```shell
[May  9 11:41:50] WARNING[15685]: pjproject: <?>:                        sdp.c .Error adding media attribute, attribute is ignored: Too many objects of the specified type (PJ_ETOOMANY)
[May  9 11:41:50] WARNING[15685]: pjproject: <?>:                        sdp.c .Error adding media attribute, attribute is ignored: Too many objects of the specified type (PJ_ETOOMANY)
[May  9 11:41:50] WARNING[15685]: pjproject: <?>:                        sdp.c .Error adding media attribute, attribute is ignored: Too many objects of the specified type (PJ_ETOOMANY)
[May  9 11:41:50] WARNING[15685]: pjproject: <?>:                        sdp.c .Error adding media attribute, attribute is ignored: Too many objects of the specified type (PJ_ETOOMANY)
[May  9 11:41:50] WARNING[15685]: pjproject: <?>:                        sdp.c .Error adding media attribute, attribute is ignored: Too many objects of the specified type (PJ_ETOOMANY)
[May  9 11:41:50] WARNING[15685]: pjproject: <?>:                        sdp.c .Error adding media attribute, attribute is ignored: Too many objects of the specified type (PJ_ETOOMANY)
[May  9 11:41:50] WARNING[15685]: pjproject: <?>:                        sdp.c .Error adding media attribute, attribute is ignored: Too many objects of the specified type (PJ_ETOOMANY)
[May  9 11:41:50] ERROR[15685]: pjproject: <?>:                      sip_inv.c .Error parsing/validating SDP body: Missing SDP rtpmap for dynamic payload type (PJMEDIA_SDP_EMISSINGRTPMAP)
    -- Remote UNIX connection
```

[...]

PR #1249, opened 2025-05-15
pjproject: Increase maximum SDP attribute limit

jjsearle:
Fixes: https://github.com/asterisk/asterisk/issues/1240

Since Chrome 136, using Windows, when initiating a video call the INVITE SDP exceeds the maximum number of allowed attributes, resulting in the INVITE being rejected. This PR increases the limit when using bundled pjproject.


jjsearle:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

third-party/pjproject/patches/config_site.h
```diff
@@ -86,7 +86,7 @@
 /* Increase limits to allow more formats */
 #define	PJMEDIA_MAX_SDP_FMT   64
```
gtjoseph, line 87: The SDP in your log has 50 formats which is getting close to t4he max of 64.  How about increasing PJMEDIA_MAX_SDP_FMT from 64 to 72 while you're at it.  This way we'll be safe for a while.

gtjoseph (changes requested):

third-party/pjproject/patches/config_site.h
jjsearle: I have increased the number of allowed formats to 72 and updated the commit / description.

gtjoseph (approved):

PR #1257, opened 2025-05-22
asterisk.c: Add option to restrict shell access from remote consoles.

gtjoseph:
UserNote: A new asterisk.conf option 'disable_remote_console_shell' has
been added that, when set, will prevent remote consoles from executing
shell commands using the '!' prefix.

Resolves: #GHSA-c7p6-7mvq-8jq2


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

PR #1261, opened 2025-05-26
app_queue: queue rules – Add support for QUEUE_RAISE_PENALTY=rN to raise penalties only for members within min/max range

phoneben:
This update adds support for a new QUEUE_RAISE_PENALTY format: rN

When QUEUE_RAISE_PENALTY is set to rN (e.g., r4), only members whose current penalty
is greater than or equal to the defined min_penalty and less than or equal to max_penalty
will have their penalty raised to N.

Members with penalties outside the min/max range remain unchanged.

Example behaviors:

QUEUE_RAISE_PENALTY=4     → Raise all members with penalty < 4 (existing behavior)
QUEUE_RAISE_PENALTY=r4    → Raise only members with penalty in [min_penalty, max_penalty] to 4

Implementation details:

Adds parsing logic to detect the r prefix and sets the raise_respect_min flag

Modifies the raise logic to skip members outside the defined penalty range when the flag is active

UserNote: This change introduces QUEUE_RAISE_PENALTY=rN, allowing selective penalty raises
only for members whose current penalty is within the [min_penalty, max_penalty] range.
Members with lower or higher penalties are unaffected.
This behavior is backward-compatible with existing queue rule configurations.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -1880,6 +1880,7 @@ struct queue_ent {
 	int max_penalty;                       /*!< Limit the members that can take this call to this penalty or lower */
 	int min_penalty;                       /*!< Limit the members that can take this call to this penalty or higher */
 	int raise_penalty;                     /*!< Float lower penalty members to a minimum penalty */
+	int raise_respect_min;                 /*!< A switch raise_penalty should respect min_penalty not just max_penalty */
 	int linpos;                            /*!< If using linear strategy, what position are we at? */
 	int linwrapped;                        /*!< Is the linpos wrapped? */
 	time_t start;                          /*!< When we started holding */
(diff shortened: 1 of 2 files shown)


alex2grad:
You should also reflect this change in the comments of the queuerules.conf file.

apps/app_queue.c
```diff
@@ -3306,6 +3313,11 @@ static int insert_penaltychange(const char *list_name, const char *content, cons
 	}
 
 	if (!ast_strlen_zero(raisestr)) {
+		rule->raise_respect_min = 0;  // Initialize to 0
```
seanbright, line 3316: Don't use `//` style comments.

apps/app_queue.c
```diff
@@ -3306,6 +3313,11 @@ static int insert_penaltychange(const char *list_name, const char *content, cons
...
+		if (*raisestr == 'r') {
+			rule->raise_respect_min = 1;               // Set the flag
```
seanbright, line 3318: Don't use `//` style comments.

seanbright (dismissed):

phoneben:
missed a place 
```
if ((raise_penalty_str = pbx_builtin_getvar_helper(chan, "QUEUE_RAISE_PENALTY"))) {
		if (sscanf(raise_penalty_str, "%30d", &raise_penalty) == 1) {
			ast_debug(1, "%s: Got raise penalty %d from ${QUEUE_RAISE_PENALTY}.\n", ast_channel_name(chan), raise_penalty);
		} else {
			ast_log(LOG_WARNING, "${QUEUE_RAISE_PENALTY}: Invalid value (%s), channel %s.\n",
				raise_penalty_str, ast_channel_name(chan));
			raise_penalty = INT_MAX;
		}
	}
```

jcolp (approved):

gtjoseph:
@seanbright Can you confirm your review issues are resolved?

seanbright:
> @seanbright Can you confirm your review issues are resolved?

They are

Issue #1254, opened 2025-05-21
[bug]: ActiveChannels not reported when using AMI command PJSIPShowEndpoint


massimocode:
### Severity

Minor

### Versions

22

### Components/Modules

AMI, PJSIPShowEndpoint

### Operating Environment

Debian 12

### Frequency of Occurrence

Constant

### Issue Description

As per docs here there's supposed to be a property ActiveChannels https://docs.asterisk.org/Latest_API/API_Documentation/AMI_Events/EndpointDetail/

However, the reported value is always blank

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

massimocode:
It's not high priority. I'm using CoreShowChannels instead which works great for my use case.

jcolp:
Please provide output of your request and the corresponding AMI response.

massimocode:
Request
```
PJSIPShowEndpoint
ActionID: QuerySIPEndpoint
Endpoint: test_endpoint
```

Response
```
Response: Success
ActionID: QuerySIPEndpoint
EventList: start
Message: Following are Events for each object associated with the Endpoint

Event: EndpointDetail
ActionID: QuerySIPEndpoint
ObjectType: endpoint
ObjectName: test_endpoint
RpidImmediate: false
Webrtc: no
Ignore183WithoutSdp: false
SendAoc: false
DeviceStateBusyAt: 0
T38UdptlMaxdatagram: 0
CodecPrefsIncomingAnswer: prefer:pending, operation:intersect, keep:all, transcode:allow
SecurityNegotiation: no
DtlsRekey: 0
NamedPickupGroup:
DirectMediaMethod: invite
SendRpid: false
PickupGroup:
SdpSession: Asterisk
DtlsVerify: No
MessageContext:
Tenantid:
Mailboxes:
RecordOnFeature: automixmon
DtlsPrivateKey:
DtlsFingerprint: SHA-256
OverlapContext:
OutgoingCallOfferPref: remote_merge
FromDomain:
TimersSessExpires: 1800
NamedCallGroup:
DtlsCipher:
MediaEncryptionOptimistic: false
SuppressQ850ReasonHeaders: false
Aors: test_aor
IdentifyBy: username,ip
CalleridPrivacy: allowed_not_screened
MwiSubscribeReplacesUnsolicited: no
CosAudio: 0
FollowEarlyMediaFork: true
Context: my-context
[...]

PR #1262, opened 2025-05-27
res_pjsip: Fix empty `ActiveChannels` property in AMI responses.

seanbright:
The logic appears to have been reversed since it was introduced in 05cbf8df.

Resolves: #1254


seanbright:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

jcolp:
I think it would indeed get unwieldy. This also fixes it for EndpointList, which would be an even worse result since it would be all the endpoints. Depending on usage it would get big... fast.

I think since the documentation states it is just the number of active channels we stick with that for a fix. If there comes a point where channels are needed an AMI action can instead be added to explicitly do that.

jcolp (approved):

Issue #1251, opened 2025-05-18
[improvement]: PJSIP shouldn't require SIP Date header to process full shaken passport which includes iat


shaneyoung:
### Improvement Description

According to https://docs.asterisk.org/Deployment/STIR-SHAKEN/#verification
  * The "iat" date parameter in the header must be within a few seconds of "now".
  * The SIP Date header must be present and also within a few seconds of "now". 
However, when the passport contains the iat, there should be no reason to use the SIP date header.

https://access.atis.org/higherlogic/ws/public/download/67436/ATIS-1000074.v003.pdf states: 

5.3.3 Use of the Full Form of PASSporT
IETF RFC 8224 [Ref 13] supports the use of both full and compact forms of the PASSporT in the Identity header.
The full form of the PASSporT shall be used to avoid any potential SIP network element interaction with headers,
in particular the Date header field, which could lead to large numbers of errors being generated (the “iat” value in
the payload that is protected by the signatures should be considered a more reliable indicator of PASSporT
freshness than any time value in the SIP Date header).

The improvement could be to eliminate the use of the SIP Date header and rely only on the IAT, or an option could be added to the verification profile to use only the IAT.

rberlin01:
Bump

There should be a configuration option to ignore Date header verification. IAT should be the sole point of truth for call time. This seems to be breaking verification if no Date header is present in the INVITE.

gtjoseph:
I'll be working this on Monday.

seanbright:
bump thanks bump

gtjoseph:
PR #1270 has been pushed that adds a new "ignore_sip_date_header" option.  PLEASE TEST!!

rberlin01:
> PR [#1270](https://github.com/asterisk/asterisk/pull/1270) has been pushed that adds a new "ignore_sip_date_header" option. PLEASE TEST!!

This setting does appear to be working. Set to no, I get errors:

```
[Jun 17 10:12:56] ERROR[117751]: res_stir_shaken/verification.c:748 check_date_header:  PJSIP/xxx-00000001: No date header provided
[Jun 17 10:12:56] ERROR[117751]: res_stir_shaken/verification.c:948 ast_stir_shaken_vs_verify:  PJSIP/xxx-00000001: Date header verification failed
```

But when set to yes, these errors do not appear. I can't confirm beyond that if it's working, as I'm hitting what appears to be a different bug in the verification process. But the date field error does not appear when ignore_sip_date_header is set to yes.

[...]

PR #1270, opened 2025-06-15
res_stir_shaken: Add "ignore_sip_date_header" config option.

gtjoseph:
UserNote: A new STIR/SHAKEN verification option "ignore_sip_date_header" has
been added that when set to true, will cause the verification process to
not consider a missing or invalid SIP "Date" header to be a failure.  This
will make the IAT the sole "truth" for Date in the verification process.
The option can be set in the "verification" and "profile" sections of
stir_shaken.conf.

Resolves: #1251


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

res/res_stir_shaken/verification.c
```diff
@@ -630,6 +630,12 @@ int ast_stir_shaken_vs_get_use_rfc9410_responses(
 	return ctx->eprofile->vcfg_common.use_rfc9410_responses;
 }
 
+int ast_stir_shaken_vs_get_ignore_sip_date_header(
```
mbradeen: Does this function get used?

gtjoseph: Oops.  Left in by mistake.

Issue #1272, opened 2025-06-17
[improvement]: STIR/SHAKEN handle X5U downloaded certificate chains


rberlin01:
### Improvement Description

Calls can arrive signed with a chain of certificates. Some intermediate certificates are not provided by STI-PA and have to be downloaded and verified before the call can pass verification. Adding in ability for this to happen in Asterisk's verification workflow would solve this.

gtjoseph:
@rberlin01 I've created a PR (#1274) that enables loading certificate chains from the X5U URL.  I tested it with the att cert you referenced in the other issue.

Please test it yourself.

gtjoseph:
@rberlin01 Are you able to test?

rberlin01:
> [@rberlin01](https://github.com/rberlin01) Are you able to test?

Sorry for the late reply. I am just returning from vacation. I can test this on Friday when I return to office.

gtjoseph:
@rberlin01 No worries.  This PR has been merged and will be included in release candidates to be published today.  The GA releases are scheduled for next Thursday so if you can test before Tuesday that'd be great because we'll still have time to address any issues before the releases.

PR #1274, opened 2025-06-18
res_stir_shaken.so: Handle X5U certificate chains.

gtjoseph:
The verification process will now load a full certificate chain retrieved
via the X5U URL instead of loading only the end user cert.

* Renamed crypto_load_cert_from_file() and crypto_load_cert_from_memory()
to crypto_load_cert_chain_from_file() and crypto_load_cert_chain_from_memory()
respectively.

* The two load functions now continue to load certs from the file or memory
PEMs and store them in a separate stack of untrusted certs specific to the
current verification context.

* crypto_is_cert_trusted() now uses the stack of untrusted certs that were
extracted from the PEM in addition to any untrusted certs that were passed
in from the configuration (and any CA certs passed in from the config of
course).

Resolves: #1272

UserNote: The STIR/SHAKEN verification process will now load a full
certificate chain retrieved via the X5U URL instead of loading only
the end user cert.

### DIFF
--- a/res/res_stir_shaken/attestation_config.c
+++ b/res/res_stir_shaken/attestation_config.c
@@ -164,8 +164,8 @@ int as_check_common_config(const char *id, struct attestation_cfg_common *acfg_c
 				acfg_common->public_cert_url);
 		}
 
-		public_cert = crypto_load_cert_from_memory(public_cert_data,
-			public_cert_len);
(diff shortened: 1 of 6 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

Issue #1273, opened 2025-06-18
[bug]: When executed with GotoIf, the action Redirect does not take effect and causes confusion in dialplan execution.


Tinet-mucw:
### Severity

Trivial

### Versions

18.10

### Components/Modules

main

### Operating Environment

centos8

### Frequency of Occurrence

One Time

### Issue Description

Log of normal execution

```c
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx.c: Executing [s@ivr_websocket_robot:48] UserEvent("PJSIP/sipforward01-00004a01", "audioSocketConnectStart,CallType: 1,Params: ,CustomerNumber: 1773464,IvrId: 44907,IvrNode: g1OIc4pS,Prologue: ,PrologueSsml: ,SpyFlag: ,MonitorFile: 8010321-20250617103141-1773464-01012345678--record-medias_6-1750127501.22969") in new stack
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx.c: Executing [s@ivr_websocket_robot:49] GotoIf("PJSIP/sipforward01-00004a01", "0?50:82") in new stack
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx_builtins.c: Goto (ivr_websocket_robot,s,82)
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx.c: Executing [s@ivr_websocket_robot:82] GotoIf("PJSIP/sipforward01-00004a01", "1?83:85") in new stack
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx_builtins.c: Goto (ivr_websocket_robot,s,83)
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx.c: Executing [s@ivr_websocket_robot:83] Set("PJSIP/sipforward01-00004a01", "cur_node_wait_time=10") in new stack
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx.c: Executing [s@ivr_websocket_robot:84] Goto("PJSIP/sipforward01-00004a01", "86") in new stack
[Jun 17 10:32:04] VERBOSE[7875][C-00003742] pbx_builtins.c: Goto (ivr_websocket_robot,s,86)
[...]

PR #1275, opened 2025-06-19
pbx.c: when set flag AST_SOFTHANGUP_ASYNCGOTO, ast_explicit_goto should return -1.

Tinet-mucw:
Under certain circumstances the context/extens/prio are set in the ast_async_goto, for example action Redirect.
In the situation that action Redirect is broken by GotoIf this info is changed.
that will causes confusion in dialplan execution.
    
Resolves: #1273


Tinet-mucw:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

Issue #927, opened 2024-10-01
[bug]: no audio when media source changed during the call


n2p-litvinov:
### Severity

Major

### Versions

18, 20

### Components/Modules

res_rtp_asterisk

### Operating Environment

Debian 11 “bullseye” x64

### Frequency of Occurrence

Frequent

### Issue Description

There is no audio issue with some of the carriers in case when media source changed during the call (e.g. call transfer, audio prompt and then some sip endpoint, etc.). Usually audio appears after some time.
Steps to reproduce: 
1. Make call from Alice to Carrier
2. Answer the call on Carrier side
3. Make transfer (any type of transfer) from Alice to Bob
4. Bob answers the call and start speaking
5. On the carrier side there will be silence instead of Bob's speech

Issue related to big timestamp skew. Solution is to set Mark when detect such case.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

gaaf:
> Issue related to big timestamp skew.

Have you checked the SSRC is the same? The timestamps for different SSRC's are not related in any way. The receiver must separately track the timestamps for every unique SSRC it receives.

tomec-martin:
In the patch above, we implemented an alternative method for SSRC change - similar to the approach previously used in chan_sip. This resolved the issue in our scenario, as described in this post:
https://community.asterisk.org/t/rtp-stream-timestamp-ssrc-issue-after-blind-transfer-on-chan-pjsip-asterisk-22-4-1/108771

We were not able to test video, as mentioned in issue #1135, but from my perspective, patch appears to be unrelated to the issue addressed here.

meengu:
Fix for this broke webrtc audio stream *reception* in Firefox and Safari (not Chrome, though), so there's something subtly wrong about it. Sorry that I don't have details, yet. Reverting it makes webrtc calls work in all browsers again.

gtjoseph:
> Fix for this broke webrtc audio stream _reception_ in Firefox and Safari (not Chrome, though), so there's something subtly wrong about it. Sorry that I don't have details, yet. Reverting it makes webrtc calls work in all browsers again.

@meengu When you have the details and can reproduce using a current version of Asterisk, please create a new issue.

meengu:
The cause is that this patch causes SSRC to not match what was negotiated via SDP offer. Firefox seems to care for SSRC mismatch. Chrome doesn't.

PR #1277, opened 2025-06-26
chan_pjsip.c: Change SSRC after media source change

tomec-martin:
When the RTP media source changes, such as after a blind transfer, the new source introduces a discontinuous timestamp. According to RFC 3550, Section 5.1, an RTP stream's timestamp for a given SSRC must increment monotonically and linearly.
To comply with the standard and avoid a large timestamp jump on the existing SSRC, a new SSRC is generated for the new media stream.
This change resolves known interoperability issues with certain SBCs (like Sonus/Ribbon) that stop forwarding media when they detect such a timestamp violation. This code uses the existing implementation from chan_sip.

Resolves: #927

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -1820,6 +1820,14 @@ static int chan_pjsip_indicate(struct ast_channel *ast, int condition, const voi
 	case AST_CONTROL_SRCUPDATE:
 		break;
 	case AST_CONTROL_SRCCHANGE:
+		/* Generate a new SSRC due to media source change and RTP timestamp reset.
+		   Ensures RFC 3550 compliance and avoids SBC interoperability issues (Sonus/Ribbon)*/
+		for (i = 0; i < AST_VECTOR_SIZE(&channel->session->active_media_state->sessions); ++i) {
+			media = AST_VECTOR_GET(&channel->session->active_media_state->sessions, i);
+			if (media && media->rtp) {
+				ast_rtp_instance_change_source(media->rtp);
+			}
+		}
 		break;
 	case AST_CONTROL_REDIRECTING:
 		if (ast_channel_state(ast) != AST_STATE_UP) {


tomec-martin:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

seanbright:
> The main reason for this change is incompatibility with some SBCs

Please provide the names/versions of those SBCs if possible. Is the incompatibility due to something out-of-spec that Asterisk is doing or the SBCs?

channels/chan_pjsip.c
```diff
@@ -1818,8 +1818,20 @@ static int chan_pjsip_indicate(struct ast_channel *ast, int condition, const voi
 		}
 		break;
 	case AST_CONTROL_SRCUPDATE:
+		for (i = 0; i < AST_VECTOR_SIZE(&channel->session->active_media_state->sessions); ++i) {
+				media = AST_VECTOR_GET(&channel->session->active_media_state->sessions, i);
+				if (media && media->rtp) {
+					ast_rtp_instance_update_source(media->rtp);
+				}
```
seanbright: One too many tabs at the beginning of these lines.

channels/chan_pjsip.c
```diff
@@ -1818,8 +1818,20 @@ static int chan_pjsip_indicate(struct ast_channel *ast, int condition, const voi
...
+		}
 		break;
 	case AST_CONTROL_SRCCHANGE:
+		for (i = 0; i < AST_VECTOR_SIZE(&channel->session->active_media_state->sessions); ++i) {
+				media = AST_VECTOR_GET(&channel->session->active_media_state->sessions, i);
+				if (media && media->rtp) {
+					ast_rtp_instance_change_source(media->rtp);
+				}
```
seanbright: One too many tabs at the beginning of these lines.

seanbright (changes requested):

channels/chan_pjsip.c
tomec-martin: Fixed, thank you.

tomec-martin:
> > The main reason for this change is incompatibility with some SBCs
> 
> Please provide the names/versions of those SBCs if possible. Is the incompatibility due to something out-of-spec that Asterisk is doing or the SBCs?

This is because Asterisk violates RFC 3550. I have amended the commit message to make it clearer. I don't have any information about the SBCs. Thanks

seanbright:
> I don't have any information about the SBCs.

You’re asserting there is an incompatibility with some SBCs but you can’t name an SBC that demonstrates this incompatibility?

janecek:
Unfortunately, I do not have the exact make/model/version of the SBC, as the provider did not share this information (it is one of the larger European providers).

However, the main point is that Asterisk currently violates RFC 3550 by not changing the SSRC after an RTP source change (such as after a transfer or change in transport address). This out-of-spec behavior is the root cause of the incompatibility; the SBC simply stops sending RTP when it detects a large timestamp jump without an SSRC change, as allowed by the RFC. The patch brings Asterisk chan\_pjsip in line with both the RFC and with the correct behavior already implemented in chan\_sip.

For more background, this issue was also discussed with the community here:
[https://community.asterisk.org/t/rtp-stream-timestamp-ssrc-issue-after-blind-transfer-on-chan-pjsip-asterisk-22-4-1/108771](https://community.asterisk.org/t/rtp-stream-timestamp-ssrc-issue-after-blind-transfer-on-chan-pjsip-asterisk-22-4-1/108771)

jcolp:
@seanbright It's probably Sonus. It's almost always them with things like this.

seanbright:
I don’t actually care which SBC it is I just want it documented so when the next person comes along and wants to remove this code because it’s not compatible with their particular equipment we can back it up with something other than because-this-one-user-said-so. 

I also find it odd that this is only just now a problem.

jcolp:
I also disagree with this:

"If a source changes its source transport address, it must also choose a new SSRC identifier to avoid being interpreted as a looped source"

Source transport address has meaning within RFC3550, and from the perspective of Asterisk to upstream it is the same source transport address. That's not a reason within this to change the SSRC.

seanbright:
Section 8.2 goes on to say (contradictingly):

> Therefore, if a source changes its source transport address, it MAY
   also choose a new SSRC identifier to avoid being interpreted as a
   looped source.  (This is not MUST because in some applications of RTP
   sources may be expected to change addresses during a session.)

janecek:
As defined in RFC 3550, Section 5.1 (RTP Fixed Header Fields), the timestamp “MUST be derived from a clock that increments monotonically and linearly in time.” Forwarding a new media stream, which has its own independent clock, under the old SSRC directly violates this principle.

This leads me to believe that changing the SSRC is the correct approach here. To comply with the standard, we must either change the SSRC or prevent the timestamp from changing.

The practical consequence of this violation is illustrated in the attached graph. The massive spike, showing Jitter (ms): 4292573 and Skew: 68681097, is the direct, measurable result of the receiving end trying to process a stream with a discontinuous timestamp under a single SSRC.

rtp_analyze

jcolp:
The commit message needs to be updated with that, taking away the reference to source address as that is not relevant. A comment should also be added around the code so if someone is looking they will see it, and it can be mentioned that Sonus is likely the cause as I remember them from when this was done in chan_sip.

luislza:
I believe we've also encountered this issue under the following additional scenarios (other than SBC compatibility):

1. VOIP monitoring systems incorrectly report high max jitter because of the significant timestamp jump without SSRC change - the specific example in our case is www.voipmonitor.org
2. Packet sniffing call recorders don't record streams accurately because they regard the significant timestamp jump without SSRC change as non-compliant with RFC3550

seanbright:
Admittedly I have not looked too deeply into this but I assume fixing the timestamp skew is not feasible?

luislza:
I'm definitely not qualified to comment on fixing the timestamp skew so I'll leave that to the experts. :-)

This scenario seems to occur on call transfer to a new endpoint - the timestamp jumps after the transfer when the audio stream from the new endpoint is bridged to the SBC. 

I'm happy to share PCAPs privately if that helps anybody.

janecek:
@seanbright  That's a valid question. My feeling is that actively rewriting the timestamps would be a more complex and potentially fragile solution.

This pull request aligns `chan_pjsip`'s behavior with the established and proven logic from `chan_sip`. Introducing a new, stateful mechanism for timestamp manipulation seems riskier than restoring a well-understood behavior that is known to work.

However, I also admit that I do not feel qualified enough to definitively assess all the technical implications of such a change.

luislza:
@tomec-martin, I notice that chan_sip also applies the same logic for AST_CONTROL_SRCUPDATE. 

I wonder if the same shouldn't apply here?

Actually I see in chan_sip ast_rtp_instance_update_source is also called in sip_write as well as for AST_CONTROL_HOLD, AST_CONTROL_UNHOLD.

I don't know enough to know if that would apply for chan_pjsip as well.

tomec-martin:
> I notice that chan_sip also applies the same logic for AST_CONTROL_SRCUPDATE.

The logic is different (_ast_rtp_instance_update_source_ vs. _ast_rtp_instance_change_source_).
Code for ast_rtp_instance_update_source does not change SSRC because "... SRCUPDATE control frames are sent many times where we don't want a new ssrc ..." Described more in commit 68d1ded8dd40f646017da31c1fd3b4a0cc23c375.

tomec-martin:
> Admittedly I have not looked too deeply into this but I assume fixing the timestamp skew is not feasible?

tomec-martin:
@seanbright The change of SSRC seems legitimate to me, as the comment for AST_CONTROL_SRCCHANGE is:
https://github.com/asterisk/asterisk/blob/b158011c4e1032e77a40fdefac67a79617ed5e8d/include/asterisk/frame.h#L303
However, since I'm not experienced in Asterisk development, I may have missed something...

luislza:
I see that ast_rtp_instance_change_source will also set the marker bit on source change. I wonder if that in turn will affect video again.... 

ast_rtp_instance_update_source seems to only set the marker bit as its only function.

Pure speculation of course since I don't know enough about the ins and outs of asterisk code.

seanbright:
@gtjoseph I have no comments on the merits of this change. I'm satisfied that the code formatting changes that I requested were made. I can't "approve" this PR because I don't know enough about the problem space to say if this is correct or not.

Issue #1289, opened 2025-07-07
[bug]: sorcery - duplicate objects from multiple backends and backend divergence on update


alex2grad:
### Severity

Minor

### Versions

20.14, 20.15.0-rc2

### Components/Modules

sorcery, res_sorcery_astdb, res_sorcery_memory, res_sorcery_realtime

### Operating Environment

Slackware 15.0 x86_64

### Frequency of Occurrence

Constant

### Issue Description

This issue describes two closely related bugs in Asterisk's Sorcery framework when multiple backends (mappings) are configured for the same object type.

1. Duplicate objects across multiple backends

When an object (e.g., a PJSIP contact) exists in more than one Sorcery backend (e.g., realtime and astdb), Asterisk currently returns multiple instances of the same object during lookup. This leads to incorrect behavior in modules that expect uniqueness — most notably, the PJSIP registrar.

For example, if max_contacts=1 and remove_existing=yes are set for an endpoint, and a matching contact exists in both realtime and astdb, Sorcery returns a container with two objects. The registrar then incorrectly concludes that more than max_contacts are in use, and may delete valid contacts, instead of updating the expiration time on the existing one.

This leads to unexpected registration removal and broken contact tracking.

2. Backend divergence during update across multiple writable backends

When multiple writable mappings are defined for a Sorcery object type, like:

[res_pjsip]
contact=astdb,registrar
contact=realtime,ps_contacts

Asterisk will call the .update() method of each backend when an object is modified. However, if the object does not exist in a given backend, all backend implementations simply skip the update silently, rather than creating the missing object.

This results in a divergence scenario:

    If a backend (e.g., realtime) is unavailable when a contact is initially created, the object may be added only to astdb.

    Later updates to the contact will call .update() on both backends — but the realtime backend will skip it because the object is not present.

    The object is therefore never inserted in the missing backend, and the data becomes permanently inconsistent.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1290, opened 2025-07-07
sorcery: Prevent duplicate objects and ensure missing objects are created on update

alex2grad:
This patch resolves two issues in Sorcery objectset handling with multiple
backends:

1. Prevent duplicate objects:
   When an object exists in more than one backend (e.g., a contact in both
   'astdb' and 'realtime'), the objectset previously returned multiple instances
   of the same logical object. This caused logic failures in components like the
   PJSIP registrar, where duplicate contact entries led to overcounting and
   incorrect deletions, when max_contacts=1 and remove_existing=yes.

   This patch ensures only one instance of an object with a given key is added
   to the objectset, avoiding these duplicate-related side effects.

2. Ensure missing objects are created:
   When using multiple writable backends, a temporary backend failure can lead
   to objects missing permanently from that backend.
   Currently, .update() silently fails if the object is not present,
   and no .create() is attempted.
   This results in inconsistent state across backends (e.g. astdb vs. realtime).

   This patch introduces a new global option in sorcery.conf:
     [general]
     update_or_create_on_update_miss = yes|no

   Default: no (preserves existing behavior).

   When enabled: if .update() fails with no data found, .create() is attempted
   in that backend. This ensures that objects missing due to temporary backend
   outages are re-synchronized once the backend is available again.

   Added a new CLI command:
     sorcery show settings
   Displays global Sorcery settings, including the current value of
   update_or_create_on_update_miss.

   Updated tests to validate both flag enabled/disabled behavior.

Fixes: #1289

UserNote: Users relying on Sorcery multiple writable backends configurations
(e.g., astdb + realtime) may now enable update_or_create_on_update_miss = yes
in sorcery.conf to ensure missing objects are recreated after temporary backend
failures. Default behavior remains unchanged unless explicitly enabled.

### DIFF
--- a/configs/samples/sorcery.conf.sample
+++ b/configs/samples/sorcery.conf.sample
@@ -1,5 +1,16 @@
 ; Sample configuration file for Sorcery Data Access Layer
 
+[general]
+
+; When true, writable Sorcery backends may attempt to insert the object
+; if the object is missing.
+; This helps re-populate missing backends when using multiple writable mappings
+; (e.g., AstDB + Realtime) if a backend was temporarily unavailable.
+; NOTE: This is best-effort and does not guarantee atomicity across backends.
+; Default: no (preserves legacy behavior).
+;
+;update_or_create_on_update_miss = no
+
 ;
 ; Wizards
 ;
--- a/res/res_sorcery_astdb.c
+++ b/res/res_sorcery_astdb.c
@@ -373,7 +373,7 @@ static int sorcery_astdb_update(const struct ast_sorcery *sorcery, void *data, v
 	snprintf(family, sizeof(family), "%s/%s", prefix, ast_sorcery_object_get_type(object));
 
 	/* It is okay for the value to be truncated, we are only checking that it exists */
-	if (ast_db_get(family, ast_sorcery_object_get_id(object), value, sizeof(value))) {
+	if (ast_db_get(family, ast_sorcery_object_get_id(object), value, sizeof(value)) && !ast_sorcery_update_or_create_on_update_miss) {
 		return -1;
 	}
 
--- a/include/asterisk/sorcery.h
+++ b/include/asterisk/sorcery.h
@@ -1623,6 +1623,24 @@ int ast_sorcery_is_object_field_registered(const struct ast_sorcery_object_type 
  */
 const char *ast_sorcery_get_module(const struct ast_sorcery *sorcery);
 
+/**
+ * \brief Global control for optional update->create fallback in backends.
+ *
+ * When non-zero, writable Sorcery backends may attempt to call create()
+ * if update() indicates the object is missing (e.g., 0 rows affected or
+ * NOT_FOUND). This helps re-populate backends that temporarily missed an
+ * insert.
+ *
+ * Default is 0 (off) to maintain legacy behavior.
+ *
+ * Configured via sorcery.conf:
+ *   [general]
+ *   update_or_create_on_update_miss = yes|no
+ *
+ * NOTE: Backends MUST gate their fallback logic on this variable.
+ */
+extern int ast_sorcery_update_or_create_on_update_miss;
+
 /*!
  * \section AstSorceryConvenienceMacros Simple Sorcery Convenience Macros
  *
--- a/main/sorcery.c
+++ b/main/sorcery.c
@@ -42,6 +42,7 @@
 #include "asterisk/threadpool.h"
 #include "asterisk/json.h"
 #include "asterisk/vector.h"
+#include "asterisk/cli.h"
 
 /* To prevent DEBUG_FD_LEAKS from interfering with things we undef open and close */
 #undef open
@@ -283,6 +284,9 @@ struct ao2_container *observers;
 /*! \brief Registered sorcery instances */
 static struct ao2_container *instances;
 
+/* \brief Global config flag (declared in sorcery.h) */
+int ast_sorcery_update_or_create_on_update_miss = 0;
+
 static int int_handler_fn(const void *obj, const intptr_t *args, char **buf)
 {
 	int *field = (int *)(obj + args[0]);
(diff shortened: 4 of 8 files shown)


alex2grad:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

alex2grad:
@gtjoseph,
do I need to do anything about cherry-pick-checks-failed?

gtjoseph:
> @gtjoseph, do I need to do anything about cherry-pick-checks-failed?

Nope.  It appears github is having issues again with their caching framework.  I'll deal with it and re-run the tests when they get it fixed.

res/res_sorcery_astdb.c
```diff
@@ -367,17 +367,7 @@ static void sorcery_astdb_retrieve_prefix(const struct ast_sorcery *sorcery, voi
 
 static int sorcery_astdb_update(const struct ast_sorcery *sorcery, void *data, void *object)
 {
-	const char *prefix = data;
```
jcolp: I do not agree with this fundamental behavior change being the default. It should be optional, and default to off.

It should also be documented that this is can still fail, due to something else manipulating the backend between the update and the create fallback.

alex2grad: I'd like to respectfully clarify why I consider this patch is a bug fix, not a fundamental behavior change,
why it should be the default behavior, not optional, and explain the reasoning behind it:

1. Current behavior is inconsistent and misleading
    When multiple writable backends are configured, Asterisk attempts to update an object in all of them.
    However, all backends silently refuse to update if the object doesn't exist in that backend.
    This silently causes divergence between backends - and no indication is logged or returned to the caller.
    The current behavior is surprising, undocumented, and damaging.

2. This bug is observably harmful
    As shown in the use case with ps_contacts and astdb, divergence can break core modules like res_pjsip_registrar.
    Real-world impact:
        Contacts disappear from one backend permanently.
        Registrar miscalculates contact counts.
    This is not a theoretical edge case — it happens in practice, and is difficult to detect or correct after the fact.

3. Why optional behavior won’t help
    Making the fallback behavior optional (e.g., via config flag) shifts responsibility to the user/admin — but they don’t even know divergence is occurring.
    There’s no logging, no error, no warning. Asterisk acts like the update succeeded.
    So offering an “optional fix” doesn’t prevent the harm; it just hides the broken behavior behind complexity.
    Making this fix optional would leave users exposed to silent corruption they can't detect.

The logic patched here simply adjusts how Sorcery handles failed .update() attempts across multiple writable backends.
The patch ensures that if an object is not present, we gracefully fall back to .create(), which is the same logic used during object initialization.
No new semantic is being added.

The current behavior causes divergence and silent failure, especially in multi-backend configurations like:

[res_pjsip]
contact=astdb,registrar
contact=realtime,ps_contacts

If the realtime DB is temporarily down during a contact add, only AstDB gets the contact.
When the realtime DB comes back up, it will never be updated or populated, even after repeated registration updates, because .update() is silently skipped, and no .create() is attempted.

The behavior being fixed here is not documented - the idea that Sorcery should silently skip missing objects during update is not stated in Asterisk documentation.
On the other hand, users reasonably expect that using multiple writable backends results in synchronized data.
That expectation is currently broken.

Because this patch aligns actual behavior with expected and safe operation, without introducing new risk, I continue to believe it should be the default and not optional. 

If you believe the fallback .create() behavior should be optional, I'm willing to add a configuration setting to enable/disable this behavior, with appropriate documentation.

jcolp: I'll ask for input from others, but I do want to state that this is still not an atomic operation - meaning it is possible, albeit unlikely, that between update and create something else can manipulate the backend, resulting in failure still occurring during that slim window.

alex2grad: The current implementation also is not atomic - between check and update something else can manipulate the backend.

jcolp: My comment was not specific to this implementation. It was a general comment applicable to others as well, realtime more so.

alex2grad: @jcolp,
Just following up kindly to see if there's been any further input or discussion regarding this PR.

jcolp: Any further input or discussion would be on this issue, so - no. If that changes then it'll appear on here.

mbradeen: What are the underlying scenarios where the second resolved issue occurs ie; why would the insert fail on one back end and not another (and why would we assume adding it this time wouldn't fail again?)

If we are already in a mismatched state due to an error like a temporary back end disconnect, we are in undefined territory.  Adding missing elements on the fly is not going to resolve that deterministically.

I think this has to be behind an optional flag if we are talking about non-deterministic scenarios.

alex2grad: @mbradeen,
Thank you for the reply. I believe I already described this scenario in detail in my earlier comment to @jcolp, but let me clarify again:

In multi-backend configurations like:

[res_pjsip]  
contact=astdb,registrar  
contact=realtime,ps_contacts

If the Realtime DB is temporarily down during a contact registration, the contact is only inserted into AstDB. Once the Realtime DB comes back online, it never receives the object, not even after multiple subsequent updates, because .update() silently skips the backend if the object is missing, and no .create() is attempted. This leads to permanent divergence.

So to directly answer your questions:

    Why would insert fail on one backend and not another?
    Because AstDB is local (SQLite), it is always available. Realtime DBs, however, are typically remote (e.g., MySQL/Postgres), and can become temporarily unavailable due to server restarts, network hiccups, or DB maintenance.

    Why assume that adding it this time wouldn’t fail again?
    Because this is about recovery from a temporary failure. If the backend is now available (after reboot or reconnection), it should accept the insert.

This isn’t a theoretical edge case. It has occurred in real-world scenarios where Asterisk-based services experienced backend divergence and incorrect behavior, such as missing contacts due to incomplete backend state.

You're right that a temporary backend disconnect introduces a non-deterministic window where mismatched state can occur. However, I’d argue that doing nothing and silently skipping updates only guarantees continued divergence. The system will never self-heal unless manual intervention is taken, which is not scalable or practical in production.

Adding missing elements on-the-fly does not claim to fully resolve all divergence scenarios, but it does close the most common and damaging gap: a writable backend missing a contact because it was temporarily unavailable during registration.

This patch simply ensures that writable backends that can now accept data will be updated, instead of being silently ignored. It's still subject to existing backend error handling, if the insert fails, it will still fail gracefully. But if it succeeds, the system moves closer to consistency.
I'm not trying to define a full sync protocol, just to avoid "write once, forget forever" behavior.
This is a safe and bounded improvement: it only inserts when .update() fails due to a missing object.
Without this fix or option, divergence becomes permanent by design.

That’s why I see this as a correctness fix, not a fundamental behavior change, but I remain open to guarding it behind a flag if that’s the consensus. But I strongly believe the current silent failure is dangerous and should be addressed in some form.

alex2grad:
It’s been over 3 weeks since @jcolp requested reviews from @gtjoseph, @bkford and @mbradeen. I’m wondering if the lack of feedback means the consensus is to guard this behavior behind an optional flag. I’m happy to implement that if it will help move this PR forward.

jcolp:
Yes, as an optional flag would move it along.

alex2grad:
I’ll proceed with adding an optional flag for this behavior.

One design question before I start.
Where would you prefer the flag to live?

Option 1: As a global option in asterisk.conf (simpler).
Option 2: In sorcery.conf, which currently only has mappings. This would require adding a new [general] section for global Sorcery options (more complex change).

jcolp:
I would prefer sorcery.conf as it is sorcery specific and isolated to it.

configs/samples/sorcery.conf.sample
```diff
@@ -1,5 +1,16 @@
 ; Sample configuration file for Sorcery Data Access Layer
 
+[general]
+
+; When non-zero, writable Sorcery backends may attempt to insert the object
```
gtjoseph: "true" rather than "non-zero".

Issue #1288, opened 2025-07-07
[bug]: Alternate Channel Storage Backends


phoneben:
### Severity

Trivial

### Versions

20.15.0-rc1

### Components/Modules

 Alternate Channel Storage Backends

### Operating Environment

centos 7

### Frequency of Occurrence

None

### Issue Description

it is keping on coreing what do i attach?

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

jcolp:
If you believe it is coreing, then:

https://docs.asterisk.org/Development/Debugging/Getting-a-Backtrace/?h=backtrace

phoneben:
can i send the files securely without having to scrub them ?

![Image](https://github.com/user-attachments/assets/503e8eca-6fb3-4b12-b5e2-99550368833b)

jcolp:
It is highly preferred that you provide the files publicly.

phoneben:
is there a script to scrub the files ?any idea?

phoneben:
IF THIS IS TO MUTCH SCRUBBED PLEASE TELL ME THANK YOU 
[core-asterisk-2025-07-07T12-33-26Z-info_scrubbed_20250707_224503.txt](https://github.com/user-attachments/files/21109281/core-asterisk-2025-07-07T12-33-26Z-info_scrubbed_20250707_224503.txt)
[core-asterisk-2025-07-07T12-33-26Z-thread1_scrubbed_20250707_224503.txt](https://github.com/user-attachments/files/21109278/core-asterisk-2025-07-07T12-33-26Z-thread1_scrubbed_20250707_224503.txt)
[processing_summary_20250707_224503.txt](https://github.com/user-attachments/files/21109284/processing_summary_20250707_224503.txt)
[core-asterisk-2025-07-03T09-20-05Z-brief_scrubbed_20250707_224503.txt](https://github.com/user-attachments/files/21109288/core-asterisk-2025-07-03T09-20-05Z-brief_scrubbed_20250707_224503.txt)
[core-asterisk-2025-07-03T09-20-05Z-locks_scrubbed_20250707_224503.txt](https://github.com/user-attachments/files/21109285/core-asterisk-2025-07-03T09-20-05Z-locks_scrubbed_20250707_224503.txt)
[...]

PR #1295, opened 2025-07-09
channelstorage_cpp_map_name_id: Fix callback returning non-matching channels.

gtjoseph:
When the callback() API was invoked but no channel passed the test, callback
would return the last channel tested instead of NULL.  It now correctly
returns NULL when no channel matches.

Resolves: #1288


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1301, opened 2025-07-14
[bug]: sig_analog: fgccamamf doesn't handle STP, STP2, or STP3


InterLinked1:
### Severity

Major

### Versions

Git

### Components/Modules

sig_analog

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

When using Feature Group C CAMA MF, if using STP, ST2P, or ST3P (which are represented by A, B, and C DTMF frames), the start digit is not parsed correctly and is treated as part of the dialed number, which is incorrect.

### Relevant log output

```shell
[2025-05-31 19:56:14.633] DEBUG[1447][C-00000011]: sig_analog.c:1610 analog_handle_dtmf: Begin DTMF digit: 0x34 '4' on DAHDI/23-1  
[2025-05-31 19:56:14.633] DEBUG[1447][C-00000011]: chan_dahdi.c:1921 my_handle_dtmf: Begin DTMF digit: 0x34 '4' on DAHDI/23-1  
[2025-05-31 19:56:14.653] DEBUG[1447][C-00000011]: sig_analog.c:1610 analog_handle_dtmf: End DTMF digit: 0x34 '4' on DAHDI/23-1  
[2025-05-31 19:56:14.653] DEBUG[1447][C-00000011]: chan_dahdi.c:1921 my_handle_dtmf: End DTMF digit: 0x34 '4' on DAHDI/23-1  
[2025-05-31 19:56:14.773] DEBUG[1447][C-00000011]: sig_analog.c:1610 analog_handle_dtmf: Begin DTMF digit: 0x36 '6' on DAHDI/23-1  
[2025-05-31 19:56:14.773] DEBUG[1447][C-00000011]: chan_dahdi.c:1921 my_handle_dtmf: Begin DTMF digit: 0x36 '6' on DAHDI/23-1  
[2025-05-31 19:56:14.793] DEBUG[1447][C-00000011]: sig_analog.c:1610 analog_handle_dtmf: End DTMF digit: 0x36 '6' on DAHDI/23-1  
[2025-05-31 19:56:14.793] DEBUG[1447][C-00000011]: chan_dahdi.c:1921 my_handle_dtmf: End DTMF digit: 0x36 '6' on DAHDI/23-1  
[2025-05-31 19:56:14.913] DEBUG[1447][C-00000011]: sig_analog.c:1610 analog_handle_dtmf: Begin DTMF digit: 0x33 '3' on DAHDI/23-1  
[...]

PR #1302, opened 2025-07-14
sig_analog: Properly handle STP, ST2P, and ST3P for fgccamamf.

InterLinked1:
Previously, we were only using # (ST) as a terminator, and not handling A (STP), B (ST2P), or C (ST3P), which erroneously led to it being treated as part of the dialed number. Parse any of these as the start digit.

Resolves: #1301


InterLinked1:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

Issue #1305, opened 2025-07-14
[bug]: Realtime incorrectly falls back to next backend on record-not-found (SQL_NO_DATA), causing incorrect behavior and delay


alex2grad:
### Severity

Major

### Versions

 20.14.1

### Components/Modules

main/config, res/res_config_odbc

### Operating Environment

Slackware 15.0 x86_64"

### Frequency of Occurrence

Constant

### Issue Description

When using Asterisk Realtime with multiple backends configured via extconfig.conf using priorities, Asterisk incorrectly attempts to fall back to the next backend even when the current backend is healthy and simply returns no matching data (e.g., SQL_NO_DATA). This breaks expected behavior and introduces logical errors and performance issues.

**Reproduction Steps:**
    1. Configure Realtime mapping, for example:
```
       extensions => odbc,primary,extensions,1
       extensions => odbc,backup,extensions,2
```

    2. In the primary ODBC database, remove an extension (e.g., exten => 1234), ensuring it still exists in the backup DB.

    3. Trigger a lookup of that extension via the dialplan (Dial, Goto, etc.).

**Expected Behavior:**
Asterisk should:
    Query the primary backend
    Receive a SQL_NO_DATA result (no such extension)
    Stop there, and return "not found"
    Only try backup if primary is unreachable or produces a backend failure

**Actual Behavior:**
    Asterisk receives SQL_NO_DATA from the primary backend
    Interprets it as a generic NULL return from realtime_func()
    Proceeds to query the backup backend unnecessarily
    If backup is unreachable/down, it introduces significant delays before ultimately failing

**This behavior leads to:**
    Unexpected fallback to outdated/stale data
    Inconsistencies when deleting or modifying data in the primary
    Delays in call setup or dialplan execution, especially if the backup database is unreachable or slow to respond

**Root Cause:**

In main/config.c, function ast_load_realtime_all_fields() loops through all configured engines:

```
for (i = 1; ; i++) {
    if ((eng = find_engine(family, i, db, sizeof(db), table, sizeof(table)))) {
        if (eng->realtime_func && (res = eng->realtime_func(db, table, fields))) {
            return res;
        }
    } else {
        return NULL;
    }
}
```

The realtime_func() call in the ODBC backend (res/res_config_odbc.c) returns NULL on both error and "no data":

```
res = SQLFetch(stmt);
if (res == SQL_NO_DATA) {
    SQLFreeHandle(SQL_HANDLE_STMT, stmt);
[...]

PR #1306, opened 2025-07-15
res_config_odbc: Prevent Realtime fallback on record-not-found (SQL_NO_DATA)

alex2grad:
This patch fixes an issue in the ODBC Realtime engine where Asterisk incorrectly
falls back to the next configured backend when the current one returns
SQL_NO_DATA (i.e., no record found).
This is a logical error and performance risk in multi-backend configurations.

Solution:
Introduced CONFIG_RT_NOT_FOUND ((void *)-1) as a special return marker.
ODBC Realtime backend now return CONFIG_RT_NOT_FOUND when no data is found.
Core engine stops iterating on this marker, avoiding unnecessary fallback.

Notes:
Other Realtime backends (PostgreSQL, LDAP, etc.) can be updated similarly.
This patch only covers ODBC.

Fixes: #1305

### DIFF
--- a/include/asterisk/config.h
+++ b/include/asterisk/config.h
@@ -126,6 +126,12 @@ typedef int realtime_require(const char *database, const char *table, va_list ap
  */
 typedef int realtime_unload(const char *database, const char *table);
 
+/*! Special return value indicating a successful query that returned no data.
+ * Used by realtime backends to signal "not found" vs an actual backend failure.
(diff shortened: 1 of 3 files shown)


alex2grad:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1307, opened 2025-07-16
[improvement]: ast_tls_cert: Allow certificate validity to be configurable


InterLinked1:
### Improvement Description

Currently, the script is hardcoded to 365 days, make this configurable.

PR #1308, opened 2025-07-16
ast_tls_cert: Make certificate validity configurable.

InterLinked1:
Currently, the ast_tls_cert script is hardcoded to produce certificates with a validity of 365 days, which is not generally desirable for self- signed certificates. Make this parameter configurable.

Resolves: #1307


InterLinked1:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

jcolp (approved):

Issue #1259, opened 2025-05-22
[bug]: New TenantID feature doesn't seem to set CDR for incoming calls


ssether-usinternet:
### Severity

Major

### Versions

20.14.0

### Components/Modules

CDR, PJSIP

### Operating Environment

Ubuntu 22.04, Asterisk 20.14.0

### Frequency of Occurrence

Constant

### Issue Description

I noticed the new TenantID feature blogged about at [Tenant ID, a Flexible Asterisk Tool ⋆ Asterisk](https://www.asterisk.org/tenant-id-a-flexible-asterisk-tool/) and found this would be a useful feature for us, as we have tenants and want the tenantID to show up in CDRs, and currently use a workaround by setting a custom CDR field.

I was able to get this working well for outgoing calls, via setting the tenantID in the pjsip.conf. That winds up in our newly created tenantID column in our CDRs.

However, I’m not able to get the tenantID to show up in the CDR for incoming calls. In this case I’m setting the tenantID in the dialplan, since I won’t know who the tenant is until I can look up who owns the DID in the dialplan. So I do something like this:

same => n,Set(CHANNEL(tenantid)=${companyID})

(companyID is set previously)

However, this doesn’t show up in my DB. Furthermore, this returns blank:

same => n,NoOp(tenantID: ${CDR(tenantid)})

NoOp(tenantID: )

I’ve also tried setting the CDR explicitly,

same => n,Set(CDR(tenantid)=${CHANNEL(tenantid)})

Which when I query CDR, shows up:

NoOp(tenantID: uatTest)

But doesn’t show in my DB column in CDR, it’s still null.

Just to repeat, this all works perfectly for an outgoing call, when tenantID is set in my pjsip.conf for an outgoing call, DB column and all.

It’s just not terribly useful to us unless we can get the tenantID into all calls, outgoing, incoming, fax, etc. I think I followed the blog entry correctly, but otherwise can’t get this feature to work.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

bkford:
Adding the community link here for reference:
https://community.asterisk.org/t/new-tenantid-feature-doesnt-seem-to-set-cdr-for-incoming-calls/108466/4

ssether-usinternet:
[...]

PR #1312, opened 2025-07-17
cdr.c: Set tenantid from party_a->base instead of chan->base.

gtjoseph:
The CDR tenantid was being set in cdr_object_alloc from the channel->base
snapshot.  Since this happens at channel creation before the dialplan is even
reached, calls to `CHANNEL(tenantid)=<something>` in the dialplan were being
ignored.  Instead we now take tenantid from party_a when
cdr_object_create_public_records() is called which is after the call has
ended and all channel snapshots rebuilt.  This is exactly how accountcode
and amaflags, which can also be set in tha dialplpan, are handled.

Resolves: #1259


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph:
> Tested and working, I can now use the CHANNEL function to set tenantid for my incoming calls.

Thanks @SB-JohnK.  Double check that it still works in all your other scenarios as well (if you haven't already done so).

PR #1314, opened 2025-07-21
options:  Change ast_options from ast_flags to ast_flags64.

gtjoseph:
DeveloperNote: The 32-bit ast_options has no room left to accomodate new
options and so has been converted to an ast_flags64 structure. All internal
references to ast_options have been updated to use the 64-bit flag
manipulation macros.  External module references to the 32-bit ast_options
should continue to work on little-endian systems because the
least-significant bytes of a 64 bit integer will be in the same location as a
32-bit integer.  Because that's not the case on big-endian systems, we've
swapped the bytes in the flags manupulation macros on big-endian systems
so external modules should still work however you are encouraged to test.


gtjoseph:
cherry-pick-to: 21
cherry-pick-to: 22

gtjoseph:
Hmmm.  I'll submit a separate PR for the 20 branch when this one gets approved.

bkford:
Lgtm. Are those flaky tests or flukes, or is there a change needed for this pull request or the tests themselves?

gtjoseph:
I'm pretty sure they're just flakey but I just kicked them off again to be sure.

bkford (approved):

Issue #1315, opened 2025-07-22
[bug]: [bug]: When executed with dialplan, the action Redirect does not take effect.


Tinet-mucw:
### Severity

Trivial

### Versions

22.0.0

### Components/Modules

main

### Operating Environment

centos8 

### Frequency of Occurrence

One Time

### Issue Description

Log of Abnormal Execution

```shell
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx.c: Executing [~~s~~@ivr_websocket_robot:49] GotoIf("PJSIP/sipforward01-0000f11c", "0?50:82") in new stack
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx_builtins.c: Goto (ivr_websocket_robot,~~s~~,82)
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx.c: Executing [~~s~~@ivr_websocket_robot:82] GotoIf("PJSIP/sipforward01-0000f11c", "1?83:85") in new stack
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx_builtins.c: Goto (ivr_websocket_robot,~~s~~,83)
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx.c: Executing [~~s~~@ivr_websocket_robot:83] Set("PJSIP/sipforward01-0000f11c", "cur_node_wait_time=10") in new stack
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx.c: Executing [~~s~~@ivr_websocket_robot:84] Goto("PJSIP/sipforward01-0000f11c", "86") in new stack
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx_builtins.c: Goto (ivr_websocket_robot,~~s~~,85)
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx.c: Executing [~~s~~@ivr_websocket_robot:84] Goto("PJSIP/sipforward01-0000f11c", "86") in new stack
[Jul 11 15:17:58] VERBOSE[25143][C-0000afe8] pbx_builtins.c: Goto (ivr_websocket_robot,~~s~~,86)
[...]

PR #1316, opened 2025-07-22
pbx.c: When the AST_SOFTHANGUP_ASYNCGOTO flag is set, pbx_extension_helper should return directly.

Tinet-mucw:
Under certain circumstances the context/extens/prio are set in the ast_async_goto, for example action Redirect.
In the situation that action Redirect is broken by pbx_extension_helper this info is changed.
This will cause the current dialplan location to be executed twice. 
In other words, the Redirect action does not take effect.

Resolves: #1315


Tinet-mucw:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

Issue #1317, opened 2025-07-22
[bug]: AGI command buffer overflow with long variables


viktike:
### Severity

Minor

### Versions

all

### Components/Modules

agi

### Operating Environment

Ubuntu LTS versions including 20.04, 22.04, 24.04

### Frequency of Occurrence

Constant

### Issue Description

No sure if it's a bug or misuse instead, but the AGI and FastAGI command truncated due to command buffer overflow.

I have managed to fix this by changing the buffer site in res/res_agi.c:1519 from 2K to 16K.
`#define AGI_BUF_LEN 16384`

Now I'm not sure if this is a correct value or not, but I'm sure setting variable with this long value works from the dialplan. Let me know if I should create a PR for this.

<img width="1902" height="667" alt="Image" src="https://github.com/user-attachments/assets/9a29ab25-12f9-4872-8696-4967429cc81f" />

### Relevant log output

```shell
res_agi.c: AGI Rx << temp buffer SET VARIABLE TEST "some really-really long value not copied here ... " - errno Operation now in progress
No \n received, checking again.
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
@viktike Yeah go ahead and submit a PR.  2k -> 16K might be a bit much though.  Maybe start with 8k?

viktike:
@gtjoseph I have a question as well, maybe you can help.

The only reason why I do these long JSON variables instead of multiple shorter text ones, is that under high usage, the SET VARIABLE commands become progressively slower. I've tracked it down to pbx_builtin_setvar_helper(), and the statis taskprocessor queue getting overloaded, which is shown in the logs as well.

I've tried many configurations in stasis.conf, pjsip.conf related to taskprocessors, but the only thing that helped was to modify `include/taskprocessor.h` and change `AST_TASKPROCESSOR_HIGH_WATER_LEVEL` from _500_ to all the way up to **25000** and recompile.

How it become 500? Git blame said it was last modified (increased) because one of the test cases didn't even run.
Should I create another issue/PR for this? What would be a reasonable number in your opinion? Or is there anything I'm overseeing?

Hardware is a commodity 16 core x64 server, and AGI is FastAGI served by [Swoole TCP server](https://openswoole.com/docs/modules/swoole-server-doc) written in PHP, which has <5ms response time, so it's definitely not the issue.

jcolp:
[...]

PR #1318, opened 2025-07-22
res_agi: Increase AGI command buffer size from 2K to 8K

viktike:
Fixes: #1317


viktike:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

PR #1325, opened 2025-07-30
resource_channels.c: Don't call ast_channel_get_by_name on empty optional arguments

SvenKube:
`ast_ari_channels_create` and `ast_ari_channels_dial` called the `ast_channel_get_by_name` function with optional arguments. Since 8f1982c4d6, this function logs an error for empty channel names. This commit adds checks for empty optional arguments that are used to call `ast_channel_get_by_name` to prevent these error logs.

cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22


res/ari/resource_channels.c
```diff
@@ -1850,7 +1850,10 @@ void ast_ari_channels_create(struct ast_variable *headers,
 		return;
 	}
 
-	originator = ast_channel_get_by_name(args->originator);
+	if(!ast_strlen_zero(args->originator)) {
```
silentindark: space, if( -> if (

res/ari/resource_channels.c
```diff
@@ -1952,7 +1955,9 @@ void ast_ari_channels_dial(struct ast_variable *headers,
 		return;
 	}
 
-	caller = ast_channel_get_by_name(args->caller);
+	if(!ast_strlen_zero(args->caller)) {
```
silentindark: space, if( -> if (

res/ari/resource_channels.c
SvenKube: Thx, changed it.

res/ari/resource_channels.c
SvenKube: same

res/ari/resource_channels.c
```diff
@@ -1790,7 +1790,7 @@ void ast_ari_channels_create(struct ast_variable *headers,
 	char *stuff;
 	int cause;
 	struct ast_format_cap *request_cap;
-	struct ast_channel *originator;
+	RAII_VAR(struct ast_channel *, originator, NULL, ast_channel_cleanup);
```
jcolp: Why the switch to RAII_VAR? originator could just be set to NULL here. As well there are already ast_channel_cleanup calls for originator, does this not cause a double unref?

SvenKube: :man_facepalming: Yes, you are totally right.

jcolp (approved):

gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

PR #1328, opened 2025-07-30
options:  Change ast_options from ast_flags to ast_flags64.

gtjoseph:
DeveloperNote: The 32-bit ast_options has no room left to accomodate new
options and so has been converted to an ast_flags64 structure. All internal
references to ast_options have been updated to use the 64-bit flag
manipulation macros.  External module references to the 32-bit ast_options
should continue to work on little-endian systems because the
least-significant bytes of a 64 bit integer will be in the same location as a
32-bit integer.  Because that's not the case on big-endian systems, we've
swapped the bytes in the flags manupulation macros on big-endian systems
so external modules should still work however you are encouraged to test.


gtjoseph:
cherry-pick-to: none

gtjoseph:
This is the "20-only" version of the change.

jcolp (approved):

Issue #1327, opened 2025-07-30
[bug]: res_stasis_device_state: can't delete ARI Devicestate after asterisk restart


jose-lopes-nfon:
### Severity

Trivial

### Versions

22.5.0

### Components/Modules

res_stasis_device_state

### Operating Environment

Linux Debian 11.11

### Frequency of Occurrence

Constant

### Issue Description

I am using ARI Devicestates and noticed that after an asterisk restart, when I delete ARI Devicestate, it doesn't return an error, but the Devicestate is still there.

I tried to do some analysis, and I got the following conclusions
* When I delete devicestate, the devicestate was not found in cache, so it returned ok
* I believe I found the problem in the populate_cache function. I believe there is a typo there ("%s%s\n" -> "%s%s"): https://github.com/asterisk/asterisk/blame/009e3ef3f5fae65700515739d8f9d8ff522eed37/res/res_stasis_device_state.c#L286  
* I changed the previous line, and it worked for me. I will create a PR shortly.

```
$ curl -s -X PUT -u "$ari_user:$ari_password" localhost:8088/ari/deviceStates/Stasis:403?deviceState=INUSE

$ curl -s -u "$ari_user:$ari_password" localhost:8088/ari/deviceStates
[{"name":"Stasis:403","state":"INUSE"}]

$ sudo systemctl restart asterisk

jose.lopes@pbx011:~$ curl -s -u "$ari_user:$ari_password" localhost:8088/ari/deviceStates
[{"name":"Stasis:403","state":"INUSE"}]

$ curl -v -s -X DELETE -u "$ari_user:$ari_password" localhost:8088/ari/deviceStates/Stasis:403
* Connected to localhost (127.0.0.1) port 8088 (#0)
* Server auth using Basic with user 'admin'
> DELETE /ari/deviceStates/Stasis:403 HTTP/1.1
> Host: localhost:8088
> Authorization: Basic xxx
> User-Agent: curl
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 204 No Content
< Server: Asterisk
< Date: Wed, 30 Jul 2025 12:33:39 GMT
< Cache-Control: no-cache, no-store
< Content-Length: 0
<
* Connection #0 to host localhost left intact


$ curl -s -u "$ari_user:$ari_password" localhost:8088/ari/deviceStates
[{"name":"Stasis:403","state":"INUSE"}]
```

### Relevant log output

```shell
[...]

PR #1329, opened 2025-07-30
res_stasis_device_state: Fix delete ARI Devicestates after asterisk restart.

jose-lopes-nfon:
After an asterisk restart, the deletion of ARI Devicestates didn't
return error, but the devicestate was not deleted.
Found a typo on populate_cache function that created wrong cache for
device states.
This bug caused wrong assumption that devicestate didn't exist,
since it was not in cache, so deletion didn't returned error.

Fixes: #1327


jose-lopes-nfon:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

PR #1331, opened 2025-07-31
safe_asterisk: Add ownership checks for /etc/asterisk/startup.d and its files.

gtjoseph:
UpgradeNote: The safe_asterisk script now checks that, if it was run by the
root user, the /etc/asterisk/startup.d directory and all the files it contains
are owned by root.  If the checks fail, safe_asterisk will exit with an error
and Asterisk will not be started.  Additionally, the default logging
destination is now stderr instead of tty "9" which probably won't exist
in modern systems.

Resolves: #GHSA-v9q8-9j8m-5xwp


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

jcolp (approved):

Issue #1332, opened 2025-08-01
[new-feature]: func_curl: Allow auth methods to be set


InterLinked1:
### Feature Description

Currently, func_curl only supports basic authentication - also allow digest authentication to be used.

PR #1333, opened 2025-08-01
func_curl: Allow auth methods to be set.

InterLinked1:
Currently the CURL function only supports Basic Authentication, the default auth method in libcurl. Add an option that also allows enabling digest authentication.

Resolves: #1332

UserNote: The httpauth field in CURLOPT now allows the authentication methods to be set.


InterLinked1:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

funcs/func_curl.c
```diff
@@ -131,6 +131,14 @@
 						<para>Include header information in the result
 						(boolean)</para>
 					</enum>
+					<enum name="httpauth">
+						<para>Type of authentication method to use. The default is Basic Authentication.</para>
+						<para>Multiple values can be specified to enable multiple auth methods.</para>
```
seanbright: How? I assume by invoking `CURLOPT` multiple times?

Also "auth" -> "authentication"

funcs/func_curl.c
```diff
@@ -489,6 +500,19 @@ static int acf_curlopt_write(struct ast_channel *chan, const char *cmd, char *na
 				if ((new = ast_calloc(1, sizeof(*new)))) {
 					new->value = (void *)ptype;
 				}
+			} else if (key == CURLOPT_HTTPAUTH) {
+				long authtype = 0;
+				if (strcasestr(value, "basic")) {
```
seanbright: Did you mean to use `strcasestr`? `strcasecmp` would be more portable.

seanbright (changes requested):

funcs/func_curl.c
InterLinked1: It can (and has to) be done in one go, e.g. `Set(CURLOPT(httpauth)=basic,digest)`

funcs/func_curl.c
InterLinked1: I did, I initially wrote it using strcasecmp, but then it occured to me a user might want to set multiple, and strcasestr made more sense for that.

What are the portability issues with strcasestr? It's used in the tree already.

seanbright: > I did, I initially wrote it using strcasecmp, but then it occured to me a user might want to set multiple, and strcasestr made more sense for that.

You would need to use separate `if` statements instead of `else if`s, correct? It's also a little gross, since `basicallydigestive` would be valid.

> What are the portability issues with strcasestr? It's used in the tree already.

I wasn't aware of that. It's a GNU extension apparently.

InterLinked1: > You would need to use separate if statements instead of else ifs, correct?

Yeah, that's a good catch.

> It's also a little gross, since basicallydigestive would be valid.

If you don't think there's a compelling reason to allow multiple methods, I can undo that part. For my use case, I only needed to change it to digest, not set both. Not familiar with how common requiring both would be in the real world.

seanbright: > If you don't think there's a compelling reason to allow multiple methods, I can undo that part.

I think if you're going to allow multiples the parsing should be better. Do with that information what you will.

funcs/func_curl.c
```diff
@@ -131,6 +131,14 @@
 						<para>Include header information in the result
 						(boolean)</para>
 					</enum>
+					<enum name="httpauth">
+						<para>Type of authentication method to use. The default is Basic Authentication.</para>
+						<para>Multiple values can be specified to enable multiple authentication methods. To do so, invoke CURLOPT once using comma-separated values.</para>
```
seanbright: Wrap this.

seanbright (changes requested):
One minor thing. Otherwise looks fine.

seanbright (approved):

PR #1334, opened 2025-08-01
channelstorage_cpp_map_name_id.cc: Refactor iterators for thread-safety.

gtjoseph:
The fact that deleting an object from a map invalidates any iterator
that happens to currently point to that object was overlooked in the initial
implementation.  Unfortunately, there's no way to detect that an iterator
has been invalidated so the result was an occasional SEGV triggered by modules
like app_chanspy that opens an iterator and can keep it open for a long period
of time.  The new implementation doesn't keep the underlying C++ iterator
open across calls to ast_channel_iterator_next() and uses a read lock
on the map to ensure that, even for the few microseconds we use the
iterator, another thread can't delete a channel from under it.  Even with
this change, the iterators are still WAY faster than the ao2_legacy
storage driver.

Full details about the new implementation are located in the comments for
iterator_next() in channelstorage_cpp_map_name_id.cc.

Resolves: 1309


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

alex2grad:
@gtjoseph,

Just wondering how the modified "C++ Maps on Name and Uniqueid" compares to "Sqlite3 In-memory Database". Could it be that "Sqlite3 In-memory Database" will be better or have the same performance, but won't require using C++?

gtjoseph:
> @gtjoseph,
> 
> Just wondering how the modified "C++ Maps on Name and Uniqueid" compares to "Sqlite3 In-memory Database". Could it be that "Sqlite3 In-memory Database" will be better or have the same performance, but won't require using C++?

Excellent question so I added back in the sqlite3 driver and ran the performance unit test...

Except for the summary lines, all times are microseconds.

| Test  | C++ Map<br>With PR | C++ Map<br>Original | Sqlite3 | Legacy AO2 |
| ----:  | -------: | -----: | ------: | ----------: |
|        Channel Count:  |  10000 |   10000| 10000 |   10000|
|      create channels:  |  17425 |   23886| 78871 |   33716|
|        by name exact:  |   5370 |    7003|  9482 |    8526|
|    by uniqueid exact:  |   3783 |    5435|  7387 | 3310760|
|   by uniqueid via nm:  |   6780 |    8083| 13321 | 3268882|
|       by name prefix:  |   3549 |    4181|  9896 | 3825405|
|     by context/exten:  |  17701 |   24176| 58046 |   52247|
|        iter all chan:  |   1299 |     638|  1238 |     988|
| iter 10 partial name:  |      6 |       4|    27 |     833|
|   iter context/exten:  |    980 |    1345|  1924 |    1262|
|     del all channels:  |   9753 |    9258| 51546 |    4363|
|                TOTAL:  |   66ms |    88ms| 233ms | 10507ms|

Although the sqliteq and ao2 drivers look slightly better than the C++ map for the "iter all chan" test, it's a difference of only 300 _microseconds_ over 10,000 channels and its performance in the other categories more than makes up for that.  So, C++ Map still wins!

EDIT:  I added a column with the stats for C++ Map without this PR for comparison.  Don't ask me why the original C++ Map implementation is slightly slower than this PR in most categories.  I used exactly the same configure and compile options.

jcolp (approved):

Issue #1260, opened 2025-05-24
[bug]: Asterisk sends RTP audio stream before ICE/DTLS completes


meengu:
### Severity

Minor

### Versions

22

### Components/Modules

res_rtp_asterisk

### Operating Environment

Debian 11, manual asterisk build from source code without any special options.

### Frequency of Occurrence

Constant

### Issue Description

I have a setup with ICE/DTLS-SRTP/mandatory encryption enabled on an endpoint:

```
 dtls_auto_generate_cert            : No
 dtls_ca_file                       : 
 dtls_ca_path                       : 
 dtls_cert_file                     : /etc/certs/host.crt
 dtls_cipher                        : 
 dtls_fingerprint                   : SHA-256
 dtls_private_key                   : /etc/certs/host.key
 dtls_rekey                         : 0
 dtls_setup                         : actpass
 dtls_verify                        : Yes

 media_encryption                   : dtls
 media_encryption_optimistic        : false

 webrtc                             : yes
 ice_support                        : true

 rtcp_mux                           : true
```

I looked at the code, and I did not find any place in the media write path, where there is any check that would block delivery of RTP audio packets to the network until DTLS handshake completes. Indeed, RTP packets are sent out right from the moment SDP from INVITE response is processed, even if encryption is configured to be mandatory.

Until ICE completes and finds a working pair of candidates, the packets are also sent to an address and port from `o=` and `m=` lines, instead of to the valid candidate. Not sure if this is intentional behavior. I'd assume, that when both ends are detected to support ICE, asterisk should wait for ICE completion, before sending any RTP data. And should not send unencrypted RTP data before DTLS handshake completes.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

bkford:
Hi @meengu , there's a pull request up for this which can be found here:
https://github.com/asterisk/asterisk/pull/1337

If you're able to, would you be willing to test this and confirm if it prevents RTP from being sent before DTLS is complete?

nappsoft:
This commit breaks WebRTC video conferencing in our setup (causing several seconds of delay). Reverting the change fixes the issue.

jcolp:
Please create a new issue with accompanying debug logs.

meengu:
@nappsoft 

[...]

PR #1337, opened 2025-08-04
res_rtp_asterisk: Don't send RTP before DTLS has negotiated.

bkford:
There was no check in __rtp_sendto that prevented Asterisk from sending
RTP before DTLS had finished negotiating. This patch adds logic to do
so.

Fixes: #1260

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -3451,6 +3451,16 @@ static int __rtp_sendto(struct ast_rtp_instance *instance, void *buf, size_t siz
 	struct ast_rtp *transport_rtp = ast_rtp_instance_get_data(transport);
 	struct ast_srtp *srtp = ast_rtp_instance_get_srtp(transport, rtcp);
 	int res;
+#if defined(HAVE_OPENSSL) && (OPENSSL_VERSION_NUMBER >= 0x10001000L) && !defined(OPENSSL_NO_SRTP)
+	char *out = buf;
+	struct dtls_details *dtls = !rtcp ? &rtp->dtls : &rtp->rtcp->dtls;
+
+	/* Don't send RTP if DTLS hasn't finished yet */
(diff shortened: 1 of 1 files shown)


res/res_rtp_asterisk.c
```diff
@@ -3451,6 +3451,14 @@ static int __rtp_sendto(struct ast_rtp_instance *instance, void *buf, size_t siz
 	struct ast_rtp *transport_rtp = ast_rtp_instance_get_data(transport);
 	struct ast_srtp *srtp = ast_rtp_instance_get_srtp(transport, rtcp);
 	int res;
+#if defined(HAVE_OPENSSL) && (OPENSSL_VERSION_NUMBER >= 0x10001000L) && !defined(OPENSSL_NO_SRTP)
+	struct dtls_details *dtls = !rtcp ? &rtp->dtls : &rtp->rtcp->dtls;
+
+	/* Handshake not complete yet */
+	if (dtls->ssl && dtls->connection == AST_RTP_DTLS_CONNECTION_NEW) {
```
jcolp: This blocks the DTLS packets themselves from being sent, so this needs to be more intelligent to allow those through.

res/res_rtp_asterisk.c
```diff
@@ -4458,7 +4468,7 @@ static int ast_rtp_dtmf_continuation(struct ast_rtp_instance *instance)
 	int hdrlen = 12, res = 0;
 	char data[256];
 	unsigned int *rtpheader = (unsigned int*)data;
-	int ice;
+	int ice = 0;
```
jcolp: If you're setting it to 0 here, then it doesn't need to be in __rtp_sendto as well - pick one place or the other.

bkford:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

IgorrG:
@jcolp comparing branches found, that this change were not cherry-picked to 23

jcolp:
@IgorrG Thanks for the information! We'll do an audit and see what all is missing.

PR #1338, opened 2025-08-05
res_srtp: Add menuselect options to enable AES_192, AES_256 and AES_GCM

gtjoseph:
UserNote: Options are now available in the menuselect "Resource Modules"
category that allow you to enable the AES_192, AES_256 and AES_GCM
cipher suites in res_srtp. Of course, libsrtp and OpenSSL must support
them but modern versions do.  Previously, the only way to enable them was
to set the CFLAGS environment variable when running ./configure.
The default setting is to disable them preserving existing behavior.


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

res/res_srtp.c
```diff
@@ -35,6 +35,29 @@
 	<support_level>core</support_level>
 ***/
 
+/*** MAKEOPTS
+<category name="MENUSELECT_RES" displayname="Resource Modules">
+	<member name="ENABLE_SRTP_AES_192" displayname="Enable AES 192 cipher suite in res_srtp" touch_on_change="res/res_srtp.c">
```
jcolp, line 40: Does this allow a user to select this when the functionality is not supported? If so, I don't think it should. If we're going to the length of adding this as menuselect options then they should have dependency checking too.

jcolp (dismissed):

res/res_srtp.c
gtjoseph: Hmmm.  No but I can add it (I think).

gtjoseph: Done.

jcolp (approved):

PR #1339, opened 2025-08-06
taskpool: Add taskpool API, switch Stasis+PJSIP to using it.

jcolp:
This change introduces a new API called taskpool. This is a pool of taskprocessors. It provides the following functionality:

1. Task pushing to a pool of taskprocessors
2. Synchronous tasks
3. Serializers for execution ordering of tasks
4. Growing/shrinking of number of taskprocessors in pool

This functionality already exists through the combination of threadpool+taskprocessors but through investigating I determined that this carries substantial overhead for short to medium duration tasks. The threadpool uses a single queue of work, and for management of threads it involves additional tasks.

I wrote taskpool to eliminate the extra overhead and management as much as possible. Instead of a single queue of work each taskprocessor has its own queue and at push time a selector chooses the taskprocessor to queue the task to. Each taskprocessor also has its own thread like normal. This spreads out the tasks immediately and reduces contention on shared resources.

Using the included efficiency tests the number of tasks that can be executed per second in a taskpool is 6-12 times more than an equivalent threadpool+taskprocessor setup.

Serializers and synchronous tasks have already been made more intelligent, allowing PJSIP to operate with only a single thread if desired.


jcolp:
Random thoughts:

1. My investigation has shown that taskprocessors themselves are actually pretty efficient and fast. Problems with them come from the USERS of taskprocessors.
2. Since this is built on taskprocessors the taskpool ones will appear in "core show taskprocessors" with static ones prefixed with "s:" and dynamic ones prefixed with "d:". Dynamic being the part of the pool that can grow/shrink depending on usage and configuration.

include/asterisk/shutdown_group.h
alex2grad, line 1: The file name "shutdown_group.h" does not correctly display the file contents. I think should be serializer_shutdown_group.h

include/asterisk/shutdown_group.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * \return Number of serializers that did not get shutdown within the timeout.
+ */
+int ast_serializer_shutdown_group_join(struct ast_serializer_shutdown_group *shutdown_group, int timeout);
+
+/*!
+ * \brief Increment the number of serializer members in the group.
+ * \since 13.5.0
+ *
+ * \param shutdown_group Group shutdown controller.
```
alex2grad: It's new function, so "since 13.5.0" is incorrect

include/asterisk/shutdown_group.h
```diff
@@ -0,0 +1,61 @@
...
+ */
+ void ast_serializer_shutdown_group_inc(struct ast_serializer_shutdown_group *shutdown_group);
+
+ /*!
+ * \brief Decrement the number of serializer members in the group.
+ * \since 13.5.0
```
alex2grad: It's new function, so "since 13.5.0" is incorrect

main/taskpool.c
```diff
@@ -0,0 +1,820 @@
...
+{
+	unsigned int taskprocessor_num = taskprocessors->taskprocessor_num;
+
+	if (!AST_VECTOR_SIZE(&taskprocessors->taskprocessors)) {
+		*growth_threshold_reached = 1;
+		return;
+	}
+
+	taskprocessors->taskprocessor_num++;
+	if (taskprocessors->taskprocessor_num == AST_VECTOR_SIZE(&taskprocessors->taskprocessors)) {
```
alex2grad: I suggest modifying the task processor scheduling algorithm from round-robin to least-size, as it is now.
This change aligns with the approach I implemented in ast_serializer_pool_get() some years ago, where the least-loaded serializer is selected to ensure better load distribution.

Applying the same least-size strategy here would improve overall concurrency handling and maintain consistency across the codebase.

alex2grad (changes requested):

main/taskpool.c
jcolp: During initial experimentation I did implement least-loaded and the result was actually worse by 15-20% than this simple selector implementation. It ultimately depends on the execution speed of the underlying tasks themselves, since finding the least-loaded also incurs a cost. I'll likely do both implementations so that profiling can be done to see.

alex2grad: Thanks for the context. I agree that synthetic benchmarks can be useful for isolating behavior, but they often don't capture real-world performance characteristics, especially under uneven workloads.

One key issue with round-robin is that it doesn't account for task execution time. Imagine a scenario where one task processor is handling a long-running task, round-robin will continue assigning new tasks to it, causing those new tasks to wait unnecessarily. In contrast, a least-loaded strategy avoids this by assigning new tasks to the queues that are more likely to be free sooner, improving throughput and latency in mixed or unpredictable workloads.

The cost of checking queue lengths is usually negligible compared to the benefit of avoiding overloaded queues, particularly in systems where task durations can vary significantly.

Having both implementations for profiling is a good approach. I'd just caution against optimizing solely based on synthetic test results, as they might hide performance pitfalls that show up only under real-world concurrency patterns.

alex2grad: Another concern with the round-robin approach is the logic around task processor pool growth.

With round-robin, this check only inspects a single task processor in the pool, not the overall load across all task processors. If that one happens to have a backlog (e.g., due to a long-running task), while others are idle or underutilized, the system may incorrectly assume the pool is overloaded and spin up unnecessary additional task processors.

This leads to inefficient resource usage and skews the scheduling behavior even further. A least-loaded approach would both balance the load more effectively and give a more accurate picture of when pool growth is actually needed.

jcolp: I understand the difference between the two, and logically those statements are true. Experience optimizing other parts has shown me that what we theoretically think as being better does not always translate as being better in practice, which is why the profiling is needed.

Swapping out the selector is easy on purpose to allow us to try different approaches and understand the ramifications.

main/taskpool.c
```diff
@@ -0,0 +1,820 @@
...
+	/*! Taskpool-specific options */
+	struct ast_taskpool_options options;
+	/*! Dynamic pool shrinking scheduled item */
+	int shrink_sched_id;
+	/*! The name of the taskpool */
+	char name[0];
+};
+
+/*! \brief The threshold for a taskprocessor at which we consider the pool needing to grow (50% of high water threshold) */
+#define TASKPOOL_GROW_THRESHOLD (AST_TASKPROCESSOR_HIGH_WATER_LEVEL * 5) / 10
```
alex2grad, line 80: What about custom settings mwi_tps_queue_high?
See ast_taskprocessor_alert_set_levels

jcolp: Those apply to the serializer taskprocessors which are separate from the taskpool taskprocessors, and they continue to work.

alex2grad: I believe there might be a misunderstanding. I'm not referring to the serializer taskprocessors specifically.

My point is that AST_TASKPROCESSOR_HIGH_WATER_LEVEL is just the default threshold, but many modules override this via ast_taskprocessor_alert_set_levels(), for example, using the mwi_tps_queue_high setting from configuration or ast_sorcery_object_set_congestion_levels()

        ast_sorcery_object_set_congestion_levels(sorcery, "contact", -1,
                3 * AST_TASKPROCESSOR_HIGH_WATER_LEVEL);

However, this hardcodes the growth threshold based on a default global constant, ignoring cases where High/Congestion level may have different values set by module or via config.

I think TASKPOOL_GROW_THRESHOLD should be derived from the actual high_water level assigned via ast_taskprocessor_alert_set_levels(), not from the default AST_TASKPROCESSOR_HIGH_WATER_LEVEL.

This would ensure consistency with the rest of the system, where thresholds are often tuned per-module for performance and resource control.

jcolp: There is currently no control over the thresholds for taskpool taskprocessors. There is no access to the taskprocessors directly outside of taskpool, so it can only ever be the default. Changing the thresholds may be added via options, in which case this would be changed accordingly.

jcolp: I have added a todo list to my initial comment to update as things progress, including the items from your comments.

phoneben:
just asking 
when I run  `top -Hp (pid of asterisk)`
the top thread consuming cpu 
would be  `taskprocessor`
```
asterisk -rx "core show threads "|grep 21415
0x7fddb0afd700 21415 default_tps_processing_function started at [  229] taskprocessor.c default_listener_start()
```
is this what you are trying to solve ?

this is after using channelstorage_cpp still getting high load

jcolp:
That's an extremely broad question. I am trying to eliminate the overhead that threadpool+taskprocessor incurs. This can appear as a heavily loaded/used taskprocessor. Whether that is the cause of your high load and whether this would help, I have no idea. Ultimately there is always a point where load is high - work has to be done.

phoneben:
thank you 
when ready for testing i would want to participate .

include/asterisk/shutdown_group.h
```diff
@@ -0,0 +1,61 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2012-2013, Digium, Inc.
```
bkford: Outdated copyright

include/asterisk/shutdown_group.h
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Mark Michelson <mmmichelson@digium.com>
```
bkford: Credit yourself :)

include/asterisk/shutdown_group.h
```diff
@@ -0,0 +1,61 @@
...
+ * \brief Create a serializer group shutdown control object.
+ * \since 13.5.0
+ *
+ * \return ao2 object to control shutdown of a serializer group.
+ */
+struct ast_serializer_shutdown_group *ast_serializer_shutdown_group_alloc(void);
+
+/*!
+ * \brief Wait for the serializers in the group to shutdown with timeout.
+ * \since 13.5.0
```
bkford: Same comments as @alex2grad about version. Looks to be in multiple places, so just adding a comment as a reminder for when the time comes to determine which version this will go in.

main/shutdown_group.c
```diff
@@ -0,0 +1,108 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2012-2013, Digium, Inc.
+ *
+ * Mark Michelson <mmmichelson@digium.com>
```
bkford: Copyright and author

main/taskpool.c
```diff
@@ -0,0 +1,820 @@
...
+{
+	struct ast_taskpool *pool = ast_taskpool_get_current();
+	struct ast_taskprocessor *tps = data;
+	struct ast_taskprocessor_listener *listener = ast_taskprocessor_listener(tps);
+	struct serializer *ser = ast_taskprocessor_listener_get_user_data(listener);
+	size_t remaining, requeue = 0;
+
+	/* In a normal scenario this lock will not be in contention with
+	 * anything else. It is only if a synchronous task is pushed to
+	 * the serializer that it may be blocked on the the synchronous
```
bkford: Double "the"

jcolp:
I have rescoped this change to strictly the taskpool implementation and the first consumer of it, stasis. The complexity of PJSIP has proven to be difficult to reconcile in this single change. It uses characteristics of an unbound threadpool to carry out its work. It may be possible to adapt both PJSIP and taskpool to cover its full usage, but that will not be done as part of this change. Stasis has been by far the heaviest user of threadpool+taskprocessor in reality and benefits from this more. Its usage is also substantially simpler and straightforward.

There have also been comments on the shutdown_group code. I added version information for the two API calls I exposed, but the rest was split out from threadpool and the work was done by Mark Michelson - so I'm leaving that aspect unchanged.

This has also been targeted against 24.0.0 only for now, though I lean towards putting it in earlier branches too I think.

jcolp:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

phoneben:
can i test it as is on 20.15.2?

jcolp:
> can i test it as is on 20.15.2?

This was written and tested on master. I have not tried earlier versions and can not guarantee it will apply.

phoneben:
my question was if i test and find a issue will it be dealt with it ?
it compiled but i had to delete first 
`rm -f menuselect.makeopts` it should compile .
will report if a issue arises .

jcolp:
If it's in the implementation, yes.

kodxxl:
@jcolp , are You Sure?

phoneben:
I have tested on version 20.15.2. Previously, the regular average load on a 20-CPU system was higher, with spikes reaching 25-30 during peak times. With the new update, the regular average load improved to 10, and peak usage did not exceed 18.5. During testing at peak load with 1,200+ registered endpoints, 425+ channels, and 200+ calls, over 250,000 total calls were processed without any irregular issues, and Asterisk performed exceptionally well.
Note: This is day 1 data and may be incomplete for full assessment.
 What additional benchmarks or metrics should I focus on tomorrow to get a more comprehensive evaluation?
 

> This has also been targeted against 24.0.0 only for now, though I lean towards putting it in earlier branches too I think.

Given the excellent performance results, I vote it should be included in earlier branches as well.

jcolp:
There's nothing explicitly for benchmarks/metrics or usage. My primary focus was CPU consumption and time spent on stasis messages, which are used throughout the code base so just using Asterisk exercises it.

viktike:
> I have tested on version 20.15.2. Previously, the regular average load on a 20-CPU system was higher, with spikes reaching 25-30 during peak times. With the new update, the regular average load improved to 10, and peak usage did not exceed 18.5. During testing at peak load with 1,200+ registered endpoints, 425+ channels, and 200+ calls, over 250,000 total calls were processed without any irregular issues, and Asterisk performed exceptionally well. Note: This is day 1 data and may be incomplete for full assessment. What additional benchmarks or metrics should I focus on tomorrow to get a more comprehensive evaluation?
> 
> > This has also been targeted against 24.0.0 only for now, though I lean towards putting it in earlier branches too I think.
> 
> Given the excellent performance results, I vote it should be included in earlier branches as well.

I would really like to see this merged back to the v20 branch.

jcolp:
If after code review other developers are comfortable with it, then it will be put into all supported branches.

include/asterisk/serializer.h
```diff
@@ -56,6 +57,23 @@ int ast_serializer_pool_destroy(struct ast_serializer_pool *pool);
 struct ast_serializer_pool *ast_serializer_pool_create(const char *name,
 	unsigned int size, struct ast_threadpool *threadpool, int timeout);
 
+/*!
+ * \brief Create a serializer pool on taskpool.
+ *
```
gtjoseph, line 65: Need a "since"?

include/asterisk/taskpool.h
```diff
@@ -0,0 +1,270 @@
...
+	 * \note This must be equal to or greater than the initial_size,
+	 * otherwise the taskpool will adjust this to the initial_size.
+	 */
+	int max_size;
+	/*!
+	 * \brief The threshold for when to grow the pool
+	 *
+	 * This is the number of tasks that must be in queue before the pool will grow.
+	 *
+	 * \note If not specified a default of the 50% of the high water threshold will be used.
```
gtjoseph: Where is "high water threshold" set?

include/asterisk/taskpool.h
```diff
@@ -0,0 +1,270 @@
...
+ */
+
+
+#ifndef _ASTERISK_TASKPOOL_H
+#define _ASTERISK_TASKPOOL_H
+
+struct ast_taskpool;
+struct ast_taskprocessor;
+struct ast_serializer_shutdown_group;
+
```
gtjoseph, line 58: You might want to add a documentation section here with an overview of taskpools and the difference with threadpools.

include/asterisk/taskpool.h
```diff
@@ -0,0 +1,270 @@
...
+size_t ast_taskpool_taskprocessors_count(struct ast_taskpool *pool);
+
+/*!
+ * \brief Get the current number of queued tasks in the taskpool
+ * \since 24.0.0
+ *
+ * \param pool The taskpool to query
+ * \retval The number of queued tasks in the taskpool
+ */
+long ast_taskpool_queue_size(struct ast_taskpool *pool);
```
gtjoseph, line 184: Why not size_t?

main/taskpool.c
```diff
@@ -0,0 +1,938 @@
...
+	*taskprocessor = least_full;
+}
+
+struct ast_taskpool *ast_taskpool_create(const char *name,
+		const struct ast_taskpool_options *options)
+{
+	struct ast_taskpool *pool;
+
+	/* Enforce versioning on the passed-in options */
+	if (options->version != AST_TASKPOOL_OPTIONS_VERSION) {
```
gtjoseph, line 330: ast_assert() maybe?

main/taskpool.c
```diff
@@ -0,0 +1,938 @@
...
+		return NULL;
+	}
+
+	pool = ao2_alloc(sizeof(*pool) + strlen(name) + 1, NULL);
```
gtjoseph, line 334: No destructor?  If it doesn't need one, does it need to be an ao2 object?

jcolp: The reference counting aspect is still used, as is the lock, to ensure the object remains valid.

main/taskpool.c
jcolp: Threadpool used the same if approach, and I think I prefer it from a user perspective just in case.

include/asterisk/taskpool.h
jcolp: Mirrors that of taskprocessor which used a long as well.

gtjoseph: odd but OK.

main/taskpool.c
```diff
@@ -0,0 +1,938 @@
...
+	 return 0;
+ }
+
+/*! \internal */
+static void taskpool_taskprocessor_dtor(void *obj)
+{
+	struct taskpool_taskprocessor *taskprocessor = obj;
+
+	if (taskprocessor->taskprocessor && ast_taskprocessor_push(taskprocessor->taskprocessor, taskpool_taskprocessor_stop, NULL)) {
+		/* We can't actually do anything if this fails, so just accept reality */
```
mbradeen, line 121: Log a warning?  Or will that induce logger confusion?

main/taskpool.c
```diff
@@ -0,0 +1,938 @@
...
+		ao2_ref(pool, -1);
+		return NULL;
+	}
+
+	/* Create the static taskprocessors based on the passed-in options */
+	for (int i = 0; i < pool->options.minimum_size; i++) {
+		struct taskpool_taskprocessor *taskprocessor;
+
+		taskprocessor = taskpool_taskprocessor_alloc(pool, 's');
+		if (!taskprocessor) {
```
mbradeen, line 395: What happens to the interim allocated tp's if 1 < minimum.size < n and this fails somewhere between 1 and n?

main/taskpool.c
```diff
@@ -0,0 +1,938 @@
...
+		ast_taskpool_shutdown(pool);
+		return NULL;
+	}
+
+	/* Create the dynamic taskprocessor based on the passed-in options */
+	for (int i = 0; i < (pool->options.initial_size - pool->options.minimum_size); i++) {
+		struct taskpool_taskprocessor *taskprocessor;
+
+		taskprocessor = taskpool_taskprocessor_alloc(pool, 'd');
+		if (!taskprocessor) {
```
mbradeen, line 420: Same as previous question.

mbradeen (commented):
just some q's

main/taskpool.c
jcolp: There's nothing actionable, and nothing you can output to be useful.

mbradeen (approved):

include/asterisk/taskpool.h
```diff
@@ -0,0 +1,316 @@
...
+ * any of the maintainers of this project for assistance;
+ * the project provides a web site, mailing lists and IRC
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
+ * at the top of the source tree.
+ */
+
+/*
```
gtjoseph: You should make this a doxygen block like astobj2.h

include/asterisk/taskpool.h
```diff
@@ -0,0 +1,316 @@
...
+	 * otherwise the taskpool will adjust this to the initial_size.
+	 */
+	int max_size;
+	/*!
+	 * \brief The threshold for when to grow the pool
+	 *
+	 * This is the number of tasks that must be in queue before the pool will grow.
+	 *
+	 * \note If not specified a default of the 50% of the high water threshold defined
+         * in taskprocessor.h will be used.
```
gtjoseph: Bad indent.

gtjoseph (approved):

alex2grad:
@jcolp 

Suggested Questions / Requests

1. Documentation / XML Schema Versions
Since this PR was cherry-picked into branches 20, 22, and 23, should the XML <configObject name="taskpool"> schema be versioned accordingly?
For example:
 20.17.0
 22.7.0
 23.1.0
This would make it clear in the documentation when taskpool options became available in each release line.

2. Monitoring and Statistics
I didn’t find any CLI commands for showing, getting, or checking taskpool usage.
There are many useful statistics functions already implemented (e.g., ast_taskpool_taskprocessors_count, ast_taskpool_queue_size), but they don’t seem to be exposed anywhere.
Would it make sense to add CLI options to monitor taskpool usage and thresholds? This would be very helpful for administrators to see when pools are under pressure or misconfigured.

3. Configured Options Visibility
For Stasis taskpools would it be possible to add a CLI command to display the currently configured taskpool options (selector, sizes, timeouts, etc.)?

jcolp:
> @jcolp
> 
> Suggested Questions / Requests
> 
>     1. Documentation / XML Schema Versions
>        Since this PR was cherry-picked into branches 20, 22, and 23, should the XML  schema be versioned accordingly?
>        For example:
>        20.17.0
>        22.7.0
>        23.1.0
>        This would make it clear in the documentation when taskpool options became available in each release line.

There is now a PR up to update it.
 
>     2. Monitoring and Statistics
>        I didn’t find any CLI commands for showing, getting, or checking taskpool usage.
>        There are many useful statistics functions already implemented (e.g., ast_taskpool_taskprocessors_count, ast_taskpool_queue_size), but they don’t seem to be exposed anywhere.
>        Would it make sense to add CLI options to monitor taskpool usage and thresholds? This would be very helpful for administrators to see when pools are under pressure or misconfigured.

It's built on taskprocessor, so it all shows up in "core show taskprocessors". Specific CLI commands could be done, but would duplicate a lot of that information.
 
>     3. Configured Options Visibility
>        For Stasis taskpools would it be possible to add a CLI command to display the currently configured taskpool options (selector, sizes, timeouts, etc.)?

It's probably possible, but my goal was to make a drop-in as much as possible. This would be a separate thing.

alex2grad:
> > ```
> > 2. Monitoring and Statistics
> >    I didn’t find any CLI commands for showing, getting, or checking taskpool usage.
> >    There are many useful statistics functions already implemented (e.g., ast_taskpool_taskprocessors_count, ast_taskpool_queue_size), but they don’t seem to be exposed anywhere.
> >    Would it make sense to add CLI options to monitor taskpool usage and thresholds? This would be very helpful for administrators to see when pools are under pressure or misconfigured.
> > ```
> 
> It's built on taskprocessor, so it all shows up in "core show taskprocessors". Specific CLI commands could be done, but would duplicate a lot of that information.

The "core show taskprocessors" shows tasks in taskprocessors. However, a taskpool contains multiple taskprocessors and manages them as a higher-level construct. What would be useful to see is the pool-level view, e.g., the current number of taskprocessors in the pool, whether the pool has grown or shrunk from its initial size, and thresholds being applied.

So exposing taskpool statistics wouldn’t duplicate existing CLI output, but rather provide different (aggregate) data that complements the per-taskprocessor view.

jcolp:
> > > ```
> > > 2. Monitoring and Statistics
> > >    I didn’t find any CLI commands for showing, getting, or checking taskpool usage.
> > >    There are many useful statistics functions already implemented (e.g., ast_taskpool_taskprocessors_count, ast_taskpool_queue_size), but they don’t seem to be exposed anywhere.
> > >    Would it make sense to add CLI options to monitor taskpool usage and thresholds? This would be very helpful for administrators to see when pools are under pressure or misconfigured.
> > > ```
> > 
> > 
> > It's built on taskprocessor, so it all shows up in "core show taskprocessors". Specific CLI commands could be done, but would duplicate a lot of that information.
> 
> The "core show taskprocessors" shows tasks in taskprocessors. However, a taskpool contains multiple taskprocessors and manages them as a higher-level construct. What would be useful to see is the pool-level view, e.g., the current number of taskprocessors in the pool, whether the pool has grown or shrunk from its initial size, and thresholds being applied.

Taskpool taskprocessors are named explicitly, so you can still see this information to a degree now - that is how many exist, how many are static, how many are dynamic, their use, etc.

> 
> So exposing taskpool statistics wouldn’t duplicate existing CLI output, but rather provide different (aggregate) data that complements the per-taskprocessor view.

Sure. File an issue. I don't know when I'll get to it.

alex2grad:
I was checking "core show taskprocessors like stasis" and noticed that many of the most heavily used taskprocessors don’t actually use a taskpool, for example:
 stasis/m:cdr:aggregator
 stasis/m:channel:all
 stasis/m:devicestate:all
 stasis/m:manager:core

Is this expected? Wouldn’t using a taskpool for these high-traffic taskprocessors improve efficiency?

jcolp:
Each subscriber decides what they subscribe using, a mailbox or pool. That's not something stasis itself decides. When written these subscriptions were written to use mailboxes, a dedicated direct taskprocessor. I expect using taskpool would not really help for such subscriptions that are under such heavy use, because they would just be bouncing around between taskpool taskprocessors constantly. A dedicated direct taskprocessor doesn't do that, it goes straight to the dedicated task queue and then to the dedicated thread.

alex2grad:
tests/test_stasis.c - many info about threadpool which should be replaced by taskpool or pool

alex2grad:
threadpool  -> taskpool

include/asterisk/stasis.h: * \brief Create a subscription whose callbacks occur on a thread pool
include/asterisk/stasis.h: * Unlike \ref stasis_subscribe, this function will explicitly use a threadpool to

include/asterisk/stasis_message_router.h: * callbacks on a thread pool.

jcolp:
Please create a single issue for these findings.

alex2grad:
https://github.com/asterisk/asterisk/issues/1478

Issue #1343, opened 2025-08-07
[improvement]: cli.c: Remove deprecated/redundant CLI command.


InterLinked1:
### Improvement Description

Remove deprecated/redundant "no debug channel" CLI command.

PR #1344, opened 2025-08-07
cli.c: Remove deprecated and redundant CLI command.

InterLinked1:
The "no debug channel" command has been deprecated since 1.6 (commit 691363656fbdc83edf04b125317aebae6525c9e7), as it is replaced by "core set debug channel", which also supports tab-completion on channels. Remove the redundant command.

Resolves: #1343

UpgradeNote: The deprecated "no debug channel" command has now been removed; use "core set debug channel" instead.


InterLinked1:
cherry-pick-to: none

jcolp (approved):

Issue #1349, opened 2025-08-07
[bug]: Race condition on redirect can cause missing Diversion header


mbradeen:
### Severity

Minor

### Versions

all

### Components/Modules

res_pjsip_diversion

### Operating Environment

all

### Frequency of Occurrence

None

### Issue Description

Due to a race condition on the PJSIP callbacks it is possible for the redirect to occur before the Diversion header is set.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1350, opened 2025-08-07
res_pjsip_diversion: resolve race condition between Diversion header processing and redirect

mbradeen:
Based on the firing order of the PJSIP call-backs on a redirect, it was possible for
the Diversion header to not be included in the outgoing 181 response to the UAC and
the INVITE to the UAS.

This change moves the Diversion header processing to an earlier PJSIP callback while also
preventing the corresponding update that can cause a duplicate 181 response when processing
the header at that time.

Resolves: #1349

### DIFF
--- a/res/res_pjsip_diversion.c
+++ b/res/res_pjsip_diversion.c
@@ -39,6 +39,14 @@ static const pj_str_t diversion_name = { "Diversion", 9 };
 static const pj_str_t history_info_name = { "History-Info", 12 };
 static pj_str_t HISTINFO_SUPPORTED_NAME = { "histinfo", 8 };
 
+/*
+ * Should we queue a frame with the updated redirecting information.
+ */
+typedef enum  {
+	PJSIP_DIVERSION_NOSEND_UPDATE = 0,
+	PJSIP_DIVERSION_SEND_UPDATE = 1,
+} pjsip_diversion_send_update;
+
 /*!
  * \internal
  * \brief Determine if the given string is a SIP token.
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22

res/res_pjsip_diversion.c
```diff
@@ -688,34 +703,48 @@ static void diversion_outgoing_response(struct ast_sip_session *session, pjsip_t
...
 	.outgoing_response = diversion_outgoing_response,
 	.response_priority = AST_SIP_SESSION_BEFORE_MEDIA,
 };
 
+static struct ast_sip_session_supplement diversion_supplement_redirecting = {
+	.method = "INVITE",
+	/* this supplement needs to be called after caller id
+           and after the channel has been created */
+	.priority = AST_SIP_SUPPLEMENT_PRIORITY_CHANNEL + 100,
+	.incoming_request = diversion_incoming_request,
```
jcolp, line 732: I don't see why both supplements need to have everything. Only one needs everything, the other just needs incoming_response since it is invoked at a different response_priority.

mbradeen: my thinking was to have it close to the way it was originally written before the series of changes to try and resolve the 181/missing header issues, but it tests fine without the pre media callbacks, so removed and pushed.

res/res_pjsip_diversion.c
```diff
@@ -385,7 +385,8 @@ static void set_redirecting_reason(pjsip_fromto_hdr *from_info, pjsip_name_addr
 
 static void set_redirecting(struct ast_sip_session *session,
 			    pjsip_fromto_hdr *from_info,
-			    pjsip_name_addr *to_info)
+			    pjsip_name_addr *to_info,
+			    int send_update)
```
seanbright: An enum for this would make the call sites self-documenting.

jcolp (approved):

Issue #1364, opened 2025-08-14
[bug]: bridge.c: BRIDGE_NOANSWER not always obeyed


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

bridge.c

### Operating Environment

N/A

### Frequency of Occurrence

None

### Issue Description

N/A

### Relevant log output

```shell
Even if `BRIDGE_NOANSWER` is set to suppress answering, the variable is not always obeyed.
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

Issue #401, opened 2023-10-30
[bug]: app_dial: Answer Gosub option passthrough regression


InterLinked1:
### Severity

Major

### Versions

21.0.0

### Components/Modules

app_dial

### Operating Environment

Debian 12

### Frequency of Occurrence

Constant

### Issue Description

I seem to have encountered a strange bug that affects Asterisk 21.0.0 but not Asterisk 20.0.0. As of yet, I have not determined a specific cause; it's probably nothing to do with app_dial itself but something internal.

In Asterisk 20, the `U` option works as expected. Answers propagate through channels:

```
[aa-ans]
exten => dial,1,Dial(Local/s@aa-ans/n)
	same => n,Hangup
exten => s,1,Wait(2)
	same => n,Answer()
	same => n,Hangup()

[answer-sub]
exten => s,1,NoOp(Answered)
	same => n,Return()

[aa-call]
exten => s,1,Dial(Local/s@aa-ans/n,,U(answer-sub,s,1))
	same => n,Hangup()
exten => dial,1,Dial(Local/dial@aa-ans/n,,U(answer-sub,s,1))
	same => n,Hangup()
```

The "Answered" NoOp is triggered for both of the following:

```
channel originate Local/s@aa-call application Wait 10
channel originate Local/dial@aa-call application Wait 10
```

However, in Asterisk 21, the second one does not trigger this.

This has broken "dialplan ABI" since dialplan does not behave the same in 20 and 21.

### Asterisk Issue Guidelines

- [X] Yes, I have read the Asterisk Issue Guidelines

jcolp:
What am I expecting to see broken out of this:

`*CLI> channel originate Local/dial@aa-call application Wait 10
*CLI>     -- Called dial@aa-call
    -- Executing [dial@aa-call:1] Dial("Local/dial@aa-call-00000000;2", "Local/dial@aa-ans/n,,U(answer-sub,s,1)") in new stack
    -- Called Local/dial@aa-ans/n
    -- Executing [dial@aa-ans:1] Dial("Local/dial@aa-ans-00000001;2", "Local/s@aa-ans/n") in new stack
    -- Called Local/s@aa-ans/n
    -- Executing [s@aa-ans:1] Wait("Local/s@aa-ans-00000002;2", "2") in new stack
    -- Executing [s@aa-ans:2] Answer("Local/s@aa-ans-00000002;2", "") in new stack
[...]

PR #1365, opened 2025-08-14
bridge.c: Obey BRIDGE_NOANSWER variable to skip answering channel.

InterLinked1:
If the BRIDGE_NOANSWER variable is set on a channel, it is not supposed to answer when another channel bridges to it using Bridge(), and this is checked when ast_bridge_call* is called. However, another path exists (bridge_exec -> ast_bridge_add_channel) where this variable was not checked and channels would be answered. We now check the variable there.

Resolves: #401
Resolves: #1364


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

PR #1368, opened 2025-08-14
xmldoc.c: Fix rendering of CLI output.

gtjoseph:
If you do a `core show application Dial`, you'll see it's kind of a mess.
Indents are wrong is some places, examples are printed in black which makes
them invisible on most terminals, and the lack of line breaks in some cases
makes it hard to follow.

* Fixed the rendering of examples so they are indented properly and changed
the color so they can be seen.
* There is now a line break before each option.
* Options are now printed on their own line with all option content indented
below them.

Example from Dial before fixes:
```
    Example: Dial 555-1212 on first available channel in group 1, searching
    from highest to lowest

    Example: Ringing FXS channel 4 with ring cadence 2

    Example: Dial 555-1212 on channel 3 and require answer confirmation

...

    O([mode]):
        mode - With <mode> either not specified or set to '1', the originator
        hanging up will cause the phone to ring back immediately.
 - With <mode> set to '2', when the operator flashes the trunk, it will ring
 their phone back.
Enables *operator services* mode.  This option only works when bridging a DAHDI
channel to another DAHDI channel only. If specified on non-DAHDI interfaces, it
will be ignored. When the destination answers (presumably an operator services
station), the originator no longer has control of their line. They may hang up,
but the switch will not release their line until the destination party (the
operator) hangs up.

    p: This option enables screening mode. This is basically Privacy mode
    without memory.
```

After:
```
    Example: Dial 555-1212 on first available channel in group 1, searching
    from highest to lowest

     same => n,Dial(DAHDI/g1/5551212)

    Example: Ringing FXS channel 4 with ring cadence 2

     same => n,Dial(DAHDI/4r2)

    Example: Dial 555-1212 on channel 3 and require answer confirmation

     same => n,Dial(DAHDI/3c/5551212)

...

    O([mode]):
        mode - With <mode> either not specified or set to '1', the originator
        hanging up will cause the phone to ring back immediately.
        With <mode> set to '2', when the operator flashes the trunk, it will
        ring their phone back.
        Enables *operator services* mode.  This option only works when bridging
        a DAHDI channel to another DAHDI channel only. If specified on
        non-DAHDI interfaces, it will be ignored. When the destination answers
        (presumably an operator services station), the originator no longer has
        control of their line. They may hang up, but the switch will not
        release their line until the destination party (the operator) hangs up.

    p:
        This option enables screening mode. This is basically Privacy mode
        without memory.
```

There are still things we can do to make this more readable but this is a
start.

### DIFF
--- a/main/xmldoc.c
+++ b/main/xmldoc.c
@@ -94,7 +94,7 @@ static const struct strcolorized_tags {
 	{ "", "", COLOR_YELLOW, "<note>",   "</note>" },
 	{ "", "", COLOR_RED,   "<warning>", "</warning>" },
 	{ "", "", COLOR_WHITE, "<example>", "</example>" },
-	{ "", "", COLOR_GRAY, "<exampletext>", "</exampletext>"},
+	{ "", "", COLOR_WHITE, "<exampletext>", "</exampletext>"},
 };
 
 static const struct strspecial_tags {
@@ -1359,7 +1359,8 @@ static int xmldoc_parse_para(struct ast_xml_node *node, const char *tabs, const 
  * \retval 0 if no example node is parsed.
  * \retval 1 if an example node is parsed.
  */
-static int xmldoc_parse_example(struct ast_xml_node *fixnode, struct ast_str **buffer)
+static int xmldoc_parse_example(struct ast_xml_node *fixnode, const char *tabs,
+	struct ast_str **buffer)
 {
 	struct ast_xml_node *node = fixnode;
 	const char *tmptext;
@@ -1387,9 +1388,11 @@ static int xmldoc_parse_example(struct ast_xml_node *fixnode, struct ast_str **b
 	for (node = ast_xml_node_get_children(node); node; node = ast_xml_node_get_next(node)) {
 		tmptext = ast_xml_get_text(node);
 		if (tmptext) {
-			xmldoc_string_cleanup(tmptext, &stripped_text, 0, 1);
+			const char *skipped = ast_skip_blanks(tmptext);
+			xmldoc_string_cleanup(skipped, &stripped_text, 0, 1);
 			if (stripped_text) {
-				ast_str_append(buffer, 0, "<exampletext>%s</exampletext>\n", ast_str_buffer(stripped_text));
+				ast_str_append(buffer, 0, "\n %s<exampletext>%s</exampletext>\n",
+					tabs, ast_str_buffer(stripped_text));
 				ast_xml_free_text(tmptext);
 				ast_free(stripped_text);
 			}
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

main/xmldoc.c
```diff
@@ -1902,19 +1907,19 @@ static int xmldoc_parse_option(struct ast_xml_node *fixnode, const char *tabs, s
 		return ret;
 	}
 	for (node = ast_xml_node_get_children(fixnode); node; node = ast_xml_node_get_next(node)) {
+		/* if this is the first data appended to buffer, print a \n*/
```
InterLinked1: `\n */` ?

main/xmldoc.c
```diff
@@ -1359,7 +1359,8 @@ static int xmldoc_parse_para(struct ast_xml_node *node, const char *tabs, const
  * \retval 0 if no example node is parsed.
  * \retval 1 if an example node is parsed.
  */
-static int xmldoc_parse_example(struct ast_xml_node *fixnode, struct ast_str **buffer)
+static int xmldoc_parse_example(struct ast_xml_node *fixnode, const char *tabs,
```
InterLinked1, line 1362: I'm not seeing why this needs to be wrapped, the whole line is 87 chars

main/xmldoc.c
gtjoseph: The line was just moved up as is but I'll add the space.

main/xmldoc.c
gtjoseph: Because I happen to use a decent IDE...

Image

InterLinked1: I find premature wrapping like this to make the code harder to read, but I'm sure everyone has a different opinion on this.

Are there any real guidelines in this project on line wrapping? The only example that seems to exist is 90, and this is under that. Personal preference of the commiter?

gtjoseph: Yes, leaning towards clarity for future readers.  I'll add more guidance to the contribution and code style pages shortly.

gtjoseph: > I find premature wrapping like this to make the code harder to read,

Why would it make it harder to read?

InterLinked1: Because now you have to read up and down as well as left and right - also, less code fits on the screen vertically.

I thought somewhere I saw 100-120 as guidance too but can't find that anymore. Maybe it was in the old wiki and not moved over.

gtjoseph: > Because now you have to read up and down as well as left and right - also, less code fits on the screen vertically.

You're lucky we don't require putting every function argument on a separate line like some other projects do. :)

Issue #1373, opened 2025-08-15
[new-feature]: Specify a musiconhold class for transferee on the transferer's channel


viktike:
### Feature Description

I'm having an issue, where the configured music-on-hold class in not applied during a built in feature, attended transfer. During a transfer, it's always the 'default' class. I've tried with setting it in the 'moh_suggest' option for the transferee's PJSIP endpoint, and by CHANNEL(musicclass) on the transferree's channel.

Unfortunately, I was not able fix this issue. I found out, that it has something to do with the bridging framework and when an AST_CONTROL_HOLD frame is pushed onto my simple_bridge. This maybe is a bug, but I was looking for something else actually, where I can set this dynamically on the transfer**er**'s channel.

With the PR, two optional variables built-in to specify a MoH class:
- `BLIND_TRANSFER_MUSIC_CLASS`
- `ATTENDED_TRANSFER_MUSIC_CLASS`
These variables can even be prefixed with __ for inheritance.

When a transfer is initiated, the code will look for one of these depending on which kind of a transfer is requested. It will iterate all the other channels of the bridge and set the required MoH class on that channel. When the variable is not set however, the current operation mode is in effect.

I think this functionality may be possible with custom DYNAMIC_FEATURES, but not sure, and it's way easier to implement.

alex2grad:
@viktike
Did you try setting a custom TRANSFER_CONTEXT and then overriding the transferee’s MUSICCLASS within that context?
That way you can apply a different music-on-hold class dynamically during the transfer.

PR #1374, opened 2025-08-15
features: Ability to set music-on-hold class on the transferer's channel

viktike:
UserNote: When a built-in transfer is initiated,
Asterisk now look for the BLIND_TRANSFER_MUSIC_CLASS
or ATTENDED_TRANSFER_MUSIC_CLASS variables on the
transferer's channel, depending on which kind of transfer
was initiated.

Resolves: #1373


viktike:
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

main/bridge_basic.c
```diff
@@ -3392,7 +3394,25 @@ static int feature_attended_transfer(struct ast_bridge_channel *bridge_channel,
...
+	/* Check for ATTENDED_TRANSFER_MUSIC_CLASS variable */
+	musicclass = pbx_builtin_getvar_helper(bridge_channel->chan, "ATTENDED_TRANSFER_MUSIC_CLASS");
+	if (musicclass != NULL) {
+		ast_bridge_lock(bridge_channel->bridge);
+		AST_LIST_TRAVERSE(&bridge_channel->bridge->channels, other_bridge_channel, entry) {
+			if (other_bridge_channel->chan != bridge_channel->chan) {
+				ast_verb(3, "DTMF attended transfer by '%s' of channel '%s' suggests moh class '%s' by variable, replacing '%s'.\n",
+					ast_channel_name(bridge_channel->chan), ast_channel_name(other_bridge_channel->chan), musicclass, ast_channel_musicclass(other_bridge_channel->chan));
+				ast_channel_lock(other_bridge_channel->chan);
+				ast_channel_musicclass_set(other_bridge_channel->chan, musicclass);
```
jcolp, line 3408: Setting the music class results in changing the music class indefinitely, not just for this occurrence. Depending on what happens with the channel/where it goes this can then result in unexpected behavior with people expecting the music class to be what it was set prior. This may be fine in your environment, but not for others.

I think analysis of the underlying issue should first occur, and then once that is actually understood it may drive this dynamic behavior you seek.

main/bridge_basic.c
```diff
@@ -3392,7 +3394,25 @@ static int feature_attended_transfer(struct ast_bridge_channel *bridge_channel,
 		return 0;
 	}
 
-	ast_bridge_channel_write_hold(bridge_channel, NULL);
+	/* Check for ATTENDED_TRANSFER_MUSIC_CLASS variable */
+	musicclass = pbx_builtin_getvar_helper(bridge_channel->chan, "ATTENDED_TRANSFER_MUSIC_CLASS");
```
jcolp, line 3399: This is not safe. Access to channel variables and their contents are only valid while the channel is locked, and I do not believe it is here.

jcolp (changes requested):

main/bridge_basic.c
viktike: Yes, it wasn't, I've locked the transferer's channel as well.

main/bridge_basic.c
viktike: I understood your concern about the musicclass changed for transferee channel for the lifetime of the channel. The original musicclass of the channel should be stored and restored after the transfer.

> I think analysis of the underlying issue should first occur, and then once that is actually understood it may drive this dynamic behavior you seek.

Unfortunately I'm unable to proceed with that :(
I do remember it working in older versions _(pre bridging framework)_

viktike: > Setting the music class results in changing the music class indefinitely, not just for this occurrence. Depending on what happens with the channel/where it goes this can then result in unexpected behavior with people expecting the music class to be what it was set prior. This may be fine in your environment, but not for others.

I wonder if a UserNote about this would be satisfactory. Without using these new variables, the behavior does not change.

> The original musicclass of the channel should be stored and restored after the transfer.

I've taken a look into this, storing the original musicclass of the transferee's channel is possible, but restoring after the transfer if a different story, especially in the attended transfer case, where there isn't a callback, but a state machine.

kobaz:
You can accomplish this without modifying Asterisk source by setting/reading variables.
Asterisk already tells you via BLIND_TRANSFER channel variable for example.

dub-lin:
pr

asteriskteam:

asteriskteam:

Issue #1375, opened 2025-08-15
[improvement]: dsp.c: Improve logging in tone_detect().


InterLinked1:
### Improvement Description

Improve the logging in tone_detect to make the DSP easier to troubleshoot and understand.

PR #1376, opened 2025-08-15
dsp.c: Improve debug logging in tone_detect().

InterLinked1:
The debug logging during DSP processing has always been kind of overwhelming and annoying to troubleshoot. Simplify and improve the logging in a few ways to aid DSP debugging:

* If we had a DSP hit, don't also emit the previous debug message that was always logged. It is duplicated by the hit message, so this can reduce the number of debug messages during detection by 50%.
* Include the hit count and required number of hits in the message so on partial detections can be more easily troubleshot.
* Use debug level 9 for hits instead of 10, so we can focus on hits without all the noise from the per-frame debug message.
* 1-index the hit count in the debug messages. On the first hit, it currently logs '0', just as when we are not detecting anything, which can be confusing.

Resolves: #1375

### DIFF
--- a/main/dsp.c
+++ b/main/dsp.c
@@ -618,12 +618,14 @@ static int tone_detect(struct ast_dsp *dsp, tone_detect_state_t *s, int16_t *amp
 		tone_energy *= 2.0;
 		s->energy *= s->block_size;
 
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

PR #1377, opened 2025-08-17
app_queue: Add wrapup-as-pause feature for queue member visibility

phoneben:
## Summary
Adds an optional `wrapup-as-pause` feature that automatically pauses queue members during their wrapup time and schedules auto-unpause when the period expires. This provides external monitoring systems with accurate visibility into agent availability during post-call work.

## Problem
Currently, queue members appear "available" to external systems (wallboards, reporting tools) even when they're in wrapup time and cannot take calls. This creates misleading availability metrics and makes it difficult to track actual agent utilization.

## Solution
- **New queue option**: `wrapup-as-pause` (disabled by default)
- **Global default**: `wrapup-as-pause` in `[general]` section
- **Automatic behavior**: Sets member pause reason to "WRAPUP" after call completion
- **Smart scheduling**: Auto-unpause after wrapuptime expires
- **Conflict handling**: Cancels existing wrapup schedules for new calls
- **Safety checks**: Only unpause if still paused for "WRAPUP" reason

## Configuration
```ini
[general]
wrapup-as-pause = no    ; Global default

[queue1]
wrapuptime = 30
wrapup-as-pause = yes    ; Per-queue override

### CODE REVIEW

**Comment by phoneben** [comment]
still incomplete want to get insight 
need alembic updated config file

**Comment by phoneben** [comment]
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

**Comment by jcolp** [comment]
Uhhhhhhhhh, you used AI to write this - didn't you? Or did AI itself write it entirely?

**Comment by phoneben** [comment]
Yes, I used AI to help structure the description and assist with implementation. The feature addresses a real problem our team encountered, and I've thoroughly tested it. Happy to discuss the technical approach or make any needed adjustments

**Comment by jcolp** [comment]
In the future disclose your usage of AI.

How much review did you do yourself? Did you examine it to see if it implemented things already available in Asterisk for lower level things? By "assist with implementation" what do you mean?

File: apps/app_queue.c [hunk 1]
```diff
@@ -1916,6 +1917,16 @@ struct penalty_rule {
 #define ANNOUNCEPOSITION_MORE_THAN 3 /*!< We say "Currently there are more than <limit>" */
 #define ANNOUNCEPOSITION_LIMIT 4 /*!< We not announce position more than \<limit\> */
 
+/* Wrapup-as-pause globals */
+static struct ast_sched_context *wrapup_sched = NULL;
+static ast_mutex_t wrapup_sched_lock;
```
jcolp: AST_MUTEX_DEFINE_STATIC should be used

apps/app_queue.c
```diff
@@ -1916,6 +1917,16 @@ struct penalty_rule {
...
+static int wrapup_as_pause_default = 0;
+
+/* Job tracking */
+struct wrapup_sched_entry; /* forward decl for list head */
+AST_LIST_HEAD_STATIC(scheduled_wrapups, wrapup_sched_entry);
+static ast_mutex_t scheduled_wrapups_lock;
```
jcolp: Here as well

apps/app_queue.c
```diff
@@ -2016,6 +2027,13 @@ struct call_queue {
 	struct queue_ent *head;             /*!< Head of the list of callers */
 	AST_LIST_ENTRY(call_queue) list;    /*!< Next call queue */
 	AST_LIST_HEAD_NOLOCK(, penalty_rule) rules; /*!< The list of penalty rules to invoke */
+	int wrapup_as_pause;                /*!< Expose wrapup as PAUSED to external apps */
+
+
+/*! \brief Data for wrapup auto-unpause callback */
```
jcolp: These are documenting nothing, as in they are in the wrong place

apps/app_queue.c
```diff
@@ -2029,6 +2047,162 @@ static AST_LIST_HEAD_STATIC(rule_lists, rule_list);
 static struct ao2_container *queues;
 
 static void update_realtime_members(struct call_queue *q);
+
+struct wrapup_unpause_data {
+	char *queuename;
+	char *interface;
+	time_t scheduled_time;
+	int ref_count;
```
jcolp: This reimplements reference counting which already exists through the usage of an ao2 object

apps/app_queue.c
```diff
@@ -1916,6 +1917,16 @@ struct penalty_rule {
...
 #define ANNOUNCEPOSITION_LIMIT 4 /*!< We not announce position more than \<limit\> */
 
+/* Wrapup-as-pause globals */
+static struct ast_sched_context *wrapup_sched = NULL;
+static ast_mutex_t wrapup_sched_lock;
+static int wrapup_as_pause_default = 0;
+
+/* Job tracking */
+struct wrapup_sched_entry; /* forward decl for list head */
+AST_LIST_HEAD_STATIC(scheduled_wrapups, wrapup_sched_entry);
```
jcolp: Why does there need to be a linked list? Why is just using a scheduled item insufficient? They can be cancelled.

apps/app_queue.c
```diff
@@ -11934,6 +12222,34 @@ static int load_module(void)
 	struct ast_config *member_config;
 	struct stasis_topic *queue_topic;
 	struct stasis_topic *manager_topic;
+	
+	/* Initialize wrapup-as-pause infrastructure */
+	ast_mutex_init(&wrapup_sched_lock);
+	ast_mutex_init(&scheduled_wrapups_lock);
+
+	ast_mutex_lock(&wrapup_sched_lock);
```
jcolp: Why does this lock? It should be impossible for it to get manipulated

apps/app_queue.c
```diff
@@ -3874,6 +4064,8 @@ static void destroy_queue(void *obj)
 
 static struct call_queue *alloc_queue(const char *queuename)
 {
+	/* default wrapup-as-pause */
```
jcolp: why is this here...

phoneben:
@jcolp 
I wanted wrap-up to reflect **real unavailability**: when an agent finishes a call and enters wrap-up, they shouldn’t look “available” to wallboards or external tools. So this change adds an **opt-in, per-queue** switch that pauses the member with reason `WRAPUP` after a call, then **auto-unpauses** them when the wrap-up window ends—no manual clicks, no stuck states. It’s off by default, and it only activates when `wrapuptime > 0`. We also added a small **offer-time heal** so that if Asterisk restarts during wrap-up (the in-memory timer is lost), the very next time the queue considers that member, it notices the wrap-up window has elapsed and unpauses them safely.&#x20;

# What changed and why (each piece)

* **Scheduler include** – `#include "asterisk/sched.h"`
  Needed to use `ast_sched_add/del` & the scheduler thread for auto-unpause timers.&#x20;

* **Member state (2 small fields)** – `char *wrapup_qname`, `int wrapup_sched_id`
  We snapshot the **exact queue name** that scheduled wrap-up so we unpause only that queue later, and we keep the scheduler **job id** so we can cancel/replace it if another call ends before the timer fires. Prevents unpausing “all queues” or the wrong queue.&#x20;

* **Feature globals** – `wrapup_sched` (scheduler context) and `wrapup_as_pause_default`
  Dedicated scheduler context keeps timers isolated; the global default lets admins enable behavior from `[general]`.&#x20;

* **Queue option** – `wrapup_as_pause` on `struct call_queue`
  Per-queue toggle; keeps behavior opt-in and backward-compatible.&#x20;

* **Guard helper** – `validate_wrapup_as_pause_config(q)`
  Centralizes “enabled **and** effective (`wrapuptime > 0`)” so call-end logic stays readable and we don’t flip pauses for no-op configs.&#x20;

* **Scheduler callback** – `wrapup_unpause_cb(member*)`
  Receives an **AO2-bumped** `member*`. Under lock, it verifies the member is **still** paused for `WRAPUP`, steals & clears `wrapup_qname`, clears `wrapup_sched_id`; outside the lock it unpauses that **one** queue and drops the AO2 ref. This replaces the old global list/locks with a simpler, leak-free flow.&#x20;

* **Availability heal (crash/restart case)** – inline in `is_member_available`
  If `reason_paused=="WRAPUP"` and `now - lastcall >= wrapuptime`, we unpause **this** queue immediately; we also cancel any stale job id and drop its AO2 ref to prevent leaks. No effect during normal operation.&#x20;

* **Member lifecycle hooks** – init & destroy
  On create: set `wrapup_qname=NULL`, `wrapup_sched_id=-1`. On destroy: cancel a pending job (and drop its AO2 ref if deletion succeeds), free `wrapup_qname`. Prevents dangling timers/refs across reloads.&#x20;

* **Config parsing** – `queue_set_param`
  Reads `wrapup-as-pause`, and auto-disables it if `wrapuptime <= 0`; also re-validates if `wrapuptime` changes later. Avoids a switch that can’t do anything.&#x20;

* **Queue defaults** – `alloc_queue()` seeds from `[general]`
  New queues inherit the global default in a predictable way.&#x20;

* **Call-end path** – `update_queue()`
  Under the queue lock we update member stats and **decide** if wrap-up pause applies; after unlocking we:

  1. cancel any older timer (drop its AO2 ref if `ast_sched_del` returns 0),
  2. store `wrapup_qname=strdup(q->name)`,
  3. set `WRAPUP` pause,
  4. schedule the new timer and store `wrapup_sched_id` (or revert the pause if scheduling fails).
     This sequence is race-safe and avoids lock inversions.&#x20;

* **Global defaults** – `queue_set_global_params`
  Reads `[general].wrapup-as-pause` once and logs a hint that `wrapuptime > 0` is required.&#x20;

* **Scheduler lifecycle** – `load_module()` / `unload_module()`
  We **create + start** the scheduler thread on load, and **destroy** it on unload (which cancels any pending jobs). Keeps module reloads clean.&#x20;

* **Docs** – `queues.conf.sample` snippet
  Tells admins what the switch does and that it needs `wrapuptime > 0`.&#x20;

* **Realtime migration** – add `wrapup_as_pause` column
  Makes the Realtime schema reflect the new toggle (bool-ish enum consistent with other flags), so DB-backed queues behave like file-backed queues.&#x20;

jcolp:
Why did you comment with a large AI generated blob of text and tag me with it?

It didn't answer my original questions, which aren't even applicable anymore since this has been changed so many times at this point. Did you just post it to show why the AI thought it should do things?

I'll add that I really really really don't like just having AI output given to me.

How much AI did you use to do this? What involvement did YOU have?

phoneben:
Joshua , you're absolutely right and I apologize.
 I used AI to try to explain the changes better, which completely backfired .
 I spent the day cleaning up the code based on earlier feedback.

> You're right that the original questions don't apply anymore since I've changed the approach significantly.

I eliminated the linked list and custom reference counting like you suggested - now using the scheduler directly with ao2 objects.  

If there are still issues with the current implementation, I'm ready to address them. I genuinely want to learn the right way to do this in Asterisk
> How much AI did you use to do this

I identified the problem (external systems can't see agent availability during wrapup) and figured out the solution approach (pause members with 'WRAPUP' reason, then auto-unpause after wrapuptime expires), trying to stay consistent with previous PR.

I used AI to help implement the technical details - the ao2 reference counting, scheduler integration, and error handling patterns. But I reviewed the code thoroughly today, tested it locally, and made sure I understood how it works before submitting.


That said, this PR is still a work in progress - I want to get your input on the approach before finalizing it. 
The core functionality works (pause with 'WRAPUP' → schedule auto-unpause → ao2 safety), but I have concerns about edge case handling, particularly around scheduler reliability and the healing mechanism.
I'm happy to refactor or take a completely different approach based on your feedback. 
Would rather get the architecture right than polish something that's heading in the wrong direction.
What are your thoughts on this overall direction?

apps/app_queue.c
```diff
@@ -1876,6 +1877,8 @@ struct member {
 	unsigned int delme:1;                /*!< Flag to delete entry on reload */
 	char rt_uniqueid[80];                /*!< Unique id of realtime member entry */
 	unsigned int ringinuse:1;            /*!< Flag to ring queue members even if their status is 'inuse' */
+	char *wrapup_qname;                  /* strdup'd name of the queue that scheduled wrapup */
```
InterLinked1: The comment style here is inconsistent with preceding variables in the struct

apps/app_queue.c
```diff
@@ -2016,6 +2023,7 @@ struct call_queue {
 	struct queue_ent *head;             /*!< Head of the list of callers */
 	AST_LIST_ENTRY(call_queue) list;    /*!< Next call queue */
 	AST_LIST_HEAD_NOLOCK(, penalty_rule) rules; /*!< The list of penalty rules to invoke */
+	int wrapup_as_pause;                /*!< Expose wrapup as PAUSED to external apps */
```
InterLinked1: Is this only ever 0 or 1? Maybe make it an `unsigned int:1`?

apps/app_queue.c
```diff
@@ -2034,6 +2042,46 @@ static int set_member_paused(const char *queuename, const char *interface, const
 static int update_queue(struct call_queue *q, struct member *member, int callcompletedinsl, time_t starttime);
 
 static struct member *find_member_by_queuename_and_interface(const char *queuename, const char *interface);
+/* Wrapup-as-pause forward declarations (must be visible before update_queue) */
+static int	validate_wrapup_as_pause_config(struct call_queue *q)
+{
+	return (q && q->wrapup_as_pause && q->wrapuptime > 0);
+}
+/* Auto-unpause a member after wrapup. Data is an AO2-bumped struct member*. */
```
InterLinked1: Add spaces before each of these functions

apps/app_queue.c
```diff
@@ -2034,6 +2042,46 @@ static int set_member_paused(const char *queuename, const char *interface, const
 static int update_queue(struct call_queue *q, struct member *member, int callcompletedinsl, time_t starttime);
 
 static struct member *find_member_by_queuename_and_interface(const char *queuename, const char *interface);
+/* Wrapup-as-pause forward declarations (must be visible before update_queue) */
+static int	validate_wrapup_as_pause_config(struct call_queue *q)
+{
+	return (q && q->wrapup_as_pause && q->wrapuptime > 0);
```
InterLinked1: The parentheses here aren't necessary

apps/app_queue.c
```diff
@@ -2034,6 +2042,46 @@ static int set_member_paused(const char *queuename, const char *interface, const
...
+}
+/* Auto-unpause a member after wrapup. Data is an AO2-bumped struct member*. */
+static int wrapup_unpause_cb(const void *data)
+{
+	struct member *mem = (struct member *)data;
```
InterLinked1: You don't need to cast from void

apps/app_queue.c
```diff
@@ -2034,6 +2042,46 @@ static int set_member_paused(const char *queuename, const char *interface, const
...
+	char *qname = NULL;
+	int should_unpause = 0;
+
+	if (!mem)
```
InterLinked1: This needs braces around it. Please refer to the coding guidelines: https://docs.asterisk.org/Development/Policies-and-Procedures/Coding-Guidelines/

apps/app_queue.c
```diff
@@ -2034,6 +2042,46 @@ static int set_member_paused(const char *queuename, const char *interface, const
...
+	}
+	if (mem->wrapup_qname) {
+		qname = mem->wrapup_qname;	/* take ownership */
+		mem->wrapup_qname = NULL;
+	}
+	mem->wrapup_sched_id = -1;
+	ao2_unlock(mem);
+
+	/* Do the unpause outside the lock */
+	if (should_unpause && qname)
```
InterLinked1: Ditto here and below on braces

apps/app_queue.c
```diff
@@ -2739,8 +2787,45 @@ static void update_status(struct call_queue *q, struct member *m, const int stat
...
+			now = time(NULL);
+			last = mem->lastcall;
+
+			if (wrapuptime > 0 && last && (now - last) >= wrapuptime) {
+				/* Cancel stale scheduled job (if any) and balance its AO2 ref */
+				if (wrapup_sched && mem->wrapup_sched_id != -1) {
+					rc = ast_sched_del(wrapup_sched, mem->wrapup_sched_id);
+					mem->wrapup_sched_id = -1;
+					if (rc == 0) {
+						/* Job deleted; its callback will NOT run -> drop its ao2_bump */
```
InterLinked1: "drop its ao2_bump" doesn't really make sense to me. "drop its reference" maybe, this... no

apps/app_queue.c
```diff
@@ -2739,8 +2787,45 @@ static void update_status(struct call_queue *q, struct member *m, const int stat
...
+		if (mem->paused &&
+		    mem->reason_paused[0] != '\0' &&
+		    !strcasecmp(mem->reason_paused, "WRAPUP")) {
+
+			wrapuptime = get_wrapuptime(q, mem);
+			now = time(NULL);
+			last = mem->lastcall;
+
+			if (wrapuptime > 0 && last && (now - last) >= wrapuptime) {
+				/* Cancel stale scheduled job (if any) and balance its AO2 ref */
```
InterLinked1: What is "balance its AO2 ref" supposed to mean? Are these also AI-generated comments?

apps/app_queue.c
```diff
@@ -6150,13 +6270,73 @@ static int update_queue(struct call_queue *q, struct member *member, int callcom
...
+			int rc = -1;
+			int id = -1;
+			struct member *mref = NULL;
+
+			/* Cancel any existing wrapup job for this member (balance its ref if we delete it) */
+			if (member->wrapup_sched_id != -1) {
+				rc = ast_sched_del(wrapup_sched, member->wrapup_sched_id);
+				member->wrapup_sched_id = -1;
+				if (rc == 0) {
+					/* Old job will NOT run -> drop the ao2_bump held by that job */
```
InterLinked1: Again, I don't like the wording you are using here... it's convoluted.
I would keep the first and delete all of your comments on locking. The code speaks for itself already there.

apps/app_queue.c
```diff
@@ -11855,6 +12041,13 @@ static struct stasis_forward *topic_forwarder;
 
 static int unload_module(void)
 {
+
+/* Destroy scheduler */
+if (wrapup_sched) {
```
InterLinked1: Your indentation here is messed up

apps/app_queue.c
```diff
@@ -11935,6 +12128,22 @@ static int load_module(void)
 	struct stasis_topic *queue_topic;
 	struct stasis_topic *manager_topic;
 
+/* create + start dedicated scheduler thread */
```
InterLinked1: Your indentation here is also messed up

InterLinked1 (changes requested):

apps/app_queue.c
phoneben: done

apps/app_queue.c
phoneben: fixed

apps/app_queue.c
phoneben: done

apps/app_queue.c
phoneben: done

apps/app_queue.c
phoneben: done

apps/app_queue.c
phoneben: done

apps/app_queue.c
phoneben: it meaning is "release the reference taken at schedule time"
fixed the comments

apps/app_queue.c
phoneben: it meaning is "release the reference taken at schedule time"
fixed the comments

apps/app_queue.c
phoneben: again changed " release the reference taken at schedule time "
also fixes two races by updating wrapup_sched_id and wrapup_qname under the member lock

apps/app_queue.c
phoneben: fixed

apps/app_queue.c
phoneben: i think i fixed this also

apps/app_queue.c
phoneben: i will do more research over the weekend on this

phoneben: 1. **The callback signature is fixed** by Asterisk's scheduler as `int (*)(const void *)`, so you can't change the parameter type.

2. **`const void *` → `struct member *` requires an explicit cast** to drop the `const` qualifier. Without the cast, you'd get a "discarded qualifiers" compiler warning/error.

3. **You need a non-const pointer** because the function modifies the member (setting `wrapup_sched_id = -1`, clearing `wrapup_qname`, etc.).

4. **The cast is safe here** because the original object passed to the scheduler wasn't const - it's just that the API uses `const void *` as a generic parameter type.

So the original code is correct:
```c
struct member *mem = (struct member *)data;
```

jcolp:
Is this actually up to date and reviewable at this point?

alex2grad:
Using pause for wrap-up conflates two different concepts and pollutes pause reporting.
The current patch feels overly complex and increases maintenance burden.
A simpler approach is to check lastcall + wrapuptime (see function is_member_available()), which naturally makes members unavailable during wrap-up without abusing pause.

phoneben:
@alex2grad 

The issue is that checking `lastcall + wrapuptime` in `is_member_available()` only makes members unavailable within the queue itself - it's not visible to external applications. 

I have an auto-dialer that checks if agents are available before dialing out, but when a member is in wrapup, the external application has no way to know this. The member appears available externally even though they're in their wrapup period.

The pause approach was specifically to solve this external visibility problem. Without some mechanism to expose wrapup state outside the queue, external dialers will attempt to use agents who are still in their post-call wrapup time.

What would be the recommended way to make wrapup state visible to external applications that need to check agent availability? I'm open to rewriting this for any good solution that addresses the external visibility requirement.

phoneben:
> Is this actually up to date and reviewable at this point?

doing some testing on a production node  want to see the load and if behaving as described .

alex2grad:
You don’t actually need to overload pause to make wrapup visible externally. There are already mechanisms to expose LastCall and Wrapuptime that external applications can use:

1. AMI Actions QueueStatus
```
Action: QueueStatus
    ActionID: <value>
    Queue: <value>
    Member: <value>

```
The response contains LastCall and Wrapuptime:
```
Event: QueueMember
Queue:
Name:
Location:
StateInterface:
Membership:
Penalty:
CallsTaken:
LastCall:
LastPause:
LoginTime:
InCall:
Status:
Paused:
PausedReason:
Wrapuptime:

```

2. AMI Event QueueMemberStatus Processing
This event also include LastCall and Wrapuptime
```
    Event: QueueMemberStatus
    Queue: <value>
    MemberName: <value>
    Interface: <value>
    StateInterface: <value>
    Membership: <value>
    Penalty: <value>
    CallsTaken: <value>
    LastCall: <value>
    LastPause: <value>
    LoginTime: <value>
    InCall: <value>
    Status: <value>
    Paused: <value>
    PausedReason: <value>
    Ringinuse: <value>
    Wrapuptime: <value>
```

3. Queue Log Processing
You can also track wrapup externally by processing ATTENDEDTRANSFER, BLINDTRANSFER, TRANSFER, COMPLETEAGENT and COMPLETECALLER events in queue_log. For realtime queues, a trigger on queue_log can update LastCall in queue_members so external apps always have the right data.

phoneben:
Good point on AMI—those fields are available today. The concern on my side is scale: in a large deployment with many agents/queues, continuously polling `QueueStatus` or consuming all queue events adds noticeable overhead and complexity to the dialer.

For realtime queues, a lighter alternative would be to **persist the call-completion timestamp** directly to the realtime row:

* **Schema**: add `queue_member.lastcall INT UNSIGNED` (epoch seconds).
* **Write path**: when `mem->lastcall` is set in `update_queue()`, also call `ast_update_realtime("queue_members", "queue_name", q->name, "interface", member->interface, "lastcall", …)`.
* **External logic**: dialer/wallboard can compute wrapup with a simple SQL predicate:

  ```
  (UNIX_TIMESTAMP() - lastcall) < wrapuptime
  ```

This mirrors the same calculation `is_member_available()` already uses, but avoids AMI polling/event handling. No changes to pause semantics, QueueLog, or selection logic inside Asterisk.

If that approach is acceptable, I can guard it behind a config knob (e.g., `persist-member-lastcall = yes|no`, default `no`) so installations that don’t need it incur no extra writes.

alex2grad:
I don’t think persisting lastcall inside Asterisk’s realtime code is the right direction. That functionality belongs in the external application layer, not in Asterisk core.

If you want a persisted lastcall column for your dialer/wallboard, you can simply extend your own queue_member table schema (extra fields won’t interfere with Asterisk) and maintain it externally. One straightforward way is with a SQL trigger on queue_log, for example in MySQL:
```
CREATE TRIGGER queue_log_lastcall AFTER INSERT ON queue_log
FOR EACH ROW
BEGIN
  IF NEW.event IN ('ATTENDEDTRANSFER','BLINDTRANSFER','TRANSFER','COMPLETEAGENT','COMPLETECALLER')
     AND NEW.queuename IS NOT NULL AND NEW.queuename != '' AND NEW.queuename != 'NONE'
     AND NEW.agent IS NOT NULL AND NEW.agent != '' AND NEW.agent != 'NONE'
  THEN
    UPDATE queue_member
       SET lastcall = UNIX_TIMESTAMP()
     WHERE queue_name = NEW.queuename
       AND membername = NEW.agent;
  END IF;
END;
```
This way, you get the persisted lastcall data your external system needs, without adding writes in Asterisk itself or complicating the core schema.

alex2grad:
If the core reviewers are comfortable with persisting lastcall directly in Asterisk realtime, I won’t argue further.

phoneben:
jcolp
what do you say with persisting lastcall directly in Asterisk realtime ?
I prefer it more over using a trigger -less chance of unexpected side effects and easier to troubleshoot.

jcolp:
As long as it is configurable and default behavior is preserved, then I have no problem with it.

phoneben:
Thanks .
will do.
i think i will open a pr for it

jcolp:
Closing this PR per discussion.

Issue #1378, opened 2025-08-18
[bug]: chan_dahdi: dialmode feature is not properly reset between calls


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

chan_dahdi

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

If the dialmode feature is toggled using the CHANNEL function, the change is supposed to be temporary for the channel, but actually "persists" such that in future calls, the normal settings are not in effect. This can lead to the loss of ability to make any calls.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1379, opened 2025-08-18
chan_dahdi: Fix erroneously persistent dialmode.

InterLinked1:
It is possible to modify the dialmode setting in the chan_dahdi/sig_analog private using the CHANNEL function, to modify it during calls. However, it was not being reset between calls, meaning that if, for example, tone dialing was disabled, it would never work again unless explicitly enabled.

This fixes the setting by pairing it with a "perm" version of the setting, as a few other features have, so that it can be reset to the permanent setting between calls. The documentation is also clarified to explain the interaction of this setting and the digitdetect setting more clearly.

Resolves: #1378

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -277,11 +277,16 @@
 			</enum>
 			<enum name="dialmode">
 				<para>R/W Pulse and tone dialing mode of the channel.</para>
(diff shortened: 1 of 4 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

Issue #1380, opened 2025-08-18
[bug]: sig_analog: Segfault due to calling strcmp on NULL


InterLinked1:
### Severity

Major

### Versions

Git

### Components/Modules

sig_analog

### Operating Environment

N/A

### Frequency of Occurrence

Occasional

### Issue Description

It is possible for the channel application to be NULL, which can cause a segfault.

The scenario in which this happened seems to be when a masquerade occurred as a result of calling the `Bridge` application - though this only occurred once out of a handful of times, so this may be sensitive to timing. In any case, we should check if the application is NULL and simply not do anything special in such circumstances.

### Relevant log output

```shell
[09:39:47.120] VERBOSE[984076][C-00000047] pbx.c: Executing [REDACTED@pots-ccsa:18] Bridge("DAHDI/1-1", "DAHDI/25-1") in new stack
[09:39:47.124] VERBOSE[984075][C-00000047] pbx.c: Spawn extension (pots-ccsa, REDACTED, 8) exited non-zero on 'Surrogate/DAHDI/25-1'
[09:39:47.128] VERBOSE[984078][C-00000047] bridge_channel.c: Channel DAHDI/25-1 joined 'simple_bridge' basic-bridge <ed2a6efb-3256-4902-a68d-8be0f6374b82>
[09:39:47.130] VERBOSE[984076][C-00000047] bridge_channel.c: Channel DAHDI/1-1 joined 'simple_bridge' basic-bridge <ed2a6efb-3256-4902-a68d-8be0f6374b82>
[09:39:47.133] VERBOSE[984077][C-00000047] app_chanspy.c: Done Spying on channel DAHDI/1-1
[09:39:47.133] VERBOSE[984077][C-00000047] pbx.c: Spawn extension (pots-ccsa-spy, s, 2) exited non-zero on 'Local/s@pots-ccsa-spy-00000022;2'
[09:39:47.134] VERBOSE[982274] cdr_adaptive_odbc.c: CDR column 'number' was not set and does not match filter of !''.  Cancelling this CDR.
<-- crash -->
[09:47:05.010]


gdb) print p->subs[0].owner
$4 = (struct ast_channel *) 0x7f63cc1b3440
(gdb) print p->subs[0].owner.appl
$5 = 0x0


Thread 1 (Thread 0x7f63e79826c0 (LWP 984078)):
#0  __strcmp_sse42 () at ../sysdeps/x86_64/multiarch/strcmp-sse4_2.S:227
No locals.
[...]

PR #1381, opened 2025-08-18
sig_analog: Fix SEGV due to calling strcmp on NULL.

InterLinked1:
Add an additional check to guard against the channel application being NULL.

Resolves: #1380


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

Issue #1384, opened 2025-08-19
[bug]: chan_websocket: asterisk crashes on hangup after STOP_MEDIA_BUFFERING command with id


SvenKube:
### Severity

Trivial

### Versions

master

### Components/Modules

chan_websocket

### Operating Environment

linux

### Frequency of Occurrence

Frequent

### Issue Description

When I send a `STOP_MEDIA_BUFFERING` command over the websocket with an id, e.g. `foo`, asterisk crashes when my call is hung up. In my tests this issue only occurred when the passed id has more than two characters, so `fo` does not trigger the crash on hangup. 

In my test scenario i have sipp make a call. My asterisk uses the following Dialplan:
```
[from-sipp]
exten => 1000,1,NoOp(SIPP call received)
 same => n,Dial(WebSocket/connection1/c(slin))
```

My `websocket_client.conf` has one connection configured:
```
[connection1]
type = websocket_client
uri = ws://localhost:8787
protocols = media
connection_type = per_call_config
connection_timeout = 500
reconnect_interval = 500
reconnect_attempts = 5
tls_enabled = no
```

Sample Websocket-Server in go that sends the command:
```go
package main

import (
	"log"
	"net/http"
	"time"

	"github.com/gorilla/websocket"
)

func main() {
	http.HandleFunc("/", func(w http.ResponseWriter, r *http.Request) {
		upgrader := websocket.Upgrader{}
		wsConn, _ := upgrader.Upgrade(w, r, nil)

		log.Println("Received new websocket connection")

		time.Sleep(2 * time.Second)

		wsConn.WriteMessage(websocket.TextMessage, []byte("START_MEDIA_BUFFERING"))

		time.Sleep(1 * time.Second)

		wsConn.WriteMessage(websocket.TextMessage, []byte("STOP_MEDIA_BUFFERING foo"))
		log.Println("Sent stop media buffering command.")
	})
	http.ListenAndServe(":8787", nil)
}

```

### Relevant log output

```shell
*CLI> malloc(): invalid size (unsorted)
fish: Job 1, 'sudo asterisk -C $PWD/asterisk.…' terminated by signal SIGABRT (Abort)

or

*CLI> realloc(): invalid pointer
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
I'm investigating.

gtjoseph:
Confirmed.  I'm working on a fix.

[...]

PR #1385, opened 2025-08-19
chan_websocket: Fix buffer overrun when processing TEXT websocket frames.

gtjoseph:
ast_websocket_read() receives data into a fixed 64K buffer then continually
reallocates a final buffer that, after all continuation frames have been
received, is the exact length of the data received and returns that to the
caller.  process_text_message() in chan_websocket was attempting to set a
NULL terminator on the received payload assuming the payload buffer it
received was the large 64K buffer.  The assumption was incorrect so when it
tried to set a NULL terminator on the payload, it could, depending on the
state of the heap at the time, cause heap corruption.

process_text_message() now allocates its own payload_len + 1 sized buffer,
copies the payload received from ast_websocket_read() into it then NULL
terminates it prevent the possibility of the overrun and corruption.

Resolves: #1384

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -421,12 +421,14 @@ static int process_text_message(struct websocket_pvt *instance,
 	}
 
 	/*
-	 * This is safe because the payload buffer is always >= 8K
-	 * even with LOW_MEMORY defined and we've already made sure the
-	 * command is less than 128 bytes.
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

mbradeen (approved):

Issue #1390, opened 2025-08-22
[improvement]: res_tonedetect: Add option to automatically end detection in TONE_DETECT


InterLinked1:
### Improvement Description

Add an option to auto-disable detection after the required matches, which can help prevent race conditions in the dialplan from trying to manually accomplish this.

PR #1391, opened 2025-08-22
res_tonedetect: Add option for TONE_DETECT detection to auto stop.

InterLinked1:
One of the problems with TONE_DETECT as it was originally written is that if a tone is detected multiple times, it can trigger the redirect logic multiple times as well. For example, if we do an async goto in the dialplan after detecting a tone, because the detector is still active until explicitly disabled, if we detect the tone again, we will branch again and start executing that dialplan a second time. This is rarely ever desired behavior, and can happen if the detector is not removed quickly enough.

Add a new option, 'e', which automatically disables the detector once the desired number of matches have been heard. This eliminates the potential race condition where previously the detector would need to be disabled immediately, but doing so quickly enough was not guaranteed. This also allows match criteria to be retained longer if needed, so the detector does not need to be destroyed prematurely.

Resolves: #1390

UserNote: The 'e' option for TONE_DETECT now allows detection to be disabled automatically once the desired number of matches have been fulfilled, which can help prevent race conditions in the dialplan, since TONE_DETECT does not need to be disabled after a hit.

### DIFF
--- a/res/res_tonedetect.c
+++ b/res/res_tonedetect.c
@@ -214,13 +214,16 @@
 					<option name="d">
 						<para>Custom decibel threshold to use. Default is 16.</para>
 					</option>
+					<option name="e">
+						<para>Automatically end detection after desired number of hits (as specified in the <literal>n</literal> option.</para>
+					</option>
 					<option name="g">
 						<para>Go to the specified context,exten,priority if tone is received on this channel.
-						Detection will not end automatically.</para>
+						Detection will not end automatically, unless the <literal>e</literal> option is also specified.</para>
 					</option>
 					<option name="h">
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22

jcolp (approved):

InterLinked1:
> Cherry-picks to 20 and 21 failed.

What does this mean exactly? They need to be manually cherry picked as separate PRs?

jcolp:
Yes.

InterLinked1:
Separate PR for 20/21 added here: https://github.com/asterisk/asterisk/pull/1404

Issue #1394, opened 2025-08-25
[improvement]: sig_analog: Skip Caller ID spill if Caller ID is disabled


InterLinked1:
### Improvement Description

If `usecallerid=no` for an FXS port, then don't send any Caller ID spill on the line (there isn't anything that can or should be sent anyways)

PR #1395, opened 2025-08-25
sig_analog: Skip Caller ID spill if usecallerid=no.

InterLinked1:
If Caller ID is disabled for an FXS port, then we should not send any Caller ID spill on the line, as we have no Caller ID information that we can/should be sending.

Resolves: #1394


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

PR #1407, opened 2025-08-28
res_pjsip_authenticator_digest: Fix SEGV if get_authorization_hdr returns NULL.

gtjoseph:
In the highly-unlikely event that get_authorization_hdr() couldn't find an
Authorization header in a request, trying to get the digest algorithm
would cauase a SEGV.  We now check that we have an auth header that matches
the realm before trying to get the algorithm from it.

Resolves: #GHSA-64qc-9x89-rx5j


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

PR #1409, opened 2025-08-29
res_pjsip_header_funcs.c: Add new PJSIP_HANGUP_HEADER(S) functions

IgorrG:
These new functions allow retrieving information from headers in
BYE message of add additional headers in outbound BYE. Implementation
based on existing PJSIP_HEADER and PJSIP_RESPONSE_HEADER
implementation. Functions could be used for better integration with
SIP proxies and sending custom call data to remote party in SIP on
call hangup.

UserNote: Functions PJSIP_HANGUP_HEADERS() and PJSIP_HANGUP_HEADER()
added. To get headers from incoming BYE functions could be called from
the hangup handler. To set headers use add or update options before call
hangup.

### DIFF
--- a/res/res_pjsip_header_funcs.c
+++ b/res/res_pjsip_header_funcs.c
@@ -263,6 +263,95 @@
 			<ref type="function">PJSIP_HEADERS</ref>
 		</see-also>
 	</function>
+	<function name="PJSIP_HANGUP_HEADER" language="en_US">
+		<synopsis>
+			Gets headers from an inbound PJSIP channel. Adds, updates or removes the
+			specified SIP header from an outbound PJSIP channel.
+		</synopsis>
+		<syntax>
+			<parameter name="action" required="true">
+				<enumlist>
+					<enum name="read"><para>Returns instance <replaceable>number</replaceable>
+					of header <replaceable>name</replaceable>. A <literal>*</literal>
+					may be appended to <replaceable>name</replaceable> to iterate over all 
+					headers <emphasis>beginning with</emphasis> <replaceable>name</replaceable>.
+					</para></enum>
+
+					<enum name="add"><para>Adds a new header <replaceable>name</replaceable>
+					to this session.</para></enum>
+
+					<enum name="update"><para>Updates instance <replaceable>number</replaceable>
+					of header <replaceable>name</replaceable> to a new value.
+					The header must already exist.</para></enum>
+
+					<enum name="remove"><para>Removes all instances of previously added headers
+					whose names match <replaceable>name</replaceable>. A <literal>*</literal>
+					may be appended to <replaceable>name</replaceable> to remove all headers
+					<emphasis>beginning with</emphasis> <replaceable>name</replaceable>.
+					<replaceable>name</replaceable> may be set to a single <literal>*</literal>
+					to clear <emphasis>all</emphasis> previously added headers. In all cases,
+					the number of headers actually removed is returned.</para></enum>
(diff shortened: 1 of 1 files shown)


IgorrG:
This feature 100% based on existing PJSIP_HEADER and PJSIP_RESPONSE_HEADER implementation.

gtjoseph (dismissed):
The cherry-pick-to lines must be in a comment by themselves and must include the 23 branch.

You should also add a UserNote so the new capability shows up at the top of the ChangeLog.

See:

IgorrG:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

gtjoseph:
@IgorrG You need to fix up the PR description to match the commit message.  Just copy and paste.

IgorrG:
> @IgorrG You need to fix up the PR description to match the commit message. Just copy and paste.

Thanks. Updated.

alex2grad:
What about the CANCEL method?

CANCEL is used when the caller hangs up before the callee answers.

With the current naming (PJSIP_HANGUP_HEADERS() / PJSIP_HANGUP_HEADER()), it isn’t clear whether these functions cover only BYE or also CANCEL.

To avoid confusion, maybe:
Rename functions to PJSIP_BYE_HEADER(S) if the scope is explicitly BYE only.
Or, if feasible, extend support to CANCEL as well, so the functions really apply to all hangup scenarios.

IgorrG:
> What about the CANCEL method?
> CANCEL is used when the caller hangs up before the callee answers.

Yes, we think about this, but decided that 'HANGUP' name fit the case, because often used of already established call being disconnected. Neither PJSIP_HEADER nor PJSIP_RESPONSE_HEADER function use SIP method name in its name also.

jcolp:
> > What about the CANCEL method?
> > CANCEL is used when the caller hangs up before the callee answers.
> 
> Yes, we think about this, but decided that 'HANGUP' name fit the case, because often used of already established call being disconnected. Neither PJSIP_HEADER nor PJSIP_RESPONSE_HEADER function use SIP method name in its name also.

We keep accepting more functions for specific situations. I think we've reached the point where functionality needs to be defined and implemented which allows access to headers for every point of the SIP exchange, and to add headers to them. Perhaps a new dialplan function that supersedes the others (though they won't be removed or deprecated) that allows type (request or response) to be specified, and the SIP method or SIP response code alongside the header name (and value for setting). A count function could also be provided to return how many messages there are matching, and a value for which one to use could be provided to the get function (it could also accept first or last as an argument).

Thoughts?

IgorrG:
> > > What about the CANCEL method?
> > > CANCEL is used when the caller hangs up before the callee answers.
> > 
> > 
> > Yes, we think about this, but decided that 'HANGUP' name fit the case, because often used of already established call being disconnected. Neither PJSIP_HEADER nor PJSIP_RESPONSE_HEADER function use SIP method name in its name also.
> 
> We keep accepting more functions for specific situations. I think we've reached the point where functionality needs to be defined and implemented which allows access to headers for every point of the SIP exchange, and to add headers to them. Perhaps a new dialplan function that supersedes the others (though they won't be removed or deprecated) that allows type (request or response) to be specified, and the SIP method or SIP response code alongside the header name (and value for setting). A count function could also be provided to return how many messages there are matching, and a value for which one to use could be provided to the get function (it could also accept first or last as an argument).
> 
> Thoughts?

100% agree

PR #1410, opened 2025-09-02
app_queue.c: Add new global 'log_unpause_on_reason_change'

IgorrG:
In many asterisk-based systems, the pause reason is used to separate
pauses by type,and logically, changing the reason defines two intervals
that should be accounted for separately. The introduction of a new
option allows me to separate the intervals of operator inactivity in
the log by the event of unpausing.

UserNote: Add new global option 'log_unpause_on_reason_change' that
is default disabled. When enabled cause addition of UNPAUSE event on
every re-PAUSE with reason changed.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -1742,6 +1742,9 @@ static int force_longest_waiting_caller;
 /*! \brief queues.conf [general] option */
 static int log_caller_id_name; 
 
+/*! \brief queues.conf [general] option */
+static int log_unpause_on_reason_change;
+
 /*! \brief name of the ringinuse field in the realtime database */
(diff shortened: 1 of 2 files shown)


gtjoseph (dismissed):
The cherry-pick-to lines need to be in a comment by themselves and include the 23 branch.

A UserNote should be added to highlight the change.

IgorrG:
@gtjoseph thanks for the comment, missed UserNote information in guidelines and 23 branch creation. Will fix that.

IgorrG:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

Issue #1401, opened 2025-08-27
[bug]: app_waitfornoise timeout is always less then configured because of time() usage


IgorrG:
### Severity

Minor

### Versions

22.5.1

### Components/Modules

app_waitfornoise

### Operating Environment

Debian 12

### Frequency of Occurrence

Constant

### Issue Description

Applications WaitForSilence() and WaitForNoise() uses time() function to calculate application runtime and timeout logic. In that case real timeout would be always less the specified value. In our case we found this issue when use application with 1 second timeout and found cases when it return immediately after start.

Timeout calculation should use miliseconds even in cases timeout configured in seconds.

### Relevant log output

```shell
[Aug 26 13:43:54] VERBOSE[114999][C-0048afad] pbx.c: Executing [s@announce:3] WaitForNoise("PJSIP/advans-term-00914201", "200,,1") in new stack
[Aug 26 13:43:54] VERBOSE[114999][C-0048afad] app_waitforsilence.c: Waiting 1 time(s) for 200ms of noise with 1s timeout
[Aug 26 13:43:55] VERBOSE[114999][C-0048afad] pbx.c: Executing [s@announce:4] NoOp("PJSIP/dvans-term-00914201", "TIMEOUT") in new stack
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1411, opened 2025-09-04
app_waitforsilence.c: Use milliseconds to calculate timeout time

IgorrG:
The functions WaitForNoise() and WaitForSilence() use the time() functions to calculate elapsed time, which causes the timer to fire on a whole second boundary, and the actual function execution time to fire the timer may be 1 second less than expected. This fix replaces time() with ast_tvnow().

Fixes: #1401

### DIFF
--- a/apps/app_waitforsilence.c
+++ b/apps/app_waitforsilence.c
@@ -158,7 +158,7 @@ static const struct wait_type wait_for_noise = {
 	.func = ast_dsp_noise,
 };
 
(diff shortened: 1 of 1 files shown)


IgorrG:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

jcolp (approved):

PR #1412, opened 2025-09-04
func_hangupcause.c: Add access to Reason headers via HANGUPCAUSE()

IgorrG:
As soon as SIP call may end with several Reason headers, we want to make all of them available through the HAGUPCAUSE() function. This implementation uses the same ao2 hash for cause codes storage and adds a flag to make difference between last processed sip message and content of reason headers.

UserNote: Added a new option to HANGUPCAUSE to access additional information about hangup reason. Reason headers from pjsip could be read using 'tech_extended' cause type.

### DIFF
--- a/funcs/func_hangupcause.c
+++ b/funcs/func_hangupcause.c
@@ -55,6 +55,7 @@
 				<para>Parameter describing which type of information is requested. Types are:</para>
 				<enumlist>
 					<enum name="tech"><para>Technology-specific cause information</para></enum>
+					<enum name="tech_extended"><para>Technology-specific extended list of cause information</para></enum>
 					<enum name="ast"><para>Translated Asterisk cause code</para></enum>
 				</enumlist>
 			</parameter>
@@ -119,9 +120,10 @@ static int hangupcause_read(struct ast_channel *chan, const char *cmd, char *dat
 	char *parms;
 	struct ast_control_pvt_cause_code *cause_code;
 	int res = 0;
+	struct ast_str *causelist;
 	AST_DECLARE_APP_ARGS(args,
 		AST_APP_ARG(channel);	/*!< Channel name */
-		AST_APP_ARG(type);	/*!< Type of information requested (ast or tech) */
+		AST_APP_ARG(type);	/*!< Type of information requested (ast, tech or tech_extended) */
 		);
 
 	/* Ensure that the buffer is empty */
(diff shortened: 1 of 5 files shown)


IgorrG:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

funcs/func_hangupcause.c
```diff
@@ -139,25 +141,50 @@ static int hangupcause_read(struct ast_channel *chan, const char *cmd, char *dat
 		return -1;
 	}
 
-	ast_channel_lock(chan);
-	cause_code = ast_channel_dialed_causes_find(chan, args.channel);
-	ast_channel_unlock(chan);
+//	ast_log(LOG_ERROR, "Cause for channel '%s'\n", args.channel);
```
jcolp: This should be removed

funcs/func_hangupcause.c
```diff
@@ -139,25 +141,50 @@ static int hangupcause_read(struct ast_channel *chan, const char *cmd, char *dat
...
 
-	if (!cause_code) {
-		ast_log(LOG_WARNING, "Unable to find information for channel %s\n", args.channel);
+	if (!(causelist = ast_str_create(128))) {
```
jcolp: These days we do this on separate lines. First line ast_str_create, second the check.

funcs/func_hangupcause.c
```diff
@@ -139,25 +141,50 @@ static int hangupcause_read(struct ast_channel *chan, const char *cmd, char *dat
...
+		ast_log(LOG_ERROR, "Unable to allocate dynamic string buffer\n");
```
jcolp: A user will see this, so the error message should be useful and explain what will happen and the channel involved such as:

Unable to allocate cause list storage for <channel name>, cause information will be unavailable

funcs/func_hangupcause.c
```diff
@@ -139,25 +141,50 @@ static int hangupcause_read(struct ast_channel *chan, const char *cmd, char *dat
...
 		ast_copy_string(buf, cause_code->code, len);
+	} else if (!strcmp(args.type, "tech_extended")) {
+		ast_copy_string(buf, ast_str_buffer(causelist), len);
 	} else {
 		ast_log(LOG_WARNING, "Information type not recognized (%s)\n", args.type);
 		res = -1;
 	}
 
-	ao2_ref(cause_code, -1);
+	if (cause_code) {
```
jcolp, line 184: There's an ao2_cleanup for this, you just need to replace the ao2_ref call with it

include/asterisk/channel.h
```diff
@@ -4446,6 +4446,23 @@ struct ast_str *ast_channel_dialed_causes_channels(const struct ast_channel *cha
  */
 struct ast_control_pvt_cause_code *ast_channel_dialed_causes_find(const struct ast_channel *chan, const char *chan_name);
 
+/*!
+ * \brief Retrieve a ref-counted cause code information structure iterator
+ *
```
jcolp, line 4452: \since version should be added

jcolp (changes requested):

include/asterisk/channel.h
IgorrG: Should I point some versions here right now? I saw comment in Code guidelines, but no information about at what point released versions updated.

jcolp: We're currently in release candidates, so it would be the next major version after these. 20.17.0, 22.8.0, 23.1.0

IgorrG: Done

PR #1414, opened 2025-09-04
chan_websocket.c: Add DTMF messages

jcbjoe:
Added DTMF messages to the chan_websocket feature.

When a user presses DTMF during a call over chan_websocket it will send a message like:
"DTMF PRESSED digit:1"


jcbjoe:
CLA Has been signed. It doesn't seem to be updating for some reason

jcolp:
Probably because your commits are from a different user.

jcbjoe:
> Probably because your commits are from a different user.

I see, Just another member in our team. Do they need to submit a CLA too?

jcolp:
The author of the code should sign the CLA.

PR #1415, opened 2025-09-04
chan_websocket.c: Add DTMF messages

jcbjoe:
Added DTMF messages to the chan_websocket feature.

When a user presses DTMF during a call over chan_websocket it will send a message like:
"DTMF_END digit:1"

Resolves: https://github.com/asterisk/asterisk-feature-requests/issues/70

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -156,6 +156,7 @@ struct websocket_pvt {
 #define QUEUE_DRAINED "QUEUE_DRAINED"
 #define DRIVER_STATUS "STATUS"
 #define MEDIA_BUFFERING_COMPLETED "MEDIA_BUFFERING_COMPLETED"
+#define DTMF_END "DTMF_END"
 
 #define QUEUE_LENGTH_MAX 1000
 #define QUEUE_LENGTH_XOFF_LEVEL 900
@@ -168,6 +169,7 @@ static int webchan_call(struct ast_channel *ast, const char *dest, int timeout);
 static struct ast_frame *webchan_read(struct ast_channel *ast);
 static int webchan_write(struct ast_channel *ast, struct ast_frame *f);
 static int webchan_hangup(struct ast_channel *ast);
+static int webchan_send_dtmf_text(struct ast_channel *ast, char digit, unsigned int duration);
 
 static struct ast_channel_tech websocket_tech = {
 	.type = "WebSocket",
@@ -177,6 +179,7 @@ static struct ast_channel_tech websocket_tech = {
 	.read = webchan_read,
 	.write = webchan_write,
 	.hangup = webchan_hangup,
+	.send_digit_end = webchan_send_dtmf_text,
 };
 
 static void set_channel_format(struct websocket_pvt * instance,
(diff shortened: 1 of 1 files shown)


jcbjoe:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

jcbjoe:
All checklist items should be resolved.

channels/chan_websocket.c
```diff
@@ -1410,6 +1413,30 @@ static int webchan_hangup(struct ast_channel *ast)
...
+{
+    	struct websocket_pvt *instance = ast_channel_tech_pvt(ast);
+    	char *command;
+    	int res = 0;
+    	
+    	if (!instance) {
+	    	return -1;
+    	}
+
+    	res = ast_asprintf(&command, "%s digit:%c", DTMF_PRESSED, digit);
```
gtjoseph: You have a mix of spaces and tabs for indentation here.  Needs to be just tabs.

gtjoseph: You need to free command before returning.

channels/chan_websocket.c
```diff
@@ -156,6 +156,7 @@ struct websocket_pvt {
 #define QUEUE_DRAINED "QUEUE_DRAINED"
 #define DRIVER_STATUS "STATUS"
 #define MEDIA_BUFFERING_COMPLETED "MEDIA_BUFFERING_COMPLETED"
+#define DTMF_PRESSED "DTMF_PRESSED"
```
gtjoseph: Instead of DTMF_PRESSED how about DTMF_END since the frame that triggers the callback is the AST_FRAME_DTMF_END type.

gtjoseph (changes requested):
I'll be testing more next week but here are a few things to fix in teh mean time.

channels/chan_websocket.c
jcbjoe: Yep, makes sense! Changed to DTMF_END

channels/chan_websocket.c
jcbjoe: My bad, I was switching ides to make this pull and it must have messed up the spacing. I have converted them all to tabs now!

channels/chan_websocket.c
```diff
@@ -1410,6 +1413,33 @@ static int webchan_hangup(struct ast_channel *ast)
...
+		int res = 0;
+
+		if (!instance) {
+			return -1;
+		}
+
+		res = ast_asprintf(&command, "%s digit:%c", DTMF_END, digit);
+		if (res <= 0 || !command) {
+			ast_log(LOG_ERROR, "%s: Failed to create DTMF_END\n", ast_channel_name(instance->channel));
+			ast_free(command);
```
gtjoseph: Don't call ast_free here.  If ast_asprintf fails, the content of command is "undefined" so freeing it may create a segfault.

gtjoseph (changes requested):

channels/chan_websocket.c
jcbjoe: Removed this, I was honestly debating whether or not to add it. Thanks for confirming.

jcolp (approved):

PR #1431, opened 2025-09-08
manager.c: Fix presencestate object leak

azertyfun:
ast_presence_state allocates subtype and message. We straightforwardly need to clean those up.

### DIFF
--- a/main/manager.c
+++ b/main/manager.c
@@ -5524,6 +5524,9 @@ static int action_presencestate(struct mansession *s, const struct message *m)
 	}
 	astman_append(s, "\r\n");
 
+	ast_free(subtype);
+	ast_free(message);
+
 	return 0;
 }
 


azertyfun:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

azertyfun:
Mhm, the test failing is actually indicative of a logic error on my part. Currently investigating what I did wrong and how to fix this observed leak properly.

gtjoseph:
> Mhm, the test failing is actually indicative of a logic error on my part. Currently investigating what I did wrong and how to fix this observed leak properly.

@azertyfun That test has been failing on other PR's lately.  Don't assume it's your fault. :)

I'm going to trigger a recheck so we'll see what happens.

azertyfun:
@gtjoseph good to hear :D

~~I'll keep it in draft for a little bit nonetheless, internal testing has brought up some weird behaviors but it might not be related to this particular change. I'll double check everything's fine on my end then un-draft.~~

gtjoseph:
@azertyfun Yeah, the geolocation failure was a transient failure not caused by your change.

azertyfun:
I've confirmed that the issues I was experiencing on my side were unrelated.

Just FYI, I have another memleak fix for Asterisk 18 but since that isn't supported anymore I won't make a PR here: https://github.com/azertyfun/asterisk/commit/b6f5bfaeab1afeb307ccaf35f73ed62c7748dace

However the reason the leak isn't impacting asterisk 20+ seems to be an accident (the function `ast_http_parse_post_form` was introduced which copies the reference before operating on it). It's probably fine and not going to break again but now you know :)

Issue #1340, opened 2025-08-07
[bug]: comfort noise packet corrupted


moovoo:
### Severity

Minor

### Versions

18.26.3,20.15.1,22.5.1

### Components/Modules

res_rtp_asterisk

### Operating Environment

Debian Bookworm x86_64

### Frequency of Occurrence

Constant

### Issue Description

RTP comfort noise sent by Asterisk appears to be corrupted starting 18.25.0 ([not reproducible on 18.24.3](https://github.com/asterisk/asterisk/compare/18.24.3...18.25.0)) and on 20.15.1, 22.5.1

pjsip.conf:

```
...
type=endpoint
rtp_keepalive=1
disallow=all
allow=ulaw
...

```
extensions.conf:
```
[test]
exten => _X.,1,Answer()
 same =>     n,Wait(30)
 same =>     n,Hangup()

```

[rtp_cn_18.tar.gz](https://github.com/user-attachments/files/21647525/rtp_cn_18.tar.gz)

there ff ff where normal RTP header is supposed to be 

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

jcolp:
Is this using Asterisk as provided by the project, or Debian packages?

Looking at the res_rtp_asterisk changes between the two 18 versions presented shows only a change to DTMF payload handling, no other changes.

moovoo:
built from source.

attaching dockerfile which I used to reproduce the issue

[ast_cn.tar.gz](https://github.com/user-attachments/files/21666254/ast_cn.tar.gz)

```shell
docker build -t ast18_24_3 --build-arg asterisk_version=18.24.3 .

docker build -t ast18_25_0 --build-arg asterisk_version=18.25.0 .

# start tcpdump

docker run -it --rm --net=host ast18_24_3

# make 1st call

docker run -it --rm --net=host ast18_25_0

# make 2nd call

# stop tcpdump

```

[test_calls.tar.gz](https://github.com/user-attachments/files/21666300/test_calls.tar.gz)

jcolp:
And unpatched?

moovoo:
no patches, just pure tar from downloads.asterisk.org and a few non default conf files to configure pjsip/extensions/etc

moovoo:
I think the problem is that starting with 18.25.0 payload is -1 or 0xffffffff here:
```c
[...]

PR #1433, opened 2025-09-09
rtp_engine.c: Add exception for comfort noise payload.

bkford:
In a previous commit, a change was made to
ast_rtp_codecs_payload_code_tx_sample_rate to check for differing sample
rates. This ended up returning an invalid payload int for comfort noise.
A check has been added that returns early if the payload is in fact
supposed to be comfort noise.

Fixes: #1340

### DIFF
--- a/main/rtp_engine.c
+++ b/main/rtp_engine.c
@@ -2135,6 +2135,16 @@ int ast_rtp_codecs_payload_code_tx_sample_rate(struct ast_rtp_codecs *codecs, in
 		payload = find_static_payload_type(asterisk_format, format, code);
 		ast_rwlock_unlock(&static_RTP_PT_lock);
 
+		/*
+		 * Comfort noise is NOT used as an SDP negotiated format within Asterisk;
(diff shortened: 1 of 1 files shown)


bkford:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

main/rtp_engine.c
```diff
@@ -2135,6 +2135,11 @@ int ast_rtp_codecs_payload_code_tx_sample_rate(struct ast_rtp_codecs *codecs, in
 		payload = find_static_payload_type(asterisk_format, format, code);
 		ast_rwlock_unlock(&static_RTP_PT_lock);
 
+		/* This is a special case for comfort noise */
```
jcolp: This comment needs to be expanded further.

"Comfort noise is NOT used as an SDP negotiated format within Asterisk, instead it is used when it is not negotiated. This special case allows its payload to be returned when not negotiated allowing keep alive to
function as expected."

jcolp (approved):

Issue #1434, opened 2025-09-09
[improvement]: pbx_variables: Create real channel for dialplan eval CLI command


InterLinked1:
### Improvement Description

It is possible for crashes to occur when using the dialplan eval CLI command depending on the function called, since the CLI command uses a dummy channel, which lacks properties that are usually expected of channels executing dialplan. Rather than add additional checks to every function to work around this, we simply use a real channel to make the command automatically safe / avoid NULL dereferences, etc. for any function that could be called.

PR #1435, opened 2025-09-09
pbx_variables.c: Create real channel for "dialplan eval function".

InterLinked1:
"dialplan eval function" has been using a dummy channel for function evaluation, much like many of the unit tests. However, sometimes, this can cause issues for functions that are not expecting dummy channels. As an example, ast_channel_tech(chan) is NULL on such channels, and ast_channel_tech(chan)->type consequently results in a NULL dereference. Normally, functions do not worry about this since channels executing dialplan aren't dummy channels.

While some functions are better about checking for these sorts of edge cases, use a real channel with a dummy technology to make this CLI command inherently safe for any dialplan function that could be evaluated from the CLI.

Resolves: #1434

### DIFF
--- a/main/pbx_variables.c
+++ b/main/pbx_variables.c
@@ -39,6 +39,7 @@
 #include "asterisk/module.h"
 #include "asterisk/paths.h"
 #include "asterisk/pbx.h"
+#include "asterisk/format_cache.h"
 #include "asterisk/stasis_channels.h"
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

PR #1151, opened 2025-03-07
config.c: Fix inconsistent pointer logic in ast_variable_update.

InterLinked1:
Commit 3cab4e7ab4a3ae483430d5f5e8fa167d02a8128c introduced a regression by causing the wrong pointers to be used in certain (more complex) cases. We now take care to ensure the exact same pointers are used as before that commit, and simplify by eliminating the unnecessary second for loop.

Resolves: #1147

Issue #1147, opened 2025-03-04
[bug]: Commit 3cab4e7a to config.c causes segfaults and spinloops in UpdateConfig mgr action


gtjoseph:
### Severity

Blocker

### Versions

master

### Components/Modules

main/config.c

### Operating Environment

all

### Frequency of Occurrence

Constant

### Issue Description

The following commit causes the UpdateConfig manager action to either segfault or spinloop in ast_config_text_file_save2().

3cab4e7a : config.c Make ast_variable_update update last match.

The following commit is related:

c157ee62 : config.c: Make ast_variable_retrieve return last match.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

InterLinked1:
Is there a test case or do you have an example of this? I haven't encountered this behavior with that API.

gtjoseph:
@InterLinked1 I would have assigned this to you for research but unfortunately github doesn't allow you to assign a person to an issue unless they're already mentioned in the issue.

testsuite...
tests/manager/config/no_preserve_effective_context segfaults
tests/manager/config/basic segfaults or spins in ast_config_text_file_save2() continually writing to the config file.

InterLinked1:
> [@InterLinked1](https://github.com/InterLinked1) I would have assigned this to you for research but unfortunately github doesn't allow you to assign a person to an issue unless they're already mentioned in the issue.
> 
> testsuite... tests/manager/config/no_preserve_effective_context segfaults tests/manager/config/basic segfaults or spins in ast_config_text_file_save2() continually writing to the config file.

Gotcha.
Feel free to assign this to me, I'll look into it today.

InterLinked1:
> [@InterLinked1](https://github.com/InterLinked1) I would have assigned this to you for research but unfortunately github doesn't allow you to assign a person to an issue unless they're already mentioned in the issue.
> 
> testsuite... tests/manager/config/no_preserve_effective_context segfaults tests/manager/config/basic segfaults or spins in ast_config_text_file_save2() continually writing to the config file.

@gtjoseph I was able to reproduce the issue with those tests consistently. I've linked a PR that resolves the issue and now the tests pass.

Curious, how come the tests only started failing recently? Both commits were merged in November, were these tests not being run on PRs/merges? I think you've said in the past that certain tests only run on nightlies, are all the tests also run prior to any merges?

gtjoseph:
[...]

PR #1452, opened 2025-09-11
config.c: fix saving of deep/wide template configurations

### LINKED ISSUE #1451: [bug]: ast_config_text_file_save2(): incorrect handling of deep/wide template inheritance
### Severity

Minor

### Versions

22

### Components/Modules

Main

### Operating Environment

Debian 12
Asterisk 22.5.2
Found while manipulating AllStarLink (Asterisk + app_rpt) configurations.

### Frequency of Occurrence

Occasional

### Issue Description

**Description:**

ast_config_text_file_save2() incorrectly drops or duplicates variables in deep and wide template inheritance because it checks against all ancestors instead of the nearest effective parent.

Follow-on to #244 and #960.

ast_config_text_file_save2() suppresses variables if they match any ancestor template. This breaks:

- Deep chains: values may be dropped based on distant parents instead of the nearest one.
- Wide (multi-parent) inheritance: variables can be suppressed if they match an earlier parent, even though a later parent overrides (last-wins).

**Impact:**

Saved configs can lose effective values or accumulate misleading duplicates, so the file no longer reflects runtime behavior.

**Example:**

```
[base](!)
var = AAA

[parent](base)
var = BBB

[child](parent)
; inherits var = BBB
```

Changing `var` in `[base]` causes `[child]` to lose its effective `var = BBB` when saved, because the code compares against the wrong ancestor.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

Allan-N:
Follow-on to #244 and #960 regarding how the ast_config_XXX APIs handle template inheritance.

ast_config_text_file_save2() incorrectly suppressed variables if they matched any ancestor template.  This broke deep chains (dropping values based on distant parents) and wide inheritance (ignoring last-wins order across multiple parents).

The function now inspects the full template hierarchy to find the nearest effective parent (last occurrence wins).  Earlier inherited duplicates are collapsed, explicit overrides are kept unless they exactly match the parent, and PreserveEffectiveContext avoids writing redundant lines.

Resolves: #1451

### DIFF
--- a/main/config.c
+++ b/main/config.c
@@ -225,7 +225,8 @@ struct ast_category_template_instance {
 
 struct ast_category {
 	char name[80];
-	int ignored;			/*!< do not let user of the config see this category -- set by (!) after the category decl; a template */
+	int ignored:1;	/*!< do not let user of the config see this category -- set by (!) after the category decl; a template */
+	int loaded:1;	/*!< 0 = created in memory, 1 = loaded from disk */
 	int include_level;
 	/*!
 	 * \brief The file name from whence this declaration was read
@@ -2084,6 +2085,7 @@ static int process_text_line(struct ast_config *cfg, struct ast_category **cat,
 		if (!newcat) {
 			return -1;
 		}
+		(*cat)->loaded = 1;
 		(*cat)->lineno = lineno;
 
 		/* add comments */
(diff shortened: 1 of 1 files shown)


Allan-N:
cherry-pick-to: 23

gtjoseph:
@Allan-N Does this not apply to the 20,21 and 22 branches?

Allan-N:
> Does this not apply to the 20,21 and 22 branches?

This change "can" be applied to 20, 21, and 22 but I opted not to include them because the other "template" Issues (#244, #960, #1147) / PRs (#245, #961, #1151) were not cherry-pick'd back.  If we pull this fix back should we not also pull those in too?

FWIW, I've been doing all of my development / testing on 22.5.1 + patches for the noted PRs.

gtjoseph:
> This change "can" be applied to 20, 21, and 22 but I opted not to include them because the other "template" Issues (#244, #960, #1147) / PRs (#245, #961, #1151) were not cherry-pick'd back. If we pull this fix back should we not also pull those in too?

Ah, gotcha.  It's good with just 23.

Allan-N:
FYI : I have been using the attached [cami](https://github.com/InterLinked1/cami) : `amicli` script to test the changes.

[cami-updateconfig-deep-or-wide-templates.sh](https://github.com/user-attachments/files/22277665/cami-updateconfig-deep-or-wide-templates.sh)

jcolp (approved):

gtjoseph (dismissed):
Testsuite test manager/config/advanced is now failing with this PR.  You're going to have to determine whether this is an issue with this PR or the test itself and correct.

Allan-N:
>Testsuite test manager/config/advanced is now failing with this PR. You're going to have to determine whether this is an issue with this PR or the test itself and correct.

Yup.  It looks like my change will need a few tweaks.

Allan-N:
I have updated the PR to address the issue with the Testsuite manager/config/advanced test failure.

The CI PRCheck's show a channels/pjsip/geolocation/incoming/profile_precedence failure but I don't think that's related to these changes.

gtjoseph:
> The CI PRCheck's show a channels/pjsip/geolocation/incoming/profile_precedence failure but I don't think that's related to these changes.

Yeah I think you're good.  We'll see what the cherry-pick tests reveal if anything.

Allan-N:
WAIT!  I've been doing a bit more testing and have found a gotcha I'm not thrilled with.

Allan-N:
>WAIT! I've been doing a bit more testing and have found a gotcha I'm not thrilled with.

Looks like this my comment and the approval/merge crossed paths  What I found was that a call to ast_variable_update() for a category that inherited from another, and where the variable existed only in the template (parent), would have the new value added to the template (parent) and not the leaf (child).  I will create a new PR to ensure that the correct/expected behavior.

gtjoseph:
@Allan-N Oops, sorry about that.  You have time to fix it however.  The next set of releases won't be for a while.

Allan-N:
>@Allan-N Oops, sorry about that. You have time to fix it however. The next set of releases won't be for a while.

I should have a fix in the next few days.

For future reference, and I briefly thought about this when I made my comment, would it have been better for me to change the PR to a "draft" ?

jcolp:
Yes, things in progress/not ready marked as draft won't be merged.

gtjoseph:
Well, the issue in this case is that I already started the merge before the "WAIT!" comment.  I'm not sure putting the PR in DRAFT status at that point would have prevented the merge to master but that happens after the cherry-picks anyway which don't respect the DRAFT status.  I'll have to test and think on this.

Allan-N:
Well, after a bit of digging I found that the unexpected behavior I was seeing was totally expected.  I've got code that uses ast_variable_update() and was updating a variable that existed only in the template.  Fixed that code and all is better.

As for this PR, it's ALL GOOD !!!

Issue #1419, opened 2025-09-08
[bug]: static code analysis issues in app_adsiprog.c


aol-nnov:
### Severity

Major

### Versions

18..master

### Components/Modules

app_adsiprog

### Operating Environment

static code analysis

### Frequency of Occurrence

None

### Issue Description

Return value of a function 'get_token' is dereferenced at app_adsiprog.c:851 without checking for NULL, but it is usually checked for this function (53/55).

app_adsiprog.c:856 Logic operator precedence in assignment can produce an unexpected result

correct snippet follows:
```c
if ((snums[scnt] = getstatebyname(state, sname, script, lineno, 0)) == NULL)
```

There are also a bunch of `Checker: NONTERMINATED_STRING.STYLE` errors, but seems, it is by design..

snippets like app_adsiprog.c:1240 Copying from string 'tmp' to '&disp->data[0+disp->datalen]' without null termination at app_adsiprog.c:1240. The length of the source string is calculated with the call of 'strlen' function at app_adsiprog.c:1240.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

aol-nnov:
@InterLinked1 you've missed the "Logic operator precedence in assignment can produce an unexpected result" thing...

I, for one, do not like the assignments inside if due to exact this error..

incorrect:

```
if((a = getA() == NULL))
```

correct:

```
if((a = getA()) == NULL)
```

InterLinked1:
Yeah, looks like there were two parts to it - the one on line 851 and on 859. The way the description was worded I thought it was just the first one but should've read that more carefully. I can address the other one separately.

InterLinked1:
@aol-nnov I added a comment on the PR, but I think the other warning is a false positive.

PR #1453, opened 2025-09-11
app_adsiprog: Fix incorrect checking of result for NULL.

InterLinked1:
Part of #1419


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

InterLinked1:
@aol-nnov Looking at this further, I don't think this is a bug after all. I think it's intentionally trying to see if the value is NULL and store that in the array, not the return value itself in the array. The compiler error on this PR is a dead giveaway to that.

PR #1464, opened 2025-09-17
apps/app_mixmonitor.c: Add 's' option to app_mixmonitor

daoudataha:
Add an optional 's' parameter to app_mixmonitor which skips writing audio for the first N seconds (1–120) of a recording. No audio is written to the file during this period. If the call ends before this time, no audio is saved. This is useful when discarding announcements, ringback, or other non-essential early audio.

UserNote: app_mixmonitor now supports the 's' option: s(x) to skip the first x seconds of the recording.
UpgradeNote: No upgrade steps required.

Test:
1- Call a number, run MixMonitor with s(5)
2- Check recording file: first 5 seconds are skipped
3- Ensure audio after 5 seconds is recorded correctly

cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20


daoudataha:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

daoudataha:
> **Attention!** This pull request may contain issues that could prevent it from being accepted. Please review the checklist below and take the recommended action. If you believe any of these are not applicable, just add a comment and let us know.
> 
> * [x]  The PR title does not match the commit title. This can cause confusion for reviewers and future maintainers. GitHub doesn't automatically update the PR title when you update the commit message so if you've updated the commit with a force-push, please update the PR title to match the new commit message body.
> * [x]  The PR description does not match the commit message body. This can cause confusion for reviewers and future maintainers. GitHub doesn't automatically update the PR description when you update the commit message so if you've updated the commit with a force-push, please update the PR description to match the new commit message body.
> * [x]  The PR description and/or commit message has unsupported trailers after the `Resolves`, `Fixes`, `UserNote`, `UpgradeNote` and/or `DeveloperNote` trailers. Please refrain from adding unsupported trailers as they will confuse the release change log generation.  If you really need them, please move them before any of the supportred trailers and ensure there's a blank line after them.
> * [x]  The PR description and/or commit message has malformed `UserNote`, `UpgradeNote` and/or `DeveloperNote` trailers. The `UserNote`, `UpgradeNote` and `DeveloperNote` keywords MUST be predeeded by a blank line and followed immediately by a colon and a space before the actual note text.  This is to ensure that the note is properly formatted and displayed in the release change logs.
> 
> Documentation:
> 
> * [Asterisk Developer Documentation](https://docs.asterisk.org/Development/)
>   
>   * [Code Contribution](https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/)
>   * [Commit Messages](https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages)
>   * [Alembic Scripts](https://docs.asterisk.org/Development/Reference-Information/Other-Reference-Information/Alembic-Scripts/)

Done.

jcolp:
Your change appears to have also drastically changed formatting... to the degree that Github won't display the diff by default.

daoudataha:
> Your change appears to have also drastically changed formatting... to the degree that Github won't display the diff by default.

Hi @jcolp ,
Thank you for the feedback. I have updated app_mixmonitor.c and disable my editor auto formatting . The PR now only contains the logic for the new s option, and the diff should be much smaller and easier to review.

seanbright:
> The PR now only contains the logic for the new s option, and the diff should be much smaller and easier to review.

[That is not the case.](https://github.com/asterisk/asterisk/pull/1464/files)

seanbright:
I also think that this is something trivially done with a `MixMonitor` post-process script.

daoudataha:
> > The PR now only contains the logic for the new s option, and the diff should be much smaller and easier to review.
> 
> [That is not the case.](https://github.com/asterisk/asterisk/pull/1464/files)

Please take a look [here](https://github.com/asterisk/asterisk/pull/1464/files)

daoudataha:
> I also think that this is something trivially done with a `MixMonitor` post-process script.

Thanks @seanbright  for the feedback!

You’re right that this can be achieved with a MixMonitor post-process script. However, the benefit of having it natively inside Asterisk is that the skipped audio is never written to disk in the first place. This makes it more efficient and also safer for compliance scenarios (e.g. PCI, HIPAA, GDPR) where storing even a temporary copy of sensitive information is not acceptable.

It also simplifies usage for dialplan authors — instead of configuring and maintaining an external trimming script, they can just enable the s option directly in MixMonitor.

Given that call centers and regulated environments frequently need to skip initial greetings or sensitive input, I think this feature adds practical value while keeping the implementation small and self-contained.

apps/app_mixmonitor.c
```diff
@@ -393,6 +400,10 @@
 
 #define get_volfactor(x) x ? ((x > 0) ? (1 << x) : ((1 << abs(x)) * -1)) : 0
 
+#define MIN_SKIP_SECONDS 1
+
+#define MAX_SKIP_SECONDS 120
```
InterLinked1: Why a limit of 120 seconds? It seems like an arbitrary limitation

daoudataha: > Why a limit of 120 seconds? It seems like an arbitrary limitation

That's a good point. The 120-second limit was added as a safeguard to prevent excessively long skips that could result in empty or useless recordings. In practice, use cases like skipping announcements, ringback, or IVRs typically only need a few seconds, rarely more than a minute. If you think a higher or unlimited value is more appropriate, I can adjust or remove the cap.

daoudataha:
> The new code is indented incorrectly.

Thanks for pointing this out. It looks like astyle.options is no longer included in the tree, so I corrected the indentation manually to match the existing style. Please let me know if further adjustment is needed

apps/app_mixmonitor.c
jcolp: I think the cap should be removed. It's arbitrary as InterLinked1 mentions, and if a user is requesting to skip longer than I see no reason to not allow them to.

daoudataha: Cap removed as suggested. Thanks for the feedback.

apps/app_mixmonitor.c
```diff
@@ -772,6 +787,10 @@ static void *mixmonitor_thread(void *obj)
 	int errflag = 0;
 	struct ast_format *format_slin;
 
+	unsigned int skip_samples;
+	unsigned int skipped_samples;
+	unsigned int frame_samples = 0;
```
InterLinked1: You've initialized just this variable here, but it's also initialized on line 823, so this seems unnecessary

apps/app_mixmonitor.c
```diff
@@ -1334,6 +1364,16 @@ static int mixmonitor_exec(struct ast_channel *chan, const char *data)
 				return -1;
 			}
 		}
+
+		if (ast_test_flag(&flags, MUXFLAG_SKIP)) {
+			if (ast_strlen_zero(opts[OPT_ARG_SKIP])) {
+				ast_log(LOG_WARNING, "No skip value was provided for the skip ('s') option.\n");
+			} else if ((sscanf(opts[OPT_ARG_SKIP], "%3d", &x) != 1) || (x < MIN_SKIP_SECONDS)) {
+				ast_log(LOG_NOTICE, "The skip will be ignored, the value must be a number > %d, not '%s'\n", MIN_SKIP_SECONDS, opts[OPT_ARG_SKIP]);
```
InterLinked1: `>=` not `>`, right?

apps/app_mixmonitor.c
daoudataha: Removed.

apps/app_mixmonitor.c
daoudataha: Exactly.

apps/app_mixmonitor.c
```diff
@@ -1334,6 +1363,16 @@ static int mixmonitor_exec(struct ast_channel *chan, const char *data)
 				return -1;
 			}
 		}
+
+		if (ast_test_flag(&flags, MUXFLAG_SKIP)) {
+			if (ast_strlen_zero(opts[OPT_ARG_SKIP])) {
+				ast_log(LOG_WARNING, "No skip value was provided for the skip ('s') option.\n");
+			} else if ((sscanf(opts[OPT_ARG_SKIP], "%3d", &x) != 1) || (x < MIN_SKIP_SECONDS)) {
+				ast_log(LOG_NOTICE, "The skip will be ignored, the value must be a number >= %d, not '%s'\n", MIN_SKIP_SECONDS, opts[OPT_ARG_SKIP]);
```
InterLinked1: This should be at least a warning (or maybe even an error?), not a notice, since this means that something is wrong that the user needs to fix

daoudataha: That makes sense. Since this situation indicates a usage problem that the user should address, I’ll update the log level to WARNING instead of NOTICE. 
Thanks for the suggestion.

apps/app_mixmonitor.c
```diff
@@ -1334,6 +1363,16 @@ static int mixmonitor_exec(struct ast_channel *chan, const char *data)
 				return -1;
 			}
 		}
+
+		if (ast_test_flag(&flags, MUXFLAG_SKIP)) {
+			if (ast_strlen_zero(opts[OPT_ARG_SKIP])) {
+				ast_log(LOG_WARNING, "No skip value was provided for the skip ('s') option.\n");
```
jcolp: State that no default exists so skip will be ignored

seanbright (changes requested):
This should accept fractional seconds as well.

apps/app_mixmonitor.c
daoudataha: Thanks, fixed. Updated the log message to clarify that no default exists, so the skip option will be ignored.

daoudataha:
> This should accept fractional seconds as well.

Added — skip now accepts fractional seconds. Thank you

apps/app_mixmonitor.c
```diff
@@ -1334,6 +1366,22 @@ static int mixmonitor_exec(struct ast_channel *chan, const char *data)
...
+
+		if (ast_test_flag(&flags, MUXFLAG_SKIP)) {
+			if (ast_strlen_zero(opts[OPT_ARG_SKIP])) {
+				ast_log(LOG_WARNING, "No skip value provided for the 's' (skip) option; skipping will be ignored as no default exists.\n");
+			} else {
+				char *endptr = NULL;
+				double val = strtod(opts[OPT_ARG_SKIP], &endptr);
+				if (endptr == opts[OPT_ARG_SKIP] || *endptr != '\0') {
+					ast_log(LOG_WARNING, "Skip value '%s' is not a valid number; ignoring skip.\n", opts[OPT_ARG_SKIP]);
+				} else if (val < (double)MIN_SKIP_SECONDS) {
```
InterLinked1: There should be a space between `(double)` and `MIN_SKIP_SECONDS`

apps/app_mixmonitor.c
```diff
@@ -792,12 +810,18 @@ static void *mixmonitor_thread(void *obj)
 
 	ast_mutex_unlock(&mixmonitor->mixmonitor_ds->lock);
 
+	/* The number of samples to skip (round to nearest) when fractional seconds provided */
+	skip_samples =  (unsigned int) (mixmonitor->skip_seconds * (double)mixmonitor->mixmonitor_ds->samp_rate + 0.5);
```
InterLinked1: Have you tested this approach on channels without any media flowing? Because of the audiohook timing limitation in Asterisk, audiohooks do not run when there is no media, so if we have to wait for a certain number of samples, it seems like this could skip a much greater amount of time than desired potentially.

I think the only reliable way to do this would be actually comparing timestamps,  using timers like in `app_dial` for the dial/progress timeouts, but someone else correct me if I'm wrong about that.

InterLinked1 (changes requested):

apps/app_mixmonitor.c
daoudataha: > There should be a space between `(double)` and `MIN_SKIP_SECONDS`

Added. Thank you.

apps/app_mixmonitor.c
daoudataha: > Have you tested this approach on channels without any media flowing? Because of the audiohook timing limitation in Asterisk, audiohooks do not run when there is no media, so if we have to wait for a certain number of samples, it seems like this could skip a much greater amount of time than desired potentially.
> 
> I think the only reliable way to do this would be actually comparing timestamps, using timers like in `app_dial` for the dial/progress timeouts, but someone else correct me if I'm wrong about that.

Thanks! I’ve updated the patch to address this — it now handles channels without media more reliably.

daoudataha:
> This is fairly easy to test, and should have an accompanying testsuite test with it.

Thanks for the suggestion! I’ve added a corresponding test in tests/test_app_mixmonitor_skip.c to validate the new s option behavior and ensure correct skipping of initial seconds.

tests/test_app_mixmonitor_skip.c
```diff
@@ -0,0 +1,225 @@
+/*
```
jcolp: Er, how was this test written? Normally tests like this go as a functional testsuite test, not as a unit test.

daoudataha: Thanks for clarifying — I initially implemented the test as a unit test (tests/test_app_mixmonitor.c) to validate the skip logic directly.
I’ve removed it for now and will try to move it to the functional testsuite test later, as I’m currently quite loaded with other tasks.

apps/app_mixmonitor.c
```diff
@@ -158,6 +158,13 @@
 						separated by commas eg. m(1111@default,2222@default,...).  Folders can be optionally specified using
 						the syntax: mailbox@context/folder</para>
 					</option>
+					<option name="s">
+						<argument name="seconds" required="true" />
+						<para>Skip the first <replaceable>seconds</replaceable> (can be fractional) of audio in the recording.
```
jcolp: This explanation needs changing.

"Don't record until seconds have elapsed since MixMonitor is invoked" is more accurate, and the same goes for the commit message.

apps/app_mixmonitor.c
```diff
@@ -817,18 +840,32 @@ static void *mixmonitor_thread(void *obj)
...
+			/* Skip writing audio for the first N seconds */
+			if (mixmonitor->skip_seconds > 0.0) {
+				struct timeval now = ast_tvnow();
+				double elapsed = ast_tvdiff_ms(now, skip_start) / 1000.0;
+
+				if (elapsed < mixmonitor->skip_seconds) {
+					ast_mutex_unlock(&mixmonitor->mixmonitor_ds->lock);
+					/* Skip this frame and continue */
+					goto frame_cleanup;
+				} else if (!skip_done_logged) {
```
jcolp: skip_done_logged doesn't need to exist. Once skipped you can set mixmonitor->skip_seconds to 0.

jcolp (changes requested):

daoudataha:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

apps/app_mixmonitor.c
daoudataha: Thanks for the suggestion — I’ve updated the documentation and the commit message.

apps/app_mixmonitor.c
daoudataha: Done, skip_done_logged removed and mixmonitor->skip_seconds is set to 0 after skipping.

jcolp (approved):

daoudataha:
> This should accept fractional seconds as well.

Hi @seanbright ,
I’ve addressed all the requested changes and re-requested your review.
Please let me know if anything else is needed. Thanks!

apps/app_mixmonitor.c
daoudataha: > > Have you tested this approach on channels without any media flowing? Because of the audiohook timing limitation in Asterisk, audiohooks do not run when there is no media, so if we have to wait for a certain number of samples, it seems like this could skip a much greater amount of time than desired potentially.
> > I think the only reliable way to do this would be actually comparing timestamps, using timers like in `app_dial` for the dial/progress timeouts, but someone else correct me if I'm wrong about that.
> 

Hello @InterLinked1 ,
I’ve addressed all the requested changes and re-requested your review.
Please let me know if anything else is needed. Thank you!

daoudataha:
Hello @InterLinked1 ,
I’ve addressed all the requested changes and re-requested your review.
Please let me know if anything else is needed. Thank you!

daoudataha:
> I think a UserNote in the commit message and PR description announcing the additional option is warranted.

Thanks @gtjoseph  for the suggestion! I’ve added a UserNote both in the commit message and in the PR description.

daoudataha:
The reported test failures (pjsip/geolocation/incoming/basic, pjsip/subscriptions/rls/lists_of_lists/nominal/mwi/batched, and res_pjsip/scheduler/unserialized_scheduler) don’t appear related to the changes in this PR, which only modify MixMonitor option handling.
When convenient, please re-run the CI checks to confirm.

gtjoseph:
> The reported test failures (pjsip/geolocation/incoming/basic, pjsip/subscriptions/rls/lists_of_lists/nominal/mwi/batched, and res_pjsip/scheduler/unserialized_scheduler) don’t appear related to the changes in this PR, which only modify MixMonitor option handling. When convenient, please re-run the CI checks to confirm.

Yeah, they're flukes, don't worry about it.

daoudataha:
Thanks, everyone, for the review, feedback, and merge!

PR #1465, opened 2025-09-18
stasis_channels.c: Add null check for referred_by in ast_ari_transfer_message_create

SvenKube:
When handling SIP transfers via ARI, the `referred_by` field in `transfer_ari_state` may be null, since SIP REFER requests are not required to include a `Referred-By` header. Without this check, a null value caused the transfer to fail and triggered a NOTIFY with a 500 Internal Server Error.

### DIFF
--- a/main/stasis.c
+++ b/main/stasis.c
@@ -97,30 +97,40 @@
 			</configObject>
 			<configObject name="taskpool">
 				<since>
-					<version>24.0.0</version>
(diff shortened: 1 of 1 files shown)


SvenKube:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

Issue #1462, opened 2025-09-17
[bug]: chan_websocket isn't handling the "opus" codec correctly.


gtjoseph:
### Severity

Major

### Versions

all

### Components/Modules

chan_websocket

### Operating Environment

Any

### Frequency of Occurrence

Constant

### Issue Description

chan_websocket isn't handling the "opus" codec correctly.  Currently it throws the following warning when `c(opus)` is used to create the WebSocket channel:

```
[Sep 17 08:08:38] WARNING[365649]: chan_websocket.c:847 webchan_write: WebSocket/media_ws/0x728ed400e270: This WebSocket channel only supports the ‘opus’ format
```

If we get by that then it throws the following:

```
ERROR[1081217]: codec_opus.c:503 opus_samples_count: Opus: Unable to parse packet for number of samples: corrupted stream
 ERROR[1081217]: codec_opus.c:503 opus_samples_count: Opus: Unable to parse packet for number of samples: corrupted stream
 ERROR[1081216][C-00000001]: codec_opus.c:335 should_fec: Opus: Unable to parse packet for frame data: corrupted stream
 ERROR[1081216][C-00000001]: codec_opus.c:386 opus_dec_decode: Opus: decoding: corrupted stream
```

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1466, opened 2025-09-18
chan_websocket: Fix codec validation and add passthrough option.

gtjoseph:
* Fixed an issue in webchan_write() where we weren't detecting equivalent
  codecs properly.
* Added the "p" dialstring option that puts the channel driver in
  "passthrough" mode where it will not attempt to re-frame or re-time
  media coming in over the websocket from the remote app.  This can be used
  for any codec but MUST be used for codecs that use packet headers or whose
  data stream can't be broken up on arbitrary byte boundaries. In this case,
  the remote app is fully responsible for correctly framing and timing media
  sent to Asterisk and the MEDIA text commands that could be sent over the
  websocket are disabled.  Currently, passthrough mode is automatically set
  for the opus, speex and g729 codecs.
* Now calling ast_set_read_format() after ast_channel_set_rawreadformat() to
  ensure proper translation paths are set up when switching between native
  frames and slin silence frames.  This fixes an issue with codec errors
  when transcode_via_sln=yes.

Resolves: #1462

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -59,6 +59,18 @@
 						/channels/answer ARI endpoint.
 						</para>
 					</enum>
+					<enum name="p - Passthrough mode">
+						<para>In passthrough mode, the channel driver won't attempt
+						to re-frame or re-time media coming in over the websocket from
+						the remote app.  This can be used for any codec but MUST be used
+						for codecs that use packet headers or whose data stream can't be
+						broken up on arbitrary byte boundaries. In this case, the remote
+						app is fully responsible for correctly framing and timing media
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
Updated the PR to automatically set the "p" option for opus, speex and g729a.

gtjoseph:
Now calling ast_set_read_format() after ast_channel_set_rawreadformat() to ensure proper translation paths are set up when switching between native frames and slin silence frames. This fixes an issue with codec errors when transcode_via_sln=yes.

jcolp (approved):

Issue #1474, opened 2025-09-22
[bug]: Media doesn't flow for video conference after res_rtp_asterisk change to stop media flow before DTLS completes


jcolp:
### Severity

Blocker

### Versions

20.16.0-rc1, 21.11.0-rc1, 22.6.0-rc1, 23.0.0-rc1

### Components/Modules

res_rtp_asterisk

### Operating Environment

.

### Frequency of Occurrence

None

### Issue Description

Opened based on https://github.com/asterisk/asterisk/issues/1260#issuecomment-3320238883

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

nappsoft:
The issue seems to have to do with sending rtcp packets. I've added a line of code to check in which scenarios the code prevents rtp_sendto from sending out the packet and I've got the following output (-113 and -127 is the value of *out):

[Sep 23 06:35:14] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:15] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:16] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:16] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -127
[Sep 23 06:35:17] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:18] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:19] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:20] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:21] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113
[Sep 23 06:35:21] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -127
[Sep 23 06:35:22] WARNING[13727]: res_rtp_asterisk.c:3462 __rtp_sendto: rtcp -113

[...]

PR #1475, opened 2025-09-23
res_rtp_asterisk.c: Use rtp->dtls in __rtp_sendto when rtcp mux is used.

gtjoseph:
In __rtp_sendto(), the check for DTLS negotiation completion for rtcp packets
needs to use the rtp->dtls structure instead of rtp->rtcp->dtls when
AST_RTP_INSTANCE_RTCP_MUX is set.

Resolves: #1474


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1479, opened 2025-09-25
[bug]: The ARI bridge play and record APIs limit audio bandwidth by forcing the slin8 format.


gtjoseph:
### Severity

Minor

### Versions

All

### Components/Modules

res_ari

### Operating Environment

All

### Frequency of Occurrence

Constant

### Issue Description

The ARI bridge play and record APIs are forcing the Announcer/Recorder channel to slin8 which means that if you play or record audio with a sample rate > 8K, it's being downsampled to 8K limiting the bandwidth.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1480, opened 2025-09-25
ARI: The bridges play and record APIs now handle sample rates > 8K correctly.

gtjoseph:
The bridge play and record APIs were forcing the Announcer/Recorder channel
to slin8 which meant that if you played or recorded audio with a sample
rate > 8K, it was downsampled to 8K limiting the bandwidth.

* The /bridges/play REST APIs have a new "announcer_format" parameter that
  allows the caller to explicitly set the format on the "Announcer" channel
  through which the audio is played into the bridge.  If not specified, the
  default depends on how many channels are currently in the bridge.  If
  a single channel is in the bridge, then the Announcer channel's format
  will be set to the same as that channel's.  If multiple channels are in the
  bridge, the channels will be scanned to find the one with the highest
  sample rate and the Announcer channel's format will be set to the slin
  format that has an equal to or greater than sample rate.

* The /bridges/record REST API has a new "recorder_format" parameter that
  allows the caller to explicitly set the format on the "Recorder" channel
  from which audio is retrieved to write to the file.  If not specified,
  the Recorder channel's format will be set to the format that was requested
  to save the audio in.

Resolves: #1479

DeveloperNote: The ARI /bridges/play and /bridges/record REST APIs have new
parameters that allow the caller to specify the format to be used on the
"Announcer" and "Recorder" channels respecitvely.

### DIFF
--- a/res/ari/resource_bridges.c
+++ b/res/ari/resource_bridges.c
@@ -311,7 +311,8 @@ static void *bridge_channel_control_thread(void *data)
 	return NULL;
 }
 
-static struct ast_channel *prepare_bridge_media_channel(const char *type)
+static struct ast_channel *prepare_bridge_media_channel(const char *type,
+	struct ast_format *channel_format)
 {
 	RAII_VAR(struct ast_format_cap *, cap, NULL, ao2_cleanup);
 	struct ast_channel *chan;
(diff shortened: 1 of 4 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
BTW, I haven't tested fully yet which is why the PR is in DRAFT mode.

gtjoseph:
Issue reporter confirms that the process is working so I'm moving to ready.  Still working on a test though.

gtjoseph:
I've tested locally but there's no way to create a testsuite test for this because there's no way to get the format currently on the Announcer or Recorder channels.

jcolp (approved):

Issue #1483, opened 2025-09-29
[improvement]: sig_analog: Eliminate possible timeout for Last Number Redial


InterLinked1:
### Improvement Description

When Last Number Redial is used, it is possible for a timeout to occur if the number in question is a prefix of another valid number. This doesn't make sense for redial, so ensure redialed calls complete immediately in all circumstances.

PR #1484, opened 2025-09-29
sig_analog: Eliminate potential timeout with Last Number Redial.

InterLinked1:
If Last Number Redial is used to redial, ensure that we do not wait for further digits. This was possible if the number that was last dialed is a prefix of another possible dialplan match. Since all we did is copy the number into the extension buffer, if other matches are now possible, there would thus be a timeout before the call went through. We now complete redialed calls immediaetly in all cases.

Resolves: #1483


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1487, opened 2025-09-30
[improvement]: app_dial: Allow partial seconds to be used for dial timeouts


InterLinked1:
### Improvement Description

Currently, the dial timeouts must be whole numbers. Allow timeouts to truly have millisecond-precision.

PR #1488, opened 2025-09-30
app_dial: Allow fractional seconds for dial timeouts.

InterLinked1:
Even though Dial() internally uses milliseconds for its dial timeouts, this capability has been mostly obscured from users as the argument is only parsed as an integer, thus forcing the use of whole seconds for timeouts.

Parse it as a decimal instead so that timeouts can now truly have millisecond precision.

Resolves: #1487

UserNote: The answer and progress dial timeouts now have millisecond precision, instead of having to be whole numbers.

### DIFF
--- a/apps/app_dial.c
+++ b/apps/app_dial.c
@@ -97,6 +97,7 @@
 				<para>If a second argument is specified, this controls the number of seconds we attempt to dial the specified devices
 				without receiving early media or ringing. If neither progress, ringing, nor voice frames have been received when this
 				timeout expires, the call will be treated as a CHANUNAVAIL. This can be used to skip destinations that may not be responsive.</para>
+				<para>The timeouts need not be whole numbers; both arguments accept fractional seconds.</para>
 			</parameter>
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
This test failure appears to be valid and caused by this change.

apps/app_dial.c
```diff
@@ -2947,9 +2949,9 @@ static int dial_exec_full(struct ast_channel *chan, const char *data, struct ast
 			to_answer = -1;
 		}
 		if (!ast_strlen_zero(args.timeout)) {
-			to_progress = atoi(args.timeout);
-			if (to_progress > 0) {
-				to_progress *= 1000;
+			if (sscanf(anstimeout, "%30lf", &tmp) == 1 && tmp > 0) {
```
seanbright: Copypasta

Issue #1489, opened 2025-09-30
[improvement]: config_options.c: Improve misleading error message


InterLinked1:
### Improvement Description

If requesting config documentation for a module that has none, it says "Module was not found", even though that's not really accurate. Example:

```
*CLI> config show help app_queue
Module app_queue not found.
```

Improve the clarity of this message.

PR #1490, opened 2025-09-30
config_options.c: Improve misleading warning.

InterLinked1:
When running "config show help <module>", if no XML documentation exists for the specified module, "Module <module> not found." is returned, which is misleading if the module is loaded but simply has no XML documentation for its config. Improve the message to clarify that the module may simply have no config documentation.

Resolves: #1489


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

PR #1492, opened 2025-10-01
res_audiosocket: add message types for all slin sample rates

SvenKube:
Extend audiosocket messages with types 0x11 - 0x18 to create asterisk
frames in slin12, slin16, slin24, slin32, slin44, slin48, slin96, and
slin192 format, enabling the transmission of audio at a higher sample
rates. For audiosocket messages sent by Asterisk, the message kind is
determined by the format of the originating asterisk frame.

UpgradeNote: New audiosocket message types 0x11 - 0x18 has been added
for slin12, slin16, slin24, slin32, slin44, slin48, slin96, and
slin192 audio. External applications using audiosocket may need to be
updated to support these message types if the audiosocket channel is
created with one of these audio formats.

### DIFF
--- a/include/asterisk/res_audiosocket.h
+++ b/include/asterisk/res_audiosocket.h
@@ -39,19 +39,43 @@ extern "C" {
 
 enum ast_audiosocket_msg_kind {
 	/*! \brief Message indicates the channel should be hung up, direction: Sent only. */
-	AST_AUDIOSOCKET_KIND_HANGUP = 0x00,
+	AST_AUDIOSOCKET_KIND_HANGUP         = 0x00,
 
 	/*! \brief Message contains the connection's UUID, direction: Received only. */
-	AST_AUDIOSOCKET_KIND_UUID   = 0x01,
+	AST_AUDIOSOCKET_KIND_UUID           = 0x01,
 
 	/*! \brief Message contains a DTMF digit, direction: Received only. */
-	AST_AUDIOSOCKET_KIND_DTMF   = 0x03,
+	AST_AUDIOSOCKET_KIND_DTMF           = 0x03,
 
-	/*! \brief Messages contains audio data, direction: Sent and received. */
-	AST_AUDIOSOCKET_KIND_AUDIO  = 0x10,
+	/*! \brief Messages contains audio data, format: slin, direction: Sent and received. */
+	AST_AUDIOSOCKET_KIND_AUDIO          = 0x10,
+
+	/*! \brief Messages contains audio data, format: slin12, direction: Sent and received. */
+	AST_AUDIOSOCKET_KIND_AUDIO_SLIN12   = 0x11,
+
+	/*! \brief Messages contains audio data, format: slin16, direction: Sent and received. */
+	AST_AUDIOSOCKET_KIND_AUDIO_SLIN16   = 0x12,
+
+	/*! \brief Messages contains audio data, format: slin24, direction: Sent and received. */
+	AST_AUDIOSOCKET_KIND_AUDIO_SLIN24   = 0x13,
+
+	/*! \brief Messages contains audio data, format: slin32, direction: Sent and received. */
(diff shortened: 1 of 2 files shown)


SvenKube:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 21
cherry-pick-to: 20

seanbright:
It might be more flexible to include the sample rate in kHz in the message payload (1 byte should be enough?), then you could send any signed-linear data regardless of rate.

SvenKube:
I understand the idea of including the sample rate in the payload, but I think introducing a dedicated message type for slin16 audio messages is the better approach here.

The main reasons are:
- Non-breaking change: This approach doesn’t modify the existing 0x10 packet semantics, so existing clients and libraries remain compatible.
- Protocol consistency: The audiosocket spec already defines message kinds to describe payload formats. Using a new kind for slin16 aligns with that model, whereas overloading the payload format of 0x10 would diverge from it.
- Specification accuracy: The spec currently (incorrectly) describes 0x10 as 16-bit 8 kHz signed linear data. By adding 0x11 for 16 kHz slin16, we bring the implementation in line with the documented meaning.
- Predictable parsing: Keeping the 3-byte header + payload structure untouched ensures that existing parsers continue to work as-is, without needing to introspect payload bytes to determine format.

seanbright:
That… is a lot of AI generated text. 

I’m suggesting that 0x11 have the sample rate as part of the payload. I’m not suggesting any change to 0x10.

SvenKube:
That... is what i've written with some rewording by AI.

@seanbright To make sure I understand your suggestion correctly, which message type would you expect for 8kHz signed linear audio data? The 0x10 that would currently be sent or your proposed message containing the sample rate?

seanbright:
8kHz would continue to be 0x10. 16kHz and up would be 0x11. You also don’t have to do anything - it was just a suggestion.

jcolp:
I agree with @seanbright - it's going to be new, so may as well cover all the slin rates so that the next person who comes along that wants a new one doesn't have to do anything.

seanbright:
> * Specification accuracy: The spec currently (incorrectly) describes 0x10 as 16-bit 8 kHz signed linear data.

Out of curiosity, what is incorrect about the specification?

SvenKube:
> > * Specification accuracy: The spec currently (incorrectly) describes 0x10 as 16-bit 8 kHz signed linear data.
> 
> Out of curiosity, what is incorrect about the specification?

The specification states that the 0x10 Packet Type contains 16-bit 8 kHz signed linear data, but when the AudioSocket channel is created with the codec `slin16` the Audiosocket sends also a 0x10 packet containing 640 bytes of 16-bit 16 kHz signed linear data.

So to get 16 kHz audio data out of asterisk this "works", the receiver could determine by the payload size that the content is probably 16kHz audio. Sending 0x10 Packets with 640 bytes of 16kHz audio data on the other hand does not work as the format of the generated asterisk frame is hardcoded to slin in the module.

It seemed to me that the original specification intended for a packet type to only transport one audio format, which is where I got the idea for the new packet for 16kHz audio only.

SvenKube:
I'm ok with changing the new packet to contain the sample rate in the first byte of the payload but in my opinion it is a bit weird that the 8kHz audio comes in the 0x10 packet and 12,16,24,32,44,48,96,192 kHz audio in the same 0x11 packet which has to be parsed differently. For example i have writte a wireshark lua dissector to improve the wireshark experience where parsing is easier with the one-to-one relation between packet type and sample rate. 

@seanbright @jcolp Would the criticism that changes would be necessary for other sample rates be resolved if I would introduce a new packet type for each sample rate? Or would you still prefer the 8kHz=0x10 other=0x11 split?

jcolp:
If all slin are covered in the change in some manner I'm fine either way.

seanbright:
> If all slin are covered in the change in some manner I'm fine either way.

Same

SvenKube:
I updated the PR with message types for all supported slin sample rates.

Issue #1493, opened 2025-10-01
[new-feature]: app_sf: Add post-digit timer option


InterLinked1:
### Feature Description

Add the ability in `ReceiveSF` to have a timer since the last digit was detected, as opposed to just for receiving all digits. This is pretty much a requirement if you are receiving an unknown number of digits, since there is no "start" signal as with MF.

PR #1494, opened 2025-10-01
app_sf: Add post-digit timer option to ReceiveSF.

InterLinked1:
Add a sorely needed option to set a timeout between digits, rather than for receiving the entire number. This is needed if the number of digits being sent is unknown by the receiver in advance. Previously, we had to wait for the entire timer to expire.

Resolves: #1493

UserNote: The 't' option for ReceiveSF now allows for a timer since
the last digit received, in addition to the number-wide timeout.

### DIFF
--- a/apps/app_sf.c
+++ b/apps/app_sf.c
@@ -78,6 +78,20 @@
 					<option name="m">
 						<para>Mute conference.</para>
 					</option>
+					<option name="t">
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

Issue #1496, opened 2025-10-01
[improvement]: dsp.c: Minor fixes to debug log messages


InterLinked1:
### Improvement Description

There is a debug message that should be inside a conditional, since it is not applicable otherwise.

PR #1497, opened 2025-10-01
dsp.c: Make minor fixes to debug log messages.

InterLinked1:
Commit dc8e3eeaaf094a3d16991289934093d5e7127680 improved the debug log messages in dsp.c. This makes two minor corrections to it:

* Properly guard an added log statement in a conditional.
* Don't add one to the hit count if there was no hit (however, we do still want to do this for the case where this is one).

Resolves: #1496


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

Issue #1491, opened 2025-10-01
[bug]: Segfault in channelstorage_cpp during app_queue hangup lookup (get_by_name_exact → __ao2_ref)


phoneben:
### Severity

Trivial

### Versions

20.15.2

### Components/Modules

channelstorage_cpp

### Operating Environment

centos 7

### Frequency of Occurrence

One Time

### Issue Description

On Asterisk 20.15.2, a crash occurs when app_queue handles a hangup event and calls ast_channel_get_by_name for a Local/... channel. The lookup in channelstorage_cpp::get_by_name_exact returns a corrupted/freed entry, and the subsequent call to __ao2_ref() in astobj2.c dereferences invalid memory. This points to a race/use-after-free in channelstorage_cpp when a channel entry is removed while still being referenced by the hangup handler.

[core-asterisk-2025-09-30T22-37-00Z-locks-sanitized.txt](https://github.com/user-attachments/files/22630364/core-asterisk-2025-09-30T22-37-00Z-locks-sanitized.txt)
[core-asterisk-2025-09-30T22-37-00Z-thread1-sanitized.txt](https://github.com/user-attachments/files/22630365/core-asterisk-2025-09-30T22-37-00Z-thread1-sanitized.txt)
[core-asterisk-2025-09-30T22-37-00Z-info-sanitized.txt](https://github.com/user-attachments/files/22630363/core-asterisk-2025-09-30T22-37-00Z-info-sanitized.txt)
[core-asterisk-2025-09-30T22-37-00Z-full-sanitized.txt](https://github.com/user-attachments/files/22630362/core-asterisk-2025-09-30T22-37-00Z-full-sanitized.txt)
[core-asterisk-2025-09-30T22-37-00Z-brief-sanitized.txt](https://github.com/user-attachments/files/22630366/core-asterisk-2025-09-30T22-37-00Z-brief-sanitized.txt)

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

phoneben:
gtjoseph 
chat gpt analysis
is this correct ?
how bad 


* **Root cause:** `channelstorage_cpp`’s fast lookup paths (`get_by_name_exact`, `get_by_uniqueid`) access the `std::map` without taking a read lock, then call `ao2_bump()` on the retrieved pointer. A concurrent `erase` (e.g., during hangup) can free the entry, causing a use-after-free and segfault.
[...]

PR #1498, opened 2025-10-01
channelstorage_cpp_map_name_id: Add read locking around retrievals.

gtjoseph:
When we retrieve a channel from a C++ map, we actually get back a wrapper
object that points to the channel then right after we retrieve it, we bump its
reference count.  There's a tiny chance however that between those two
statements a delete and/or unref might happen which would cause the wrapper
object or the channel itself to become invalid resulting in a SEGV.  To avoid
this we now perform a read lock on the driver around those statements.

Resolves: #1491


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

mbradeen (approved):

jcolp (approved):

PR #1504, opened 2025-10-02
func_channel: Allow manually changing the audio format during a call.

### LINKED ISSUE #1503: [new-feature]: func_channel: Allow manually changing the audio format during a call
### Feature Description

Currently, the APIs to change channel formats during a call are not
exposed, but it may be necessary to manually change the format for
certain things to work, so add the ability to change the format.

InterLinked1:
Currently, the APIs to change channel formats during a call are not exposed, but it may be necessary to manually change the format for certain things to work, so add the ability to change the format.

Normally, this should not be necessary as formats are transcoded as needed, but it may be necessary to control when this happens.

As an example:

IAX [lpc] <-> Local [lpc] <-> Local [lpc] <-> PJSIP [ulaw]

While channel drivers will transcode automatically on the last Local -> PJSIP hop in this scenario, this may be too late if audio from the Local channel itself needs to be in ulaw (and normally, there is no reason to transcode between Local channels). By setting CHANNEL(audionativeformat)=ulaw in the appropriate place, it becomes:

IAX [lpc] <-> Local [lpc] <-> Local [ulaw] <-> PJSIP [ulaw]

The transcoding happens automatically in the desired place.

Resolves: #1503

UserNote: Writing CHANNEL(audionativeformat) can now manually set the channel format during a call if needed.

### DIFF
--- a/funcs/func_channel.c
+++ b/funcs/func_channel.c
@@ -127,7 +127,11 @@
 						<para>R/O format currently being read.</para>
 					</enum>
 					<enum name="audionativeformat">
-						<para>R/O format used natively for audio.</para>
+						<para>R/W format used natively for audio.</para>
+						<para>You can write a new value to change the audio format during
+						a call and exert more control over where transcoding happens, e.g.
+						if you have Local channels between channels with different formats.</para>
+						<para>Writing is only supported on Local channels.</para>
 					</enum>
 					<enum name="audiowriteformat">
 						<para>R/O format currently being written.</para>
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20
testsuite-test-pr: 154

funcs/func_channel.c
```diff
@@ -626,7 +629,30 @@ static int func_channel_write_real(struct ast_channel *chan, const char *functio
...
+		ast_format_cap_update_by_allow_disallow(cap, value, 1);
+		fmt = ast_format_cap_get_format(cap, 0);
+		if (!fmt) {
+			ast_log(LOG_ERROR, "No format for capabilities '%s'\n", value);
+			ao2_ref(cap, -1);
+			return -1;
+		}
+
+		ast_channel_lock(chan);
+		ast_channel_nativeformats_set(chan, cap);
```
jcolp, line 672: Nativeformats alongside the raw format variants are between the channel driver implementation and the core. They are NOT something to be set arbitrarily like this in the CHANNEL function for all channels.

The non-read and write formats are fine to set as they are between what is reading frames from the channel or writing frames, and the core.

Something for Local channels specifically would be fine, because Asterisk itself has full control and they are internal. The raw variants would also need to be set.

jcolp (changes requested):

funcs/func_channel.c
InterLinked1: I started with that, but this didn't work for me until I also set the native formats. Open to a workaround for that though, this is just what worked for me.

InterLinked1: Also, I should clarify I have been using this on Local channels only so far

jcolp: Well for non-Local channels setting nativeformats is an absolute no.

InterLinked1: Do you think it would make more sense to skip certain logic for non-Local channels, or just restrict the capability entirely to Local channels?

Based on what you've said, I'm not sure how much sense it makes for non-Locals, and although I don't have a use case for non-Locals, I wonder if there are possible use cases someone might have.

jcolp: The only case for non-Local is to change the readformat and writeformat. Realistically I don't see why users would care about those, and an application could subsequently change them.

InterLinked1: So I've been doing some further testing based on this feedback and adding these two lines makes it not work:

```
ast_channel_set_rawreadformat(chan, fmt);
ast_channel_set_rawwriteformat(chan, fmt);
```

That's it, no other changes.

I don't entirely understand the intricacies of the raw vs native vs regular formats, not to mention you can manually set translation paths, but it does seem to be very picky in terms of calling only certain ones but not others.

jcolp: nativeformats = what has been negotiated by the underlying channel driver (in SIP sense this would be the result of the SDP negotiation), raw = what the channel driver is expecting from the core and what the channel driver is providing to the core (in SIP sense what comes/goes over RTP), non-raw = what the application is expecting to read from the channel and what the application is writing to the channel (what something like Playback or Record is providing/expecting)

jcolp: and Local channels complicate things because there can be two of them, and they may expect that each one matches the other so internally within the Local/unreal implementation frames can pass between them without transcoding

InterLinked1: Thanks, that helps a little. I try to visualize how the frames are flowing and where the transcoding is but with Local channels and multiple formats involved it gets a little complicated.

For my particular use case, the way the code is now seems to be the only way it will work. If you think the raw formats *should* be set, maybe it makes sense to break this out into multiple set operations using the channel function, since it sounds like depending on use case the user may or may not want to set the raw formats, just the native and non-raw. It seems I need to leave the raw alone.

I wanted to put it in one function call though because I know the three different formats are confusing and requiring the user to choose which ones get set and which don't will probably confuse people even more, but maybe that's the price of the flexibility needed?

jcolp: I think it's "working" because by not touching raw you are ensuring that the formats that get passed around inside Local/unreal stuff remains what is expected, however the raw formats are derived from native formats. So if nativeformats is set to g722, but this option sets the raw ones to ulaw - you've now fundamentally violated the contract though it just happens to work for your usage.

So I guess the question is... why are you setting native formats?

InterLinked1: > So I guess the question is... why are you setting native formats?

Because it didn't actually work (do anything) unless I did.

If I just set the non-raw formats, reading the audio format using CHANNEL, it said ulaw instead of what it was before (lpc10), but the audio sent to the SIP trunk from the channel sounded LPC. (This is a good example, since you can easily hear if something sounds "wrong" somewhere.). This audio in question is audio being generated on the local channel itself.

Setting native formats was sufficient to ensure the audio on the channel that reached the SIP trunk was truly ulaw and not transcoded after it was already too late, i.e. I believe what this did is force the transcoding to happen "earlier" in the channel sequence going towards the SIP trunk, ensuring audio generated on the channel was post-transcoding and already native ulaw at that point. Not sure if that's correct, it's just what it appeared from observation.

jcolp: It really sounds as if you've just found a combination that violates the contract but stuff compensates/is broken in such a way that it works for you and what you want.

This needs to go back to the drawing board to understand the full flow and interaction for Local channels while respecting the contract of nativeformats and raw formats.

asteriskteam:

asteriskteam:

funcs/func_channel.c
InterLinked1: I finally sat down and put together a test suite case for this as I've been meaning to, here: https://github.com/asterisk/testsuite/pull/154

I went ahead and tested each combination of formats to see if changing them would work properly or not:

nativeformats only -> works
nativeformats + read/write -> works
nativeformats + rawread/rawwrite -> works
nativeformats + read/write + rawread/rawwrite -> fails
rawread/rawwrite -> fails
read/write -> fails

I think that's all the combinations that would make sense...

The 3 at the top seem to work, and the 3 at the bottom don't. Right now the PR is written with #2.

So it will need to be one of the first 3. If you have a preference on which case to use, I can change it to that, I don't really know what else can be changed beyond this.

InterLinked1:
Oh, and if someone could reopen this PR as well, that would be great

Issue #1510, opened 2025-10-03
[new-feature]: sig_analog: Allow '#' to end the inter-digit timeout when dialing.


InterLinked1:
### Feature Description

Currently, if there are dialplan matches that prefix other valid matches, the user is forced to wait for the inter-digit timeout to expire. Dialing '#' is treated literally and is rejected. In cases where there is no dialplan match with the '#', allow this to terminate the timeout.

PR #1511, opened 2025-10-03
sig_analog: Allow '#' to end the inter-digit timeout when dialing.

InterLinked1:
It is customary to allow # to terminate digit collection immediately when there would normally be a timeout. However, currently, users are forced to wait for the timeout to expire when dialing numbers that are prefixes of other valid matches, and there is no way to end the timeout early. Customarily, # terminates the timeout, but at the moment, this is just rejected unless there happens to be a matching extension ending in #.

Allow # to terminate the timeout in cases where there is no dialplan match. This ensures that the dialplan is always respected, but if a valid extension has been dialed that happens to prefix other valid matches, # can be used to dial it immediately.

Resolves: #1510

### DIFF
--- a/channels/sig_analog.c
+++ b/channels/sig_analog.c
@@ -2422,6 +2422,7 @@ static void *__analog_ss_thread(void *data)
 		while (len < AST_MAX_EXTENSION-1) {
 			int is_exten_parking = 0;
 			int is_lastnumredial = 0;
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

PR #1513, opened 2025-10-06
app_mixmonitor: Prevent recording multiple times to the same file.

### LINKED ISSUE #1512: [improvement]: app_mixmonitor: Prevent MixMonitor from recording to the same file multiple times
### Improvement Description

It is possible to hose an Asterisk system simply by accidentally making a MixMonitor call in a loop. Multiple simultaneous recordings are permitted; however, we can prevent this by preventing attempts to record multiple times to the same file:

```
    -- Executing [s@mm:2] MixMonitor("Local/s@mm-00000001;2", "/tmp/test.wav") in new stack
    -- Executing [s@mm:3] Set("Local/s@mm-00000001;2", "i=988") in new stack
  == Begin MixMonitor Recording Local/s@mm-00000001;2
    -- Executing [s@mm:4] Wait("Local/s@mm-00000001;2", "0.01") in new stack
    -- Executing [s@mm:5] Goto("Local/s@mm-00000001;2", "s,2") in new stack
    -- Goto (mm,s,2)
    -- Executing [s@mm:2] MixMonitor("Local/s@mm-00000001;2", "/tmp/test.wav") in new stack
  == Begin MixMonitor Recording Local/s@mm-00000001;2
    -- Executing [s@mm:3] Set("Local/s@mm-00000001;2", "i=989") in new stack
    -- Executing [s@mm:4] Wait("Local/s@mm-00000001;2", "0.01") in new stack
    -- Executing [s@mm:5] Goto("Local/s@mm-00000001;2", "s,2") in new stack
    -- Goto (mm,s,2)
    -- Executing [s@mm:2] MixMonitor("Local/s@mm-00000001;2", "/tmp/test.wav") in new stack
    -- Executing [s@mm:3] Set("Local/s@mm-00000001;2", "i=990") in new stack
  == Begin MixMonitor Recording Local/s@mm-00000001;2
    -- Executing [s@mm:4] Wait("Local/s@mm-00000001;2", "0.01") in new stack
    -- Executing [s@mm:5] Goto("Local/s@mm-00000001;2", "s,2") in new stack
    -- Goto (mm,s,2)
    -- Executing [s@mm:2] MixMonitor("Local/s@mm-00000001;2", "/tmp/test.wav") in new stack
  == Begin MixMonitor Recording Local/s@mm-00000001;2
[Oct  5 00:22:06] WARNING[861406][C-00000002]: file.c:1588 ast_writefile: Unable to open file /tmp/test.wav: Too many open files
[Oct  5 00:22:06] ERROR[861406][C-00000002]: app_mixmonitor.c:740 mixmonitor_save_prep: Cannot open /tmp/test.wav
    -- Executing [s@mm:3] Set("Local/s@mm-00000001;2", "i=991") in new stack
    -- Executing [s@mm:4] Wait("Local/s@mm-00000001;2", "0.01") in new stack
    -- Executing [s@mm:5] Goto("Local/s@mm-00000001;2", "s,2") in new stack
    -- Goto (mm,s,2)
    -- Executing [s@mm:2] MixMonitor("Local/s@mm-00000001;2", "/tmp/test.wav") in new stack
  == Begin MixMonitor Recording Local/s@mm-00000001;2
[Oct  5 00:22:06] WARNING[861407][C-00000002]: file.c:1588 ast_writefile: Unable to open file /tmp/test.wav: Too many open files
[Oct  5 00:22:06] ERROR[861407][C-00000002]: app_mixmonitor.c:740 mixmonitor_save_prep: Cannot open /tmp/test.wav
    -- Executing [s@mm:3] Set("Local/s@mm-00000001;2", "i=992") in new stack
    -- Executing [s@mm:4] Wait("Local/s@mm-00000001;2", "0.01") in new stack
    -- Executing [s@mm:5] Goto("Local/s@mm-00000001;2", "s,2") in new stack
    -- Goto (mm,s,2)
    -- Executing [s@mm:2] MixMonitor("Local/s@mm-00000001;2", "/tmp/test.wav") in new stack
  == Begin MixMonitor Recording Local/s@mm-00000001;2
[Oct  5 00:22:06] WARNING[861408][C-00000002]: file.c:1588 ast_writefile: Unable to open file /tmp/test.wav: Too many open files
[Oct  5 00:22:06] ERROR[861408][C-00000002]: app_mixmonitor.c:740 mixmonitor_save_prep: Cannot open /tmp/test.wav
```

InterLinked1:
MixMonitor has no restrictions on making multiple calls to it, which allows recording to different files simultaneously. However, nothing prevents users from writing to the same file multiple times. There would be no reason to do this intentionally, but if a user accidentally makes the same MixMonitor call in a loop (perhaps due to a bad branch), it will happily open the file for each iteration, quickly exhausting all available file descriptors and eventually crashing the process.

If we detect an attempt to record to a file that is already active, prevent it and log an error. While the user still has a dialplan bug that needs to be addressed, we can prevent the system from running away to oblivion.

Resolves: #1512

### DIFF
--- a/apps/app_mixmonitor.c
+++ b/apps/app_mixmonitor.c
@@ -1232,6 +1232,51 @@ static char *filename_parse(char *filename, char *buffer, size_t len)
 	return buffer;
 }
 
+static int is_duplicate_mixmonitor(struct ast_channel *chan, const char *filename, const char *fn_write, const char *fn_read)
+{
+	struct ast_datastore *datastore = NULL;
+	struct mixmonitor_ds *mixmonitor_ds;
+	int duplicate;
+	const char *dup_file;
+
+	ast_channel_lock(chan);
+
+	datastore = ast_channel_datastore_find(chan, &mixmonitor_ds_info, NULL);
+	if (!datastore) {
+		ast_channel_unlock(chan);
+		return 0;
+	}
+	mixmonitor_ds = datastore->data;
+
+	/*
+	 * Nothing prevents a user calling MixMonitor() multiple times.
+	 * It makes sense if the user is recording to multiple files simultaneously,
+	 * but it doesn't make sense to record to the same file twice.
+	 * If we are able to catch that, prevent it. */
+	ast_mutex_lock(&mixmonitor_ds->lock);
+	/* Duplicates mixed file? */
+	dup_file = mixmonitor_ds->filename;
+	duplicate = !strcmp(filename, dup_file);
+	/* Duplicates write file? */
+	if (!duplicate && mixmonitor_ds->fs_write && !ast_strlen_zero(mixmonitor_ds->fs_write->filename) && !ast_strlen_zero(fn_write)) {
+		dup_file = mixmonitor_ds->fs_write->filename;
+		duplicate = !strcmp(fn_write, dup_file);
+	}
+	/* Duplicates read file? */
+	if (!duplicate && mixmonitor_ds->fs_read && !ast_strlen_zero(mixmonitor_ds->fs_read->filename) && !ast_strlen_zero(fn_read)) {
+		dup_file = mixmonitor_ds->fs_read->filename;
+		duplicate = !strcmp(fn_read, dup_file);
+	}
+	ast_mutex_unlock(&mixmonitor_ds->lock);
+
+	ast_channel_unlock(chan);
+
+	if (duplicate) {
+		ast_log(LOG_ERROR, "MixMonitor already recording to %s on %s, not recording again!\n", dup_file, ast_channel_name(chan));
+	}
+	return duplicate;
+}
+
 static int mixmonitor_exec(struct ast_channel *chan, const char *data)
 {
 	int x, readvol = 0, writevol = 0;
@@ -1345,6 +1390,9 @@ static int mixmonitor_exec(struct ast_channel *chan, const char *data)
 	/* If filename exists, try to create directories for it */
 	if (!(ast_strlen_zero(args.filename))) {
 		args.filename = ast_strdupa(filename_parse(args.filename, filename_buffer, sizeof(filename_buffer)));
+		if (is_duplicate_mixmonitor(chan, args.filename, filename_write, filename_read)) {
+			return 0;
+		}
 	}
 
 	pbx_builtin_setvar_helper(chan, "MIXMONITOR_FILENAME", args.filename);


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20
testsuite-test-pr: 106

seanbright:
If you're going to do this, it should check for duplicates across channels.

seanbright:
Also the arguments to the `r(...)` and `t(...)` options which are filenames

InterLinked1:
> If you're going to do this, it should check for duplicates across channels.

I don't see that sort of issue being high risk (I can't conceive how that would cause the issues mentioned here at least), but I don't have an objection to doing it if it's straightforward enough.

Currently, there isn't any kind of linked list that keeps track of all the MixMonitor sessions globally, since any modifications require the channel as an argument. Would there be a way to do this without adding and now maintaining a linked list for all MixMonitor operations? Seems kind of inefficient to do that purely for this avoidance warning, especially in an environment where maybe every channel is recording something. Just wondering if it's worth the performance hit.

> Also the arguments to the r(...) and t(...) options which are filenames

That can be done easily enough.

seanbright:
> I don't have an objection to doing it if it's straightforward enough.

Ah, OK. It may not be straightforward so nevermind.

InterLinked1:
> Also the arguments to the `r(...)` and `t(...)` options which are filenames

Completed this, and added test for it as well.

apps/app_mixmonitor.c
```diff
@@ -1232,6 +1232,51 @@ static char *filename_parse(char *filename, char *buffer, size_t len)
...
+	if (!duplicate && mixmonitor_ds->fs_read && !ast_strlen_zero(mixmonitor_ds->fs_read->filename) && !ast_strlen_zero(fn_read)) {
+		dup_file = mixmonitor_ds->fs_read->filename;
+		duplicate = !strcmp(fn_read, dup_file);
+	}
+	ast_mutex_unlock(&mixmonitor_ds->lock);
+
+	ast_channel_unlock(chan);
+
+	if (duplicate) {
+		ast_log(LOG_ERROR, "MixMonitor already recording to %s on %s\n", dup_file, ast_channel_name(chan));
```
jcolp: extend this to state the actual effect - that is: it won't record

jcolp (changes requested):

alex2grad:
Adding several locks to handle a very rare user error seems excessive.
This increases code complexity, potential for deadlocks, and may impact performance.

I'd suggest making this behavior optional, e.g. via a new MixMonitor flag, so users who need it can enable it explicitly.

seanbright:
> I'd suggest making this behavior optional, e.g. via a new MixMonitor flag, so users who need it can enable it explicitly.

If you’re savvy enough to know you need this new flag, you’re savvy enough to not need it.

seanbright:
I think this might make more sense if there was just a flag that prevented MixMonitor from starting if MixMonitor was already running on a given channel.

Otherwise this is too niche for inclusion IMO.

jcolp:
That would be less intensive, for sure. It does raise the question for me that aside from MixMonitor, what other ways can people shoot themselves in the foot in dialplan? Tons I'm sure so is this the start of trying to stop all those holes, or was this just a stumbled upon niche instance?

phoneben:
Having 2–3 recordings for one call is a common scenario by us , and the dial plan isn’t that bad.
I’ll try to get the exact dial plan over the weekend.

alex2grad:
In our setup, multiple MixMonitor instances per channel are valid, they write to different files.
We use separate recordings for different tenants in a multi-tenant setup, so blocking a second instance isn’t suitable.

seanbright:
> I think this might make more sense if there was just a flag

InterLinked1:
> In our setup, multiple MixMonitor instances per channel are valid, they write to different files. We use separate recordings for different tenants in a multi-tenant setup, so blocking a second instance isn’t suitable.

This doesn't block separate recordings on the same channels, it blocks attempts to record to the same exact file(s) multiple times on the same channel.

alex2grad:
> > In our setup, multiple MixMonitor instances per channel are valid, they write to different files. We use separate recordings for different tenants in a multi-tenant setup, so blocking a second instance isn’t suitable.
> 
> This doesn't block separate recordings on the same channels, it blocks attempts to record to the same exact file(s) multiple times on the same channel.

My comment was on @seanbright comment 
"I think this might make more sense if there was just a flag that prevented MixMonitor from starting if MixMonitor was already running on a given channel."

jcolp:
> That would be less intensive, for sure. It does raise the question for me that aside from MixMonitor, what other ways can people shoot themselves in the foot in dialplan? Tons I'm sure so is this the start of trying to stop all those holes, or was this just a stumbled upon niche instance?

I'd like some answers on this as to the actual background of this.

InterLinked1:
> > That would be less intensive, for sure. It does raise the question for me that aside from MixMonitor, what other ways can people shoot themselves in the foot in dialplan? Tons I'm sure so is this the start of trying to stop all those holes, or was this just a stumbled upon niche instance?
> 
> I'd like some answers on this as to the actual background of this.

I guess more along the lines of a stumbled upon niche instance, though I don't think it's particularly hard to stumble into it.

For context, this issue actually happened to me and I hosed a production system in doing so and had to kill/restart Asterisk. The natural post mortem question is "how can I prevent this from ever happening again"?

Sure, there are other ways people can shoot themselves and we're not going to prevent those, but most "dialplan loop" issues people can accidentally get themselves into are typically harmless and can be remedied by hanging up the channel from the console, whereas this particular scenario is game over without this patch. Aside from preventing a runaway system, the scenario is also wrong and a warning would be good anyways.

seanbright:
> Aside from preventing a runaway system, the scenario is also wrong

It's wrong when it crosses channels boundaries too. If we're going to do this we should cover that scenario too.

@jcolp what about doing this check in `ast_writestream()`? Too big a hammer?

jcolp:
@seanbright ast_writefile would be the place, and since that just happens when opening the filestream the impact should be minimal for the nominal case. I thinkkkkkkk that would be fine.

InterLinked1:
> @seanbright ast_writefile would be the place, and since that just happens when opening the filestream the impact should be minimal for the nominal case. I thinkkkkkkk that would be fine.

So I understand properly, the ask here is to move all this logic from `app_mixmonitor` to `ast_writefile`?

It seems like `mixmonitor_save_prep` saves the `writefile` return value but I don't see this really being used in a way that would prevent the issue, unless I'm missing an abort somewhere.

jcolp:
It doesn't currently. The code would have to be changed to actually pay attention to the error value, which it probably already should be doing.

asteriskteam:

asteriskteam:

InterLinked1:
> It doesn't currently. The code would have to be changed to actually pay attention to the error value, which it probably already should be doing.

There's a small hitch with doing this in `ast_writefile`, which is that `ast_writefile` doesn't receive the channel, which is needed to do this.

The good news is that `ast_writefile` is only called in a small number of places, and all call sites have access to the channel:
- `res_convert`
- `res_agi`
- `app_mixmonitor`
- `app_dicate`
- `app_record`
- `app_meetme`
- `file.c`

So, if we do need to break ABI and add an argument, it could be done with somewhat minor impact, although I think done the simplest way, that would preclude this change from being made prior to October, unless we add a new API variant and switch all call sites to the new API, just for ABI. All call sites also need to check for failure, which is not done at the moment, so modifications would be needed regardless.

The other half of this is the `mixmonitor` datastore was convenient since it kept track of all active recordings - they're added when we start and removed when we stop. `ast_writefile` would need a new datastore, and we would need to remove the recording from the datastore or it could cause problems. It *should* still be possible, as I *think* all calls to `ast_writefile` are paired with a call to `ast_closestream`.

All this considered, does this still seem like a reasonable path forward?

The other consideration is that if we want to just store the files globally, as opposed to per-channel, then the changes would be much simpler. However, based on this thread, it seems like there *might* be use cases where different channels record to the same file - that's not something that I aimed to address, but maybe worth considering:

- Should we also aim to prevent that?
- Are there legitimate use cases for different channels recording to the same file? (I can't think of any, but maybe it's an edge case in some systems that would have appeared to work previously, but would break if prevented)

seanbright:
> However, based on this thread, it seems like there _might_ be use cases where different channels record to the same file

There are no use cases for that

InterLinked1:
@asteriskteam Could somebody reopen this PR? It looks like I don't have permissions.
Took me a while but I have an updated version of this now.

jcolp:
Nope. "The master-issue-1512 branch was force-pushed or recreated."

I can not reopen.

InterLinked1:
> Nope. "The master-issue-1512 branch was force-pushed or recreated."
> 
> I can not reopen.

Hmm, okay. Should I try to fix that or is it fine if I just create a new PR?

jcolp:
Up to you. I don't care if it's a new PR, just add a separate comment linking to this one for reference.

Issue #1517, opened 2025-10-06
[improvement]: core_unreal: Preserve ADSI capability when dialing Local channels


InterLinked1:
### Improvement Description

In core_local, we were not copying the ADSI capabilities to the second channel in the channel pair, meaning the setting didn't inherit through Local channels - copy it over so it's preserved through Local channels.

PR #1518, opened 2025-10-06
core_unreal: Preserve ADSI capability when dialing Local channels.

InterLinked1:
Dial() already preserves the ADSI capability by copying it to the new channel, but since Local channel pairs consist of two channels, we also need to copy the capability to the second channel.

Resolves: #1517


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

Issue #1519, opened 2025-10-06
[improvement]: app_dial / func_callerid: DNIS information is not propagated by Dial


InterLinked1:
### Improvement Description

The DNIS (DNID) is not propagated by the Dial application, which is unlike the other Caller ID fields.

I believe this is possibly intentional for some reason, but it isn't spelled out in the documentation. I plan to add a documentation note to that effect, unless someone thinks it should be getting propagated (I am not sure myself).

PR #1520, opened 2025-10-06
func_callerid: Document limitation of DNID fields.

InterLinked1:
The Dial() application does not propagate DNID fields, which is counter to the behavior of the other Caller ID fields. This behavior is likely intentional since the use of Dial theoretically suggests a new dialed number, but document this caveat to inform users of it.

Resolves: #1519


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

PR #1522, opened 2025-10-07
res_audiosocket: fix temporarily unavailable

romkazor:
Operations on non-blocking sockets may return a resource temporarily unavailable error (EAGAIN or EWOULDBLOCK). This is not a fatal error but a normal condition indicating that the operation would block.

This patch corrects the handling of this case. Instead of incorrectly treating it as a reason to terminate the connection, the code now waits for data to arrive on the socket.

### DIFF
--- a/res/res_audiosocket.c
+++ b/res/res_audiosocket.c
@@ -277,10 +277,29 @@ struct ast_frame *ast_audiosocket_receive_frame_with_hangup(const int svc,
 		*hangup = 0;
 	}
 
-	n = read(svc, &header, 3);
-	if (n == -1) {
-		ast_log(LOG_WARNING, "Failed to read header from AudioSocket because: %s\n", strerror(errno));
-		return NULL;
+	while (i < 3) {
+		n = read(svc, header, 3);
+		if (n == -1) {
+			if (errno == EAGAIN || errno == EWOULDBLOCK) {
+				int poll_result = ast_wait_for_input(svc, 5);
+
+				if (poll_result == 1) {
+					continue;
+				} else if (poll_result == 0) {
+					ast_debug(1, "Poll timed out while waiting for header data\n");
+					continue;
+				} else {
+					ast_log(LOG_WARNING, "Poll error: %s\n", strerror(errno));
+				}
+			}
+
+			ast_log(LOG_ERROR, "Failed to read header from AudioSocket because: %s\n", strerror(errno));
+			return NULL;
+		}
+		if (n == 0) {
+			break;
+		}
+		i += n;
 	}
 
 	if (n == 0 || *kind == AST_AUDIOSOCKET_KIND_HANGUP) {


romkazor:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

res/res_audiosocket.c
```diff
@@ -278,6 +278,14 @@ struct ast_frame *ast_audiosocket_receive_frame_with_hangup(const int svc,
 	}
 
 	n = read(svc, &header, 3);
+	/* 
+	 *	A non-blocking socket sometimes returns resource temporarily unavailable due to a lack of data.
+	 *	But this is no reason to close the connection. 
+	 */
+	if (n == -1 && (errno == EAGAIN || errno == EWOULDBLOCK)) {
+		return &ast_null_frame;
```
seanbright: [Further down in this function we wait for data to be available](https://github.com/asterisk/asterisk/pull/1522/files#diff-106ec6e8285f319e9a44a3f1259a1e40d251cb5736234c352e5318c879bbdea2R324-R327). Why couldn't we do this same here?

romkazor: I ran a few tests on this specific code block. The code you mentioned inside the loop not working because of the [code block](https://github.com/asterisk/asterisk/pull/1522/files#diff-106ec6e8285f319e9a44a3f1259a1e40d251cb5736234c352e5318c879bbdea2R289-R292) right before it—that's where the socket is disconnecting. So, the options are to either remove that preceding code block entirely, take out the return NULL statement, or move the code into the loop. What do you think we should do?

```c
if (n == -1) {
		ast_log(LOG_WARNING, "Failed to read header from AudioSocket because: %s\n", strerror(errno));
		return NULL;
	}
```

seanbright: The code I referenced handles the `EAGAIN`/`EWOULDBLOCK` by polling and waiting. Your code handles it by returning the null frame. I'm suggesting you change your handling to poll and wait instead.

SvenKube: I also stumbled across the abovementioned issue and wrote a patch that also handles partial reads of the header data. I saw that beeing an issue when multiple audiosocket packets were sent in the same tcp packet and fragmentation occurs. I have it running in production for some time now and it seems to fix the issue for me. 

```diff
-	n = read(svc, &header, 3);
-	if (n == -1) {
-		ast_log(LOG_WARNING, "Failed to read header from AudioSocket because: %s\n", strerror(errno));
-		return NULL;
+	i = 0;
+	while (i < 3) {
+		n = read(svc, header + i, 3 - i);
+		if (n == -1) {
+			if (errno == EAGAIN || errno == EWOULDBLOCK) {
+				/* If we have not partially read the header it's ok to return an empty frame. */
+				if (i == 0) {
+					return &ast_null_frame;
+				}
+
+				int poll_result = ast_wait_for_input(svc, 5);
+
+				if (poll_result == 1) {
+					continue;
+				} else if (poll_result == 0) {
+					ast_log(LOG_WARNING, "Poll timed out while waiting for header data\n");
+				} else {
+					ast_log(LOG_WARNING, "Poll error: %s\n", strerror(errno));
+				}
+			}
+
+			ast_log(LOG_ERROR, "Failed to read header from AudioSocket because: %s\n", strerror(errno));
+			return NULL;
+		}
+		if (n == 0) {
+			break;
+		}
+		i += n;
 	}

```

romkazor: Yes, it seems this is what @seanbright  was talking about. Is SvenKubes's option acceptable in the end? If so, I'll fix the PR. I've tested this version, and it's also working.

SvenKube:
@romkazor Is there a reason you have removed the check if there has been read nothing from the header yet so that it just returns an null_frame in that case?

romkazor:
> Is there a reason you have removed the check if there has been read nothing from the header yet so that it just returns an null_frame in that case?

Right, Sean didn't like the idea of an empty frame. Indeed, there's no particular need for it. We will simply pause and polling and waiting for data from the socket again until it appears within a few attempts.

seanbright:
> Right, Sean didn't like the idea of an empty frame.

I didn’t have an issue with the null frame, I was trying to understand why we would return a null frame in one error path and `NULL` in another.

SvenKube:
> > Right, Sean didn't like the idea of an empty frame.
> 
> I didn’t have an issue with the null frame, I was trying to understand why we would return a null frame in one error path and `NULL` in another.

If we have already read some part of the header we cannot return a null_frame because when we try again to read a packet n bytes of the header were already consumed an we read parts of the actual payload as the header. That's why we could only return a null frame in case the number of bytes already read is zero.

romkazor:
> I was trying to understand why we would return a null frame in one error path and `NULL` in another.

Presumably, the initial assumption was that the socket should be closed upon a header read error

```c
/*!
 * \brief Receive an Asterisk frame from an AudioSocket server
 *
 * This returned object is a pointer to an Asterisk frame which must be
 * manually freed by the caller.
 *
 * \param svc The file descriptor of the network socket to the AudioSocket server.
 *
 * \retval A \ref ast_frame on success
 * \retval NULL on error
 */
struct ast_frame *ast_audiosocket_receive_frame(const int svc);

gtjoseph:
@seanbright Are you OK with this PR now?

res/res_audiosocket.c
SvenKube: I'm in the process of updating to Asterisk 23.1.0 and noticed that from my suggestion in a previous comment the line `n = read(svc, header + i, 3 - i);` was changed to not include i in the header offset and amount of data still needed to read for the header. 

In case of fragmented packets this could lead to the read call returning just one or two bytes and three bytes again in the next iteration of the loop overwriting the previously read bytes of the header and also taking some bytes of the payload as header bytes.

Why did this part got changed?

romkazor: I thought this was redundant, but now I see it needs to be added to handle edge cases. Could you open a new PR to update this?

PR #1529, opened 2025-10-10
app_queue: Allow stasis message filtering to work.

jcolp:
The app_queue module subscribes on a per-dialed agent basis to both the bridge all and channel all topics to keep apprised of things going on involving them. This subscription has associated state that must be cleaned up when the subscription ends. This was done by setting a default router callback that only had logic to handle the case where the subscription ends. By using the default router callback all filtering for the subscription was disabled, causing unrelated messages to get published and handled by it.

This change makes it so that an explicit route is added for the message type used for the message indicating the subscription has ended and removes the default router callback. This allows message filtering to occur on publishing reducing the messages to app_queue to only those it is interested in.


jcolp:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

PR #1530, opened 2025-10-10
endpoints: Remove need for stasis subscription.

jcolp:
When an endpoint is created in the core of Asterisk a subscription was previously created alongside it to monitor any channels being destroyed that were related to it. This was done by receiving all channel snapshot updates for every channel and only reacting when it was indicated that the channel was dead.

This change removes this logic and instead provides an API call for directly removing a channel from an endpoint. This is called when channels are destroyed. This operation is fast, so blocking the calling thread for a short period of time doesn't have any noticeable impact.


jcolp:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp:
Ooh, a failure! I'll look into that.

PR #1537, opened 2025-10-17
devicestate: Don't publish redundant device state messages.

jcolp:
When publishing device state check the local cache for the existing device state. If the new device state is unchanged from the prior one, don't bother publishing the update. This can reduce the work done by consumers of device state, such as hints and app_queue, by not publishing a message to them.

These messages would most often occur with devices that are seeing numerous simultaneous channels. The underlying device state would remain as in use throughout, but an update would be published as channels are created and hung up.

### DIFF
--- a/main/devicestate.c
+++ b/main/devicestate.c
@@ -309,14 +309,14 @@ enum ast_device_state ast_parse_device_state(const char *device)
 	return res;
 }
 
-static enum ast_device_state devstate_cached(const char *device)
+static enum ast_device_state devstate_cached(const char *device, const struct ast_eid *eid)
 {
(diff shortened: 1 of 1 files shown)


jcolp:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
I looked at that failure and didn't come up with an explanation of why this change would cause such a failure. Any ideas?

gtjoseph:
> I looked at that failure and didn't come up with an explanation of why this change would cause such a failure. Any ideas?

Re-ran the test and it passed so I think it's fine.

Issue #1539, opened 2025-10-17
[bug]: safe_asterisk without TTY doesn't log to file


whitmangroup220-cloud:
### Severity

Major

### Versions

22.1, 22.2, 22.3, 22.4, 22.5, 22.6

### Components/Modules

Logger

### Operating Environment

Rocky Linux

### Frequency of Occurrence

Constant

### Issue Description

Logger is no longer recording call detail to messages or full log even with verbose and debug set.
V 22.6 did not have this problem
v20.16 exhibits this problem where earlier versions did not.
In logger.conf
;messages => notice,warning,error
messages => notice,warning,error,debug,verbose
;
full => notice,warning,error,debug,verbose,dtmf,fax
;full => notice,warning,error,verbose

Refuses to log VERBOSE entries that appear in console but not written to log like this entry:
[2025-10-15 11:03:49] VERBOSE[45989][C-0000009e] pbx.c: Executing [1@trinity_call2017:1] System("PJSIP/zsiptrunk-0000013d", "/var/lib/asterisk/agi-bin/log_aster.pl "Trinity_Dialup User selected option 1"") in new stack

No Verbose entries written to log

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

jcolp:
Is safe_asterisk being used? If so if you re-enable the TTY in the script does that resolve it:

https://github.com/asterisk/asterisk/commit/0788c03f9242e1a4a2045cf03fac35a9cb271073

jcolp:
https://github.com/asterisk/asterisk/issues/1472 came up previously about it

whitmangroup220-cloud:
Yes, that worked.

Thank you!

Mike Atkinson
Director of IT
220 Parker Street, Warsaw, IN 46580
800.428.0408 - www.trinityschool.org<http://www.trinityschool.org/>

________________________________

PR #1540, opened 2025-10-17
safe_asterisk:  Fix logging and sorting issue.

gtjoseph:
Re-enabled "TTY=9" which was erroneously disabled as part of a recent
security fix and removed another logging "fix" that was added.

Also added a sort to the "find" that enumerates the scripts to be sourced so
they're sourced in the correct order.

Resolves: #1539

### DIFF
--- a/contrib/scripts/safe_asterisk
+++ b/contrib/scripts/safe_asterisk
@@ -6,7 +6,7 @@ ASTVARRUNDIR="__ASTERISK_VARRUN_DIR__"
 ASTVARLOGDIR="__ASTERISK_LOG_DIR__"
 
 CLIARGS="$*"			# Grab any args passed to safe_asterisk
-#TTY=9				# TTY (if you want one) for Asterisk to run on
+TTY=9				# TTY (if you want one) for Asterisk to run on
 CONSOLE=yes			# Whether or not you want a console
 #NOTIFY=root@localhost		# Who to notify about crashes
 #EXEC=/path/to/somescript	# Run this command if Asterisk crashes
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 18
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

phoneben:
I think `TTY` should be behind a flag or by disrto .
As I understood for debian it is bad and for rhel it is needed.

gtjoseph:
> I think `TTY` should be behind a flag or by disrto . As I understood for debian it is bad and for rhel it is needed.

Right now, we just need to get it back to the way it was before.  We can investigate distro related differences at another time.

seanbright:
The change from a glob in the `for` loops to a `$(find …)` is also weird. Not being antagonistic but was this not reviewed?

jcolp:
It did go through review, though maybe not as much as it should have or just bash script knowledge/thinking not as high as other things.

gtjoseph:
> The change from a glob in the `for` loops to a `$(find …)` is also weird. Not being antagonistic but was this not reviewed?

There was a reason for the change from the glob to a find but I don't remember what it was right now.  

I also just updated this PR with fixes to remove a left over debugging "echo" and added a sort to the second loop so the scripts would be sourced in order.

seanbright:
> and added a sort to the second loop so the scripts would be sourced in order.

[Globs are sorted.](https://github.com/asterisk/asterisk/pull/196) I would love to know the rationale on the change to `find` purely to satiate my curiosity. 

The only reason I could imagine for switching to `find` would be to visit subdirectories but I don’t see why that change would be relevant to a security fix.

gtjoseph:
Yeah it wasn't sorting that was the issue with globs.  It was something else I found during testing various scenarios that made me use find instead.  I'll see if I saved some of my test versions.  Maybe they'll jog my memory.

Issue #1536, opened 2025-10-17
[bug]: asterisk -rx connects to console instead of executing a command


akrivulya:
### Severity

Trivial

### Versions

20.16.0, 22.6.0

### Components/Modules

core

### Operating Environment

```
# uname -a
FreeBSD graal.it-profi.org.ua 14.3-RELEASE-p2 FreeBSD 14.3-RELEASE-p2 GENERIC amd64
```

### Frequency of Occurrence

Constant

### Issue Description

After upgrade to 22.6.0 or 20.16.0 LTS-versions `asterisk -rx` command not works anymore as expected. Instead of an executing a command supplied by arguments it simply attaches to asterisk's console:

```
# asterisk -rx 'core show version'
Asterisk 20.16.0, Copyright (C) 1999 - 2025, Sangoma Technologies Corporation and others.
Created by Mark Spencer <markster@digium.com>
Asterisk comes with ABSOLUTELY NO WARRANTY; type 'core show warranty' for details.
This is free software, with components licensed under the GNU General Public
License version 2 and other licenses; you are welcome to redistribute it under
certain conditions. Type 'core show license' for details.
=========================================================================
Connected to Asterisk 20.16.0 currently running on graal (pid = 86142)
graal*CLI> 

```
With both previous versions 20.15.2 and 22.5.2 all works fine.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

jcolp:
I don't seem to be able to reproduce this under Linux. If you start Asterisk in a console using "asterisk -vvvvvgc" and then use "asterisk -rx" in another, does it work or does that also fail?

Our testsuite also uses this and has had no failures. I also did a quick check of CLI related commits and found nothing recent.

okryvulia:
Very interesting thing. `asterisk -vvvvvgc` now runs asterisk in background and detaches from controlling terminal.

jcolp:
Definitely not experiencing that here and haven't heard from anyone else such things, or seen such issues in the testsuite as I mentioned.

This may be a FreeBSD specific thing in some way.

jcolp:
I'm accepting this as an extended support issue, as I can not reproduce or confirm it on Linux which is what we core support.

okryvulia:
git bisect shows 9eef99f8f39613adf477551478531ad80c77996d as first bad commit. A problem is in  

```
[...]

PR #1541, opened 2025-10-19
Fix Endianness detection in utils.h for non-Linux

moench-tegeder:
Commit 43bf8a4ded7a65203b766b91eaf8331a600e9d8d introduced endian dependend byte-swapping code in include/asterisk/utils.h, where the endianness was detected using the __BYTE_ORDER macro. This macro lives in endian.h, which on Linux is included implicitely (by the network-related headers, I think), but on FreeBSD the headers are laid out differently and we do not get __BYTE_ORDER the implicit way.

Instead, this makes the usage of endian.h explicit by including it where we need it, and switches the BYTE_ORDER/*ENDIAN macros to the POSIX-defined ones (see
https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/endian.h.html for standard compliance). Additionally, this adds a compile-time check for the endianness-logic: compilation will fail if neither big nor little endian can be detected.

Fixes: #1536

### DIFF
--- a/include/asterisk/utils.h
+++ b/include/asterisk/utils.h
@@ -28,6 +28,7 @@
 #include <time.h>	/* we want to override localtime_r */
 #include <unistd.h>
 #include <string.h>
+#include <endian.h>
 
 #include "asterisk/lock.h"
 #include "asterisk/time.h"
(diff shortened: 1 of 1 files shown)


moench-tegeder:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

Issue #1534, opened 2025-10-16
[bug]: app_queue when using gosub breaks dialplan when going from 20 to 21, What's new in 21 doesn't mention it's a breaking change,


Alec-Davis-NZ:
### Severity

Major

### Versions

21, 22, 23, master

### Components/Modules

app_queue

### Operating Environment

Debian Bookworm

### Frequency of Occurrence

Constant

### Issue Description

In asterisk 20.6.0 we used the gosub 
**same => n,Queue(${QUEUE_NAME},crhCHxX,,,60,,,sub-queue-connected)**

But now there is 1 less comma before sub-queue-connected where the macro used to be.
**same => n,Queue(${QUEUE_NAME},crhCHxX,,,60,,sub-queue-connected)**

In addition to **gosub** not being run, the **rule** and **position** are now out by one position.

code from app_queue,c now misses the sixth parameter, The macro line was just deleted without adjusting the \param comments

` * \param[in] qe the queue_ent structure which corresponds to the caller attempting to reach members
 * \param[in] opts the options passed as the third parameter to the Queue() application
 * \param[in] opt_args the options passed as the third parameter to the Queue() application
 * \param[in] announceoverride filename to play to user when waiting
 * \param[in] url the url passed as the fourth parameter to the Queue() application
 * \param[in,out] tries the number of times we have tried calling queue members
 * \param[out] noption set if the call to Queue() has the 'n' option set.
 * \param[in] agi the agi passed as the fifth parameter to the Queue() application
 * \param[in] gosub the gosub passed as the seventh parameter to the Queue() application
 * \param[in] ringing 1 if the 'r' option is set, otherwise 0
 */
static int try_calling(struct queue_ent *qe, struct ast_flags opts, char **opt_args, char *announceoverride, const char *url, int *tries, int *noption, const char *agi, const char *gosub, int ringing)
{
`

Alec

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

phoneben:
What about the other parameters after this change? They seem to have been shifted as well and may also need to be adjusted.
`Queue(${quid},${HASH(queue,qu_parameters)}${QUEUEOPTION},,,${QUTIMEOUT},${QUEUEAGI},,${QUGOSUB},${SETQUEUERULE},${priority});`

How should the dialplan look if we want it to remain backward-compatible with previous Asterisk versions?

phoneben:
[...]

PR #1543, opened 2025-10-20
app_queue.c: Fix error in Queue parameter documentation.

bkford:
When macro was removed in Asterisk 21, the parameter documentation in
code was not updated to reflect the correct numerization for gosub. It
still stated that it was the seventh parameter, but got shifted to the
sixth due to the removal of macro. This has been updated to correctly
reflect the parameter order, and a note has been added to the XML that
states this was done after the initial commit.

Fixes: #1534

UpgradeNote: As part of Asterisk 21, macros were removed from Asterisk.
This resulted in argument order changing for the Queue dialplan
application since the macro argument was removed. Upgrade notice was
missed when this was done, so this upgrade note has been added to
provide a record of such and a notice to users who may have not upgraded
yet.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -259,6 +259,10 @@
 				<para>Will run a gosub on the called party's channel (the queue member)
 				once the parties are connected.  The subroutine execution starts in the
 				named context at the s exten and priority 1.</para>
+				<note><para>Macro was removed in Asterisk 21 which resulted in an
+				argument ordering change. The upgrade notice was missed for this,
+				so a note is being made here to provide a record of the change
+				for users who have not upgraded yet.</para></note>
 			</parameter>
 			<parameter name="rule">
 				<para>Will cause the queue's defaultrule to be overridden by the rule specified.</para>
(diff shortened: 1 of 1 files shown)


bkford:
cherry-pick-to: 23
cherry-pick-to: 22

jcolp:
An UpgradeNote is user facing, so it should be written as such and be useful. For example:

"As part of Asterisk 21 macros were removed from Asterisk. This resulted in argument ordering changing for the Queue dialplan application when the macro argument was removed. Upgrade notice was missed when this was done so this upgrade note has been added to provide a record of such and notice to users who may have not yet upgraded."

apps/app_queue.c
```diff
@@ -130,6 +130,9 @@
 			Queue a call for a call queue.
 		</synopsis>
 		<syntax>
+			<note><para>Macro was removed in Asterisk 21, but gosub still stated that it was the
```
jcolp: The code parameters are not user facing, so this comment doesn't really make sense because it was never stated for users.

jcolp (approved):

Issue #1535, opened 2025-10-16
[bug]: chan_pjsip changes SSRC on WebRTC channels, which is unsupported by some browsers


meengu:
### Severity

Major

### Versions

22

### Components/Modules

rtp

### Operating Environment

Any OS (Arch and Debian tested), Asterisk 22.6.0-rc2

### Frequency of Occurrence

None

### Issue Description

After commit 2a661446c4a0c2bc7ad9 media stream is no longer received in Firefox due to wrong SSRC

The offending commit: https://github.com/asterisk/asterisk/commit/2a661446c4a0c2bc7ad99da8e8a41c214bd662c9

PJSIP endpoints are configured for webrtc, and a direct call is placed between two such endpoints via asterisk from the browser (using jssip).

SDP sent from asterisk to the browser SIP client indicates that SSRC that the browser should expect should be XXX, but the first sent RTP packet to the browser already contains a different SSRC YYY, due to the SSRC change introduced in the mentioned commit.

### Relevant log output

```shell
ParameterName                      : ParameterValue
 ===================================================================================================
 100rel                             : yes
 accept_multiple_sdp_answers        : false
 accountcode                        : 
 acl                                : 
 aggregate_mwi                      : true
 allow                              : (opus|h264)
 allow_overlap                      : true
 allow_subscribe                    : true
 allow_transfer                     : true
 allow_unauthenticated_options      : false
 aors                               : 99401
 asymmetric_rtp_codec               : false
 auth                               : 99401-auth
 bind_rtp_to_media_address          : false
 bundle                             : true
 call_group                         : 
 callerid                           : "401" <401>
 callerid_privacy                   : allowed_not_screened
 callerid_tag                       : 
 codec_prefs_incoming_answer        : prefer:pending, operation:intersect, keep:all, transcode:allow
 codec_prefs_incoming_offer         : prefer:pending, operation:intersect, keep:all, transcode:allow
 codec_prefs_outgoing_answer        : prefer:pending, operation:intersect, keep:all, transcode:allow
 codec_prefs_outgoing_offer         : prefer:pending, operation:union, keep:all, transcode:allow
 connected_line_method              : invite
 contact_acl                        : 
 context                            : from-internal
 cos_audio                          : 7
 cos_video                          : 0
 device_state_busy_at               : 0
 direct_media                       : false
 direct_media_glare_mitigation      : none
 direct_media_method                : invite
 disable_direct_media_on_nat        : false
[...]

PR #1548, opened 2025-10-21
chan_pjsip: Disable SSRC change for WebRTC endpoints.

gtjoseph:
Commit b333ee3b introduced a fix to chan_pjsip that addressed RTP issues with
blind transfers and some SBCs.  Unfortunately, the fix broke some WebRTC
clients that are sensitive to SSRC changes and non-monotonic timestamps so
the fix is now disabled for endpoints with the "bundle" parameter set to true.

Resolves: #1535

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -1820,12 +1820,14 @@ static int chan_pjsip_indicate(struct ast_channel *ast, int condition, const voi
 	case AST_CONTROL_SRCUPDATE:
 		break;
 	case AST_CONTROL_SRCCHANGE:
-		/* Generate a new SSRC due to media source change and RTP timestamp reset.
-		   Ensures RFC 3550 compliance and avoids SBC interoperability issues (Sonus/Ribbon)*/
-		for (i = 0; i < AST_VECTOR_SIZE(&channel->session->active_media_state->sessions); ++i) {
-			media = AST_VECTOR_GET(&channel->session->active_media_state->sessions, i);
-			if (media && media->rtp) {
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

channels/chan_pjsip.c
```diff
@@ -1820,12 +1820,14 @@ static int chan_pjsip_indicate(struct ast_channel *ast, int condition, const voi
 	case AST_CONTROL_SRCUPDATE:
 		break;
 	case AST_CONTROL_SRCCHANGE:
-		/* Generate a new SSRC due to media source change and RTP timestamp reset.
-		   Ensures RFC 3550 compliance and avoids SBC interoperability issues (Sonus/Ribbon)*/
-		for (i = 0; i < AST_VECTOR_SIZE(&channel->session->active_media_state->sessions); ++i) {
-			media = AST_VECTOR_GET(&channel->session->active_media_state->sessions, i);
-			if (media && media->rtp) {
-				ast_rtp_instance_change_source(media->rtp);
+		if (!channel->session->endpoint->media.webrtc) {
```
jcolp: Thinking about this some more, and since we allow bundle to be configured independently, I think if channel->session->endpoint->media.bundle isn't enabled should be the check since bundle is really the cause of the issue.

jcolp (changes requested):

channels/chan_pjsip.c
gtjoseph: Changed to bundle.

Issue #1546, opened 2025-10-20
[improvement]: Not able to pass the custom variables over the websockets using external Media with ari client library nodejs


gauravs456:
### Improvement Description

While creating the external media channel i am unable to set the custom variables in media connect message
 return await channel.externalMedia( { channelId      : externalMediaId, app            :${constants.CALL_TYPE},chanType:streaming, external_host  : 'act_media_ws', format         : "slin16", transport      : "websocket", encapsulation  : "none", connection_type: "client", direction      : "both", variables: { "name": "Alice" } });
I am using ari client library in nodejs 
and doing like this but i am not able to receive that custom varailbes in media connect message which was received when this s conneted to websocket server

gtjoseph:
In your example above, you want "name: Alice" to be sent in the MEDIA_START websocket message, is this correct?  This would require sending all channel variables and would probably mean sending a JSON payload in a new MEDIA_START_JSON message.  Is this acceptable?

gauravs456:
Hi @gtjoseph , Yes it will be perfect  that sending the channel variables in the JSON payload and getting that in MEDIA_START_JSON will work, the final use case is that  we will able to send the data if it is in JSON format then it would be perfect

PR #1555, opened 2025-10-23
chan_websocket: Add capability for JSON control messages and events.

gtjoseph:
With recent enhancements to chan_websocket, the original plain-text
implementation of control messages and events is now too limiting.  We
probably should have used JSON initially but better late than never.  Going
forward, enhancements that require control message or event changes will
only be done to the JSON variants and the plain-text variants are now
deprecated but not yet removed.

* Added the chan_websocket.conf config file that allows setting which control
message format to use globally: "json" or "plain-text".  "plain-text" is the
default for now to preserve existing behavior.

* Added a dialstring option `f(json|plain-text)` to allow the format to be
overridden on a call-by-call basis.  Again, 'plain-text' is the default for
now to preserve existing behavior.

The JSON for commands sent by the app to Asterisk must be...
`{ "command": "<command>" ... }` where `<command>` is one of `ANSWER`, `HANGUP`,
`START_MEDIA_BUFFERING`, etc.  The `STOP_MEDIA_BUFFERING` command takes an
additional, optional parameter to be returned in the corresponding
`MEDIA_BUFFERING_COMPLETED` event:
`{ "command": "STOP_MEDIA_BUFFERING", "correlation_id": "<correlation id>" }`.

The JSON for events sent from Asterisk to the app will be...
`{ "event": "<event>", "channel_id": "<channel_id>" ... }`.
The `MEDIA_START` event will now look like...

```
{
  "event": "MEDIA_START",
  "connection_id": "media_connection1",
  "channel": "WebSocket/media_connection1/0x5140001a0040",
  "channel_id": "1761245643.1",
  "format": "ulaw",
  "optimal_frame_size": 160,
  "ptime": 20,
  "channel_variables": {
    "DIALEDPEERNUMBER": "media_connection1/c(ulaw)",
    "MEDIA_WEBSOCKET_CONNECTION_ID": "media_connection1",
    "MEDIA_WEBSOCKET_OPTIMAL_FRAME_SIZE": "160"
  }
}
```

Note the addition of the channel variables which can't be supported
with the plain-text formatting.

The documentation will be updated with the exact formats for all commands
and events.

Resolves: #1546
Resolves: #1563

DeveloperNote: The chan_websocket plain-text control and event messages are now
deprecated (but remain the default) in favor of JSON formatted messages.
See https://docs.asterisk.org/Configuration/Channel-Drivers/WebSocket for
more information.

DeveloperNote: A "transport_data" parameter has been added to the
channels/externalMedia ARI endpoint which, for websocket, allows the caller
to specify parameters to be added to the dialstring for the channel.  For
instance, `"transport_data": "f(json)"`.

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -31,76 +31,6 @@
 	<support_level>core</support_level>
  ***/
 
-/*** DOCUMENTATION
-	<info name="Dial_Resource" language="en_US" tech="WebSocket">
-		<para>WebSocket Dial Strings:</para>
-		<para><literal>Dial(WebSocket/connectionid[/websocket_options])</literal></para>
-		<para>WebSocket Parameters:</para>
-		<enumlist>
-			<enum name="connectionid">
-				<para>For outgoing WebSockets, this is the ID of the connection
-				in websocket_client.conf to use for the call.  To accept incoming
-				WebSocket connections use the literal <literal>INCOMING</literal></para>
-			</enum>
-			<enum name="websocket_options">
-				<para>Options to control how the WebSocket channel behaves.</para>
-				<enumlist>
-					<enum name="c(codec) - Specify the codec to use in the channel">
-						<para></para>
-						<para> If not specified, the first codec from the caller's channel will be used.
-						</para>
-					</enum>
-					<enum name="n - Don't auto answer">
-						<para>Normally, the WebSocket channel will be answered when
-						connection is established with the remote app.  If this
-						option is specified however, the channel will not be
-						answered until the <literal>ANSWER</literal> command is
-						received from the remote app or the remote app calls the
-						/channels/answer ARI endpoint.
-						</para>
-					</enum>
-					<enum name="p - Passthrough mode">
-						<para>In passthrough mode, the channel driver won't attempt
-						to re-frame or re-time media coming in over the websocket from
-						the remote app.  This can be used for any codec but MUST be used
-						for codecs that use packet headers or whose data stream can't be
-						broken up on arbitrary byte boundaries. In this case, the remote
-						app is fully responsible for correctly framing and timing media
-						sent to Asterisk and the MEDIA text commands that could be sent
-						over the websocket are disabled.  Currently, passthrough mode is
-						automatically set for the opus, speex and g729 codecs.
-						</para>
-					</enum>
-					<enum name="v(uri_parameters) - Add parameters to the outbound URI">
-						<para>This option allows you to add additional parameters to the
-						outbound URI.  The format is:
-						<literal>v(param1=value1,param2=value2...)</literal>
-						</para>
-						<para>You must ensure that no parameter name or value contains
-						characters not valid in a URL.  The easiest way to do this is to
-						use the URIENCODE() dialplan function to encode them.  Be aware
-						though that each name and value must be encoded separately.  You
-						can't simply encode the whole string.</para>
-					</enum>
-				</enumlist>
-			</enum>
-		</enumlist>
-		<para>Examples:
-		</para>
-		<example title="Make an outbound WebSocket connection using connection 'connection1' and the 'sln16' codec.">
-		same => n,Dial(WebSocket/connection1/c(sln16))
-		</example>
-		<example title="Make an outbound WebSocket connection using connection 'connection1' and the 'opus' codec. Passthrough mode will automatically be set.">
-		same => n,Dial(WebSocket/connection1/c(opus))
-		</example>
-		<example title="Listen for an incoming WebSocket connection and don't auto-answer it.">
-		same => n,Dial(WebSocket/INCOMING/n)
-		</example>
-		<example title="Add URI parameters.">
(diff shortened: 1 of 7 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
This PR is in draft mode because we _really_ need feedback.

gauravs456:
Hi the approach looks good but the concern is   how this options will be used  using external media so the [node-ari-client](https://github.com/asterisk/node-ari-client) this option will be added in external media option and the rest JSON and plain text looks good and understandable

gtjoseph:
node-ari-client uses swagger to discover the resources so it should automatically enable setting the `transport_data` parameter.  If you're always going to use JSON, the better approach would be to simply set it in chan_websocket.conf.

gauravs456:
Hi @gtjoseph , I read
https://docs.asterisk.org/Configuration/Channel-Drivers/WebSocket/#protocol

and i have a doubt about sending the audio to the asterisk in chan_websocket

support i have using slin16
format so that means asterisk want 320bytes in 20 ms

so my question is can i send complete 3200 bytes which is a multiple of 320 bytes
 and what will asterisk do ??
because i checked at my end it do not plays that but in doc it is written that asterisk will split automatically and do all the timing thing

second is this means that my application do not need to maintain the 20 ms timer or anything ?? i can send directly larger bytes without doing the 20ms timer interval ???

jcolp:
Please don't coopt this PR for something unrelated. You can post on the forum for help, and if in the end there's an issue then a new issue would be opened.

res/ari/resource_channels.c
```diff
@@ -2213,9 +2213,9 @@ static int external_media_websocket(struct ast_ari_channels_external_media_args
 	struct ast_channel *chan;
 	struct varshead *vars;
 
-	if (ast_asprintf(&endpoint, "WebSocket/%s/c(%s)",
```
jcolp, line 2216: Doesn't this alter existing behavior due to 'c' being removed?

channels/chan_websocket.c
```diff
@@ -198,6 +260,250 @@ static struct ast_channel_tech websocket_tech = {
...
+	if (instance->control_msg_format == WEBCHAN_CONTROL_MSG_FORMAT_JSON) {
+		struct ast_json *msg = ast_json_pack("{s:s, s:s, s:s, s:s, s:s, s:i, s:i, s:o }",
+			"event", "MEDIA_START",
+			"connection_id", instance->connection_id,
+			"channel", ast_channel_name(instance->channel),
+			"channel_id", ast_channel_uniqueid(instance->channel),
+			"format", ast_format_get_name(instance->native_format),
+			"optimal_frame_size", instance->optimal_frame_size,
+			"ptime", instance->native_codec->default_ms,
+			"channel_variables", ast_json_channel_vars(ast_channel_varshead(
```
mbradeen, line 222: Should the documentation indicate that the chan vars are only presented when using the json format?  Or are they sent another way?

gtjoseph: They're only supported via JSON.  I wanted to wait a few days before updating the docs website in case things changed.  I'll do that today.

channels/chan_websocket.c
```diff
@@ -1277,13 +1588,15 @@ enum {
...
 	OPT_WS_PASSTHROUGH =  (1 << 3),
+	OPT_WS_MSG_FORMAT =  (1 << 4),
 };
 
 enum {
 	OPT_ARG_WS_CODEC,
 	OPT_ARG_WS_NO_AUTO_ANSWER,
 	OPT_ARG_WS_URI_PARAM,
 	OPT_ARG_WS_PASSTHROUGH,
+	OPT_ARG_WS_MSG_FORMAT =  (1 << 4),
```
bkford: Does this need to be shifted if none of the others are?

res/ari/resource_channels.c
bkford: Also curious about this.

gtjoseph: No.  If you look further down in the function, args->format is passed to ari_channels_handle_originate_with_id().  This in turn makes its way to chan_websocket:webchan_request() which checks for the "c" parameter in the dialstring and if it doesn't find it, it uses the format passed in.

However, you highlighted another issue...  If transport_data is empty, the dialstring will end with a '/' which isn't good.

channels/chan_websocket.c
gtjoseph: oops, cut and paste error.

res/ari/resource_channels.c
gtjoseph: Fixed.

jcolp (approved):

gtjoseph:
The latest forced-push just updated the `<since>` elements in the XML documentation.

gtjoseph:
Force-push again to add the `f(format)` dialstring parameter to the XML and separate the XML into its own file.

Issue #781, opened 2024-06-25
[improvement]: Allow call by call disabling Stir/Shaken header inclusion 


oza4h07:
### Improvement Description

When a single trunk is used for both domestic and international outbound calls, you both need to include Stir/Shaken headers for domestic calls while remove them for international ones.
Having a dialplan function allowing call by call removal of Stir/Shaken headers fulfills this use case.

oza4h07:
Attaching to attestations a patten defining for which destination numbers, Stir/Shaken attestation should be done seems very fine to me.

oza4h07:
Any update on this ?
I asked the Asterisk Forum, for experience with cross-border calls to learn if others also prefer to remove Stir/Shaken headers when such are not required. I didn't get much reply unfortunately.

gtjoseph:
@oza4h07 Please test PR #1556

PR #1556, opened 2025-10-24
res_stir_shaken: Add STIR_SHAKEN_ATTESTATION dialplan function.

gtjoseph:
Also...

* Refactored the verification datastore process so instead of having
a separate channel datastore for each verification result, there's only
one channel datastore with a vector of results.

* Refactored some log messages to include channel name and removed
some that would be redundant if a memory allocation failed. 

Resolves: #781

UserNote: The STIR_SHAKEN_ATTESTATION dialplan function has been added
which will allow suppressing attestation on a call-by-call basis
regardless of the profile attached to the outgoing endpoint.

### DIFF
--- a/res/res_pjsip_stir_shaken.c
+++ b/res/res_pjsip_stir_shaken.c
@@ -33,8 +33,9 @@
 #include "asterisk/res_pjsip_session.h"
 #include "asterisk/module.h"
 #include "asterisk/rtp_engine.h"
+#include "asterisk/datastore.h"
 
-#include "asterisk/res_stir_shaken.h"
+#include "res_stir_shaken/stir_shaken.h"
 
 static const pj_str_t identity_hdr_str = { "Identity", 8 };
 static const pj_str_t date_hdr_str = { "Date", 4 };
@@ -395,6 +396,8 @@ static void stir_shaken_outgoing_request(struct ast_sip_session *session,
 	struct ast_stir_shaken_as_ctx *ctx = NULL;
 	enum ast_stir_shaken_as_response_code as_rc;
 	const char *session_name = ast_sip_session_get_name(session);
+	struct stir_shaken_attestation_ds *attestation_ds;
+
 	SCOPE_ENTER(1, "%s: Enter\n", session_name);
 
 	if (!session) {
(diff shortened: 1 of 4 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

res/res_pjsip_stir_shaken.c
```diff
@@ -407,6 +410,26 @@ static void stir_shaken_outgoing_request(struct ast_sip_session *session,
 		SCOPE_EXIT_LOG_RTN(LOG_ERROR, "%s: No tdata\n", session_name);
 	}
 
+	/*
+	 * On an outgoing channel, there should be at most one STIR/SHAKEN
+	 * datastore.
+	 */
+	ast_channel_lock(session->channel);
+	AST_LIST_TRAVERSE(ast_channel_datastores(session->channel), chan_datastore, entry) {
```
jcolp: Add a comment explaining why this iterates and an API call can't be done

res/res_stir_shaken.c
```diff
@@ -273,9 +264,110 @@ static int func_read(struct ast_channel *chan, const char *function,
...
+		ast_log(LOG_ERROR, "%s: %s requires a boolean value\n", channel_name, function);
+		return -1;
+	}
+
+	/*
+	 * On an outgoing channel, there should be at most one STIR/SHAKEN
+	 * datastore.
+	 */
+	ast_channel_lock(chan);
+	AST_LIST_TRAVERSE(ast_channel_datastores(chan), chan_datastore, entry) {
```
jcolp: Since you have attestation_ds_info can't you use ast_channel_datastore_find?

jcolp (changes requested):

res/res_pjsip_stir_shaken.c
gtjoseph: Added an API to get the attestation datastore so the traverse is no longer needed.

res/res_stir_shaken.c
gtjoseph: Yep, done.

jcolp (approved):

Issue #1559, opened 2025-10-27
[improvement]: Handle TLS handshake attacks in order to resolve the issue of exceeding the maximum number of HTTPS sessions.


Tinet-mucw:
### Improvement Description

Handle TLS handshake attacks in order to resolve the issue of exceeding the maximum number of HTTPS sessions.

[core-asterisk-2025-10-24T22-34-58Z-info.txt](https://github.com/user-attachments/files/23156903/core-asterisk-2025-10-24T22-34-58Z-info.txt)
[core-asterisk-2025-10-24T22-34-58Z-locks.txt](https://github.com/user-attachments/files/23156905/core-asterisk-2025-10-24T22-34-58Z-locks.txt)
[core-asterisk-2025-10-24T22-34-58Z-thread1.txt](https://github.com/user-attachments/files/23156904/core-asterisk-2025-10-24T22-34-58Z-thread1.txt)

Failed to upload "core-asterisk-2025-10-24T22-34-58Z-full.txt
Failed to upload "core-asterisk-2025-10-24T22-34-58Z-brief.txt

[root@iZuf60twxxjszuta2jksneZ tmp]# ls -l core-asterisk-2025-10-24T22-34-58Z*txt -h
-rw-r--r-- 1 root root  27M Oct 27 11:51 core-asterisk-2025-10-24T22-34-58Z-brief.txt
-rw-r--r-- 1 root root  65M Oct 27 12:20 core-asterisk-2025-10-24T22-34-58Z-full.txt
-rw-r--r-- 1 root root  46K Oct 27 12:20 core-asterisk-2025-10-24T22-34-58Z-info.txt
-rw-r--r-- 1 root root  518 Oct 27 12:20 core-asterisk-2025-10-24T22-34-58Z-locks.txt
-rw-r--r-- 1 root root 1.5K Oct 27 11:30 core-asterisk-2025-10-24T22-34-58Z-thread1.txt

[root@iZuf60twxxjszuta2jksneZ tmp]# more core-asterisk-2025-10-24T22-34-58Z-brief.txt
!@!@!@! brief.txt !@!@!@!

$2 = {si_signo = 19, si_errno = 0, si_code = 128, _sifields = {_pad = {0 <repeats 28 times>}, _kill = {si_pid = 0, si_uid = 0}, _timer = {si_tid = 0, si_overrun = 0, si_sigval = {si
[...]

PR #1560, opened 2025-10-27
iostream.c: Handle TLS handshake attacks in order to resolve the issue of exceeding the maximum number of HTTPS sessions.

Tinet-mucw:
The TCP three-way handshake completes, but if the server is under a TLS handshake attack, asterisk will get stuck at SSL_do_handshake(). In this case, a timeout mechanism should be set for the SSL/TLS handshake process to prevent indefinite waiting during the SSL handshake.

Resolves: #1559


Tinet-mucw:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1566, opened 2025-10-28
[improvement]: Improve Taskprocessor logging


mbradeen:
### Improvement Description

Improve taskprocessor logging:

Keep track of and optionally log task origin.
Keep track of the time to execute and origin of slowest executing task per taskprocessor.
Add the ability to dump a given taskprocessor's active queue.
Additional logging changes to improve clarity.

PR #1567, opened 2025-10-28
taskprocessors: Improve logging and add new cli options

mbradeen:
This change makes some small changes to improve log readability in
addition to the following changes:

Modified 'core show taskprocessors' to now show Low time and High time
for task execution.

New command 'core show taskprocessor name <taskprocessor-name>' to dump
taskprocessor info and current queue.

Addionally, a new test was added to demonstrate the 'show taskprocessor
name' functionality:
test execute category /main/taskprocessor/ name taskprocessor_cli_show

Setting 'core set debug 3 taskprocessor.c' will now log pushed tasks.
(Warning this is will cause extremely high levels of logging at even
low traffic levels.)

Resolves: #1566

UserNote: New CLI command has been added -
core show taskprocessor name <taskprocessor-name>

### DIFF
--- a/include/asterisk/res_pjsip.h
+++ b/include/asterisk/res_pjsip.h
@@ -2076,7 +2076,11 @@ struct ast_sip_endpoint *ast_sip_dialog_get_endpoint(pjsip_dialog *dlg);
  * \retval 0 Success
  * \retval -1 Failure
  */
-int ast_sip_push_task(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data);
+int __ast_sip_push_task(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data,
+	const char *file, int line, const char *function);
+
+#define ast_sip_push_task(serializer, sip_task, task_data) \
+	__ast_sip_push_task(serializer, sip_task, task_data, __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
 /*!
  * \brief Push a task to SIP servants and wait for it to complete.
@@ -2112,13 +2116,19 @@ int ast_sip_push_task(struct ast_taskprocessor *serializer, int (*sip_task)(void
  * \return sip_task() return value on success.
  * \retval -1 Failure to push the task.
  */
-int ast_sip_push_task_wait_servant(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data);
+int __ast_sip_push_task_wait_servant(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data,
+	const char *file, int line, const char *function);
+#define ast_sip_push_task_wait_servant(serializer, sip_task, task_data) \
+	__ast_sip_push_task_wait_servant(serializer, sip_task, task_data, __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
 /*!
  * \brief Push a task to SIP servants and wait for it to complete.
  * \deprecated Replaced with ast_sip_push_task_wait_servant().
  */
-int ast_sip_push_task_synchronous(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data);
+int __ast_sip_push_task_synchronous(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data,
+	const char *file, int line, const char *function);
+#define ast_sip_push_task_synchronous(serializer, sip_task, task_data) \
+	__ast_sip_push_task_synchronous(serializer, sip_task, task_data, __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
 /*!
  * \brief Push a task to the serializer and wait for it to complete.
@@ -2162,7 +2172,10 @@ int ast_sip_push_task_synchronous(struct ast_taskprocessor *serializer, int (*si
  * \return sip_task() return value on success.
  * \retval -1 Failure to push the task.
  */
-int ast_sip_push_task_wait_serializer(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data);
+int __ast_sip_push_task_wait_serializer(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data,
+	const char *file, int line, const char *function);
+#define ast_sip_push_task_wait_serializer(serializer, sip_task, task_data) \
+	__ast_sip_push_task_wait_serializer(serializer, sip_task, task_data, __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
 /*!
  * \brief Determine if the current thread is a SIP servant thread
--- a/include/asterisk/taskprocessor.h
+++ b/include/asterisk/taskprocessor.h
@@ -214,8 +214,10 @@ void *ast_taskprocessor_unreference(struct ast_taskprocessor *tps);
  * \retval -1 failure
  * \since 1.6.1
  */
-int ast_taskprocessor_push(struct ast_taskprocessor *tps, int (*task_exe)(void *datap), void *datap)
-	attribute_warn_unused_result;
+int __ast_taskprocessor_push(struct ast_taskprocessor *tps, int (*task_exe)(void *datap), void *datap,
+	const char *file, int line, const char *function) attribute_warn_unused_result;
+#define ast_taskprocessor_push(tps, task_exe, datap) \
+	__ast_taskprocessor_push(tps, task_exe, datap, __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
 /*! \brief Local data parameter */
 struct ast_taskprocessor_local {
@@ -240,9 +242,11 @@ struct ast_taskprocessor_local {
  * \retval -1 failure
  * \since 12.0.0
  */
-int ast_taskprocessor_push_local(struct ast_taskprocessor *tps,
-	int (*task_exe)(struct ast_taskprocessor_local *local), void *datap)
-	attribute_warn_unused_result;
+int __ast_taskprocessor_push_local(struct ast_taskprocessor *tps,
+	int (*task_exe)(struct ast_taskprocessor_local *local), void *datap,
+	const char *file, int line, const char *function) attribute_warn_unused_result;
+#define ast_taskprocessor_push_local(tps, task_exe, datap) \
+	__ast_taskprocessor_push_local(tps, task_exe, datap, __FILE__, __LINE__, __PRETTY_FUNCTION__)
 
 /*!
  * \brief Indicate the taskprocessor is suspended.
--- a/include/asterisk/threadpool.h
+++ b/include/asterisk/threadpool.h
@@ -188,8 +188,11 @@ void ast_threadpool_set_size(struct ast_threadpool *threadpool, unsigned int siz
  * \retval 0 success
  * \retval -1 failure
  */
-int ast_threadpool_push(struct ast_threadpool *pool, int (*task)(void *data), void *data)
-	attribute_warn_unused_result;
+int __ast_threadpool_push(struct ast_threadpool *pool, int (*task)(void *data), void *data,
+	const char *file, int line, const char *function) attribute_warn_unused_result;
+
+#define ast_threadpool_push(pool, task, data) \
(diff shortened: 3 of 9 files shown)


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

include/asterisk/taskprocessor.h
```diff
@@ -214,8 +214,10 @@ void *ast_taskprocessor_unreference(struct ast_taskprocessor *tps);
  * \retval -1 failure
  * \since 1.6.1
  */
-int ast_taskprocessor_push(struct ast_taskprocessor *tps, int (*task_exe)(void *datap), void *datap)
```
jcolp, line 217: This breaks ABI.

include/asterisk/taskprocessor.h
```diff
@@ -240,9 +242,11 @@ struct ast_taskprocessor_local {
  * \retval -1 failure
  * \since 12.0.0
  */
-int ast_taskprocessor_push_local(struct ast_taskprocessor *tps,
-	int (*task_exe)(struct ast_taskprocessor_local *local), void *datap)
-	attribute_warn_unused_result;
+int __ast_taskprocessor_push_local(struct ast_taskprocessor *tps,
```
jcolp, line 245: This also breaks ABI.

jcolp (changes requested):

seanbright (dismissed):
ABI is still broken. The original functions with their original arguments need to remain. You can create new functions and update everything to call them, but the old ones need to still exist.

seanbright:
> ABI is still broken. The original functions with their original arguments need to remain. You can create new functions and update everything to call them, but the old ones need to still exist.

Disregard. I see that you are still exporting the functions.

tests/test_taskprocessor.c
```diff
@@ -963,6 +966,174 @@ AST_TEST_DEFINE(serializer_pool)
...
+		goto cleanup;
+	}
+
+	ast_test_status_update(test, "CLI command output validated successfully\n");
+	res = AST_TEST_PASS;
+
+cleanup:
+	/* Signal all tasks to complete */
+	if (task_data1) {
+		ast_mutex_lock(&task_data1->lock);
```
jcolp: Why does this logic exist to set wait_time to 0? The tasks aren't in a loop, they wait a single time and altering wait_time isn't going to make them wake up early.

tests/test_taskprocessor.c
```diff
@@ -963,6 +966,174 @@ AST_TEST_DEFINE(serializer_pool)
...
+		ast_test_status_update(test, "Failed to queue task 2\n");
+		goto cleanup;
+	}
+	if (ast_taskprocessor_push(tps, task, task_data3)) {
+		ast_test_status_update(test, "Failed to queue task 3\n");
+		goto cleanup;
+	}
+
+	/* Give the taskprocessor a moment to start processing */
+	usleep(100000); /* 100ms */
```
jcolp: This feels like a race condition waiting to happen... does it actually need to wait?

tests/test_taskprocessor.c
```diff
@@ -963,6 +966,174 @@ AST_TEST_DEFINE(serializer_pool)
...
+	}
+	if (task_data3) {
+		ast_mutex_lock(&task_data3->lock);
+		task_data3->wait_time = 0;
+		ast_mutex_unlock(&task_data3->lock);
+	}
+
+	/* Wait for tasks to complete */
+	if (task_data1) {
+		task_wait(task_data1);
```
jcolp: If the task fails to queue, this will wait indefinitely

main/taskprocessor.c
```diff
@@ -1212,22 +1338,27 @@ static int taskprocessor_push(struct ast_taskprocessor *tps, struct tps_task *t)
...
 	}
 
 	if (!t) {
-		ast_log(LOG_ERROR, "t is NULL!\n");
+		ast_log(LOG_ERROR, "Task is NULL!\n");
 		return -1;
 	}
 
+	if (log_task_origin && t->file) {
+		ast_log(LOG_DEBUG, "Taskprocessor '%s': Task pushed from %s:%d (%s)\n",
```
jcolp: This will be very chatty, I think it should be ast_debug(3

main/taskprocessor.c
```diff
@@ -1212,22 +1338,27 @@ static int taskprocessor_push(struct ast_taskprocessor *tps, struct tps_task *t)
...
+	}
+
 	ao2_lock(tps);
 	AST_LIST_INSERT_TAIL(&tps->tps_queue, t, list);
 	previous_size = tps->tps_queue_size++;
 
 	if (tps->tps_queue_high <= tps->tps_queue_size) {
 		if (!tps->high_water_alert) {
-			ast_log(LOG_WARNING, "The '%s' task processor queue reached %ld scheduled tasks%s.\n",
+			ast_log(LOG_WARNING, "Taskprocessor '%s' queue reached %ld scheduled tasks (high water mark)%s.\n",
```
jcolp: Let's go ahead and add the high water mark to this message too

main/taskprocessor.c
```diff
@@ -136,6 +150,9 @@ static AST_VECTOR_RW(subsystem_alert_vector, struct subsystem_alert *) overloade
 /*! \brief tps_singletons is the astobj2 container for taskprocessor singletons */
 static struct ao2_container *tps_singletons;
 
+/*! \brief Flag to enable logging of task origin (file/line/function) */
+static int log_task_origin = 0;
```
seanbright: Initialization isn't necessary

main/taskprocessor.c
```diff
@@ -390,43 +424,51 @@ int ast_tps_init(void)
 }
 
 /* allocate resources for the task */
-static struct tps_task *tps_task_alloc(int (*task_exe)(void *datap), void *datap)
+static struct tps_task *tps_task_alloc(int (*task_exe)(void *datap), void *datap
+	, const char *file, int line, const char *function)
```
seanbright: Move the leading comma to the end of the previous line

main/taskprocessor.c
```diff
@@ -390,43 +424,51 @@ int ast_tps_init(void)
...
+	t->file = file;
+	t->line = line;
+	t->function = function;
 
 	return t;
 }
 
-static struct tps_task *tps_task_alloc_local(int (*task_exe)(struct ast_taskprocessor_local *local), void *datap)
+static struct tps_task *tps_task_alloc_local(int (*task_exe)(struct ast_taskprocessor_local *local), void *datap
+	, const char *file, int line, const char *function)
```
seanbright: Move the leading comma to the end of the previous line

main/taskprocessor.c
```diff
@@ -1242,14 +1373,26 @@ static int taskprocessor_push(struct ast_taskprocessor *tps, struct tps_task *t)
 	return 0;
 }
 
+int ast_taskprocessor_push_origin(struct ast_taskprocessor *tps, int (*task_exe)(void *datap), void *datap
+	, const char *file, int line, const char *function)
```
seanbright: Move the leading comma to the end of the previous line

main/taskprocessor.c
```diff
@@ -1242,14 +1373,26 @@ static int taskprocessor_push(struct ast_taskprocessor *tps, struct tps_task *t)
...
+}
+
 int ast_taskprocessor_push(struct ast_taskprocessor *tps, int (*task_exe)(void *datap), void *datap)
 {
-	return taskprocessor_push(tps, tps_task_alloc(task_exe, datap));
+	return ast_taskprocessor_push_origin(tps, task_exe, datap, NULL, 0, NULL);
+}
+
+int ast_taskprocessor_push_local_origin(struct ast_taskprocessor *tps, int (*task_exe)(struct ast_taskprocessor_local *datap), void *datap
+	, const char *file, int line, const char *function)
```
seanbright: Move the leading comma to the end of the previous line

seanbright (changes requested):

main/taskprocessor.c
mbradeen: It is currently behind it's own cli command, do we want both?

jcolp: The log message itself could be useful to understand trends by examining the log, but just not at this level. It either needs to be enabled separately OR debug level bumped up high. I'd suggest seeing how chatty it is in the first place on an active system, and if it just makes the log unusable - a separate option.

mbradeen: It is indeed very chatty. Running it locally, I can only see it being useful when `core show taskprocessor name XXX` isn't showing the stack because the tasks are being executed quickly.

mbradeen:
Converting to draft while taskpool and threadpool are addressed

include/asterisk/res_pjsip.h
```diff
@@ -2076,7 +2076,11 @@ struct ast_sip_endpoint *ast_sip_dialog_get_endpoint(pjsip_dialog *dlg);
  * \retval 0 Success
  * \retval -1 Failure
  */
-int ast_sip_push_task(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data);
+int ast_sip_push_task_origin(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data,
```
gtjoseph: What's the significance of "origin"?  Seems an odd choice of words.

gtjoseph (changes requested):
The commit message has a typo...  "Seting 'core set..." should be "Setting".

The new unit test shouldn't be a UserNote.  It's not really a "user" thing.  Also, the text needs to start on the same line as the UserNote.

include/asterisk/res_pjsip.h
mbradeen: because it tracks the origin of the call.

jcolp: Looking at the other header files the predominant approach is to instead prefix calls that accept file/line/function with __

include/asterisk/taskprocessor.h
```diff
@@ -95,8 +95,12 @@ struct ast_taskprocessor_listener_callbacks {
 	 *
 	 * \param listener The listener
 	 * \param was_empty If non-zero, the taskprocessor was empty prior to the task being pushed
+	 * \param file The file where the task was pushed from
+	 * \param line The line number where the task was pushed from
+	 * \param function The function where the task was pushed from
 	 */
-	void (*task_pushed)(struct ast_taskprocessor_listener *listener, int was_empty);
+	void (*task_pushed)(struct ast_taskprocessor_listener *listener, int was_empty,
+		const char *file, int line, const char *function);
```
gtjoseph: Adding these parameters to the callback breaks ABI.

res/res_pjsip.c
```diff
@@ -2096,13 +2096,50 @@ struct ast_taskprocessor *ast_sip_create_serializer(const char *name)
 	return ast_sip_create_serializer_group(name, NULL);
 }
 
-int ast_sip_push_task(struct ast_taskprocessor *serializer, int (*sip_task)(void *), void *task_data)
+#undef ast_sip_push_task
+#define ast_sip_push_task_internal(serializer, sip_task, task_data) \
```
gtjoseph: ast_sip_push_task_internal doesn't seem to be used anywhere.

include/asterisk/threadpool.h
```diff
@@ -45,10 +45,13 @@ struct ast_threadpool_listener_callbacks {
...
 	 * \param listener The threadpool listener
 	 * \param was_empty Indicates whether there were any tasks prior to adding the new one.
+	 * \param file The file where the task was pushed from
+	 * \param line The line number where the task was pushed from
+	 * \param function The function where the task was pushed from
 	 */
 	void (*task_pushed)(struct ast_threadpool *pool,
 			struct ast_threadpool_listener *listener,
-			int was_empty);
+			int was_empty, const char *file, int line, const char *function);
```
gtjoseph: Adding these parameters to the callback breaks ABI.

include/asterisk/taskprocessor.h
gtjoseph: I see that the extra parameters were removed but another callback wasn't added.  What was the resolution then?

include/asterisk/threadpool.h
gtjoseph: Same

PR #1571, opened 2025-10-31
build: Add menuselect options to facilitate code tracing and coverage

gtjoseph:
The following options have been added to the menuselect "Compiler Flags"
section...

CODE_COVERAGE: The ability to enable code coverage via the `--enable-coverage`
configure flag has existed for many years but changing it requires
re-running ./configure which is painfully slow.  With this commit, you can
now enable and disable it via menuselect. Setting this option adds the
`-ftest-coverage` and `-fprofile-arcs` flags on the gcc and ld command lines.
It also sets DONT_OPTIMIZE. Note: If you use the `--enable-coverage` configure
flag, you can't turn it off via menuselect so choose one method and stick to
it.

KEEP_FRAME_POINTERS: This option sets `-fno-omit-frame-pointers` on the gcc
command line which can facilitate debugging with 'gdb' and tracing with 'perf'.
Unlike CODE_COVERAGE, this option doesn't depend on optimization being
disabled.  It does however conflict with COMPILE_DOUBLE.

### DIFF
--- a/Makefile.rules
+++ b/Makefile.rules
@@ -82,17 +82,21 @@ ifeq ($(CC),gcc)
 	endif
 endif
 
-ifeq ($(findstring DONT_OPTIMIZE,$(MENUSELECT_CFLAGS))$(AST_CODE_COVERAGE),no)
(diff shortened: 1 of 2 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

mbradeen (approved):

Issue #1572, opened 2025-11-03
[improvement]: List modules at various support levels


chrsmj:
### Improvement Description

This could probably be a separate contrib shell script but for some reason I recall saying I would try to add it to the Makefile -- maybe so it gets more attention in the future IDK.

PR #1573, opened 2025-11-03
Makefile: Add support-list-* targets.

chrsmj:
Convenience wrappers for showing modules at various support levels.

* support-list-core
* support-list-extended
* support-list-deprecated

Resolves: #1572

UserNote: Try "make support-list-deprecated" to see what files are on their way out the door.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,13 +108,13 @@ export XMLLINT
 export XMLSTARLET
 
 # makeopts is required unless the goal is just {dist{-}}clean
-# or support-list-{core,extended,deprecated}
+# or module-list-{core,extended,deprecated}
 ifeq ($(MAKECMDGOALS),clean)
 else ifeq ($(MAKECMDGOALS),distclean)
 else ifeq ($(MAKECMDGOALS),dist-clean)
-else ifeq ($(MAKECMDGOALS),support-list-core)
-else ifeq ($(MAKECMDGOALS),support-list-extended)
-else ifeq ($(MAKECMDGOALS),support-list-deprecated)
+else ifeq ($(MAKECMDGOALS),module-list-core)
+else ifeq ($(MAKECMDGOALS),module-list-extended)
+else ifeq ($(MAKECMDGOALS),module-list-deprecated)
 else
 include makeopts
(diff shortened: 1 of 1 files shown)


chrsmj:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

seanbright:
> I think `list-(core|extended|deprecated)-modules` might sound a bit better.

Bikeshedding: having the variable part at the end makes it easier to up-arrow/backspace/alter/enter than having it in the middle.

seanbright:
How about `module-list-(core|extended|deprecated)`?

chrsmj:
It does grab some things, like tests, that are not modules themselves but still marked as core support level. (So my original post was a little off.)

gtjoseph:
> How about `module-list-(core|extended|deprecated)`?

That works for me.

Makefile
```diff
@@ -612,6 +616,14 @@ oldmodcheck:
 		echo " WARNING WARNING WARNING" ;\
 	fi
 
+# The support-list-* independent targets are for evaluation of the source tree
+# by users interested in which files are at which support levels, sans compile.
+support-list-core support-list-extended support-list-deprecated:
+	@s=$(subst support-list-,,$@) ;\
+	echo "Showing all files at support level '$$s':" ;\
+	find . -name *.c  -exec grep -l "<support_level>$$s</support_level>" {} + | sort ;\
```
gtjoseph: You can replace this with one call that also excludes the test and utils directories.

$(GREP) -l -r --include='*.c' --include='*.cc' --exclude-dir=tests --exclude-dir=utils "<support_level>$$s</support_level>" | sort

Depending on intention, you may also want to exclude the main directory since they're not separate modules.

chrsmj: Thanks -- made change but hard coded the grep because this runs before configure.

chrsmj:
Argh I borked the merge commit reversal.

chrsmj:
Abandoning due to merge commit borkage in favor of https://github.com/asterisk/asterisk/pull/1596

Issue #1578, opened 2025-11-06
[bug]: Deadlock with externalMedia custom channel id and cpp map channel backend


gtjoseph:
### Severity

Blocker

### Versions

all

### Components/Modules

channel.c, channelstorage_map_name_id.cc

### Operating Environment

All

### Frequency of Occurrence

Occasional

### Issue Description

After PR #1498 added read locking to channelstorage_cpp_map_name_id, if ARI channels/externalMedia is called with a custom channel id AND the cpp_map_name_id channel storage backend is in use, a deadlock can occur when hanging up the channel.  It's actually triggered in channel.c:__ast_channel_alloc_ap() when it gets a write lock on the channelstorage driver then subsequently does a lookup for channel uniqueid which now does a read lock.  This is an invalid operation and causes the lock state to get "bad".  When the channels try to hang up a write lock is attempted which hangs and causes the deadlock.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1579, opened 2025-11-06
channelstorage:  Allow storage driver read locking to be skipped.

gtjoseph:
After PR #1498 added read locking to channelstorage_cpp_map_name_id, if ARI
channels/externalMedia was called with a custom channel id AND the
cpp_map_name_id channel storage backend is in use, a deadlock can occur when
hanging up the channel. It's actually triggered in
channel.c:__ast_channel_alloc_ap() when it gets a write lock on the
channelstorage driver then subsequently does a lookup for channel uniqueid
which now does a read lock. This is an invalid operation and causes the lock
state to get "bad". When the channels try to hang up, a write lock is
attempted again which hangs and causes the deadlock.

Now instead of the cpp_map_name_id channelstorage driver "get" APIs
automatically performing a read lock, they take a "lock" parameter which
allows a caller who already has a write lock to indicate that the "get" API
must not attempt its own lock.  This prevents the state from getting mesed up.

The ao2_legacy driver uses the ao2 container's recursive mutex so doesn't
have this issue but since it also implements the common channelstorage API,
it needed its "get" implementations updated to take the lock parameter. They
just don't use it.

Resolves: #1578

### DIFF
--- a/main/channel.c
+++ b/main/channel.c
@@ -498,7 +498,7 @@ void ast_softhangup_all(void)
 /*! \brief returns number of active/allocated channels */
 int ast_active_channels(void)
 {
-	return current_channel_storage_instance ? CHANNELSTORAGE_API(current_channel_storage_instance, active_channels) : 0;
+	return current_channel_storage_instance ? CHANNELSTORAGE_API(current_channel_storage_instance, active_channels, 1) : 0;
 }
 
 int ast_undestroyed_channels(void)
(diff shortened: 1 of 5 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):
This gives me flashbacks to when Asterisk didn't have/use recursive mutexes, ha.

Issue #1444, opened 2025-09-10
[bug]: static code analysis issues in bridges/bridge_simple.c


aol-nnov:
### Severity

Major

### Versions

18..master

### Components/Modules

bridges/bridge_simple.c

### Operating Environment

static code analysis 

### Frequency of Occurrence

None

### Issue Description

Logic operator precedence in assignment can produce an unexpected result  at bridges/bridge_simple.c:160 correct snippet follows

```
if ((unhold_c1 = ast_channel_hold_state(c1)) == AST_CONTROL_HOLD) {
```

Logic operator precedence in assignment can produce an unexpected result  at bridges/bridge_simple.c:164

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1582, opened 2025-11-07
bridge_simple: Increase code verbosity for clarity.

seanbright:
There's no actual problem here, but I can see how it might by confusing.

Resolves: #1444


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1442, opened 2025-09-10
[bug]: static code analysis issues in main/bridge_basic.c


aol-nnov:
### Severity

Major

### Versions

18..master

### Components/Modules

main/bridge_basic.c

### Operating Environment

static code analysis 

### Frequency of Occurrence

None

### Issue Description

Missing break at the end of 'case STIMULUS_TIMEOUT' at line 2384 (if that was intentional, there is an attribute in gcc to mark explicit fallthrough and thus make things clear - `__attribute__((fallthrough));`)

Missing break at the end of 'case STIMULUS_TIMEOUT' at line 2695

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

jcolp:
This is on purpose.

aol-nnov:
> This is on purpose.

it should be marked as such then..

jcolp:
Yes, and I accepted this issue. I put a note in case someone decided to fix this and did not look at things/know that.

PR #1583, opened 2025-11-07
main: Explicitly mark case statement fallthrough as such.

seanbright:
Resolves: #1442


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1421, opened 2025-09-08
[bug]: static code analysis issues in apps/app_dtmfstore.c


aol-nnov:
### Severity

Major

### Versions

18..master

### Components/Modules

apps/app_dtmfstore.c

### Operating Environment

static code analysis

### Frequency of Occurrence

None

### Issue Description

An element of array '&varnamesub[0]' of size 64, declared at app_dtmfstore.c:145, is accessed by an index with values in [0, +inf] at app_dtmfstore.c:173, which may lead to a buffer overflow. Details: Format string: 'Optional[${%s}]'.Filled values can have unlimited string length.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1584, opened 2025-11-07
app_dtmfstore: Avoid a potential buffer overflow.

seanbright:
Prefer snprintf() so we can readily detect if our output was truncated.

Resolves: #1421


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

Issue #1587, opened 2025-11-11
[bug]: chan_websocket terminates websocket on CNG/non-audio


jcbjoe:
### Severity

Minor

### Versions

22.6.0

### Components/Modules

chan_websocket

### Operating Environment

Ubuntu 24.04

### Frequency of Occurrence

Constant

### Issue Description

Hi Team,

It seems with the new chan_websocket that if comfort noise is sent it terminates the websocket connection.

This is due to the following snippet
```
if (f->frametype != AST_FRAME_VOICE) {
                ast_log(LOG_WARNING, "%s: This WebSocket channel only supports AST_FRAME_VOICE frames\n",
                        ast_channel_name(ast));
                return -1;
        } 
```

The above snippet causes the websocket to be terminated if the frame is AST_FRAME_CNG - comfort noise.

From the Asterisk side we cannot ignore the comfort noise and our clients dont have the option to turn it off. To this is causing us a few issues.

### Relevant log output

```shell
This WebSocket channel only supports AST_FRAME_VOICE frames
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

jcolp:
"From the Asterisk side we cannot ignore the comfort noise"

If you mean you don't know of a way, the FRAME_DROP dialplan function[1] should be capable of doing so as an immediate workaround.

[1] https://docs.asterisk.org/Latest_API/API_Documentation/Dialplan_Functions/FRAME_DROP/

jcbjoe:
Hey @jcolp I tried to use FRAME_DROP but this didn't seem to work for chan_websocket.

PR #1588, opened 2025-11-12
chan_websocket.c: Tolerate other frame types

jcbjoe:
Currently, if chan_websocket receives an un supported frame like comfort noise it will exit the websocket. The proposed change is to tolerate the other frames by not sending them down the websocket but instead just ignoring them.

Resolves: https://github.com/asterisk/asterisk/issues/1587

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -1099,10 +1099,14 @@ static int webchan_write(struct ast_channel *ast, struct ast_frame *f)
 		return -1;
 	}
 
+	if (f->frametype == AST_FRAME_CNG) {
+		return 0;
+	}
+
 	if (f->frametype != AST_FRAME_VOICE) {
 		ast_log(LOG_WARNING, "%s: This WebSocket channel only supports AST_FRAME_VOICE frames\n",
 			ast_channel_name(ast));
-		return -1;
+		return 0;
 	}
 
 	if (ast_format_cmp(f->subclass.format, instance->native_format) == AST_FORMAT_CMP_NOT_EQUAL) {


channels/chan_websocket.c
```diff
@@ -1099,6 +1099,10 @@ static int webchan_write(struct ast_channel *ast, struct ast_frame *f)
 		return -1;
 	}
 
+	if (f->frametype == AST_FRAME_CNG) {
+		return 0;
+    }
+
 	if (f->frametype != AST_FRAME_VOICE) {
 		ast_log(LOG_WARNING, "%s: This WebSocket channel only supports AST_FRAME_VOICE frames\n",
 			ast_channel_name(ast));
```
jcolp, line 1108: I think this should also be changed to be tolerant and return 0, looking at other drivers they are tolerant of such things.

jcbjoe: Yes, this makes sense. I have committed this change

jcbjoe:
> I should have been more specific. This should have been in addition to your original change of suppressing the log message for CNG and making it tolerant for that as well.

Okay, added back my original code. This makes sense actually if we are receiving a lot of comfort noise. We don't want to spam the log.

jcolp (approved):
The PR checklist needs to be taken care of, but from a code perspective this is fine.

jcbjoe:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcbjoe:
FYI @gtjoseph Commit message updated to add the resolves.

PR #1589, opened 2025-11-12
ast_coredumper: Fix multiple issues

gtjoseph:
* Fixed an issue with tarball-coredumps when asterisk was invoked without an
absolute path.

* Fixed an issue with gdb itself segfaulting when trying to get symbols from
separate debuginfo files.  The command line arguments needed to be altered
such that the gdbinit files is loaded before anything else but the
`dump-asterisk` command is run after full initialization.

In the embedded gdbinit script:

* The extract_string_symbol function needed a `char *` cast to work properly.

* The s_strip function needed to be updated to continue to work with the
cpp_map_name_id channel storage backend.

* A new function was added to dump the channels when cpp_map_name_id was
used.

* The Channel object was updated to account for the new channel storage
backends

* The show_locks function was refactored to work correctly.

### DIFF
--- a/contrib/scripts/ast_coredumper
+++ b/contrib/scripts/ast_coredumper
@@ -266,6 +266,8 @@ for i in "${!COREDUMPS[@]}" ; do
 		libdir=$(dirname "${libfile}")
 	}
 
+	[ "$(dirname "${astbin}")" == "." ] && astbin="$(which "${astbin}")" || :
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1590, opened 2025-11-13
[bug]: Fix: Use ast instead of p->chan to get the DIALSTATUS variable


Tinet-mucw:
### Severity

Trivial

### Versions

23.0.0

### Components/Modules

main

### Operating Environment

ubuntu 24.04

### Frequency of Occurrence

None

### Issue Description

After p->chan = NULL, ast still points to the valid channel object
Using ast safely accesses the channel's DIALSTATUS variable before it's fully destroyed

``` c
[root@bogon asterisk]# git diff
diff --git a/asterisk/asterisk-22.0.0/main/core_unreal.c b/asterisk/asterisk-22.0.0/main/core_unreal.c
index 2f3fc73f2..885cc9e26 100644
--- a/asterisk/main/core_unreal.c
+++ b/asterisk/main/core_unreal.c
@@ -1045,7 +1045,7 @@ int ast_unreal_hangup(struct ast_unreal_pvt *p, struct ast_channel *ast)
                ast_clear_flag(p, AST_UNREAL_CARETAKER_THREAD);
                p->chan = NULL;
                if (p->owner) {
-                       const char *status = pbx_builtin_getvar_helper(p->chan, "DIALSTATUS");
+                       const char *status = pbx_builtin_getvar_helper(ast, "DIALSTATUS");
 
                        if (status) {
                                ast_channel_hangupcause_set(p->owner, cause);
```

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

jcolp:
Are you going to put a PR up for this?

PR #1591, opened 2025-11-14
core_unreal.c: Use ast instead of p->chan to get the DIALSTATUS variable

Tinet-mucw:
After p->chan = NULL, ast still points to the valid channel object, using ast safely accesses the channel's DIALSTATUS variable before it's fully destroyed

Resolves: #1590


Tinet-mucw:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1592, opened 2025-11-15
[bug]: app_disa: ResetCDR warning on most invocations


InterLinked1:
### Severity

Trivial

### Versions

23, 22, 21 (but not 20)

### Components/Modules

app_disa

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Any time the DISA() application is executed, the following warning is emitted:

```
[2025-11-15 01:37:11.381] WARNING[957781][C-000044f0]: app.c:3497 parse_options: Unrecognized option: 'e'
```

The culprit is its usage of `ResetCDR`:

```
if (ast_pbx_exec_application(chan, "ResetCDR", special_noanswer ? "" : "e")) {
	ast_log(AST_LOG_NOTICE, "ResetCDR application not found; CDR will not be reset\n");
}
```

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1593, opened 2025-11-15
app_disa: Avoid use of removed ResetCDR() option.

InterLinked1:
Commit a46d5f9b760f84b9f27f594b62507c1443aa661b removed the deprecated 'e' option to ResetCDR; this now causes DISA() to emit a warning if attempting to call ResetCDR() with the deprecated option (in all cases except when the no answer option is provided). Rewrite the code to do this the current way.

Resolves: #1592


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22

jcolp (approved):

Issue #1570, opened 2025-10-31
[bug]: DTMF detection issue on DAHDI channels


aol-nnov:
### Severity

Major

### Versions

master

### Components/Modules

main/dsp.c

### Operating Environment

Debian 12

### Frequency of Occurrence

Constant

### Issue Description

The is an Asterisk server with 16 DAHDI channels, divided into two groups:
- channels 1 to 8 are configured as FXS ports
- channels 9 to 16 are configured as FXO ports

Initial setup:
- Two analog telephone sets are connected to the FXS port 1 (PSTN subscriber A) and FXS port 2 (PSTN subscriber B).
- FXS port 3 is connected to FXO port 9 with copper wire

We're modelling (simulating) a long distance call from "PSTN subscriber A" to "PSTN subscriber B" and "Copper wire" is, well, plain half-a-meter wire, not a long line simulator or whatsoever and does not affect signal in any way.

Steps to reproduce:
- "PSTN subscriber A" initiates call to FXO port 9 and after the call is accepted on FXS port 3
- by now we have FXS port 1 bridged with FXO port 9 on one leg and active channel on FXS port 3.
- Then "PSTN subscriber A" starts dialing DTMF (inband) and dialed numbers should be detected in the active channel of FXS port 3.
- After "PSTN subscriber A" finishes dialing, if dialed number matches "PSTN subscriber B", they should receive incoming call and FXS port 3 should be bridged with FXS port 2 forming another leg of a long-distance call.

The issue:
When "PSTN subscriber A" sends DTMF digits, on receiving side (FXS port 3) audio signal is distorted (fragmented) which leads to detection of extra numbers as if subscriber A dialed "222" when it actually he was holding "2" all that time.

Analysis revealed that the incoming DTMF signals on the receiving channel (FXO port 9) are highly teared - it was discovered by an acoustic tap to the copper wire. Disabling the "busydetect" parameter in the chan_dahdi.conf configuration file helps to detect dialed DTMF numbers correctly, but in production we're unable to disable busydetect on ports.

Presumably, the cause of our issue is that when "busydetect" is enabled, the dtmf_detect function is triggered twice:
- The first time on the FXO port 9 that is bridged with FXS port 1. This is done to prevent false positives from the busy detection function (busydetect).
- The second time on the receiving DAHDI channel FXS port 3, where DTMF detection takes place. 
[...]

PR #1595, opened 2025-11-17
dsp.c: Avoid running digit detection twice if busydetect is enabled.

InterLinked1:
When setting busydetect=yes in chan_dahdi.conf, chan_dahdi uses a second DSP to process audio on the channel, in addition to the one that exists normally. Within ast_dsp_process, the conditional guarding digit detection (either mf_detect or dtmf_detect) was true if either DSP_FEATURE_DIGIT_DETECT or DSP_FEATURE_BUSY_DETECT was set. Because digits are normally quelched unless DSP_DIGITMODE_NOQUELCH is set, the first pass would quelch digits on the channel and the second pass would see discrete DTMF tones instead of the continuous tone that is actually present.

This second digit detection on the frame should not be happening, and DSP_FEATURE_BUSY_DETECT should not be part of this condition in the first place. While the busy detector gets reset if a digit is detected, this is done even without it being part of this conditional. This regression was introduced in commit ec0581fef4e44b029588cd1ce0539dffa36db94a.

Removing it ensures that mf_detect/dtmf_detect is only called once per frame, since busy detection doesn't need digit detection to work and digit detection shouldn't be performed twice.

Resolves: #1570

### DIFF
--- a/main/dsp.c
+++ b/main/dsp.c
@@ -1571,7 +1571,7 @@ struct ast_frame *ast_dsp_process(struct ast_channel *chan, struct ast_dsp *dsp,
 		dsp->f.frametype = AST_FRAME_CONTROL;
 		dsp->f.subclass.integer = AST_CONTROL_BUSY;
 		ast_frfree(af);
-		ast_debug(1, "Requesting Hangup because the busy tone was detected on channel %s\n", ast_channel_name(chan));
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

asteriskteam:

asteriskteam:

Issue #1597, opened 2025-11-17
[bug]: app_reload: Reload() without arguments doesn't work.


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

app_reload

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

Calling Reload() without any arguments is supposed to reload everything (equivalent to a 'core reload'), but actually does nothing. This is because it was calling ast_module_reload with an empty string, and the argument needs to explicitly be NULL.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1598, opened 2025-11-17
app_reload: Fix Reload() without arguments.

InterLinked1:
Calling Reload() without any arguments is supposed to reload everything (equivalent to a 'core reload'), but actually does nothing. This is because it was calling ast_module_reload with an empty string, and the argument needs to explicitly be NULL.

Resolves: #1597


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1599, opened 2025-11-17
[bug]: pbx.c: Running "dialplan reload" shows wrong number of contexts


InterLinked1:
### Severity

Trivial

### Versions

Git

### Components/Modules

pbx.c

### Operating Environment

N/A

### Frequency of Occurrence

Constant

### Issue Description

When running "dialplan reload", the number of contexts it prints out is the old number, not the new number. The command has to be run twice for it to correctly show how many contexts are loaded.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1600, opened 2025-11-17
pbx.c: Print new context count when reloading dialplan.

InterLinked1:
When running "dialplan reload", the number of contexts reported is initially wrong, as it is the old context count. Running "dialplan reload" a second time returns the correct number of contexts that are loaded. This can confuse users into thinking that the reload didn't work successfully the first time.

This counter is currently only incremented when iterating the old contexts prior to the context merge; at the very end, get the current number of elements in the context hash table and report that instead. This way, the count is correct immediately whenever a reload occurs.

Resolves: #1599


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

PR #1602, opened 2025-11-20
ccss:  Add option to ccss.conf to globally disable it.

gtjoseph:
The Call Completion Supplementary Service feature is rarely used but many of
it's functions are called by app_dial and channel.c "just in case".  These
functions lock and unlock the channel just to see if CCSS is enabled on it,
which it isn't 99.99% of the time.

UserNote: A new "enabled" parameter has been added to ccss.conf.  It defaults
to "yes" to preserve backwards compatibility but CCSS is rarely used so
setting "enabled = no" in the "general" section can save some unneeded channel
locking operations and log message spam.  Disabling ccss will also prevent
the func_callcompletion and chan_dahdi modules from loading.

DeveloperNote: A new API ast_is_cc_enabled() has been added.  It should be
used to ensure that CCSS is enabled before making any other ast_cc_* calls.

### DIFF
--- a/main/ccss.c
+++ b/main/ccss.c
@@ -115,6 +115,16 @@
  * them here at the top is easiest.
  */
 
+static int global_enabled;
+
+#define RETURN_IF_NOT_ENABLED(rc) \
+({ \
+	if (!global_enabled) { \
+		ast_debug(3, "CCSS disabled globally\n"); \
+		return rc; \
+	} \
+})
+
 /*!
  * The ast_sched_context used for all generic CC timeouts
  */
@@ -463,6 +473,9 @@ struct ast_cc_agent *ast_cc_agent_callback(int flags, ao2_callback_fn *function,
 {
 	struct cc_callback_helper helper = {.function = function, .args = args, .type = type};
 	struct cc_core_instance *core_instance;
+
+	RETURN_IF_NOT_ENABLED(NULL);
+
 	if ((core_instance = ao2_t_callback(cc_core_instances, flags, cc_agent_callback_helper, &helper,
 					"Calling provided agent callback function"))) {
 		struct ast_cc_agent *agent = cc_ref(core_instance->agent, "An outside entity needs the agent");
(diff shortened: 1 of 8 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
At some point, ccss should probably be moved to a separate module and use OPTIONAL_API.

alex2grad:
I think it's a bad idea to simply return NULL/0 from ast_cc_* functions.
For example, ast_cc_config_params_init will return NULL if CCSS is disabled.
Then, ast_channel_cc_params_init will return -1 (error).
Then, ast_unreal_new_channels will fail.

I think it would be better to make Asterisk global option (for example disable_ccss) or compilation option

gtjoseph:
Well, I'd prefer not to propagate ccss stuff any further than it is by adding an asterisk.conf option but I could add an `ast_cc_is_enabled()` API call and use that in ~~chan_dahdi~~, core_unreal, etc.

Actually chan_dahdi depends on ccss.c so it won't load if it's disabled.

main/ccss.c
```diff
@@ -4426,7 +4506,40 @@ static void initialize_cc_devstate_map(void)
...
+	if ((enabled = ast_variable_retrieve(cc_config, "general", "enabled"))) {
+		global_enabled = ast_true(enabled);
+	}
+
+	if (global_enabled) {
+		initialize_cc_max_requests(cc_config);
+		initialize_cc_devstate_map(cc_config);
+	} else {
+		ast_log(LOG_NOTICE, "CCSS disabled globally\n");
+	}
```
bkford, line 4534: Would this make more sense to check at the top of the function and bail early if not global_enabled? That way you don't have to allocate or destroy anything.

gtjoseph: No because we have to load the config first to get the value of `global_enabled`.

bkford: Ah, that's annoying. Makes sense.

PR #1608, opened 2025-11-25
configure.ac: use AC_PATH_TOOL for nm

nwf:
`nm` might, especially in cross-compilation scenarios, be available but prefixed with the target triple.  So: use `AC_PATH_TOOL` rather than `AC_PATH_PROG` to find it.  (See https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Programs.html .)

Found and proposed fix tested by cross-compiling Asterisk using Nixpkgs on x86_64 targeting aarch64. :)

### DIFF
--- a/configure
+++ b/configure
@@ -19025,8 +19025,9 @@ printf "%s\n" "no" >&6; }
 fi
 
 
-# Extract the first word of "nm", so it can be a program name with args.
-set dummy nm; ac_word=$2
+if test -n "$ac_tool_prefix"; then
+  # Extract the first word of "${ac_tool_prefix}nm", so it can be a program name with args.
+set dummy ${ac_tool_prefix}nm; ac_word=$2
 { printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for $ac_word" >&5
 printf %s "checking for $ac_word... " >&6; }
 if test ${ac_cv_path_NM+y}
@@ -19057,7 +19058,6 @@ done
   done
 IFS=$as_save_IFS
 
-  test -z "$ac_cv_path_NM" && ac_cv_path_NM=":"
   ;;
 esac ;;
 esac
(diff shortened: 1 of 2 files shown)


jcolp:
The PR check also needs to be fulfilled and branches specified.

nwf:
Running `./bootstrap.sh` on 5d15a42a makes changes to `./configure`.  I've pulled those out to a commit prior to my change, and then ran `./bootstrap.sh` again after `AC_PATH_TOOL`-ization of `nm` as a 3rd commit.  Please feel free to squash appropriately, or tell me that I have done something wrong because the first of those should be empty.

jcolp:
Squash it all down to one commit, it's fine.

gtjoseph:
cherry-pick-to: 22
cherry-pick-to: 20

gtjoseph:
@nwf This PR has been sitting for a while.  If you want it merged, you need to fix the commit message and PR description so they match.

nwf:
Ah, sorry.  Like that?

gtjoseph:
> Ah, sorry. Like that?

Yep, that did it.

gtjoseph (changes requested):
This PR doesn't cherry-pick cleanly to the 23 branch although it does cherry-pick cleanly to 22 and 20.  You're going to need to do the following...

Edit the `cherry-pick-to` comment in this PR and remove the `cherry-pick-to: 23` line.

In your development repo...
```shell
$ git checkout 23
$ git cherry-pick 2228239998f06185fec21
# It's going to fail but that's OK.
$ git add configure
# configure will be invalid but that's OK.
$ git cherry-pick --continue
$ ./bootstrap.sh
# configure will now be regenerated.
$ git commit -a --amend --no-edit
# Create a new PR just for 23.
$ gh pr create --fill --base 23

Now add a `cherry-pick-to: none` comment to the new PR.

Issue #1609, opened 2025-11-26
[bug]: Crash: Double free in ast_channel_destructor leading to SIGABRT (Asterisk 20.17.0) with C++ channel storage


phoneben:
### Severity

Trivial

### Versions

20.17.0

### Components/Modules

core

### Operating Environment

centos 9

### Frequency of Occurrence

None

### Issue Description

Asterisk 20.17.0 crashes with SIGABRT due to double-free memory corruption in ast_channel_destructor() triggered during channel teardown while MixMonitor threads and bridging threads are active. Crash confirmed via glibc _int_free() abort.

[core-asterisk-2025-11-26T07-59-01Z-brief-sanitized.txt](https://github.com/user-attachments/files/23766022/core-asterisk-2025-11-26T07-59-01Z-brief-sanitized.txt)
[core-asterisk-2025-11-26T07-59-01Z-full-sanitized.txt](https://github.com/user-attachments/files/23766025/core-asterisk-2025-11-26T07-59-01Z-full-sanitized.txt)
[core-asterisk-2025-11-26T07-59-01Z-info-sanitized.txt](https://github.com/user-attachments/files/23766024/core-asterisk-2025-11-26T07-59-01Z-info-sanitized.txt)
[core-asterisk-2025-11-26T07-59-01Z-locks-sanitized.txt](https://github.com/user-attachments/files/23766021/core-asterisk-2025-11-26T07-59-01Z-locks-sanitized.txt)
[core-asterisk-2025-11-26T07-59-01Z-thread1-sanitized.txt](https://github.com/user-attachments/files/23766023/core-asterisk-2025-11-26T07-59-01Z-thread1-sanitized.txt)

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

phoneben:
what i see in the logs somehow there was a mixup  pbx02-1764143921.2090277 
call C-0000f272 had a CDRcallUniqueID 
and somehow this was set on C-0000f200 which is earlier call that was ended after C-0000f272 mostprobily causing the SIGABRT 
am i correct ?
could this be related to the last cpp fix ?
 gtjoseph  could this be the issue ? https://github.com/asterisk/asterisk/pull/1579

```
[root@pbx02 asterisk]# grep pbx02-1764143921.2090277 full.0
[...]

PR #1610, opened 2025-11-26
channelstorage_cpp: Fix fallback return value in channelstorage callback

phoneben:
callback returned the last iterated channel when no match existed, causing invalid channel references and potential double frees. Updated to correctly return NULL when there is no match.

Resolves: #1609

### DIFF
--- a/main/channelstorage_cpp_map_name_id.cc
+++ b/main/channelstorage_cpp_map_name_id.cc
@@ -172,9 +172,8 @@ static struct ast_channel *callback(struct ast_channelstorage_instance *driver,
 		rdlock(driver);
 	}
 	for (it = getdb(driver).begin(); it != getdb(driver).end(); it++) {
-		chan = it->second;
-		if (cb_fn(chan, arg, data, ao2_flags) == (CMP_MATCH | CMP_STOP)) {
-			ao2_bump(chan);
+		if (cb_fn(it->second, arg, data, ao2_flags) == (CMP_MATCH | CMP_STOP)) {
+			chan = ao2_bump(it->second);
 			break;
 		}
 	}


main/channelstorage_cpp_map_name_id.cc
```diff
@@ -172,9 +173,10 @@ static struct ast_channel *callback(struct ast_channelstorage_instance *driver,
 		rdlock(driver);
 	}
 	for (it = getdb(driver).begin(); it != getdb(driver).end(); it++) {
-		chan = it->second;
-		if (cb_fn(chan, arg, data, ao2_flags) == (CMP_MATCH | CMP_STOP)) {
-			ao2_bump(chan);
+		candidate = it->second;
```
seanbright: You can declare `candidate` here. It’s only needed in this block.

seanbright (changes requested):

phoneben:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

main/channelstorage_cpp_map_name_id.cc
```diff
@@ -172,9 +172,10 @@ static struct ast_channel *callback(struct ast_channelstorage_instance *driver,
 		rdlock(driver);
 	}
 	for (it = getdb(driver).begin(); it != getdb(driver).end(); it++) {
-		chan = it->second;
-		if (cb_fn(chan, arg, data, ao2_flags) == (CMP_MATCH | CMP_STOP)) {
-			ao2_bump(chan);
+		struct ast_channel *candidate = it->second;
```
seanbright: On second thought - do we need `candidate` at all? Could we just pass `it->second` to `cb_fn` and only assign `chan` if that returns true?

main/channelstorage_cpp_map_name_id.cc
```diff
@@ -172,9 +172,9 @@ static struct ast_channel *callback(struct ast_channelstorage_instance *driver,
 		rdlock(driver);
 	}
 	for (it = getdb(driver).begin(); it != getdb(driver).end(); it++) {
-		chan = it->second;
-		if (cb_fn(chan, arg, data, ao2_flags) == (CMP_MATCH | CMP_STOP)) {
-			ao2_bump(chan);
+		if (cb_fn(it->second, arg, data, ao2_flags) == (CMP_MATCH | CMP_STOP)) {
+			ao2_bump(it->second);
+			chan = it->second;
```
seanbright: chan = ao2_bump(it->second);

Issue #1611, opened 2025-11-26
[bug]: asterisk deadlocked on start sometimes


phoneben:
### Severity

Trivial

### Versions

20.17.0

### Components/Modules

load_module

### Operating Environment

centos 9

### Frequency of Occurrence

Occasional

### Issue Description

asterisk crashed and did not restart fully .
i can not figure out what is wrong with my node why does it happen .

multiple restarts will work .

## Deadlock during res_manager_devicestate module load

**Issue:**
Asterisk deadlocks during module initialization and never completes startup.

**Evidence from core dump:**

Main thread (Thread 1) blocked during module load:
```
#0  __lll_lock_wait
#1  pthread_mutex_lock
#2  __ast_pthread_mutex_lock
#3  __ao2_lock
#4  topic_add_subscription at stasis.c:1273
#5  topic_add_subscription at stasis.c:1283
#6  stasis_forward_all at stasis.c:1680
#7  load_module() at res_manager_devicestate.c:137
#8  start_resource at loader.c:1881
```
(Heavy concurrent AMI queue show activity during startup)

Multiple other threads concurrently blocked on `__ao2_lock` while executing AMI `queue show` commands in `app_queue.c:find_queue_by_name_rt`.

**Analysis:**
`res_manager_devicestate` module initialization calls `stasis_forward_all()`, which attempts to acquire an AO2 mutex on a Stasis topic object. This mutex is held/contended by worker threads processing concurrent AMI requests, creating a deadlock. Main thread never completes `load_modules()`, causing startup hang.

[core-asterisk-running-2025-11-26T08-21-16Z-brief-sanitized.txt](https://github.com/user-attachments/files/23773018/core-asterisk-running-2025-11-26T08-21-16Z-brief-sanitized.txt)
[core-asterisk-running-2025-11-26T08-21-16Z-full-sanitized.txt](https://github.com/user-attachments/files/23773021/core-asterisk-running-2025-11-26T08-21-16Z-full-sanitized.txt)
[core-asterisk-running-2025-11-26T08-21-16Z-info-sanitized.txt](https://github.com/user-attachments/files/23773022/core-asterisk-running-2025-11-26T08-21-16Z-info-sanitized.txt)
[core-asterisk-running-2025-11-26T08-21-16Z-locks-sanitized.txt](https://github.com/user-attachments/files/23773019/core-asterisk-running-2025-11-26T08-21-16Z-locks-sanitized.txt)
[...]

PR #1614, opened 2025-11-29
stasis.c: Fix deadlock in stasis_topic_pool_get_topic during module load

phoneben:
stasis.c: Fix deadlock in stasis_topic_pool_get_topic during module load.

Deadlock occurs when res_manager_devicestate loads concurrently with
device state operations due to lock ordering violation:

Thread 1: Holds pool lock → needs topic lock (in stasis_forward_all)
Thread 2: Holds topic lock → needs pool lock (in stasis_topic_pool_get_topic)

Fix: Release pool lock before calling stasis_topic_create() and
stasis_forward_all(). Re-acquire only for insertion with race check.

Preserves borrowed reference semantics while breaking the deadlock cycle.

Fixes: #1611

### DIFF
--- a/main/stasis.c
+++ b/main/stasis.c
@@ -1776,6 +1776,22 @@ static void send_subscription_unsubscribe(struct stasis_topic *topic,
 struct topic_pool_entry {
 	struct stasis_forward *forward;
 	struct stasis_topic *topic;
+	/*
+	 * Per-entry initialization state. This lets us serialize creation of a
+	 * given topic name without holding the pool container lock while doing
+	 * the heavy lifting (topic creation, forwarding setup, etc).
+	 *
+	 * A topic_pool_entry starts life in an "in-flight" state where neither
+	 * initialized nor failed are set.  The first thread to link the entry
+	 * into the pool becomes the creator and is responsible for completing
+	 * initialization, setting one of the flags, and broadcasting init_cond.
+	 * Other threads that find the same entry simply wait for initialization
+	 * to complete and then reuse the created topic.
+	 */
+	ast_mutex_t init_lock;
+	ast_cond_t init_cond;
+	unsigned int initialized; /* terminal success state */
+	unsigned int failed; /* terminal failure state */
 	char name[0];
 };
 
@@ -1786,6 +1802,8 @@ static void topic_pool_entry_dtor(void *obj)
 	entry->forward = stasis_forward_cancel(entry->forward);
 	ao2_cleanup(entry->topic);
 	entry->topic = NULL;
+	ast_cond_destroy(&entry->init_cond);
+	ast_mutex_destroy(&entry->init_lock);
 }
 
 static struct topic_pool_entry *topic_pool_entry_alloc(const char *topic_name)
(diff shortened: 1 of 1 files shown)


phoneben:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

phoneben:
disclaimer ai was used .

phoneben:
ok tested now solves my original issue .
80 k call done

main/stasis.c
```diff
@@ -1968,27 +1989,60 @@ struct stasis_topic *stasis_topic_pool_get_topic(struct stasis_topic_pool *pool,
...
 	 * system we prefix it with the pooling topic name, which should itself already be unique.
 	 */
+	/* Build fully-qualified topic name */
 	ret = ast_asprintf(&new_topic_name, "%s/%s", stasis_topic_name(pool->pool_topic), topic_name);
 	if (ret < 0) {
+		ao2_ref(topic_pool_entry, -1);
 		return NULL;
 	}
 
 	topic_pool_entry->topic = stasis_topic_create(new_topic_name);
```
jcolp: As there is no longer a lock being held during this entire operation it is possible for stasis_topic_create to fail for some threads if multiple are requesting the same topic name at the same time. This is due to stasis_topic_create failing if the topic already exists. This would result in a race condition and unexpected behavior.

jcolp (dismissed):

phoneben:
Small race note: If thread A successfully creates the topic but hasn't yet
inserted it into the pool, thread B's stasis_topic_create() will fail and its
follow-up lookup may still not see the entry. In that tiny window B returns
NULL even though A already created the topic. This is a transient race and is
acceptable since callers must already handle NULL, and it avoids the original
deadlock risk.

main/stasis.c
phoneben: Addressed by adding a pool re-check when stasis_topic_create() fails, so that
a thread can return an existing topic if another thread created it first.

phoneben:
should i change it to ?
[better version.txt](https://github.com/user-attachments/files/24108796/better.version.txt)


A cleaner fix is to insert an in-flight placeholder entry into the pool before creating the topic.
Other threads then see the placeholder, wait for initialization, and always receive the correct topic — without reintroducing the deadlock.

phoneben:
i got more the 100 k on this .
asterisk started good 
needs a review

asteriskteam:

main/stasis.c
```diff
@@ -1797,9 +1815,11 @@ static struct topic_pool_entry *topic_pool_entry_alloc(const char *topic_name)
 	if (!topic_pool_entry) {
 		return NULL;
 	}
-
+	ast_mutex_init(&topic_pool_entry->init_lock);
+	ast_cond_init(&topic_pool_entry->init_cond, NULL);
+	topic_pool_entry->initialized = 0;
```
jcolp: ao2_alloc_options allocates with a zeroed structure, so setting these to 0 is not needed

phoneben:
> Workflow PRCheck failed [master-pjs1](https://github.com/asterisk/asterisk/actions/runs/21356831656/job/61466694308): FAILED TEST: channels/pjsip/geolocation/incoming/basic [master-pjs1](https://github.com/asterisk/asterisk/actions/runs/21356831656/job/61466694308): FAILED TEST: channels/pjsip/geolocation/incoming/profile_precedence [master-pjs1](https://github.com/asterisk/asterisk/actions/runs/21356831656/job/61466694308): FAILED TEST: channels/pjsip/message/message_redirect [master-pjs1](https://github.com/asterisk/asterisk/actions/runs/21356831656/job/61466694308): FAILED TEST: channels/pjsip/message/message_redirect_multiple

is this real ? nothing changed that should brake the test !
the pr is old could be that is the reason ?

jcolp:
Yes, it is because it is old and hasn't been rebased. It doesn't have changes/fixes that the tests cover or were adapted for.

PR #1617, opened 2025-12-01
stasis_broadcast: Add optional ARI broadcast with first-claim-wins

danieldonoghue:
Adds two optional modules:
res_stasis_broadcast.so: Infrastructure for broadcasting a single incoming
channel to multiple ARI applications with atomic first-claim-wins semantics.

app_stasis_broadcast.so: Provides the StasisBroadcast() dialplan application
which invokes the broadcast infrastructure.

Both modules are self-contained; if neither is loaded there is zero runtime
impact. Loading them does not alter existing Stasis or ARI behavior unless
explicitly used.

Key Features (only active when modules are loaded):
Fisher-Yates shuffled broadcast dispatch for fair claim races
Atomic claim operations using mutex + condition variable signaling
Configurable broadcast timeouts
Safe regex application filtering with validation to mitigate ReDoS risk
Thread-safe channel variable snapshotting (channel locked during reads)
Late-claim safety: broadcast context kept alive until after the Stasis
session ends so concurrent claimants always receive 409 Conflict rather
than 404 Not Found
Memory safety via RAII_VAR, ast_json_ref/unref, and ao2 reference counting

Components Added:
res/res_stasis_broadcast.c: Core broadcast + claim logic
apps/app_stasis_broadcast.c: StasisBroadcast() dialplan application
include/asterisk/stasis_app_broadcast.h: Public API header
res/ari/resource_events.c: Integrates POST /ari/events/claim endpoint
rest-api/api-docs/events.json: New CallBroadcast and CallClaimed events

Implementation Notes:
Broadcast contexts reside in an ao2 hash container keyed by channel id. Each
context holds atomic claim state, winner application name, timeout metadata,
and a condition variable for waiters. Broadcast contexts are kept alive until
after stasis_app_exec() returns so that concurrent claimants racing against
the timeout always receive 409 Conflict. Broadcast dispatch calls
stasis_app_send() directly for each matching application in shuffled order.
Regex filters are validated with bounded length, group depth, quantified
group count, and alternation limits to reduce pathological backtracking.
Timeout calculation uses timespec arithmetic with overflow-safe millisecond
remainder handling. Event JSON follows existing Stasis/ARI conventions;
references are managed correctly to avoid leaks or double frees.

Optional Nature / Impact:
No changes to existing APIs, events, or applications when absent.
Clean fallback: systems ignoring the modules behave identically to prior
versions.

Development was assisted by Claude (Anthropic). All generated code has been
reviewed, tested, and is understood by the author.

UserNote: New optional modules res_stasis_broadcast.so and
app_stasis_broadcast.so enable broadcasting an incoming channel to multiple
ARI applications. The first application to successfully claim (via
POST /ari/events/claim) wins channel control. StasisBroadcast() dialplan
application initiates broadcasts. CallBroadcast and CallClaimed events notify
applications. When modules are not loaded, behavior is unchanged.

DeveloperNote: New public APIs in stasis_app_broadcast.h:
stasis_app_broadcast_channel(), stasis_app_claim_channel(),
stasis_app_broadcast_winner(), and stasis_app_broadcast_wait(). New ARI event
types (CallBroadcast, CallClaimed) added to events.json. All code is isolated;
no existing ABI modified.

### DIFF
--- a/apps/app_stasis_broadcast.c
+++ b/apps/app_stasis_broadcast.c
@@ -1,0 +1,286 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2026, Aurora Innovation AB
+ *
+ * Daniel Donoghue <daniel.donoghue@aurorainnovation.com>
+ *
+ * See http://www.asterisk.org for more information about
+ * the Asterisk project. Please do not directly contact
+ * any of the maintainers of this project for assistance;
+ * the project provides a web site, mailing lists and IRC
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
+ * at the top of the source tree.
+ */
+
+/*! \file
+ *
+ * \brief Stasis broadcast dialplan application
+ *
+ * \author Daniel Donoghue <daniel.donoghue@aurorainnovation.com>
+ */
+
+/*** MODULEINFO
+	<depend type="module">res_stasis</depend>
+	<depend type="module">res_stasis_broadcast</depend>
+	<support_level>extended</support_level>
+ ***/
+
+#include "asterisk.h"
+
+#include "asterisk/app.h"
+#include "asterisk/module.h"
+#include "asterisk/pbx.h"
+#include "asterisk/stasis_app_broadcast.h"
+#include "asterisk/stasis_app_impl.h"
+
+/*** DOCUMENTATION
+	<application name="StasisBroadcast" language="en_US">
+		<since>
+			<version>20.17.0</version>
+			<version>22.7.0</version>
+			<version>23.1.0</version>
+		</since>
+		<synopsis>Broadcast a channel to multiple ARI applications for claiming,
+		then hand control to the winning application.</synopsis>
+		<syntax>
+			<parameter name="timeout">
+				<para>Timeout in milliseconds to wait for a claim.</para>
+				<para>Valid range: 0 to 60000ms</para>
+				<para>Default: 500ms</para>
+			</parameter>
+			<parameter name="app_filter">
+				<para>Regular expression to filter which ARI applications
+				receive the broadcast. Only applications with names matching
+				the regex will be notified.</para>
+				<para>Because arguments are comma-delimited, commas cannot
+				appear in the regex pattern. Use character classes
+				(e.g. <literal>[,]</literal>) if a literal comma is
+				needed, or omit the filter and handle selection in the
+				ARI application.</para>
+				<para>Default: all connected applications</para>
+			</parameter>
+			<parameter name="args">
+				<para>Optional colon-delimited arguments passed to the winning
+				application via the <literal>StasisStart</literal> event. These
+				are equivalent to the extra arguments in <literal>Stasis()</literal>.</para>
+				<para>Example: <literal>sales:priority-high</literal></para>
+			</parameter>
+			<parameter name="notify_claimed">
+				<para>Whether to send a <literal>CallClaimed</literal> event to
+				ARI applications when a channel is claimed.</para>
+				<para>When enabled, the <literal>CallClaimed</literal> event is
+				sent only to applications that matched the
+				<replaceable>app_filter</replaceable> (or all applications if no
+				filter was set).</para>
+				<para>Disabled by default to minimise WebSocket traffic under
+				high load. Losing claimants already receive a
+				<literal>409</literal> HTTP response.</para>
+				<para>Default: no</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Broadcasts the incoming channel to all connected ARI applications
+			(or a filtered subset) via a <literal>CallBroadcast</literal> event.
+			ARI applications can respond with a claim request. The first application
+			to claim the channel wins, and subsequent claims are rejected.</para>
+			<para>If an application claims the channel within the timeout, the channel
+			is automatically placed under Stasis control with the winning application,
+			exactly as if <literal>Stasis(winner_app)</literal> had been called.
+			The winning application receives a <literal>StasisStart</literal> event
+			and has full channel control until it calls <literal>continue</literal>
+			or the channel hangs up.</para>
+			<para>If no application claims the channel within the timeout, control
+			returns to the dialplan immediately, allowing fallback handling.</para>
+			<para>This application will set the following channel variables:</para>
+			<variablelist>
+				<variable name="STASISSTATUS">
+					<value name="SUCCESS">
+						An application claimed the channel and the Stasis
+						session completed without failures.
+					</value>
+					<value name="FAILED">
+						An application claimed the channel but a failure
+						occurred when executing the Stasis application.
+					</value>
+					<value name="TIMEOUT">
+						No application claimed the channel within the
+						timeout period.
+					</value>
+				</variable>
+			</variablelist>
+			<example>
+			; Broadcast with default timeout (500ms) to all apps
+			; Channel automatically enters Stasis with the winner
+			exten => _X.,1,StasisBroadcast()
+			 same => n,GotoIf($["${STASISSTATUS}"="TIMEOUT"]?no_route)
+			 same => n,Hangup()
+			 same => n(no_route),Playback(sorry-no-agent)
+			 same => n,Hangup()
+			</example>
+			<example>
+			; Broadcast with custom timeout, app filter, and args for the winner
+			exten => _X.,1,StasisBroadcast(2000,^ivr-.*,sales:priority-high)
+			 same => n,GotoIf($["${STASISSTATUS}"="TIMEOUT"]?no_route)
+			 same => n,Hangup()
+			 same => n(no_route),Playback(sorry-no-agent)
+			 same => n,Hangup()
+			</example>
+		</description>
+	</application>
+ ***/
+
+/*! \brief Dialplan application name */
+static const char *app = "StasisBroadcast";
+
+/*! \brief Default timeout in milliseconds */
+#define DEFAULT_TIMEOUT_MS 500
+
+/*! \brief Maximum timeout in milliseconds */
+#define MAX_TIMEOUT_MS 60000
+
+/*! \brief Maximum number of Stasis arguments */
+#define MAX_STASIS_ARGS 128
+
+/*! \brief StasisBroadcast dialplan application callback */
+static int stasis_broadcast_exec(struct ast_channel *chan, const char *data)
+{
+	char *parse = NULL;
+	int timeout_ms = DEFAULT_TIMEOUT_MS;
+	const char *app_filter = NULL;
+	const char *stasis_args_raw = NULL;
+	unsigned int flags = STASIS_BROADCAST_FLAG_SUPPRESS_CLAIMED;
+	char *winner = NULL;
+	int result = 0;
+	int stasis_argc = 0;
+	char *stasis_argv[MAX_STASIS_ARGS];
+
+	AST_DECLARE_APP_ARGS(args,
+		AST_APP_ARG(timeout);
+		AST_APP_ARG(app_filter);
+		AST_APP_ARG(stasis_args);
+		AST_APP_ARG(notify_claimed);
+	);
+
+	ast_assert(chan != NULL);
+
+	/* Initialize channel variable */
+	pbx_builtin_setvar_helper(chan, "STASISSTATUS", "");
+
+	/* Parse positional arguments if provided */
+	if (!ast_strlen_zero(data)) {
+		parse = ast_strdupa(data);
+		AST_STANDARD_APP_ARGS(args, parse);
+
+		if (!ast_strlen_zero(args.timeout)) {
+			if (sscanf(args.timeout, "%d", &timeout_ms) != 1
+				|| timeout_ms < 0 || timeout_ms > MAX_TIMEOUT_MS) {
+				ast_log(LOG_WARNING,
+					"Channel %s: invalid timeout value '%s' (must be 0-%dms), using default %dms\n",
+					ast_channel_name(chan), args.timeout, MAX_TIMEOUT_MS, DEFAULT_TIMEOUT_MS);
+				timeout_ms = DEFAULT_TIMEOUT_MS;
+			}
+		}
+
+		if (!ast_strlen_zero(args.app_filter)) {
+			app_filter = args.app_filter;
+		}
+
+		if (!ast_strlen_zero(args.stasis_args)) {
+			stasis_args_raw = args.stasis_args;
+		}
+
+		if (!ast_strlen_zero(args.notify_claimed) && ast_true(args.notify_claimed)) {
+			flags &= ~STASIS_BROADCAST_FLAG_SUPPRESS_CLAIMED;
+		}
+	}
+
+	/*
+	 * Parse colon-delimited Stasis arguments.  stasis_argv[] holds
+	 * pointers into the stack-allocated args_copy buffer.  This is
+	 * safe because stasis_app_exec is called within this same
+	 * function scope so the stack frame remains alive.
+	 */
+	if (!ast_strlen_zero(stasis_args_raw)) {
+		char *args_copy = ast_strdupa(stasis_args_raw);
+		char *arg;
+
+		while ((arg = strsep(&args_copy, ":")) != NULL && stasis_argc < MAX_STASIS_ARGS) {
+			stasis_argv[stasis_argc++] = arg;
+		}
+	}
+
+	ast_debug(3, "Broadcasting channel %s (timeout=%dms, filter=%s, args=%d)\n",
+		ast_channel_name(chan), timeout_ms, app_filter ? app_filter : "none",
+		stasis_argc);
+
+	/* Start the broadcast */
+	result = stasis_app_broadcast_channel(chan, timeout_ms, app_filter, flags);
+	if (result) {
+		ast_log(LOG_ERROR, "Failed to broadcast channel %s: %s\n",
+			ast_channel_name(chan),
+			result == AST_OPTIONAL_API_UNAVAILABLE ? "res_stasis_broadcast not loaded" : "internal error");
+		pbx_builtin_setvar_helper(chan, "STASISSTATUS", "FAILED");
+		return 0;
+	}
+
+	/* Wait for a claim.  A late claim can arrive between the timeout
+	 * expiring and our cleanup call, so always check for a winner
+	 * regardless of the wait result. */
+	stasis_app_broadcast_wait(chan, timeout_ms);
+	winner = stasis_app_broadcast_winner(ast_channel_uniqueid(chan));
+
+	if (winner) {
+		int ret;
+
+		ast_debug(3, "Channel %s claimed by %s, entering Stasis\n",
+			ast_channel_name(chan), winner);
+
+		/* Defer cleanup until after Stasis so concurrent claimants can still
+		 * find the context (with claimed=1) and receive 409 Conflict instead
+		 * of 404 Not Found. */
+		ret = stasis_app_exec(chan, winner, stasis_argc, stasis_argv);
+		ast_free(winner);
+
+		/* Clean up now that the Stasis session has ended */
+		stasis_app_broadcast_cleanup(ast_channel_uniqueid(chan));
+
+		if (ret) {
+			pbx_builtin_setvar_helper(chan, "STASISSTATUS", "FAILED");
+			if (ast_check_hangup(chan)) {
+				return -1;
+			}
+		} else {
+			pbx_builtin_setvar_helper(chan, "STASISSTATUS", "SUCCESS");
+		}
+	} else {
+		/* No winner: clean up immediately, nothing to race against */
+		stasis_app_broadcast_cleanup(ast_channel_uniqueid(chan));
+		ast_log(LOG_WARNING, "Channel %s: not claimed within %dms timeout\n",
+			ast_channel_name(chan), timeout_ms);
+		pbx_builtin_setvar_helper(chan, "STASISSTATUS", "TIMEOUT");
+	}
+
+	return 0;
+}
+
+static int load_module(void)
+{
+	return ast_register_application_xml(app, stasis_broadcast_exec);
+}
+
+static int unload_module(void)
+{
+	return ast_unregister_application(app);
+}
+
+AST_MODULE_INFO(ASTERISK_GPL_KEY, AST_MODFLAG_DEFAULT,
+	"Stasis application broadcast",
+	.support_level = AST_MODULE_SUPPORT_EXTENDED,
+	.load = load_module,
+	.unload = unload_module,
+	.requires = "res_stasis,res_stasis_broadcast",
+);
--- a/include/asterisk/stasis_app_broadcast.h
+++ b/include/asterisk/stasis_app_broadcast.h
@@ -1,0 +1,131 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2026, Aurora Innovation AB
+ *
+ * Daniel Donoghue <daniel.donoghue@aurorainnovation.com>
+ *
+ * See http://www.asterisk.org for more information about
+ * the Asterisk project. Please do not directly contact
+ * any of the maintainers of this project for assistance;
+ * the project provides a web site, mailing lists and IRC
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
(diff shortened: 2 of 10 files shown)


danieldonoghue:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

danieldonoghue:
testsuite-test-pr: 118

jcolp:
Can you explain the reasoning behind an approach of having a separate dialplan application that does the broadcast and then returns to the dialplan, instead of a single dialplan application that makes the operation self contained?

jcolp:
And how does this interact with the new support for outgoing websocket connections?

danieldonoghue:
> Can you explain the reasoning behind an approach of having a separate dialplan application that does the broadcast and then returns to the dialplan, instead of a single dialplan application that makes the operation self contained?

Hi @jcolp, my reasoning was fairly simple. I wanted to create a capability that would enhance the existing Stasis application without interfering with its existing functionality and making maintenance more difficult. It was easier to reason about the development work with a separate dialplan application (and how it would work in an actual dialplan) and I could leave Stasis doing what Stasis already does very well. 

It also felt as though it might be useful to be able to do additional handling in the dialplan following the broadcast - for example, handling timeouts by relaxing app_filter or trying a fallback of some sort. You could also use this just to send an arbitrary notification to all ari applications if you really wanted to.

I could have modified the Stasis application but found it difficult to reason about integrating the optional arguments with it and how to explain the functionality and increased complexity, in addition to the reasons I just outlined.

I suppose I could also have embedded the Stasis application call directly into this application but worried that would make it more difficult to maintain, moving forward, should the Stasis application's parameters change and still more difficult to reason about how that would look in the dialplan and the arguments being passed through. This option became less appealing the more I thought about it.

If you feel that this is something that should become part of the standard Stasis application I'd be happy to take another look and try to integrate it directly.

danieldonoghue:
> And how does this interact with the new support for outgoing websocket connections?

I dont think it should matter whether the websockets are inbound or outbound as this feature just uses the current list of registered applications, using stasis_app_get_all(). I'm also using the standard stasis_app_send() function and leaving the Stasis application to do the actual call communication so I dont foresee any issues.

 If I understand correctly, the websockets are handled directly by asterisk and, once connected, it then registers an application, that application will be capable receiving a broadcast event, so long as it listens to the appropriate events.

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2025, Aurora Innovation AB
```
jcolp: 2026

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+/*** MODULEINFO
+	<depend type="module">res_stasis</depend>
+	<depend type="module">res_stasis_broadcast</depend>
+	<support_level>extended</support_level>
+ ***/
+
+/*** DOCUMENTATION
+	<application name="StasisBroadcast" language="en_US">
+		<since>
+			<version>21.0.0</version>
```
jcolp: Needs to be updated

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+
+	ast_assert(chan != NULL);
+
+	/* Initialize channel variable */
+	ast_channel_lock(chan);
+	pbx_builtin_setvar_helper(chan, "BROADCAST_WINNER", "");
+	ast_channel_unlock(chan);
+
+	/* Parse arguments if provided */
+	if (!ast_strlen_zero(data)) {
```
jcolp, line 174: This large block is not needed, it can be parsed regardless.

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+		parse = ast_strdupa(data);
+		AST_STANDARD_APP_ARGS(args, parse);
+
+		if (!ast_strlen_zero(args.options)) {
+			int i;
+			options_str = ast_strdupa(args.options);
+			AST_STANDARD_APP_ARGS(options, options_str);
+
+			/* Parse key=value options */
+			for (i = 0; i < options.argc; i++) {
```
jcolp: I don't understand why you have parsing code here. The app args parser can already handle the parsing of the string and putting it into a position for access.

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+						app_filter = val;
+					} else {
+						ast_log(LOG_WARNING, "Unknown option '%s'\n", key);
+					}
+				}
+			}
+		}
+	}
+
+	ast_log(LOG_NOTICE, "Broadcasting channel %s (timeout=%dms, filter=%s)\n",
```
jcolp: This should be a debug message.

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+		ast_channel_name(chan), timeout_ms, app_filter ? app_filter : "none");
+
+	/* Start the broadcast */
+	result = stasis_app_broadcast_channel(chan, timeout_ms, app_filter);
+	if (result != 0) {
```
jcolp: We just use if (result) in the Asterisk code base

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+		ast_log(LOG_ERROR, "Failed to broadcast channel %s (return code: %d)\n",
+			ast_channel_name(chan), result);
```
jcolp: If there's an actual useful result code then a string representation so users are aware would be good.

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+		return 0;
+	}
+
+	/* Wait for a claim */
+	if (stasis_app_broadcast_wait(chan, timeout_ms) == 0) {
+		/* Channel was claimed */
+		winner = stasis_app_broadcast_winner(ast_channel_uniqueid(chan));
+		if (winner) {
+			ast_log(LOG_NOTICE, "Channel %s claimed by %s\n",
```
jcolp: debug message

apps/app_stasis_broadcast.c
```diff
@@ -0,0 +1,217 @@
...
+				ast_channel_name(chan), winner);
+			ast_channel_lock(chan);
+			pbx_builtin_setvar_helper(chan, "BROADCAST_WINNER", winner);
+			ast_channel_unlock(chan);
+			ast_free(winner);
+		}
+	} else {
+		ast_log(LOG_NOTICE, "Channel %s not claimed within timeout\n",
```
jcolp: I think this should be a warning

configs/samples/extensions_stasis_broadcast.conf.sample
```diff
@@ -0,0 +1,107 @@
+;
```
jcolp: We don't put stuff like this in the tree. There should be a new docs page in the documentation repo instead covering documentation for this and examples.

include/asterisk/stasis_app_broadcast.h
```diff
@@ -0,0 +1,112 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2025, Aurora Innovation AB
```
jcolp: 2026

include/asterisk/stasis_app_broadcast.h
```diff
@@ -0,0 +1,112 @@
...
+/*!
+ * \brief Start a broadcast for a channel
+ *
+ * Broadcasts a channel to all ARI applications (or filtered applications)
+ * allowing them to claim the channel. Only the first claim will succeed.
+ *
+ * \param chan The channel to broadcast
+ * \param timeout_ms Timeout in milliseconds to wait for a claim
+ * \param app_filter Optional regex filter for application names (NULL for all)
+ *
```
jcolp, line 54: These new API calls should have \since as well

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2025, Aurora Innovation AB
```
jcolp: 2026

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+/*! \brief Group depth threshold for alternation limits */
+#define ALTERNATION_DEPTH_THRESHOLD 2
+
+/*! \brief Maximum broadcast timeout in milliseconds (24 hours) */
+#define MAX_BROADCAST_TIMEOUT_MS (24 * 60 * 60 * 1000)
+
+/*! \brief Broadcast context stored on channel */
+struct stasis_broadcast_ctx {
+	/*! The unique ID of the channel */
+	char channel_id[AST_MAX_EXTENSION];
```
jcolp: AST_MAX_PUBLIC_UNIQUEID

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	/*! Name of the winning application */
+	char winner_app[AST_MAX_EXTENSION];
```
jcolp: I don't think there is a max for a Stasis application, and AST_MAX_EXTENSION is certainly not correct for it

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+}
+
+/*! \brief Datastore information for broadcast context */
+static const struct ast_datastore_info broadcast_datastore_info = {
+	.type = "stasis_broadcast_context",
+	.destroy = broadcast_datastore_destroy,
+};
+
+/*! \brief Hash function for broadcast context container */
+static int broadcast_ctx_hash_fn(const void *obj, const int flags)
```
jcolp: The AO2_STRING_FIELD_HASH_FN macro can be used instead.

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+		ctx = obj;
+		return ast_str_hash(ctx->channel_id);
+	default:
+		ast_assert(0);
+		return 0;
+	}
+}
+
+/*! \brief Compare function for broadcast context container */
+static int broadcast_ctx_cmp_fn(void *obj, void *arg, int flags)
```
jcolp: The AO2_STRING_FIELD_CMP_FN macro can be used instead

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	ast_mutex_destroy(&ctx->lock);
+}
+
+/*! \brief Create a new broadcast context */
+static struct stasis_broadcast_ctx *broadcast_ctx_create(
+	const char *channel_id, int timeout_ms)
+{
+	struct stasis_broadcast_ctx *ctx;
+
+	ctx = ao2_alloc(sizeof(*ctx), broadcast_ctx_destructor);
```
jcolp, line 156: This will include a mutex lock accessible using ao2_object_get_lockaddr which removes the need to put a lock on the structure

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	if (!ctx) {
+		return NULL;
+	}
+
+	ast_copy_string(ctx->channel_id, channel_id, sizeof(ctx->channel_id));
+	ast_mutex_init(&ctx->lock);
+	ast_cond_init(&ctx->cond, NULL);
+	ctx->timeout_ms = timeout_ms;
+	ctx->claimed = 0;
```
jcolp: The returned memory from ao2_alloc is zeroed out already

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	ctx->broadcast_time = ast_tvnow();
+
+	ast_debug(1, "Created broadcast context for channel %s (timeout: %dms)\n",
+		ctx->channel_id, ctx->timeout_ms);
+
+	return ctx;
+}
+
+/*! \brief Collect channel variables into JSON object */
+static struct ast_json *collect_channel_vars(struct ast_channel *chan)
```
jcolp: We have ast_json_channel_vars already, did you not use it because it did not fit your needs?

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	RAII_VAR(struct ao2_container *, apps, NULL, ao2_cleanup);
+	struct ast_json *vars = NULL;
+	struct ao2_iterator iter;
+	char *app_name;
+	regex_t regex;
+	int regex_compiled = 0;
+	const char *caller = NULL;
+	const char *called = NULL;
+
+	snapshot = ast_channel_snapshot_get_latest(ast_channel_uniqueid(chan));
```
jcolp: This is inefficient. The latest snapshot can be retrieved directly from the channel using ast_channel_snapshot directly with the chan pointer, but chan must be locked and you must bump the reference yourself to ensure it remains valid after unlock.

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	if (!snapshot) {
```
jcolp: This will never occur if accessing the channel snapshot on it.

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	event = ast_json_pack("{s: s, s: o, s: o, s: s?, s: s?, s: o?}",
+		"type", "CallBroadcast",
+		"timestamp", ast_json_timeval(ast_tvnow(), NULL),
+		"channel", ast_channel_snapshot_to_json(snapshot, NULL),
+		"caller", caller,
+		"called", called,
+		"variables", ast_json_ref(vars));
+
+	if (!event) {
+		ast_log(LOG_ERROR, "Failed to create broadcast event\n");
```
jcolp: This needs to be more useful and include channel name

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+			ao2_link(matching_apps, app_name);
+			ao2_ref(app_name, -1);
+		}
+		ao2_iterator_destroy(&iter);
+
+		ast_debug(2, "Broadcasting channel %s to %d matching applications\n",
+			ast_channel_uniqueid(chan), ao2_container_count(matching_apps));
+
+		/*
+		 * Second pass: dispatch to all matching apps in parallel via taskpool.
```
jcolp: Taskpool does not work like that as configured. It'll be sequential in the same thread unless you throw 250+ tasks at the taskpool. If you want it to be that wild, then you'd have to use threadpool.

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+		 * opportunity to claim the channel.
+		 */
+		match_iter = ao2_iterator_init(matching_apps, 0);
+		while ((match_name = ao2_iterator_next(&match_iter))) {
+			struct broadcast_task_data *task_data;
+
+			ast_debug(3, "Queueing broadcast to app '%s'\n", match_name);
+
+			/* Allocate task data */
+			task_data = ast_malloc(sizeof(*task_data));
```
jcolp: This can be allocated with space for match_name so that a separate allocation isn't needed.

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+			if (!task_data) {
+				ast_log(LOG_ERROR, "Failed to allocate broadcast task data for app '%s'\n", match_name);
```
jcolp: Include channel name

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+				ast_log(LOG_ERROR, "Failed to duplicate app name for '%s'\n", match_name);
+				ast_json_unref(task_data->event);
+				ast_free(task_data);
+				ao2_ref(match_name, -1);
+				continue;
+			}
+
+			/* Push to taskpool for parallel execution */
+			if (ast_taskpool_push(broadcast_taskpool, broadcast_send_task, task_data)) {
+				ast_log(LOG_ERROR, "Failed to push broadcast task for app '%s'\n", match_name);
```
jcolp: Include channel name

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+{
+	RAII_VAR(struct stasis_broadcast_ctx *, ctx, NULL, ao2_cleanup);
+	struct ast_datastore *datastore;
+	const char *channel_id;
+
+	if (!chan) {
+		return -1;
+	}
+
+	channel_id = ast_channel_uniqueid(chan);
```
jcolp: Why do this instead of calling ast_channel_uniqueid each time?

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+		ast_copy_string(ctx->winner_app, app_name, sizeof(ctx->winner_app));
+		result = 0;
+		ast_verb(3, "Channel %s claimed by application %s\n", channel_id, app_name);
+		/* Signal waiting thread that channel was claimed */
+		ast_cond_signal(&ctx->cond);
+	}
+	ast_mutex_unlock(&ctx->lock);
+
+	/* If claim successful, set channel variable */
+	if (result == 0) {
```
jcolp: Reduce indentation, if not successful then just return early.

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+		chan = ast_channel_get_by_name(channel_id);
+		if (chan) {
+			ast_channel_lock(chan);
+			pbx_builtin_setvar_helper(chan, "BROADCAST_WINNER", app_name);
+			ast_channel_unlock(chan);
+
+			/* Send CallClaimed event to all apps */
+			snapshot = ast_channel_snapshot_get_latest(channel_id);
```
jcolp: Same as I said before, you can access the channel directly for the snapshot and it will be non-NULL

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+ */
+void AST_OPTIONAL_API_NAME(stasis_app_broadcast_cleanup)(const char *channel_id)
+{
+	RAII_VAR(struct stasis_broadcast_ctx *, ctx, NULL, ao2_cleanup);
+
+	if (ast_strlen_zero(channel_id)) {
+		return;
+	}
+
+	ctx = ao2_find(broadcast_contexts, channel_id, OBJ_SEARCH_KEY);
```
jcolp: The unlink can be done as part of ao2_find by passing OBJ_UNLINK

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+ * \brief HTTP handler for /ari/events/claim endpoint
+ */
+/* HTTP endpoint /ari/events/claim is handled by res_ari through resource_events.c */
+
+static int load_module(void)
+{
+	struct ast_taskpool_options taskpool_options = {
+		.version = AST_TASKPOOL_OPTIONS_VERSION,
+		.selector = AST_TASKPOOL_SELECTOR_DEFAULT,
+		.idle_timeout = 0,              /* No timeout, keep threads alive */
```
jcolp: The fact that taskpool won't do what you want aside, I am not comfortable having unused threads existing for everyone when only a handful of people will ever use this functionality

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	broadcast_contexts = ao2_container_alloc_hash(AO2_ALLOC_OPT_LOCK_MUTEX, 0,
+		BROADCAST_BUCKETS, broadcast_ctx_hash_fn, NULL, broadcast_ctx_cmp_fn);
+
+	if (!broadcast_contexts) {
+		ast_taskpool_shutdown(broadcast_taskpool);
+		broadcast_taskpool = NULL;
+		return AST_MODULE_LOAD_DECLINE;
+	}
+
+	ast_log(LOG_NOTICE, "Stasis broadcast module loaded\n");
```
jcolp: Remove notice

res/res_stasis_broadcast.c
```diff
@@ -0,0 +1,945 @@
...
+	ao2_cleanup(broadcast_contexts);
+	broadcast_contexts = NULL;
+
+	/* Shutdown taskpool */
+	if (broadcast_taskpool) {
+		ast_taskpool_shutdown(broadcast_taskpool);
+		broadcast_taskpool = NULL;
+	}
+
+	ast_log(LOG_NOTICE, "Stasis broadcast module unloaded\n");
```
jcolp: Remove notice

rest-api/api-docs/events.json
```diff
@@ -110,6 +110,50 @@
...
 			]
+		},
+		{
+			"path": "/events/claim",
+			"description": "Broadcast channel claim operations",
+			"operations": [
+				{
+					"httpMethod": "POST",
+					"since": [
+						"21.0.0"
```
jcolp: Needs to be updated

jcolp (dismissed):

danieldonoghue:
> I suppose I could also have embedded the Stasis application call directly into this application but worried that would make it more difficult to maintain, moving forward, should the Stasis application's parameters change and still more difficult to reason about how that would look in the dialplan and the arguments being passed through. 

on reflection, I have made this change as it improves overall performance and clarity

res/res_stasis_broadcast.c
danieldonoghue: switched to ast_channel_snapshot_to_json() to produce a standard ARI Channel object consistent with all other ARI events. It already includes channelvars, so ast_json_channel_vars wasn't needed

res/res_stasis_broadcast.c
danieldonoghue: have updated send_broadcat_event so that it now locks the channel, calls ast_channel_snapshot(chan) directly, bumps the ref, then unlocks, as per suggestion.

not sure that's possible in stasis_app_claim_channel, though, since the function is called from the ARI HTTP handler and that doesnt have the chan pointer. I suppose I could look up the channel by id and use that to get the snapshot using ast_channel_snapshot, if that's preferred?

danieldonoghue:
I've removed the taskpool entirely, following the concerned tone I note from the comments. stasis_app_send() writes directly to each app's WebSocket socket synchronously anyway so the taskpool was adding overhead without any significant parallelism benefit. The application name shuffle before the dispatch loop ensures no single app is consistently first to receive the event, preserving fairness across broadcasts.

This also means the module doesnt spawn any threads, addressing the concern about idle threads for anybody loading this module.

One potential issue with removing the task pool, though, is that with the taskpool, a slow ARI client blocking on socket write would only have affected its own task thread. With this modification to serially dispatch, a client whose socket send buffer is full could delay delivery to subsequent apps in the loop. In practice this is unlikely given typical ARI client characteristics and TCP buffer sizes, but wanted to flag it.

danieldonoghue:
> Workflow PRCheck failed [master-pjs5](https://github.com/asterisk/asterisk/actions/runs/22405406540/job/64864768473): FAILED TEST: channels/pjsip/subscriptions/rls/lists_of_lists/nominal/mwi/batched

I dont believe this is related to my change

danieldonoghue:
documentation pr: 180

asteriskteam:

asteriskteam:

danieldonoghue:
hmm, I see there are still failing tests. I will take a look at those as soon as possible :)

gtjoseph:
> hmm, I see there are still failing tests. I will take a look at those as soon as possible :)

I don't think they're related to your change.  I'll re-run them to make sure.

gtjoseph:
@danieldonoghue Don't worry about the two pjsip tests right now but the rest_api/authentication_user_acl test isn't one I've seen failing before.  Can you run it locally both with and without this PR and see what happens?

jcolp:
Is this review based on an old version of master before the ACL support went in for ARI and HTTP?

gtjoseph:
> Is this review based on an old version of master before the ACL support went in for ARI and HTTP?

Oh yeah, no kidding.

@danieldonoghue If you rebase this PR on current master you should be good.

danieldonoghue:
@gtjoseph, I did the rebase and it fixed those tests, as expected. however, I note that caller_attended fails now. looking at it, I dont believe that to be related to my changes.

gtjoseph:
> @gtjoseph, I did the rebase and it fixed those tests, as expected. however, I note that caller_attended fails now. looking at it, I dont believe that to be related to my changes.

Any time you see a transfer test failing there's a 99% chance it's because the test is timing dependent and GitHub is being stingy with resources. :)

jcolp (approved):

Issue #1417, opened 2025-09-08
[bug]: static code analysis issues in abstract_jb


aol-nnov:
### Severity

Minor

### Versions

18..master

### Components/Modules

abstract_jb

### Operating Environment

static code analysis

### Frequency of Occurrence

None

### Issue Description

After having been compared to a NULL value at abstract_jb.c:969, pointer 'framedata->timer' is passed as 1st parameter in call to function 'ast_timer_set_rate' at abstract_jb.c:1093, where it is dereferenced at timing.c:168.

After having been assigned to a NULL value at abstract_jb.c:1047, pointer 'data->early_frames->last' is dereferenced at abstract_jb.c:1060.

Return value of a function 'ast_channel_bridge_peer' is dereferenced at abstract_jb.c:463 without checking for NULL, but it is usually checked for this function (19/20).

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

aol-nnov:
@jcolp as I'v [stated earlier](https://github.com/asterisk/asterisk/issues/1403#issue-3362877424), I'm starting to file static analysis issues.

Should I create one gh issue per found vulnerability or per source code file?

jcolp:
One issue and PR per source code file would be ideal to keep things manageable.

The issue description also makes this sound like this isn't actually an issue?

Edit: Oh, I see, 1093.

This doesn't actually look like it would be possible. If framedata->timer fails to be created then framedata is destroyed immediately as it is an error case, and framedata is never in a partial state where it would be used/that callback would occur.

aol-nnov:
okay, I'll create one issue per file then.

> The issue description also makes this sound like this isn't actually an issue?

You should follow the code along the lines in question to get the idea. Code flow proceeds beyond the null pointer check making it useless and/or allows to dereference the null pointer.

After all, issue descriptions are not mine - they are generated by the tool.

jcolp:
I edited my comment, it's not actually possible for framedata->timer to be NULL there in the first place.

aol-nnov:
> it's not actually possible 

why there is a null pointer check then?

jcolp:
I assume the original developer put it in there out of habit, or thought it could be, or wanted to be extra cautious, or the code was subsequently changed.

aol-nnov:
> developer put it in there out of habit, or thought it could be, or

a slippery road, I must say...

[...]

PR #1618, opened 2025-12-01
abstract_jb.c: Remove redundant timer check per static analysis.

seanbright:
While this check is technically unnecessary, it also was not harmful.

The 2 other items mentioned in the linked issue are false positives and require no action.

Resolves: #1417


seanbright:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1427, opened 2025-09-08
[bug]: static code analysis issues in apps/app_stream_echo.c


aol-nnov:
### Severity

Trivial

### Versions

18..master

### Components/Modules

apps/app_stream_echo.c

### Operating Environment

static code analysis 

### Frequency of Occurrence

None

### Issue Description

Return value of a function 'ast_channel_get_default_stream' is dereferenced at app_stream_echo.c:93 without checking for NULL, but it is usually checked for this function (4/5).

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1619, opened 2025-12-01
app_stream_echo.c: Check that stream is non-NULL before dereferencing.

seanbright:
Also re-order and rename the arguments of `stream_echo_write_error` to match those of `ast_write_stream` for consistency.

Resolves: #1427


seanbright:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1626, opened 2025-12-03
[bug]: res_phoneprov


systemcrash:
### Severity

Minor

### Versions

23.1.0

### Components/Modules

res_phoneprov

### Operating Environment

-

### Frequency of Occurrence

Constant

### Issue Description

Is it possible that `phoneprov_users.conf` should be named `phoneprov_users.conf.sample`?

Running `res_phoneprov` complains that it is absent, and the compile stage does not copy it, only `phoneprov.conf`.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1627, opened 2025-12-03
configs: rename phoneprov_users.conf to .sample so make installs it

systemcrash:
This conf file should be suffixed .sample so that make installs it at compile time. Otherwise res_phoneprov complains at runtime as to its absence and refuses to start.

Fixes: #1626


jcolp (approved):

gtjoseph:
@systemcrash Please read the following and correct the issues.

systemcrash:
Please detail what the problem is that I should fix. I've read both pages and I am none the wiser.

seanbright:
cherry-pick-to: 23

seanbright:
> Please detail what the problem is that I should fix. I've read both pages and I am none the wiser.

I took care of it. In short:

1. Updated the commit message (removed the `cherry-pick-to` and added a colon after `Fixes` - although the colon isn't actually needed) and force-pushed
2. Added a `cherry-pick-to` comment to this PR.

systemcrash:
OK - got it. In the actual PR messages... I see it now. :/

gtjoseph:
> although the colon isn't actually needed)

It's not needed by GitHub but it is by the release changelog process.

gtjoseph:
@systemcrash Is there some reason you don't want this to be cherry-picked to the 20 and 22 branches?

systemcrash:
> @systemcrash Is there some reason you don't want this to be cherry-picked to the 20 and 22 branches?

The phoneprov module doesn't exist there...? It's still called the deprecated user module IIRC.

Edit: my bad - it's there. 
Edit 2: well, the config file doesn't exist in those older branches, and this is only a rename...

gtjoseph:
Ah gotcha.

PR #1632, opened 2025-12-05
cel_manager.c: Correct manager event mask for CEL events.

seanbright:
There is no EVENT_FLAG_CEL and these events are raised with as EVENT_FLAG_CALL.


seanbright:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

Issue #1430, opened 2025-09-08
[bug]: static code analysis issues in res/stasis/app.c


aol-nnov:
### Severity

Major

### Versions

18..master

### Components/Modules

res/stasis/app.c

### Operating Environment

static code analysis 

### Frequency of Occurrence

None

### Issue Description

After having been compared to a NULL value at app.c:431, pointer 'new_snapshot' is dereferenced at app.c:436.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1633, opened 2025-12-07
Fix false null-deref warning in channel_state

phoneben:
Resolve analyzer warning in channel_state by checking AST_FLAG_DEAD on snapshot, which is guaranteed non-NULL.

Resolves: #1430

### DIFF
--- a/res/stasis/app.c
+++ b/res/stasis/app.c
@@ -435,21 +435,25 @@ static struct ast_json *channel_state_change_event(
 	return simple_channel_event("ChannelStateChange", snapshot, tv);
 }
 
-/*! \brief Handle channel state changes */
+/*!
+*\brief Handle channel state changes 
+*
+*\note For channel snapshot updates, new_snapshot is guaranteed to be
+*	   non-NULL. A channel going away is indicated by AST_FLAG_DEAD
+*	   on the new snapshot. See stasis_channels.h for the invariant.
+*/
 static struct ast_json *channel_state(
 	struct ast_channel_snapshot *old_snapshot,
 	struct ast_channel_snapshot *new_snapshot,
 	const struct timeval *tv)
 {
-	struct ast_channel_snapshot *snapshot = new_snapshot ?
-		new_snapshot : old_snapshot;
 
 	if (!old_snapshot) {
-		return channel_created_event(snapshot, tv);
+		return channel_created_event(new_snapshot, tv);
 	} else if (ast_test_flag(&new_snapshot->flags, AST_FLAG_DEAD)) {
-		return channel_destroyed_event(snapshot, tv);
+		return channel_destroyed_event(new_snapshot, tv);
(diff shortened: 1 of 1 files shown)


phoneben:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seanbright:
I don't think it it possible for `new_snapshot` to actually be `NULL` here. The other 3 `channel_snapshot_monitor`s that are run (`channel_dialplan`, `channel_callerid`, and `channel_connected_line`) after `channel_state` all dereference `new_snapshot` without a `NULL` check.

Also, `sub_channel_update_handler` which calls `channel_state` dereferences `new_snapshot` on line 575.

So it's the `if` at lines 444/445 that is the problem, not `new_snapshot`.

seanbright:
Also... we'd be crashing all the time

seanbright:
Last thing - @jcolp documented this invariant when he made this change:

https://github.com/asterisk/asterisk/blob/3ce5308ac9bdd4b82c4c41f508ab594f2d75c1c7/include/asterisk/stasis_channels.h#L173-L176

phoneben:
The analyzer was warning because the ternary (new_snapshot ? new_snapshot : old_snapshot) made it think new_snapshot might be NULL, even though the Asterisk invariant guarantees it never is. By switching the DEAD-flag check to use snapshot instead of new_snapshot, we use a pointer that is provably non-NULL in this branch. This keeps the logic exactly the same, stays consistent with the invariant, and removes the false NULL-dereference warning .

jcolp:
What analyzer hallucinated this in the first place?

phoneben:
This is from the static code review .
 Did I miss something?
just ignore it ?

jcolp:
Well, in the existing code it's not a problem. Change for the sake of change I don't like, and so would like to know what analyzer or tool is being satisfied.

seanbright:
See the linked issue

jcolp:
oh it's THAT thing

meh, I'm not thrilled about a commit just to appease it when it has had so much trouble over all understanding things

phoneben:
Yep, understood. The tweak doesn’t change behavior — it just uses the already-non-NULL snapshot to quiet the warning.
Happy to drop it if you’d rather avoid commits for analyzer noise.

just trying to close the issue

seanbright:
> oh it's THAT thing
> 
> meh, I'm not thrilled about a commit just to appease it when it has had so much trouble over all understanding things

Correct me if I’m wrong, but the ternary at line 444 is meaningless because `new_snapshot` will always be non-`NULL`.

I think dropping `snapshot` entirely and just using `old_snapshot` and `new_snapshot` directly would remove confusion.

jcolp:
Looking at the code, agreed.

phoneben:
seanbright 
did i overdo it ?

jcolp (approved):

Issue #1635, opened 2025-12-08
[bug]: Regression: Fix endpoint memory leak


gtjoseph:
### Severity

Blocker

### Versions

23.1.0, 22.7.0, 20.17.0

### Components/Modules

main/endpoint.c

### Operating Environment

Any

### Frequency of Occurrence

Constant

### Issue Description

Commit 26795be52ecc313b7094d9c0ba3bc37fc137cac9 introduced a memory leak of ast_endpoint when ast_endpoint_shutdown() was called.  The leak occurs only if a configuration change removes an endpoint and isn't related to call volume or the length of time asterisk has been running.  

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1636, opened 2025-12-08
endpoint.c: Plug a memory leak in ast_endpoint_shutdown().

gtjoseph:
Commit 26795be introduced a memory leak of ast_endpoint when
ast_endpoint_shutdown() was called. The leak occurs only if a configuration
change removes an endpoint and isn't related to call volume or the length of
time asterisk has been running.  An ao2_ref(-1) has been added to
ast_endpoint_shutdown() to plug the leak.

Resolves: #1635


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

Issue #1637, opened 2025-12-09
[improvement]: force_longest_waiting_caller should also consider caller priority


serfreeman1337:
### Improvement Description

When using the `force_longest_waiting_caller` setting, placing the caller first in the queue actually prioritizes the caller after other calls in other queues based on a waiting time.

Consider this situation with 400 and 401 queues, both have equal weights.
```
400
   Callers: 
      1. Local/123@from-internal-00000015;2 (wait: 0:32, prio: 20)
      2. PJSIP/trunk-00000001 (wait: 20:44, prio: 0)
      3. PJSIP/trunk-00000002 (wait: 20:29, prio: 0)
      4. PJSIP/trunk-00000003 (wait: 19:38, prio: 0)
      5. PJSIP/trunk-00000004 (wait: 19:33, prio: 0)
      6. PJSIP/trunk-00000005 (wait: 19:27, prio: 0)
      7. PJSIP/trunk-00000006 (wait: 19:23, prio: 0)

401
   Callers: 
      1. PJSIP/trunk-00000007 (wait: 11:00, prio: 0)
      2. PJSIP/trunk-00000008 (wait: 10:59, prio: 0)
      3. PJSIP/trunk-00000009 (wait: 10:53, prio: 0)
      4. PJSIP/trunk-00000010 (wait: 10:52, prio: 0)
      5. PJSIP/trunk-00000011 (wait: 10:51, prio: 0)
      6. PJSIP/trunk-00000012 (wait: 10:47, prio: 0)
      7. PJSIP/trunk-00000013 (wait: 0:24, prio: 0)
      8. PJSIP/trunk-00000014 (wait: 0:14, prio: 0)
```
Placing the `Local/123@from-internal-00000015;2` caller with priority into the 400 queue will force operators to receive calls only from 401 queue until there is no calls with waiting time more than the caller's.

Expected result:
Operators will receive the call from the `Local/123@from-internal-00000015;2` caller immediately since it has higher priority that other calls in other queues.

jcolp:
Are you planning to submit a patch for this?

PR #1640, opened 2025-12-10
app_queue: Add 'prio' setting to the 'force_longest_waiting_caller' option

serfreeman1337:
This adds a 'prio' setting to ensure that call priority is respected across multiple queues.
Using 'yes' could cause high-priority callers to be skipped if a caller
in another queue had a longer wait time, regardless of priority.

Resolves: #1637

UserNote: The 'force_longest_waiting_caller' option now supports a 'prio' setting.
When set to 'prio', calls are offered by priority first, then by wait time.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -1942,6 +1942,10 @@ struct penalty_rule {
 #define ANNOUNCEPOSITION_MORE_THAN 3 /*!< We say "Currently there are more than <limit>" */
 #define ANNOUNCEPOSITION_LIMIT 4 /*!< We not announce position more than \<limit\> */
 
+#define FORCELONGESTWAITINGCALLER_NO 0
+#define FORCELONGESTWAITINGCALLER_YES 1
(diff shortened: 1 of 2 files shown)


serfreeman1337:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (changes requested):
The sample configuration also needs to be updated. Right now the behavior is explicitly stated that it is longest waiting caller, it makes no mention of priority being considered.

I think that also raises a question I would like feedback from others on. This changes the existing option behavior and goes counter to what it has been. From a least surprise perspective I wonder if this new behavior should be opt-in only so existing users are not taken off guard when things change.

serfreeman1337:
For me, it was actually more surprising to find that this option didn't take priorities into account. I'd expect a call with high priority to be handled first.

If changing the current behavior is not desirable, I can make the priority logic optional by setting `force_longest_waiting_caller=prio`, for example.

jcolp:
I would feel more comfortable with such a change if it was configurable, and existing behavior was preserved.

jcolp (approved):

Issue #1641, opened 2025-12-10
[bug]: res_pjsip_config_wizard: Endpoints fail to update when Named ACLs change after reload


hajekmi:
### Severity

Minor

### Versions

master, 22.7.0

### Components/Modules

res_pjsip_config_wizard, named_acl

### Operating Environment

ALMA Linux 9.6
5.14.0-570.58.1.el9_6.x86_64

### Frequency of Occurrence

Constant

### Issue Description

When using res_pjsip_config_wizard to generate PJSIP endpoints, the endpoints do not update their ACL rules if the underlying Named ACL (defined in acl.conf) changes, even after issuing both module reload acl and module reload res_pjsip.

The PJSIP Wizard relies on Sorcery optimizations (checking if the pjsip_wizard.conf file has changed or if the configuration content differs). Since changing an IP address in acl.conf does not alter the text content of pjsip_wizard.conf, the wizard assumes no changes are needed. Consequently, the Endpoint objects retain a pointer to the old/stale ACL rules.

**Steps to Reproduce**
Define a named ACL in acl.conf:
```
[my_acl]
deny=0.0.0.0/0
permit=1.2.3.4/32
```

Configure a PJSIP Wizard endpoint in pjsip_wizard.conf that uses this ACL:
```
[my_phone](type=wizard)
transport=transport-udp
acl=my_acl
contact_acl=my_acl
```

Start Asterisk or reload modules. The endpoint correctly allows 1.2.3.4.

Modify acl.conf to allow a different IP:

```
[my_acl]
deny=0.0.0.0/0
permit=5.6.7.8/32
```
Reload the ACL module: module reload acl.

Reload PJSIP: module reload res_pjsip (or pjsip reload).

Result: The endpoint my_phone still blocks 5.6.7.8 and allows 1.2.3.4. The changes are ignored because pjsip_wizard.conf itself was not touched.

**Expected Behavior**
After reloading ACL and PJSIP, the Wizard should detect that the referenced ACLs have been updated globally and force a rebuild of the affected Endpoint objects, picking up the new IP rules.

**Root Cause Analysis**
The issue lies in res_pjsip_config_wizard.c, specifically in object_type_loaded_observer. It uses optimizations:

It checks CONFIG_FLAG_FILEUNCHANGED.

It uses ast_variable_lists_match to compare the old and new wizard config.

Since neither the file timestamp nor the variable list (e.g., acl=my_acl) changes when acl.conf is edited, the reload is skipped for those objects.

[...]

PR #1642, opened 2025-12-10
res_pjsip_config_wizard: Force endpoint reload when Named ACL changes

hajekmi:
When using res_pjsip_config_wizard, endpoints fail to update their ACL rules if the underlying Named ACL (in acl.conf) changes, even after reloading both ACL and PJSIP. This happens because the wizard optimizes reloads by checking file timestamps or content changes in pjsip_wizard.conf, ignoring external ACL changes.

This patch introduces a generation counter for Named ACLs.
- main/named_acl.c: Exposes a generation counter incremented on reload.
- res_pjsip_config_wizard.c: Checks if the global ACL generation changed. If an ACL change is detected during a reload, the wizard forces a rebuild of endpoints that use 'acl' or 'contact_acl', bypassing standard optimization checks.

Fixes: #1641
Signed-off-by: Michal Hajek michal.hajek@daktela.com

cherry-pick-to: 23
cherry-pick-to: 22

### DIFF
--- a/include/asterisk/acl.h
+++ b/include/asterisk/acl.h
@@ -487,6 +487,8 @@ void ast_ha_output(int fd, const struct ast_ha *ha, const char *prefix);
  */
 void ast_acl_output(int fd, struct ast_acl_list *acl, const char *prefix);
 
+int ast_named_acl_get_generation(void);
+
 #if defined(__cplusplus) || defined(c_plusplus)
 }
 #endif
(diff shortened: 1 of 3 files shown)


jcolp (changes requested):
See my comment on the issue.

jcolp:
I'm marking this as closed as I do not believe it is the correct fix, per my comment on the original issue.

PR #1644, opened 2025-12-10
res_pjsip_config_wizard: Force reload on Named ACL change events

hajekmi:
Currently, endpoints created via the PJSIP Config Wizard do not update
their ACL rules if the underlying Named ACL (in acl.conf) changes.
This occurs because the wizard relies on file timestamp and content
caching of pjsip_wizard.conf, which remains unchanged during an external
ACL update. As a result, endpoints retain stale ACL rules even after
a reload.

This patch updates res_pjsip_config_wizard to subscribe to the
ast_named_acl_change_type Stasis event. A local generation counter is
incremented whenever an ACL change event is received.

During a reload, the wizard compares the current local generation against
the generation stored in the wizard object. If a change is detected:
1. The file cache optimization (CONFIG_FLAG_FILEUNCHANGED) is bypassed.
2. Wizard objects utilizing 'acl' or 'contact_acl' are forced to update,
   ensuring they pick up the new IP rules.

Signed-off-by: Michal Hajek michal.hajek@daktela.com

Fixes: #1641

### DIFF
--- a/res/res_pjsip_config_wizard.c
+++ b/res/res_pjsip_config_wizard.c
@@ -49,6 +49,9 @@
 #include "asterisk/pbx.h"
 #include "asterisk/sorcery.h"
 #include "asterisk/vector.h"
+#include "asterisk/stasis.h"
+#include "asterisk/acl.h"
+#include "asterisk/security_events.h"
 
 /*** DOCUMENTATION
 	<configInfo name="res_pjsip_config_wizard" language="en_US">
@@ -296,6 +299,22 @@ static AST_VECTOR_RW(object_type_wizards, struct object_type_wizard *) object_ty
 })
 
 const static char *object_types[] = {"phoneprov", "registration", "identify", "endpoint", "aor", "auth", NULL};
+
+static int acl_change_detected = 0;
+static struct stasis_subscription *acl_change_sub;
+
+/*! \brief Callback for Named ACL changed */
+static void acl_change_stasis_cb(void *data, struct stasis_subscription *sub, struct stasis_message *message)
+{
+	if (stasis_message_type(message) != ast_named_acl_change_type()) {
+		return;
+	}
+
+	ast_debug(3, "PJSIP Wizard: Named ACL change detected via Stasis. Triggering reload.\n");
+	acl_change_detected = 1;
+	ast_sorcery_reload(ast_sip_get_sorcery());
+	acl_change_detected = 0;
+}
 
 static int is_one_of(const char *needle, const char *haystack[])
 {
@@ -1064,7 +1083,9 @@ static void object_type_loaded_observer(const char *name,
 		return;
 	}
 
-	if (reloaded && otw->last_config) {
+	/* Only use the FILEUNCHANGED optimization if the ACLs haven't changed.
+	 * If ACLs changed, we force a reload of the config file to re-evaluate rules. */
+	if (reloaded && otw->last_config && !acl_change_detected) {
 		flags.flags = CONFIG_FLAG_FILEUNCHANGED;
 	}
 
@@ -1089,6 +1110,14 @@ static void object_type_loaded_observer(const char *name,
 		if (otw->last_config) {
 			last_cat = ast_category_get(otw->last_config, id, "type=^wizard$");
 			changes = !ast_variable_lists_match(ast_category_first(category), ast_category_first(last_cat), 1);
+
+			/* If the ACL has changed, we assume EVERYTHING might have changed.
+			 * We force an update for all wizard objects. */
+			if (!changes && reloaded && acl_change_detected) {
+				ast_debug(3, "Forcing update of wizard '%s' due to global ACL change.\n", id);
+				changes = 1;
+			}
+
 			if (last_cat) {
 				ast_category_delete(otw->last_config, last_cat);
 			}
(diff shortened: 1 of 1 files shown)


phoneben:
i think cherry-pick should be in a separate comment .
not in the original  .

hajekmi:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

phoneben:
can not be cherry picked to 20 ?

hajekmi:
> can not be cherry picked to 20 ?

Unfortunately, I don't know... version 20 - I don't know how good the Stasis is there... :)

jcolp:
The specific pattern of subscribing for ACL change events has been in use since 2013.

jcolp:
21 is security fix only and does not receive bug fixes. You should also not add an additional comment, but edit your existing one.

hajekmi:
Does my last commit look okay? :)

jcolp:
It has yet to be reviewed, when it is then you'll see updates/comments/etc.

res/res_pjsip_config_wizard.c
```diff
@@ -297,6 +300,9 @@ static AST_VECTOR_RW(object_type_wizards, struct object_type_wizard *) object_ty
 
 const static char *object_types[] = {"phoneprov", "registration", "identify", "endpoint", "aor", "auth", NULL};
 
+static struct stasis_subscription *acl_change_sub;
+static int acl_change_detected = 0;
```
jcolp, line 303: This is set once and never reset.

res/res_pjsip_config_wizard.c
```diff
@@ -1043,6 +1049,19 @@ struct ast_sorcery_instance_observer observer = {
...
+/*! \brief Callback for Named ACL changed */
+static void acl_change_stasis_cb(void *data, struct stasis_subscription *sub, struct stasis_message *message)
+{
+	/* We will verify whether it is an "ACL Change" message. */
+	if (stasis_message_type(message) != ast_named_acl_change_type()) {
+		return;
+	}
+
+	ast_debug(3, "PJSIP Wizard: Named ACL change detected via Stasis.\n");
+	acl_change_detected = 1;
```
jcolp: The commit message says this forces a reload on named ACL changes, but this code doesn't. It should and then reset the ACL change detected variable after.

res/res_pjsip_config_wizard.c
```diff
@@ -1053,6 +1072,8 @@ static void object_type_loaded_observer(const char *name,
 	struct ast_flags flags = { 0 };
 	struct ast_config *cfg;
 
+    int force_acl_reload = acl_change_detected;
```
jcolp: There's no need to do it this way. You can change it so the setting of flags.flags isn't done if ACL has reloaded, resulting in it not doing a file change check.

res/res_pjsip_config_wizard.c
```diff
@@ -296,6 +300,19 @@ static AST_VECTOR_RW(object_type_wizards, struct object_type_wizard *) object_ty
...
+
+/*! \brief Callback for Named ACL changed */
+static void acl_change_stasis_cb(void *data, struct stasis_subscription *sub, struct stasis_message *message)
+{
+	if (stasis_message_type(message) != ast_named_acl_change_type()) {
+		return;
+	}
+	/* We will raise the generation. This will avoid the need to manually reset the flag. */
+	current_acl_change_gen++;
+	ast_debug(3, "PJSIP Wizard: Named ACL change detected via Stasis. New generation: %d\n", current_acl_change_gen);
```
jcolp: This is still not forcing a reload of the named ACL when changed.  You need to set an acl_changed static int to 1, call ast_sorcery_reload(ast_sip_get_sorcery()), check for it where you are already, and then once ast_sorcery_reload returns set it to 0.

res/res_pjsip_config_wizard.c
```diff
@@ -284,6 +287,7 @@ struct object_type_wizard {
 	struct ast_sorcery_wizard *wizard;
 	void *wizard_data;
 	struct ast_config *last_config;
+	int last_acl_gen;
```
jcolp: There's no need to store a last_acl_gen.

res/res_pjsip_config_wizard.c
```diff
@@ -1089,6 +1115,33 @@ static void object_type_loaded_observer(const char *name,
...
+				struct ast_variable *v;
+				int uses_acl = 0;
+
+				for (v = ast_category_first(category); v; v = v->next) {
+					/* 1. Standard variables */
+					if (!strcasecmp(v->name, "acl") || !strcasecmp(v->name, "contact_acl")) {
+						uses_acl = 1;
+						break;
+					}
+					/* 2. Specific overrides for Endpoint */
```
jcolp: Do not hard code configuration names in here. Some can be added for example, which means this would most likely get missed and it would break again in specific circumstances. If the ACL changed, then assume there is always a change and reload.

jcolp (changes requested):

asteriskteam:

hajekmi:
I think you can merge it. Thanks

res/res_pjsip_config_wizard.c
```diff
@@ -1314,6 +1343,12 @@ static int load_module(void)
 	ast_sorcery_global_observer_add(&global_observer);
 	ast_cli_register_multiple(config_wizard_cli, ARRAY_LEN(config_wizard_cli));
 
+	acl_change_sub = stasis_subscribe(ast_security_topic(), acl_change_stasis_cb, NULL);
+	if (!acl_change_sub) {
+		ast_log(LOG_ERROR, "Failed to subscribe to Named ACL topic. Wizard won't auto-update on ACL changes.\n");
```
seanbright: I’d say this is one of those times where an error message is not helpful for the user. What are the practical reasons this function would fail?

hajekmi: Agreed. If stasis_subscribe fails, we are likely OOM and the system has bigger problems than logging this error. I will remove the check.

asteriskteam:

Issue #1645, opened 2025-12-10
[bug]: chan_websocket stuck channels


jcbjoe:
### Severity

Minor

### Versions

22.6.0

### Components/Modules

chan_websocket (Not sure)

### Operating Environment

Ubuntu 24.04.2

### Frequency of Occurrence

Frequent

### Issue Description

Apologies in advance for not having much information here. However it seems since implementing the new websockets for audiostreams we have stuck channels. Its not all channels and its not that often but if I run "core show channels" and then look at timestamps in logs for those calls they can be days ago. In my logs I can see the calls seems to have exited cleanly. To the point where I actually see dialplan continuing to process and make a second outbound call via our carrier.

The channels show as "UP" but I cannot run "core show channel {channel}"

If theres any more information I can provide please let me know and ill get that straight over to you! Im sure this is probably really tricky to debug. I haven't been able to notice a difference to a stuck call vs a one that ended cleanly. They seem the same

### Relevant log output

```shell
Channel is up but cannot get a status:

sipmedia1*CLI> core show channel WebSocket/{websocket}/0x798f18095410
WebSocket/{websocket}/0x798f18095410 is not a known channel
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
Interesting.  I ran into this myself a few minutes ago while I was debugging the crash issue.  I'll look at this when I'm done with the other one.

gtjoseph:
I can reproduce this.

jcbjoe:
Hey @gtjoseph It seems that this merge did not fix the stuck channels.

I have just emailed you a video showing you all the channels that still exist. Let me know if you need any more information or logs!

gtjoseph:
Ok, we're going to need a lot more info but the channels themselves probably aren't stuck.  The `core show channels` list is retrieved from channel snapshots and which don't seem to be cleaned up correctly.  I _think_ the channels themselves are gone.

* Can you run `sudo netstat -antp` or `sudo ss -antp` to see if there are actually open sockets for these channels?
* Can you create a simple scenario that we can reproduce?
* Can you run `ast_coredumper --RUNNING --tarball-coredumps` to get a snapshot of a running system?  This will pause call processing for a short time so you may not be able to do this unless you can reproduce the issue on a test system.
* What's app_audiofork?  That's not one of the distributed modules.
* What other patches are you applying to the source, if any?
[...]

Issue #1643, opened 2025-12-10
[bug]: chan_websocket crash when channel hung up before read thread is started


jcbjoe:
### Severity

Major

### Versions

22.6.0

### Components/Modules

chan_websocket

### Operating Environment

Ubuntu 24.04.2

### Frequency of Occurrence

Frequent

### Issue Description

We have come across a segfault after implementing the new websocket transport. We haven't been able to pin point exactly whats causing this. But we have been able to get a core dump. I can provide the core dump on request. It happens usually once a day if not more. Happy to test out changes and recompiles of asterisk.

### Relevant log output

```shell
kernlog
2025-12-10T11:02:55.396023+00:00 sipmedia1 kernel: asterisk[498229]: segfault at 140 ip 000079b7f7ff8fa3 sp 000079b7f4273c80 error 4 in res_http_websocket.so[79b7f7ff5000+6000] likely on CPU 0 (core 0, socket 0)
2025-12-10T11:02:55.396041+00:00 sipmedia1 kernel: Code: ec 08 4c 8d 05 be 2f 00 00 48 8d 0d e7 3a 00 00 ba 8d 01 00 00 41 55 4c 89 f6 31 ff 31 c0 e8 c4 c8 ff ff 41 5a 41 5b 44 89 e1 <48> 8b 83 40 01 00 00 83 c9 80 49 83 fd 7d 0f 86 f9 00 00 00 49 81

CoreDump:
@!@!@! thread1.txt !@!@!@!

$1 = {si_signo = 11, si_errno = 0, si_code = 1, _sifields = {_pad = {320, 0 <repeats 27 times>}, _kill = {si_pid = 320, si_uid = 0}, _timer = {si_tid = 320, si_overrun = 0, si_sigval = {sival_int = 0, sival_ptr = 0x0}}, _rt = {si_pid = 320, si_uid = 0, si_sigval = {sival_int = 0, sival_ptr = 0x0}}, _sigchld = {si_pid = 320, si_uid = 0, si_status = 0, si_utime = 0, si_stime = 0}, _sigfault = {si_addr = 0x140, _addr_lsb = 0, _addr_bnd = {_lower = 0x0, _upper = 0x0}}, _sigpoll = {si_band = 320, si_fd = 0}, _sigsys = {_call_addr = 0x140, _syscall = 0, _arch = 0}}}
Signal        Stop      Print   Pass to program Description
SIGSEGV       Yes       Yes     Yes             Segmentation fault

[...]

PR #1646, opened 2025-12-10
chan_websocket: Add locking in send_event and check for NULL websocket handle.

gtjoseph:
On an outbound websocket connection, when the triggering caller hangs up,
webchan_hangup() closes the outbound websocket session and sets the websocket
session handle to NULL.  If the hangup happened in the tiny window between
opening the outbound websocket connection and before read_thread_handler()
was able to send the MEDIA_START message, it could segfault because the
websocket session handle was NULL.  If it didn't actually segfault, there was
also the possibility that the websocket instance wouldn't get cleaned up which
could also cause the channel snapshot to not get cleaned up.  That could
cause memory leaks and `core show channels` to list phantom WebSocket
channels.

To prevent the race, the send_event() macro now locks the websocket_pvt
instance and checks the websocket session handle before attempting to send
the MEDIA_START message.

Resolves: #1643
Resolves: #1645

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -416,7 +416,8 @@ static __attribute__ ((format (gnu_printf, 2, 3))) char *_create_event_ERROR(
 ({ \
 	int _res = -1; \
 	char *_payload = _create_event_ ## _event(_instance, ##__VA_ARGS__); \
-	if (_payload) { \
+	ao2_lock(instance); \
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

bkford (approved):

PR #1656, opened 2025-12-16
chan_mobile.c: fix bluetooth mtu and add adapter/endpoint auto detection, fix different HFP versions and different bluetooth adapter version

VasylSamoilov:
chan_mobile.c: fix bluetooth mtu and add adapter/endpoint auto detection, fix different HFP versions and different bluetooth adapter version


VasylSamoilov:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1660, opened 2025-12-18
[bug]: missing hangup cause for ARI ChannelDestroyed when Originated channel times out


sarangr7:
### Severity

Minor

### Versions

20,21,22

### Components/Modules

ari,stasis

### Operating Environment

Linux

### Frequency of Occurrence

Constant

### Issue Description

**Summary**
When an originated channel times out, the hangup cause was not being properly set on the channel before hanging up. This could result in incorrect or missing hangup cause information being reported to the caller.

**Problem**
In handle_timeout_trip(), when channels are hung up due to timeout expiration, only the ast_dial_channel structure's cause field was being set to AST_CAUSE_NO_ANSWER. However, the actual ast_channel's hangup cause was never set before calling ast_hangup().

This means:

- The channel's hangup cause would not reflect AST_CAUSE_NO_ANSWER (cause code 19)
- Applications monitoring hangup causes would not see the correct reason for call termination

### Relevant log output

```shell
{
  "type": "ChannelDestroyed",
  "timestamp": "2024-10-26T19:04:48.295+0000",
  "cause": 0,
  "cause_txt": "Unknown",
  "channel": {
    "id": "54af97ff-3e76-4ace-b85d-5c2b16ced02e",
    "name": "PJSIP/kiwi-0000000f",
    "state": "Ringing",
    "protocol_id": "d2493982-7f87-4ece-9389-b0d0e7b22b5d",
    "caller": {
      "name": "Asteraceae",
      "number": "aster"
    },
    "connected": {
      "name": "Asteraceae",
      "number": "aster"
    },
    "accountcode": "kiwi",
    "dialplan": {
      "context": "softphone",
      "exten": "s",
      "priority": 1,
      "app_name": "AppDial2",
      "app_data": "(Outgoing Line)"
    },
    "creationtime": "2024-10-26T19:04:18.259+0000",
    "language": "en"
  },
  "asterisk_id": "00:00:00:00:00:00",
  "application": "mystasis-app"
}
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

sumit-chaurasi:
Before calling ast_hangup(chan), you must set the hangup cause on the channel itself.

PR #1661, opened 2025-12-18
main/dial.c: Set channel hangup cause on timeout in handle_timeout_trip

sarangr7:
When dial attempts timeout in the core dialing API, the channel's hangup cause was not being set before hanging up. Only the ast_dial_channel structure's internal cause field was updated, but the actual ast_channel hangup cause remained unset.

This resulted in incorrect or missing hangup cause information being reported through CDRs, AMI events, and other mechanisms that read the channel's hangup cause when dial timeouts occurred via applications using the dialing API (FollowMe, Page, etc.).

The fix adds proper channel locking and sets AST_CAUSE_NO_ANSWER on the channel before calling ast_hangup(), ensuring consistent hangup cause reporting across all interfaces.

Resolves: #1660


sarangr7:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

PR #1666, opened 2025-12-22
rtp/rtcp: Configure dual-stack behavior via IPV6_V6ONLY

scsiguy:
Dual-stack behavior (simultaneous listening for IPV4 and IPV6 connections on a single socket) is required by Asterisk's ICE implementation.  On systems with the IPV6_V6ONLY sockopt, set the option to 0 (dual-stack enabled) when binding to the IPV6 any address. This ensures correct behavior regardless of the system's default dual-stack configuration.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1248,6 +1248,19 @@ AC_RUN_IFELSE(
 	AC_MSG_RESULT(cross-compile)
 )
 
+AC_MSG_CHECKING([if setsockopt() accepts the IPV6_V6ONLY socket option])
+AC_COMPILE_IFELSE(
+[AC_LANG_PROGRAM([
+	   #include <sys/socket.h>
+	   #include <netinet/in.h>
+	], [
+	   int opt = IPV6_V6ONLY;
+	])],
+        AC_MSG_RESULT(yes)
+        AC_DEFINE(HAVE_SOCK_IPV6_V6ONLY, 1, [Define to 1 if your socket() implementation has the IPV6_V6ONLY socket option.]),
+	AC_MSG_RESULT(no)
+)
+
 AC_MSG_CHECKING(if we can increase the maximum select-able file descriptor)
 AC_RUN_IFELSE(
 [AC_LANG_PROGRAM([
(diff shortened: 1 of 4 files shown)


configure.ac
```diff
@@ -1248,6 +1248,19 @@ AC_RUN_IFELSE(
 	AC_MSG_RESULT(cross-compile)
 )
 
+AC_MSG_CHECKING([if socket() accepts the IPV6_V6ONLY socket option])
```
seanbright: This should probably say `if setsockopt() accepts...`

res/res_rtp_asterisk.c
```diff
@@ -3580,6 +3584,19 @@ static int create_new_socket(const char *type, int af)
 	}
 #endif
 
+#ifdef HAVE_SOCK_IPV6_V6ONLY
+	if (AF_INET6 == af && ast_sockaddr_is_any(bind_addr)) {
+		/* ICE relies on dual-stack behavior. Ensure it is enabled. */
+		int v6_only = 0;
+		int result;
+		result = setsockopt(sock, IPPROTO_IPV6, IPV6_V6ONLY, &v6_only, sizeof(v6_only));
```
seanbright: Use a compound literal instead of creating a new variable just to hold 0.

    ... IPV6_V6ONLY, &(int){0}, sizeof(int) ...

Also just do all of this is in the `if` statement:

    if (setsockopt(...)) {
        ast_log(...);
    }

scsiguy:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

scsiguy:
The last push just fixed a typo in the commit comment.

scsiguy:
I updated the PR description to match the commit, but clicking re-review for the bot doesn't seem to be triggering a re-validation of this check.

seanbright:
It’s fine. Don’t worry about it.

jcolp (approved):

mzary:
Thank you for this fix, it's highly appreciated by the Asterisk FreeBSD community.  At last, one old bug report on FreeBSD Bugzilla can be closed https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=262386
Edit: The bug mentioned above is not resolved by this change, but it improves *BSD compatibility, which is still good news.

Issue #1670, opened 2025-12-22
[new-feature]: Add new option to PJSIP_HEADER to pass headers from the inbound to outbound channel.


mbradeen:
### Feature Description

Add a new option to PJSIP_HEADER that can be used on an inbound channel to pass headers from the inbound channel to an outbound pjsip channel's INVITE upon Dial.

PR #1671, opened 2025-12-22
res_pjsip_header_funcs: Add new PJSIP_INHERITABLE_HEADER dialplan function

mbradeen:
Adds a new PJSIP_INHERITABLE_HEADER dialplan function to add
inheritable headers from the inbound channel to an outbound
bridged channel.  This works similarly to the existing
PJSIP_HEADER function, but will set the header on the bridged
outbound channel's INVITE upon Dial.

Inheritable headers can be updated or removed from the inbound
channel as well as from a pre-dial handler

Resolves: #1670

UserNote: A new PJSIP_HEADER option has been added that allows
inheriting pjsip headers from the inbound to the outbound bridged
channel.
Example- same => n,Set(PJSIP_INHERITABLE_HEADER(add,X-custom-1)=alpha)
will add X-custom-1: alpha to the outbound pjsip channel INVITE
upon Dial.

### DIFF
--- a/res/res_pjsip_header_funcs.c
+++ b/res/res_pjsip_header_funcs.c
@@ -52,7 +52,7 @@
 				<enumlist>
 					<enum name="read"><para>Returns instance <replaceable>number</replaceable>
 					of header <replaceable>name</replaceable>. A <literal>*</literal>
-					may be appended to <replaceable>name</replaceable> to iterate over all 
+					may be appended to <replaceable>name</replaceable> to iterate over all
 					headers <emphasis>beginning with</emphasis> <replaceable>name</replaceable>.
 					</para></enum>
 
@@ -141,6 +141,94 @@
 
 			[somecontext]
 			exten => 1,1,Dial(PJSIP/${EXTEN},,b(handler^addheader^1))
+			</example>
+		</description>
+	</function>
+	<function name="PJSIP_INHERITABLE_HEADER" language="en_US">
+		<since>
+			<version>20.19.0</version>
+			<version>22.9.0</version>
+			<version>23.3.0</version>
+		</since>
+		<synopsis>
+			Adds, updates or removes the specified SIP header from a PJSIP or non-PJSIP channel
+			to be inherited to an outbound PJSIP channel.
+		</synopsis>
+		<syntax>
+			<parameter name="action" required="true">
+				<enumlist>
+
+					<enum name="add"><para>Adds a new header <replaceable>name</replaceable>
+					to this channel.</para></enum>
+
+					<enum name="update"><para>Updates instance <replaceable>number</replaceable>
+					of header <replaceable>name</replaceable> to a new value.
+					The header must already exist.</para></enum>
+
+					<enum name="remove"><para>Removes all instances of previously added headers
+					whose names match <replaceable>name</replaceable>. A <literal>*</literal>
+					may be appended to <replaceable>name</replaceable> to remove all headers
+					<emphasis>beginning with</emphasis> <replaceable>name</replaceable>.
+					<replaceable>name</replaceable> may be set to a single <literal>*</literal>
+					to clear <emphasis>all</emphasis> previously added headers. In all cases,
+					the number of headers actually removed is returned.</para></enum>
+				</enumlist>
+			</parameter>
+
+			<parameter name="name" required="true"><para>The name of the header.</para></parameter>
+
+			<parameter name="number" required="false">
+				<para>If there's more than 1 header with the same name, this specifies which header
+				to read or update.  If not specified, defaults to <literal>1</literal> meaning
+				the first matching header.  Not valid for <literal>add</literal> or
+				<literal>remove</literal>.</para>
+			</parameter>
+
+		</syntax>
+		<description>
+			<para>PJSIP_INHERITABLE_HEADER allows you to write specific SIP headers on a calling
+			channel to be inherited by an outbound PJSIP channel.</para>
+			<para>Examples:</para>
+			<example title="Add an inheritable X-Myheader header with the value of myvalue">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(add,X-MyHeader)=myvalue)
+			</example>
+			<example title="Add an inheritable X-Myheader header with an empty value">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(add,X-MyHeader)=)
+			</example>
+			<example title="Add an inheritable X-MyHeader that will be inherited by child channel">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(add,X-MyHeader)=myvalue)
+			</example>
+			<example title="Update the value of the inheritable header named X-Myheader to newvalue">
+			; 'X-Myheader' must already exist or the call will fail.
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(update,X-MyHeader)=newvalue)
+			</example>
+			<example title="Remove all inheritable headers whose names exactly match X-MyHeader">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(remove,X-MyHeader)=)
+			</example>
+			<example title="Remove all inheritable headers that begin with X-My">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(remove,X-My*)=)
+			</example>
+
+			<note><para>If you call PJSIP_INHERITABLE_HEADER in a normal dialplan context you'll be
+			operating on the <emphasis>caller's</emphasis> channel which
+			will then be inherited to the <emphasis>callee's (outgoing)</emphasis>
+			channel. Inherited headers can be updated or removed via PJSIP_INHERITABLE_HEADER
+			in a pre-dial handler. </para>
+			<para>Headers added via PJSIP_INHERITABLE_HEADER are separate from headers
+			added via PJSIP_HEADER. A header added via PJSIP_INHERITABLE_HEADER can only
+			be or removed modified by PJSIP_INHERITABLE_HEADER. A header added via
+			PJSIP_HEADER can only be modified or removed by PJSIP_HEADER.</para></note>
+			<example title="Set and modify headers on callee channel">
+			[handler]
+			exten => modheader,1,Set(PJSIP_INHERITABLE_HEADER(update,X-MyHeader)=myvalue)
+				same => n,Set(PJSIP_INHERITABLE_HEADER(update,X-MyHeader2)=myvalueX)
+				same => n,Set(PJSIP_INHERITABLE_HEADER(remove,X-MyHeader2)=)
+
+			[somecontext]
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(add,X-MyHeader1)=myvalue1)
+				same => n,Set(PJSIP_INHERITABLE_HEADER(add,X-MyHeader2)=myvalue2)
+				same => n,Set(PJSIP_INHERITABLE_HEADER(add,X-MyHeader3)=myvalue3)
+				same => n,Dial(PJSIP/${EXTEN},,b(handler^modheader^1))
 			</example>
 		</description>
 	</function>
@@ -323,10 +411,39 @@ struct hdr_list_entry {
 };
 AST_LIST_HEAD_NOLOCK(hdr_list, hdr_list_entry);
 
-/*! \brief Datastore for saving headers */
-static const struct ast_datastore_info header_datastore = {
-	.type = "header_datastore",
+/*!
+ * \internal
+ * \brief Duplicate an inheritable headers list for inheritance
+ *
+ * Creates copies of the name-value pairs for inheritance to child channels.
+ */
+static void *inheritable_headers_duplicate(void *data)
+{
+	return ast_variables_dup(data);
+}
+
+/*!
+ * \internal
+ * \brief Destroy callback for inherited header datastore
+ */
+static void inheritable_headers_destroy(void *data)
+{
+	ast_variables_destroy(data);
+}
+
+/*! \brief Datastore for saving headers in session (contains hdr_list, no destroy needed) */
+static const struct ast_datastore_info session_header_datastore = {
+	.type = "session_header_datastore",
+	/* No destroy or duplicate callback - data is pool-allocated hdr_list */
 };
+
+/*! \brief Datastore for saving inheritable headers (contains hdr_list with ast_malloc'd entries) */
+static const struct ast_datastore_info inheritable_header_datastore = {
+	.type = "inheritable_header_datastore",
+	.duplicate = inheritable_headers_duplicate,
+	.destroy = inheritable_headers_destroy,
+};
+
 /*! \brief Datastore for saving response headers */
 static const struct ast_datastore_info response_header_datastore = {
 	.type = "response_header_datastore",
@@ -377,11 +494,11 @@ static int incoming_request(struct ast_sip_session *session, pjsip_rx_data * rda
 {
 	pj_pool_t *pool = session->inv_session->dlg->pool;
 	RAII_VAR(struct ast_datastore *, datastore,
-			 ast_sip_session_get_datastore(session, header_datastore.type), ao2_cleanup);
+			 ast_sip_session_get_datastore(session, session_header_datastore.type), ao2_cleanup);
 
 	if (!datastore) {
 		if (!(datastore =
-			  ast_sip_session_alloc_datastore(&header_datastore, header_datastore.type))
+			  ast_sip_session_alloc_datastore(&session_header_datastore, session_header_datastore.type))
 			||
 			!(datastore->data = pj_pool_alloc(pool, sizeof(struct hdr_list))) ||
 			ast_sip_session_add_datastore(session, datastore)) {
@@ -580,7 +697,7 @@ static int read_header(void *obj)
 
 	list = datastore->data;
 	AST_LIST_TRAVERSE(list, le, nextptr) {
-		if (data->header_name[len - 1] == '*') {
+		if (len >= 1 && data->header_name[len - 1] == '*') {
 			if (pj_strnicmp2(&le->hdr->name, data->header_name, len - 1) == 0 && i++ == data->header_number) {
 				hdr = le->hdr;
 				break;
@@ -651,11 +768,11 @@ static int add_header(void *obj)
 	struct hdr_list *list;
 
 	RAII_VAR(struct ast_datastore *, datastore,
-			 ast_sip_session_get_datastore(session, data->header_datastore->type), ao2_cleanup);
+			 ast_sip_session_get_datastore(session, session_header_datastore.type), ao2_cleanup);
 
 	if (!datastore) {
-		if (!(datastore = ast_sip_session_alloc_datastore(data->header_datastore,
-														data->header_datastore->type))
+		if (!(datastore = ast_sip_session_alloc_datastore(&session_header_datastore,
+														session_header_datastore.type))
 			|| !(datastore->data = pj_pool_alloc(pool, sizeof(struct hdr_list)))
 			|| ast_sip_session_add_datastore(session, datastore)) {
 			ast_log(AST_LOG_ERROR, "Unable to create datastore for header functions.\n");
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
testsuite-test-pr: 120

res/res_pjsip_header_funcs.c
```diff
@@ -323,10 +329,139 @@ struct hdr_list_entry {
...
+		if (!new_le) {
+			ast_log(AST_LOG_ERROR, "hdr_list_duplicate: Unable to allocate new list entry\n");
+			hdr_list_destroy(new_list);
+			return NULL;
+		}
+
+		/* Store the header name and value as a serialized string in the hdr pointer.
+		 * We'll reconstruct the actual pjsip_hdr in outgoing_request.
+		 * For now, we allocate a minimal structure to hold name/value strings. */
+		new_le->hdr = (pjsip_hdr *)ast_calloc(1, sizeof(pjsip_generic_string_hdr) +
```
InterLinked1: I don't think the cast is needed, and if it were, there should be a space before `ast_calloc`

res/res_pjsip_header_funcs.c
```diff
@@ -323,10 +329,139 @@ struct hdr_list_entry {
...
+		if (!new_le->hdr) {
+			ast_free(new_le);
+			ast_log(AST_LOG_ERROR, "hdr_list_duplicate: Unable to allocate new header\n");
+			hdr_list_destroy(new_list);
+			return NULL;
+		}
+
+		/* Copy header name and value as strings */
+		new_hdr = (pjsip_generic_string_hdr *)new_le->hdr;
+		data_ptr = (char *)new_hdr + sizeof(pjsip_generic_string_hdr);
```
InterLinked1: Cast / whitespace

res/res_pjsip_header_funcs.c
```diff
@@ -679,6 +814,111 @@ static int add_header(void *obj)
...
+static int pass_header(void *obj)
+{
+	struct header_data *data = obj;
+	struct ast_sip_session *session = data->channel->session;
+	struct hdr_list *chan_hdr_list;
+
+	ast_debug(3, "pass_header: using inheritable channel datastore\n");
+
+	/* Also add to channel datastore for inheritance support.
+		* We create a separate hdr_list to avoid including ALL headers. */
```
InterLinked1: Comment slightly misaligned here

res/res_pjsip_header_funcs.c
```diff
@@ -694,26 +934,124 @@ static int update_header(void *obj)
...
+					char *data_ptr;
+					gen_hdr = ast_calloc(1, sizeof(*gen_hdr) + old_hdr->name.slen + new_len + 2);
+					ast_debug(3, "update_header: reallocating header %s to accommodate new value %s\n",
+						data->header_name, data->header_value);
+					if (!gen_hdr) {
+						ast_log(AST_LOG_ERROR, "Failed to allocate memory for updated header\n");
+						return -1;
+					}
+
+					data_ptr = (char *)gen_hdr + sizeof(*gen_hdr);
```
InterLinked1: Cast / whitespace

res/res_pjsip_header_funcs.c
```diff
@@ -52,13 +52,16 @@
...
-					may be appended to <replaceable>name</replaceable> to iterate over all 
+					may be appended to <replaceable>name</replaceable> to iterate over all
 					headers <emphasis>beginning with</emphasis> <replaceable>name</replaceable>.
 					</para></enum>
 
 					<enum name="add"><para>Adds a new header <replaceable>name</replaceable>
 					to this session.</para></enum>
 
+					<enum name="pass"><para>Adds a new header <replaceable>name</replaceable>
+					to this channel to be passed to the outbound channel.</para></enum>
```
InterLinked1: It might be good to clarify if this only works if the outbound channel is bridged directly with the incoming channel (i.e. I imagine this wouldn't work on a PJSIP -> Local -> PJSIP call)

oza4h07:
Would it make sense to allow wildcard expressions such as PJSIP_HEADER(pass,X-MyCompany-*) ?

jcolp:
This doesn't pass through received headers from the incoming SIP leg INVITE, as a result I don't think this should be called "pass" because of that exact confusion. What it does is allow adding headers to the calling side which are INHERITED to the outgoing leg. "inherit" would be better.

res/res_pjsip_header_funcs.c
```diff
@@ -52,13 +52,16 @@
...
 					of header <replaceable>name</replaceable>. A <literal>*</literal>
-					may be appended to <replaceable>name</replaceable> to iterate over all 
+					may be appended to <replaceable>name</replaceable> to iterate over all
 					headers <emphasis>beginning with</emphasis> <replaceable>name</replaceable>.
 					</para></enum>
 
 					<enum name="add"><para>Adds a new header <replaceable>name</replaceable>
 					to this session.</para></enum>
 
+					<enum name="pass"><para>Adds a new header <replaceable>name</replaceable>
```
jcolp: inherit, not pass, and document that this does NOT automatically pass through received SIP headers

res/res_pjsip_header_funcs.c
```diff
@@ -679,6 +814,111 @@ static int add_header(void *obj)
...
+ * Create pjsip_msg and hdr_list_entry.
+ * Add the entry to the list.
+ */
+static int pass_header(void *obj)
+{
+	struct header_data *data = obj;
+	struct ast_sip_session *session = data->channel->session;
+	struct hdr_list *chan_hdr_list;
+
+	ast_debug(3, "pass_header: using inheritable channel datastore\n");
```
jcolp: this debug message if it is useful should contain more information like the session name, the same applies for all added log messages - approach them as a user, make them useful/insightful to the user including what session/channel they refer to

jcolp:
Is this test failure expected for some reason?

mbradeen:
> Is this test failure expected for some reason?

This ran before I updated the associated [testsuite PR](https://github.com/asterisk/testsuite/pull/120/changes#diff-9eba51de94fd2362d9da9e4191688f8612e72e97ebb8e0262ae905da46934b0cR67) to move from pass->inherit:
`Executing [inbound-pass@default:3] Set("PJSIP/inbound-00000003", "PJSIP_HEADER(pass,X-custom-1)=alpha") in new stack`
It should pass now.

xhuang-sangoma:
Hi guys, appreciate the effort here.

I have one question:

It's good that now there's the inherit option to pass a SIP header from an inbound PJSIP channel down to the outbound PJSIP channel. However, there's a lot of situations that we want to add a SIP header when the dialplan is running on an intermediate Local channel, but the PJSIP_HEADER func can only be run only on a PJSIP channel. This will be quite a limitation for us to use it.

The original SIPAddHeader app can be used on any type of channel including SIP and Local channels, which gives the flexibility to add a header anytime and anywhere in the dialplan without having to think which type of channel the dialplan is currently running on.

Is it possible to add similar PJSIPAddHeader and PJSIPRemoveHeader apps working the same old way?

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,91 @@
...
 		</description>
 	</function>
+	<function name="PJSIP_INHERITABLE_HEADER" language="en_US">
+		<since>
+			<version>20.17.0</version>
+			<version>22.7.0</version>
+			<version>23.1.0</version>
+		</since>
+		<synopsis>
+			Adds, updates or removes the specified SIP header from a channel to be inherited
```
jcolp: specify that the channel can be PJSIP or non-PJSIP

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,91 @@
...
+			to an outbound PJSIP channel.
+		</synopsis>
+		<syntax>
+			<parameter name="action" required="true">
+				<enumlist>
+
+					<enum name="add"><para>Adds a new header <replaceable>name</replaceable>
+					to this session.</para></enum>
```
jcolp: it's not to a session

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,91 @@
...
+			</example>
+			<example title="Remove all inheritable headers whose names exactly match X-MyHeader">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(remove,X-MyHeader)=)
+			</example>
+			<example title="Remove all inheritable headers that begin with X-My">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(remove,X-My*)=)
+			</example>
+
+			<note><para>If you call PJSIP_INHERITABLE_HEADER in a normal dialplan context you'll be
+			operating on the <emphasis>caller's (incoming)</emphasis> channel which
```
jcolp: not always incoming, so remove that aspect

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,91 @@
...
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(update,X-MyHeader)=newvalue)
+			</example>
+			<example title="Remove all inheritable headers whose names exactly match X-MyHeader">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(remove,X-MyHeader)=)
+			</example>
+			<example title="Remove all inheritable headers that begin with X-My">
+			exten => 1,1,Set(PJSIP_INHERITABLE_HEADER(remove,X-My*)=)
+			</example>
+
+			<note><para>If you call PJSIP_INHERITABLE_HEADER in a normal dialplan context you'll be
```
jcolp, line 212: Has this been tested through a Local channel, or multiple?

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,91 @@
...
+			operating on the <emphasis>caller's (incoming)</emphasis> channel which
+			will then be inherited to the <emphasis>callee's (outgoing)</emphasis>
+			channel. Inherited geaders can be updated or removed via PJSIP_INHERITABLE_HEADER
+			in a pre-dial handler. </para></note>
```
jcolp: Document how this interacts with PJSIP_HEADER in the pre-dial handler

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,91 @@
...
+				<para>If there's more than 1 header with the same name, this specifies which header
+				to read or update.  If not specified, defaults to <literal>1</literal> meaning
+				the first matching header.  Not valid for <literal>add</literal> or
+				<literal>remove</literal>.</para>
+			</parameter>
+
+		</syntax>
+		<description>
+			<para>PJSIP_INHERITABLE_HEADER allows you to write specific SIP headers on an
+			inbound channel to be inherited by an outbound PJSIP channel.
```
jcolp: on a calling channel to be inherited by an outbound PJSIP channel.

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,91 @@
...
+			added on the outbound channel.</para>
```
jcolp: stray old part from copying?

res/res_pjsip_header_funcs.c
```diff
@@ -1000,12 +1494,76 @@ static int func_write_header(struct ast_channel *chan, const char *cmd, char *da
...
+	}
+
+	header_data.channel = channel;
+	header_data.header_name = args.header_name;
+	header_data.header_number = header_number;
+	header_data.header_value = value;
+	header_data.buf = NULL;
+	header_data.len = 0;
+	header_data.header_datastore = &session_header_datastore;
+
```
jcolp: This needs a rewrite to work on non-PJSIP channels, which will probably simplify it a lot. The datastore should be put on the chan passed in. No reference to session is possible. No PJSIP function that uses a pool can be used, so the easiest option is to just have the datastore contain an ast_variable list with the name being header name, and value being header value. Manipulation will then be easy for add/remove/update. Once it hits the outgoing PJSIP leg it can then be turned into proper headers using ast_sip_add_header.

jcolp (changes requested):
The UserNote on this is also incorrect now.

res/res_pjsip_header_funcs.c
```diff
@@ -1000,12 +1410,73 @@ static int func_write_header(struct ast_channel *chan, const char *cmd, char *da
...
+	}
+	if (!args.header_number) {
+		header_number = 1;
+	} else {
+		sscanf(args.header_number, "%30d", &header_number);
+		if (header_number < 1) {
+			header_number = 1;
+		}
+	}
+
```
jcolp, line 1444: You're not queueing a task so you don't need to have a header_data structure with fields. Each of the implementations can just take in the arguments directly.

res/res_pjsip_header_funcs.c
```diff
@@ -1000,12 +1410,73 @@ static int func_write_header(struct ast_channel *chan, const char *cmd, char *da
...
+							 const char *value)
+{
+	struct inheritable_header_data header_data;
+	int header_number;
+	AST_DECLARE_APP_ARGS(args,
+						 AST_APP_ARG(action);
+						 AST_APP_ARG(header_name);
+						 AST_APP_ARG(header_number););
+	AST_STANDARD_APP_ARGS(args, data);
+
```
jcolp, line 1423: If no channel just bounce out early and remove the channel check from the implementations.

res/res_pjsip_header_funcs.c
```diff
@@ -1035,21 +1506,99 @@ static struct ast_custom_function pjsip_response_headers_function = {
...
+		int inherited_count = 0;
+		struct ast_variable *headers = inherited_datastore->data;
+		struct ast_variable *var;
+
+		for (var = headers; var; var = var->next) {
+			pj_str_t pj_name, pj_value;
+			pjsip_hdr *hdr;
+
+			/* Create header using the tdata pool */
+			pj_cstr(&pj_name, var->name);
```
jcolp: You can just use ast_sip_add_header here and remove most of this code.

res/res_pjsip_header_funcs.c
```diff
@@ -1000,12 +1383,68 @@ static int func_write_header(struct ast_channel *chan, const char *cmd, char *da
...
+{
+	int header_number;
+	AST_DECLARE_APP_ARGS(args,
+						 AST_APP_ARG(action);
+						 AST_APP_ARG(header_name);
+						 AST_APP_ARG(header_number););
+	AST_STANDARD_APP_ARGS(args, data);
+
+	if (!chan) {
+		ast_log(LOG_ERROR, "This function requires a channel.\n");
```
jcolp: Specify in this error messages what dialplan function

res/res_pjsip_header_funcs.c
```diff
@@ -732,37 +1006,145 @@ static int remove_header(void *obj)
...
+	size_t len = strlen(header_name);
+	struct ast_variable *headers = NULL;
+	struct ast_variable *var, *prev = NULL;
+	int removed_count = 0;
+	struct ast_datastore *inherited_datastore = ast_channel_datastore_find(channel, &inheritable_header_datastore, NULL);
+	ast_debug(3, "Attempting to remove header %s from channel %s\n",
+		header_name, ast_channel_name(channel));
+
+	/* Remove from inherited datastore */
+	if (inherited_datastore && inherited_datastore->data) {
```
jcolp: You can return early here if no inherited datastore, or no data, and eliminate indentation

res/res_pjsip_header_funcs.c
```diff
@@ -732,37 +1006,145 @@ static int remove_header(void *obj)
...
+
 	if (!datastore || !datastore->data) {
 		ast_log(AST_LOG_ERROR, "No headers had been previously added to this session.\n");
 		return -1;
 	}
 
 	list = datastore->data;
 	AST_LIST_TRAVERSE_SAFE_BEGIN(list, le, nextptr) {
-		if (data->header_name[len - 1] == '*') {
+		if (len >= 1 && data->header_name[len - 1] == '*') {
```
jcolp, line 1036: What prompted the addition of this?

res/res_pjsip_header_funcs.c
```diff
@@ -1000,12 +1383,68 @@ static int func_write_header(struct ast_channel *chan, const char *cmd, char *da
...
+	} else {
+		sscanf(args.header_number, "%30d", &header_number);
+		if (header_number < 1) {
+			header_number = 1;
+		}
+	}
+
+	/* TODO: we might want to check channel type and use the serializer if PJSIP */
+	if (!strcasecmp(args.action, "add")) {
+		return add_inheritable_header(chan, args.header_name, value);
```
jcolp, line 1447: I think it may be possible for these to be called from two threads at once - specifically from the dialplan and over AMI or ARI, in which case the datastore would not be protected. Locking the channel during datastore access would be safest.

res/res_pjsip_header_funcs.c
```diff
@@ -144,6 +144,94 @@
 			</example>
 		</description>
 	</function>
+	<function name="PJSIP_INHERITABLE_HEADER" language="en_US">
+		<since>
+			<version>20.17.0</version>
```
jcolp: Update the versions. This will land after the next set of releases.

jcolp (changes requested):

res/res_pjsip_header_funcs.c
mbradeen: Yes, the TS test has been updated

res/res_pjsip_header_funcs.c
mbradeen: If somehow an empty header name came through

res/res_pjsip_header_funcs.c
```diff
@@ -323,10 +411,39 @@ struct hdr_list_entry {
...
-	.type = "header_datastore",
+/*!
+ * \internal
+ * \brief Duplicate an inheritable headers list for inheritance
+ *
+ * Creates copies of the name-value pairs for inheritance to child channels.
+ */
+static void *inheritable_headers_duplicate(void *data)
+{
+	return ast_variables_dup((struct ast_variable *) data);
```
jcolp: There is no need to cast this

res/res_pjsip_header_funcs.c
```diff
@@ -323,10 +411,39 @@ struct hdr_list_entry {
...
+}
+
+/*!
+ * \internal
+ * \brief Destroy callback for inherited header datastore
+ */
+static void inheritable_headers_destroy(void *data)
+{
+	ast_variables_destroy((struct ast_variable *) data);
```
jcolp: There is no need to cast this

res/res_pjsip_header_funcs.c
```diff
@@ -679,6 +796,78 @@ static int add_header(void *obj)
...
+
+	if (!chan_datastore) {
+		/* Create new channel datastore */
+		chan_datastore = ast_datastore_alloc(&inheritable_header_datastore,
+			inheritable_header_datastore.type);
+		if (chan_datastore) {
+			chan_datastore->data = NULL;
+			chan_datastore->inheritance = DATASTORE_INHERIT_FOREVER;
+			ast_channel_datastore_add(channel, chan_datastore);
+			ast_debug(3, "Created new inheritable channel datastore for channel %s\n",
```
jcolp: Specify that it's a datastore for inheritable SIP headers

res/res_pjsip_header_funcs.c
```diff
@@ -679,6 +796,78 @@ static int add_header(void *obj)
...
+				ast_channel_name(channel));
+		} else {
+			ast_log(LOG_ERROR, "Failed to allocate channel datastore for channel %s\n",
```
jcolp: This is not specific enough for a user - "Failed to allocate channel datastore for channel %s to store inheritable SIP headers"

res/res_pjsip_header_funcs.c
```diff
@@ -679,6 +796,78 @@ static int add_header(void *obj)
...
+				ast_channel_name(channel));
+			ast_channel_unlock(channel);
+			return -1;
+		}
+	} else {
+		/* Ensure it's marked as inheritable */
+		if (!chan_datastore->inheritance) {
+			chan_datastore->inheritance = DATASTORE_INHERIT_FOREVER;
+			ast_debug(3, "Upgraded channel datastore to inheritable for channel %s\n",
```
jcolp: Same here as well

res/res_pjsip_header_funcs.c
```diff
@@ -679,6 +796,78 @@ static int add_header(void *obj)
...
+				ast_channel_name(channel));
+		}
+	}
+
+	/* Add THIS header to the channel datastore */
+	headers = chan_datastore->data;
+	new_var = ast_variable_new(header_name, header_value, "");
+	if (!new_var) {
+		ast_log(LOG_ERROR, "Failed to allocate variable for channel %s\n",
```
jcolp: And here

Issue #1675, opened 2025-12-29
[bug]: res_pjsip_mwi: off-nominal endpoint ao2 reference leak in mwi_get_notify_data()


alex2grad:
### Severity

Minor

### Versions

20.17.0,22.7.0, 23.1.0

### Components/Modules

res/res_pjsip_mwi.c

### Operating Environment

Slackware 15.0

### Frequency of Occurrence

None

### Issue Description

mwi_get_notify_data() acquires an ast_sip_endpoint reference via ast_sip_subscription_get_endpoint(sub) before error checks that may return early.

If either:
  ast_sip_subscription_get_datastore() returns NULL, or
  ao2_alloc() for counter fails,
the function returns without releasing the endpoint reference, leaking an ao2 reference.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1676, opened 2025-12-29
res_pjsip_mwi: Fix off-nominal endpoint ao2 ref leak in mwi_get_notify_data

alex2grad:
Delay acquisition of the ast_sip_endpoint reference in mwi_get_notify_data()
to avoid an ao2 ref leak on early-return error paths.

Move ast_sip_subscription_get_endpoint() to just before first use so all
acquired references are properly cleaned up.

Fixes: #1675


alex2grad:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

InterLinked1 (approved):

PR #1677, opened 2025-12-30
configure.ac: use AC_PATH_TOOL for nm

nwf:
`nm` might, especially in cross-compilation scenarios, be available but prefixed with the target triple. So: use `AC_PATH_TOOL` rather than `AC_PATH_PROG` to find it. (See https://www.gnu.org/software/autoconf/manual/autoconf-2.68/html_node/Generic-Programs.html .)

Found and proposed fix tested by cross-compiling Asterisk using Nixpkgs on x86_64 targeting aarch64. :)


nwf:
#1608 for the `23` branch.

cherry-pick-to: none

Issue #1681, opened 2025-12-31
[bug]: stasis/control.c: Memory leak of hangup_time in set-timeout


gtjoseph:
### Severity

Minor

### Versions

All

### Components/Modules

stasis/control.c

### Operating Environment

All

### Frequency of Occurrence

Frequent

### Issue Description

The set_timeout() function mallocs 16 bytes for a timeval structure and adds it to the timeout_datastore for every call made via ARI.  There's no destructor on the datastore however so that memory is not freed when the datastore is.

### Relevant log output

After 10 calls:
```text
Direct leak of 160 byte(s) in 10 object(s) allocated from:
    #0 0x7fe55ec10abd in malloc (/usr/lib64/liblsan.so.0+0x10abd) (BuildId: 700f6a0ea3110348bc9646e667215b06126a81c4)
    #1 0x00000042eabc in __ast_repl_malloc /usr/src/asterisk/asterisk/main/astmm.c:1551
    #2 0x00000042ed83 in __ast_malloc /usr/src/asterisk/asterisk/main/astmm.c:1632
    #3 0x7fe55a5358fc in set_timeout stasis/control.c:1661
    #4 0x7fe55a535af1 in app_control_dial stasis/control.c:1695
    #5 0x7fe55a53124c in command_invoke stasis/command.c:104
    #6 0x7fe55a5354cd in control_dispatch_all stasis/control.c:1559
    #7 0x7fe55a529352 in stasis_app_exec /usr/src/asterisk/asterisk/res/res_stasis.c:1602
    #8 0x7fe55a08981b in app_exec /usr/src/asterisk/asterisk/apps/app_stasis.c:116
    #9 0x000000531ad1 in pbx_exec /usr/src/asterisk/asterisk/main/pbx_app.c:492
    #10 0x7fe55a055f56 in ari_channel_thread ari/resource_channels.c:1731
    #11 0x0000005e0035 in dummy_start /usr/src/asterisk/asterisk/main/utils.c:1643
    #12 0x7fe55ec0e309 in void* ThreadStartFunc<true>(void*) (/usr/lib64/liblsan.so.0+0xe309) (BuildId: 700f6a0ea3110348bc9646e667215b06126a81c4)
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1682, opened 2025-12-31
stasis/control.c: Add destructor to timeout_datastore.

gtjoseph:
The timeout_datastore was missing a destructor resulting in a leak
of 16 bytes for every outgoing ARI call.

Resolves: #1681


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1683, opened 2025-12-31
[improvement]: chan_websocket: Use channel FD polling to read data from websocket instead of dedicated thread.


gtjoseph:
### Improvement Description

chan_websocket spins up a dedicated thread for each channel to wait on, and read data from the websocket. This requires extra locking to prevent race conditions. Instead of using a thread, we can add the websocket's file descriptor to the channel which gets polled with other channel file descriptors.  This can trigger the read from the websocket and doesn't require explicit locking because the channel core takes care of it.

PR #1684, opened 2025-12-31
chan_websocket: Use the channel's ability to poll fds for the websocket read.

gtjoseph:
We now add the websocket's file descriptor to the channel's fd array and let
it poll for data availability instead if having a dedicated thread that
does the polling. This eliminates the thread and allows removal of most
explicit locking since the core channel code will lock the channel to prevent
simultaneous calls to webchan_read, webchan_hangup, etc.

While we were here, the hangup code was refactored to use ast_hangup_with_cause
instead of directly queueing an AST_CONTROL_HANGUP frame.  This allows us
to set hangup causes and generate snapshots.

For a bit of extra debugging, a table of websocket close codes was added
to http_websocket.h with an accompanying "to string" function added to
res_http_websocket.c

Resolves: #1683

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -90,6 +90,7 @@ struct websocket_pvt {
 	pthread_t outbound_read_thread;
 	size_t bytes_read;
 	size_t leftover_len;
+	char *remote_addr;
 	char *uri_params;
 	char *leftover_data;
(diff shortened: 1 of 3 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

mbradeen (approved):

PR #1685, opened 2026-01-01
main/file.c: Smooth and stabilize video playback by clamping frame de…

bhrugu-crypto:
…lay timing

Video voicemail playback could appear inconsistent, with frames sometimes playing too fast and sometimes too slow due to irregular RTP timestamps. This patch smooths playback by clamping the calculated frame delay to a safe operational range, preventing extremely small delays (causing overly fast playback) and very large delays (causing visible pauses).

A minimum delay is enforced to avoid burst-style frame playback, while a reduced maximum delay cap helps maintain visually smooth and consistent video output.

This results in more natural, predictable, and stable video playback behavior during voicemail review.

UserNote: Video voicemail playback is now smoother and more consistent, without sudden fast or slow playback behavior.

UpgradeNote: Systems experiencing uneven, fast, slow, or jittery voicemail video playback are recommended to upgrade. With this patch applied, playback timing is stabilized using controlled frame delay clamping.

### DIFF
--- a/main/file.c
+++ b/main/file.c
@@ -1080,7 +1080,7 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
                          s->owner,
                          ast_sched_add(ast_channel_sched(s->owner), delay_ms, ast_fsread_video, s)
                  );
-                s->lasttimeout = frame_ms;
+                s->lasttimeout = whennext;
                 return FSREAD_SUCCESS_NOSCHED;
 	}
 


bhrugu-crypto:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
I have updated your cherry-pick-to comment to the current allowed branches. This change will not be cherry picked to Asterisk 21, as it no longer receives bug fixes.

main/file.c
```diff
@@ -1052,7 +1052,15 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
 	}
 
 	if (whennext != s->lasttimeout) {
-		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), whennext / (ast_format_get_sample_rate(s->fmt->format) / 1000), ast_fsread_video, s));
+                
+                int delta = whennext - s->lasttimeout;
+                int delay_ms;
+
+
+                delay_ms = (delta / 90) * 12;
```
jcolp: Where did this value come from? Add a comment to explain it.

main/file.c
```diff
@@ -1052,7 +1052,15 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+
+                ast_debug(3, "Video file playback delay=%d ms (delta=%d)\n ", delay_ms,delta);
```
jcolp: This can become extremely chatty during playback. Unless absolutely necessary (I don't think it is) it should be removed. If it is needed it should include the channel name.

main/file.c
AryanM200: The value 90 corresponds to the RTP video clock rate (90 kHz), which is standard for video payloads.
delta is therefore expressed in RTP timestamp units and must be converted to milliseconds before scheduling the next frame.

The additional scaling factor (* 12) is used to correct frame pacing during playback. Without this adjustment, frames are scheduled too frequently, causing the video to play back faster than intended. This scaling aligns the scheduler timing with the effective decoded frame duration and results in stable, real-time playback.

main/file.c
AryanM200: Agreed. The debug log is not required for normal operation and will be removed to avoid excessive logging during playback.

main/file.c
bhrugu-crypto: The value 90 corresponds to the RTP video clock rate (90 kHz), which is standard for video payloads.
delta is therefore expressed in RTP timestamp units and must be converted to milliseconds before scheduling the next frame.
The additional scaling factor (* 12) is used to correct frame pacing during playback. Without this adjustment, frames are scheduled too frequently, causing the video to play back faster than intended. This scaling aligns the scheduler timing with the effective decoded frame duration and results in stable, real-time playback.

main/file.c
bhrugu-crypto: Agreed. The debug log is not required for normal operation and will be removed to avoid excessive logging during playback.

main/file.c
jcolp: A comment should be added to the code itself, not just on here.

jcolp:
There are now two reviews from different users with the same contents. As the other review is active, I am closing this one.

PR #1688, opened 2026-01-02
asterisk.c: Allow multi-byte characters on the Asterisk CLI.

### LINKED ISSUE #60: [bug]:
### Severity

Minor

### Versions

20.2.1

### Components/Modules

CLI

### Operating Environment

Stock Ubuntu 22.04.2 LTS (Jammy Jellyfish)

### Frequency of Occurrence

Constant

### Issue Description

Can't enter any of UFT-8 character in the CLI prompt - it just beeps. If I change AST_EDITOR to 'vi' mode then I have:

```
*CLI> \U+FFFF\U+FFFF\U+FFFF\U+FFFF\U+FFFF\U+FFFF\U+FFFF\U+FFFF
```

Seems like problem with libedit. The previous asterisk versions allow to use internal libedit - current one do not.

### Relevant log output

_No response_

### Asterisk Issue Guidelines

seanbright:
Versions of libedit that support Unicode expect that the EL_GETCFN (the function that does character I/O) will fill in a `wchar_t` with a character, which may be multi-byte. The built-in function that libedit provides, but does not expose with a public API, does properly handle multi-byte sequences.

Due to the design of Asterisk's console processing loop, Asterisk provides its own implementation which does not handle multi-byte characters. Changing Asterisk to use libedit's built-in function would be ideal, but would also require changing some fundamental things about console processing which could be fairly disruptive.

Instead, we bring in libedit's `read_char` implementation and modify it to suit our specific needs.

Resolves: #60

### DIFF
--- a/main/asterisk.c
+++ b/main/asterisk.c
@@ -246,6 +246,7 @@ int daemon(int, int);  /* defined in libresolv of all places */
 
 #include "../defaults.h"
 #include "channelstorage.h"
+#include "editline_compat.h"
 
 /*** DOCUMENTATION
 	<managerEvent language="en_US" name="FullyBooted">
@@ -2725,6 +2726,12 @@ static int ast_el_read_char(EditLine *editline, CHAR_T_LIBEDIT *cp)
 		}
 
 		if (!ast_opt_exec && fds[1].revents) {
+#if HAVE_LIBEDIT_IS_UNICODE
+			num_read = editline_read_char(editline, cp);
+			if (num_read < 1) {
+				break;
+			}
+#else
 			char c = '\0';
 
 			num_read = read(STDIN_FILENO, &c, 1);
@@ -2734,6 +2741,7 @@ static int ast_el_read_char(EditLine *editline, CHAR_T_LIBEDIT *cp)
 
 			*cp = CHAR_TO_LIBEDIT(c);
 
+#endif
 			return num_read;
 		}
 
@@ -3620,6 +3628,11 @@ int main(int argc, char *argv[])
 	RAII_VAR(char *, xarg, NULL, ast_free);
 	struct rlimit l;
 	static const char *getopt_settings = "BC:cde:FfG:ghIiL:M:mnpqRrs:TtU:VvWXx:";
+
+	/* Bring in locale settings from the environment. This is needed
+	   for libedit, as the LC_CTYPE category of the locale impacts the
+	   the multi-byte character functions provided by libc */
+	setlocale(LC_ALL, "");
 
 	/* Remember original args for restart */
 	if (argc > ARRAY_LEN(_argv) - 1) {
--- a/main/editline_compat.h
+++ b/main/editline_compat.h
@@ -1,0 +1,26 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2026, Sean Bright
+ *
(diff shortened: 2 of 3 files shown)


seanbright:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seanbright:
The only thing I am _slightly_ iffy on is calling `setlocale(...)` on startup. It is needed to make the multi-byte functions actually do something, but I suppose there could be implications of doing this that I am not aware of.

InterLinked1 (changes requested):
"processing lopp" -> "processing loop"

main/asterisk.c
```diff
@@ -3621,6 +3629,8 @@ int main(int argc, char *argv[])
 	struct rlimit l;
 	static const char *getopt_settings = "BC:cde:FfG:ghIiL:M:mnpqRrs:TtU:VvWXx:";
 
+	setlocale(LC_ALL, "");
```
InterLinked1, line 3635: Since this code isn't near the other code, if we're not sure about this 100%, maybe a comment about why we need to do this might help? But I guess that can also be traced through the blame.

main/editline_compat.h
```diff
@@ -0,0 +1,25 @@
...
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
+ * at the top of the source tree.
+ */
+#ifndef EDITLINE_COMPAT_PRIVATE_H
+#define EDITLINE_COMPAT_PRIVATE_H
+
+#include <histedit.h>
```
InterLinked1, line 22: histedit.h is already included in asterisk.c, does it need to be in both?

Edit: I think you did that so editline_compat.c doesn't need it, but it seems a bit bifurcated to have one in the .c and one in the .h

main/editline_compat.h
```diff
@@ -0,0 +1,25 @@
...
+ * the Asterisk project. Please do not directly contact
+ * any of the maintainers of this project for assistance;
+ * the project provides a web site, mailing lists and IRC
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
+ * at the top of the source tree.
+ */
+#ifndef EDITLINE_COMPAT_PRIVATE_H
```
InterLinked1, line 19: If you feel like it... I think it's customary in this codebase for there to be a blank line after the header comments before anything else. Same with editline_compat.c

InterLinked1 (commented):
Mostly pedantry here...

InterLinked1 (approved):

main/editline_compat.h
seanbright: `editline_compat.h` references `EditLine *` which is provided by `histedit.h` so including it here means it will compile even if someone subsequently changes include ordering elsewhere. If you feel strongly about it I can take another look but if `histedit.h` has already been included elsewhere this is essentially a nop anyway.

That being said I did add an explicit include for `histedit.h` in `editline_compat.c` for the same reasons.

InterLinked1: That makes sense about the typedef... and I don't feel strongly about this, was just wondering. Looks good to me.

jcolp (approved):
I can't think of a real reason calling setlocale would be an issue.

PR #72, opened 2023-05-10
core/ari/pjsip: Add refer mechanism

maximilianfridrich:
This change adds support for refers that are not session based. It
includes a refer implementation for the PJSIP technology which results
in out-of-dialog REFERs being sent to a PJSIP endpoint. These can be
triggered using the new ARI endpoint `/endpoints/refer`.

Resolves: #71

UserNote: There is a new ARI endpoint `/endpoints/refer` for referring
an endpoint to some URI or endpoint.

PR #1690, opened 2026-01-05
res_pjsip_messaging: Update To URI only if it is a SIP(S) URI

### LINKED ISSUE #1357: [bug]: MessageSend WARNING “not a valid SIP/SIPS URI” when using endpoint not URI
### Severity

Minor

### Versions

16,18,20

### Components/Modules

res_pjsip_messaging.so

### Operating Environment

asterisk v16 from devuan debs (in a devuan4 container on devuan5 host)
asterisk v18 from ubuntu debs (on devuan5)
asterisk v20 from source (on devuan5)

### Frequency of Occurrence

None

### Issue Description

Getting a WARNING when using MessageSend to endpoint not URI.

same => n,MessageSend(pjsip:7006,7004)

WARNING only shows up only when going trough message_context.

Regular context with same MessageSend without WARNING.

Original issue (with more info)
https://community.asterisk.org/t/messagesend-error-not-a-valid-sip-sips-uri/109594

### Relevant log output

```shell
-- Executing [7006@astsms:8] MessageSend("Message/ast_msg_queue", "pjsip:7006,7004") in new stack
[Aug  9 12:48:09] WARNING[1476]: res_pjsip.c:3338 ast_sip_update_to_uri: To address '7006@asterisk20.misc' is not a valid SIP/SIPS URI
```

### Asterisk Issue Guidelines

maximilianfridrich:
When a message is sent via ARI, the ARI endpoint only provides a To
field which is also used as destination field. This means that the To
field might not necessarily contain a SIP URI but might instead specify
an Asterisk endpoint (in MessageDestinationInfo format). This led to
many warnings even though the message was sent correctly.

The fix is to only call `ast_sip_update_to_uri` if the To field starts
with the sip: or sips: scheme.

Resolves: #1357

### DIFF
--- a/res/res_pjsip_messaging.c
+++ b/res/res_pjsip_messaging.c
@@ -959,7 +959,15 @@ static int msg_send(void *data)
 		if (ast_begins_with(msg_to, "pjsip:")) {
 			msg_to += 6;
 		}
-		ast_sip_update_to_uri(tdata, msg_to);
+		/*
+		 * Only attempt to update the To URI if it's actually a SIP/SIPS URI.
+		 * When sending via ARI, the To field is also used as destination
+		 * (MessageDestinationInfo) and therefore might not contain a SIP URI.
+		 * ast_sip_create_request still sets the correct To header.
+		 */
+		if (ast_begins_with(msg_to, "sip:") || ast_begins_with(msg_to, "sips:")) {
+			ast_sip_update_to_uri(tdata, msg_to);
+		}
 	} else {
 		/*
 		 * If there was no To in the message, it's still possible


maximilianfridrich:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (commented):
How did you determine that these are normal code paths and acceptable?

maximilianfridrich:
In `res_pjsip_messaging:msg_send` for example the function is called for every outgoing message - basically just in case the destination is a SIP URI. If it is not a SIP URI but instead identifies an endpoint, it still works perfectly fine. The code is a bit of a mess, a lot of it has my name next to it because I made these functions available in the res_pjsip API in #72. The logic has been there before though and I didn't change it in #72 to prevent regressions. See the [MessageSend syntax docs](https://docs.asterisk.org/Asterisk_22_Documentation/API_Documentation/Dialplan_Applications/MessageSend/#syntax) for how many different things can be set as destination - at your own risk, might cause headaches.

As described in #1357, this leads to a WARNING every time a MESSAGE is sent, even though the message is sent just fine.

Let me know if you would like this to be handled differently.

mbradeen:
> In `res_pjsip_messaging:msg_send` for example the function is called for every outgoing message - basically just in case the destination is a SIP URI. If it is not a SIP URI but instead identifies an endpoint, it still works perfectly fine. The code is a bit of a mess, a lot of it has my name next to it because I made these functions available in the res_pjsip API in #72. The logic has been there before though and I didn't change it in #72 to prevent regressions. See the [MessageSend syntax docs](https://docs.asterisk.org/Asterisk_22_Documentation/API_Documentation/Dialplan_Applications/MessageSend/#syntax) for how many different things can be set as destination - at your own risk, might cause headaches.
> 
> As described in #1357, this leads to a WARNING every time a MESSAGE is sent, even though the message is sent just fine.
> 
> Let me know if you would like this to be handled differently.

Hi @maximilianfridrich - 

I did some digging on the SIP URI issue and despite originally agreeing that the warning was in error, I believe the code is correct and the docs need to be adjusted.  If we look at the [docs](https://docs.asterisk.org/Latest_API/API_Documentation/Dialplan_Applications/MessageSend/#generated-version:~:text=to%20%2D%20A%20To,an%20incoming%20message.),  the doc indeed says that:

> This can be a SIP(S) URI, such as 'Alice [sip:alice@atlanta.com](mailto:sip:alice@atlanta.com)', **or a string in the format 'alice@atlanta.com'**

However, if you use the later with PJSIP that would result in an invalid sip message because the to wouldn't be a valid SIP URI (as the error indicates).  The MESSAGE is still sent, but if you look at the actual SIP packet, what is sent in the To header is the URI of the endpoint and NOT what was set, example:
`	same => n,Set(MESSAGE(to)=sip:Banjo@homepbx.org)`
results in 
`To: <sip:Banjo@homepbx.org>`
... and no warning.  However;
`	same => n,Set(MESSAGE(to)=Banjo@homepbx.org)`
results in:
`To: <sip:Banjo@127.0.0.1;ob>`
(The URI of the endpoint is used, not the string set in dialplan and the warning is displayed)

maximilianfridrich:
Thank you for the feedback. Sorry for the sloppy fix, I looked into it more thoroughly now given your feedback. The reported issue #1357 has a different root cause I realized and I commented it in the issue report - not a problem in the code I think, you're right.

However, we see the warning in our logs too when sending a message using ARI and in that case there is a different root cause. In the ARI case, the root cause is that in `res/ari/resource_endpoints.c` both `to` and `destination` are always set to the same value. However, those two don't quite have the same semantics and the fix will have to be to use these two fields correctly in the code - if possible, because [`PUT /endpoints/sendMessage`](https://docs.asterisk.org/Asterisk_22_Documentation/API_Documentation/Asterisk_REST_Interface/Endpoints_REST_API/#sendmessage) for example only has a `To` field.

Maybe we can mark this as draft until I have a better fix.

maximilianfridrich:
I've updated the PR. The fix should be quite simple I believe and is described in the PR description and commit message:

> When a message is sent via ARI, the ARI endpoint only provides a To
> field which is also used as destination field. This means that the To
> field might not necessarily contain a SIP URI but might instead specify
> an Asterisk endpoint (in MessageDestinationInfo format). This led to
> many warnings even though the message was sent correctly.
> 
> The fix is to only call ast_sip_update_to_uri if the To field starts
> with the sip: or sips: scheme.

jcolp (approved):

Issue #1691, opened 2026-01-05
[bug]: force_longest_waiting_caller stops offering calls if a call joins at the first position


serfreeman1337:
### Severity

Trivial

### Versions

20.11.1

### Components/Modules

app_queue

### Operating Environment

Debian 12

### Frequency of Occurrence

Constant

### Issue Description

I'm not sure how I missed this during my testing in #1637

But here are the steps to reproduce:
1. Have both queues 400 and 401 with the same weight and `force_longest_waiting_caller=yes`.
2. Add a call to Queue 400.
3. Add a call to Queue 401,
4. Add a call to Queue 400 at position 1 (set the last parameter to 1 for the `Queue` app).

`queue show` should look like this:
```
400 has 2 calls (max unlimited) in 'rrmemory' strategy (0s holdtime, 0s talktime), W:0, C:0, A:15, SL:0.0%, SL2:33.3% within 60s
   Members: 
      Operator 1600 (Local/1600@from-queue/n from hint:1600@ext-local) (ringinuse disabled) (Unavailable) has taken no calls yet (login was 968 secs ago)
   Callers: 
      1. Local/400@from-internal-00000036;2 (wait: 0:04, prio: 0)
      2. Local/400@from-internal-00000034;2 (wait: 0:09, prio: 0)

401 has 1 calls (max unlimited) in 'rrmemory' strategy (0s holdtime, 0s talktime), W:0, C:0, A:14, SL:0.0%, SL2:57.1% within 60s
   Members: 
      Operator 1600 (Local/1600@from-queue/n from hint:1600@ext-local) (ringinuse disabled) (Unavailable) has taken no calls yet (login was 968 secs ago)
   Callers: 
      1. Local/401@from-internal-00000035;2 (wait: 0:06, prio: 0)
```

Calls are no longer offered to operators:
* The call `Local/400@from-internal-00000036;2` in the 400 queue gives up its turn to the call in the 401 queue because that call has the longest waiting time.
* The call `Local/401@from-internal-00000035;2` in the 401 queue gives up its turn to the 400 queue because the call at position 2 in the 400 queue has the longest waiting time.

Problem code:
https://github.com/asterisk/asterisk/blob/0bfb7c7f11060ecb7248cc146ba9c3137b91ff76/apps/app_queue.c#L4761

[...]

PR #1693, opened 2026-01-05
app_queue: Only compare calls at 1st position across queues when forcing longest waiting caller.

serfreeman1337:
This prevents a situation where a call joining at 1st position to a queue with calls leads to a state where no callers are considered the longest waiting, causing queues to stop offering calls.

Resolves: #1691


serfreeman1337:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

PR #1695, opened 2026-01-05
cli.c: Allow 'channel request hangup' to accept patterns.

seanbright:
This extends 'channel request hangup' to accept multiple channel names, a POSIX Extended Regular Expression, a glob-like pattern, or a combination of all of them.

UserNote: The 'channel request hangup' CLI command now accepts multiple channel names, POSIX Extended Regular Expressions, glob-like patterns, or a combination of all of them. See the CLI command 'core show help channel request hangup' for full details.


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

InterLinked1:
> cherry-pick-to: none

I'm not seeing how it would break current functionality, how come we can't cherry pick to existing versions?

seanbright:
> > cherry-pick-to: none
> 
> I'm not seeing how it would break current functionality, how come we can't cherry pick to existing versions?

Just being conservative. I'll let others decide.

main/cli.c
```diff
@@ -1227,57 +1230,136 @@ static char *handle_chanlist(struct ast_cli_entry *e, int cmd, struct ast_cli_ar
...
+
+static int channel_hangup_matches(
+	struct ast_cli_args *a,
+	int (*matchfn)(const char *, const void *),
+	const void *data)
+{
+	struct ast_channel_iterator *iter = NULL;
+	struct ast_channel *c;
+	int matched = 0;
+	if (!(iter = ast_channel_iterator_all_new())) {
```
gtjoseph: Instead of the live channels, would you consider iterating over channel snapshots returned by `ast_channel_cache_by_name()`?  When you get a match on name, you'd then do a direct `ast_channel_get_by_name()`.  This way you wouldn't have to lock every channel and you could just let softhangup lock the channel.

If you want to keep using the live channels, you could use  `ast_channel_callback()` instead of doing the iteration yourself.

gtjoseph (dismissed):

main/cli.c
```diff
@@ -1227,57 +1230,138 @@ static char *handle_chanlist(struct ast_cli_entry *e, int cmd, struct ast_cli_ar
...
+			char *pattern = ast_strdup(a->argv[x]);
+			regex_t re;
+			int res;
+
+			if (!pattern) {
+				return CLI_FAILURE;
+			}
+
+			pattern[strlen(pattern) - 1] = 0;
+			res = regcomp(&re, &pattern[1], REG_EXTENDED | REG_NOSUB);
```
gtjoseph: You probably want REG_ICASE here as well since the existing match is case-insensitive.

main/cli.c
```diff
@@ -1227,57 +1230,138 @@ static char *handle_chanlist(struct ast_cli_entry *e, int cmd, struct ast_cli_ar
...
 }
 
+static int channel_match_by_regex(const char *channel_name, const void *data)
+{
+	return regexec(data, channel_name, 0, NULL, 0);
+}
+
+static int channel_match_by_glob(const char *channel_name, const void *data)
+{
+	return fnmatch(data, channel_name, FNM_NOESCAPE | FNM_PATHNAME);
```
gtjoseph: ~~You probably want FNM_IGNORECASE here as well since the existing match is case-insensitive.~~

Oops, FNM_IGNORECASE is a posix 2024 extension.  I think you're going to have to convert both name and data to lower case yourself

How about FNM_EXTMATCH?  I _think_ it's portable enough.

~~And...do you really want FNM_PATHNAME since it add special meaning to `/`?~~

gtjoseph (changes requested):
Are you going to update ast_complete_channel() as well?

main/cli.c
```diff
@@ -1227,57 +1230,138 @@ static char *handle_chanlist(struct ast_cli_entry *e, int cmd, struct ast_cli_ar
 #undef VERBOSE_FORMAT_STRING2
 }
 
+static int channel_match_by_regex(const char *channel_name, const void *data)
+{
+	return regexec(data, channel_name, 0, NULL, 0);
```
gtjoseph, line 1235: How about REG_NOSUB here?  I think it saves a bit of processing since we only care about match/nomatch.

gtjoseph (changes requested):

seanbright:
> Are you going to update ast_complete_channel() as well?

You mean `ast_complete_channels()?` Update it how?

main/cli.c
seanbright: That is a `regcomp()` flag, and I'm already using it.

gtjoseph:
> > Are you going to update ast_complete_channel() as well?
> 
> You mean `ast_complete_channels()?` Update it how?

So that if you enter a regex or glob and press Tab, it expands to all the matching channels.

main/cli.c
gtjoseph: I just saw the help that explains FNM_PATHNAME so that's fine.

main/cli.c
gtjoseph: yeah sorry.  i saw it after.

seanbright:
> > > Are you going to update ast_complete_channel() as well?
> > 
> > 
> > You mean `ast_complete_channels()?` Update it how?
> 
> So that if you enter a regex or glob and press Tab, it expands to all the matching channels.

Does your shell do that? I have never seen that happen in dash or bash.

gtjoseph:
> > > > Are you going to update ast_complete_channel() as well?
> > > 
> > > 
> > > You mean `ast_complete_channels()?` Update it how?
> > 
> > 
> > So that if you enter a regex or glob and press Tab, it expands to all the matching channels.
> 
> Does your shell do that? I have never seen that happen in dash or bash.

No but the CLI does for non-regex or non-glob partial entries.

main/cli.c
seanbright: > How about FNM_EXTMATCH? I _think_ it's portable enough.

I assume you mean `FNM_CASEFOLD`? I've added it where it is available.

> And...do you really want FNM_PATHNAME since it add special meaning to `/`?

Yes, I think treating the channel tech as a "directory" is a nice metaphor.

seanbright:
> > > > > Are you going to update ast_complete_channel() as well?
> > > ...
> > Does your shell do that? I have never seen that happen in dash or bash.
> 
> No but the CLI does for non-regex or non-glob partial entries.

No, I am not going to make that change.

main/cli.c
```diff
@@ -1227,57 +1230,162 @@ static char *handle_chanlist(struct ast_cli_entry *e, int cmd, struct ast_cli_ar
...
-			"       Request that a channel be hung up. The hangup takes effect\n"
-			"       the next time the driver reads or writes from the channel.\n"
+			"Usage: channel request hangup <all>|<channel> [<channel> ...]\n"
+			"       Request that one or more channels be hung up. The hangup takes\n"
+			"       effect the next time the driver reads or writes from the channel.\n\n"
 			"       If 'all' is specified instead of a channel name, all channels\n"
-			"       will see the hangup request.\n";
+			"       will see the hangup request.\n\n"
+			"       A POSIX Extended Regular Expression can be provided in the place\n"
+			"       of a fixed channel name by wrapped it in forward slashes\n"
```
gtjoseph: "wrapping" instead of "wrapped"

gtjoseph (changes requested):

gtjoseph:
I'm fine with this being cherry-picked BTW.

gtjoseph (approved):

jcolp (approved):

Issue #1689, opened 2026-01-04
[bug]: Remote media cache race under concurrent ARI playback causes ENOENT + spurious playback failures


pekkaar:
### Severity

Trivial

### Versions

20.17.0, 20.16.0, 20.15.2

### Components/Modules

Core/Media Cache, res_http_media_cache, res_curl, ARI, Stasis Playback

### Operating Environment

Ubuntu 24 LTS, Azure VM

### Frequency of Occurrence

Frequent

### Issue Description

When playing remote media via ARI (sound:https://…) under load, Asterisk frequently (in my use case: almost always) logs “Unable to open … No such file or directory” and "Playback failed" even though the media exists and is generally playable. This appears to be a race condition between concurrent refreshes of the HTTP media cache and in‑flight open attempts.

The suspected root cause is in main/media_cache.c `ast_media_cache_retrieve()`, which can replace/delete cached files while other threads are opening the same URI. Under concurrency, this creates a window where the cache entry points to a file that has just been removed or not yet readable.

Investigation revealed additional problems with missing UTF-8 validation in `ast_streamfile()` when building the FILE_STREAM_BEGIN JSON event, as well as unnecesary warnings in res_stasis_playback.c `playback_final_update()` for "hangup while playing" scenarios.

My proposed PR is to make cache refresh safe for concurrent opens, validate UTF‑8 in `ast_streamfile()` when building the JSON event to avoid errors under load, and also downgrade “Playback failed” log entry when the channel is already hung up.

### Relevant log output

```shell
WARNING[...] file.c: ast_streamfile: Unable to open https://... (format (alaw)): No such file or directory
WARNING[...] res_stasis_playback.c: playback_final_update: Playback failed for sound:https://...
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

pekkaar:
@jcolp I had high hopes of getting completely rid of those "Unable to open" / "Playback failed" messages, but the best I could get was a compromise with a somewhat decreased figure but no stale items vs completely eliminating but having to introduce separate cleanup process for on-disk stales (that I'd avoid as long as possible)... after a focused run it seems the actual gain of lowered messages is simply not worth the level of incursion into media_cache.c. Sorry for the noise, had promising test results initially.

BTW, those missing UTF-8 validations are still worth the fix, maybe next time.

seanbright:
Let’s keep it open in case someone else wants to investigate

[...]

PR #1697, opened 2026-01-05
media_cache: harden remote media cache refresh under concurrency

pekkaar:
When playing remote media via ARI, concurrent refreshes could delete or
replace cached files while other channels were opening them, leading to
ENOENT in ast_streamfile() and spurious playback failures. Rework the
cache update flow to keep readable files on disk until a new file is
confirmed readable and safely linked, and avoid deleting shared paths.

Also sanitize FILE_STREAM_BEGIN JSON payloads to avoid invalid UTF-8
errors under load, and downgrade playback failure log level when the
channel is already hung up.

Fixes: #1689

### DIFF
--- a/main/file.c
+++ b/main/file.c
@@ -1382,9 +1382,9 @@ int ast_streamfile(struct ast_channel *chan, const char *filename,
 	cel_event = ast_json_pack("{ s: s, s: {s: s, s: s, s: s}}",
 		"event", "FILE_STREAM_BEGIN",
 		"extra",
-			"sound", tmp_filename,
+			"sound", AST_JSON_UTF8_VALIDATE(tmp_filename),
 			"format", ast_format_get_name(ast_channel_writeformat(chan)),
-			"language", preflang ? preflang : "default"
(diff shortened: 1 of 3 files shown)


pekkaar:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seanbright:
> Also sanitize FILE_STREAM_BEGIN JSON payloads to avoid invalid UTF-8 errors under load

Can you explain how load is impacting this?

pekkaar:
> > Also sanitize FILE_STREAM_BEGIN JSON payloads to avoid invalid UTF-8 errors under load
> 
> Can you explain how load is impacting this?

With the improved retry logic in my load tests, the number of ast_streamfile() calls increased, exposing the latent non‑UTF‑8 input into that JSON payload.

jcolp:
how does that cause non-UTF8 to appear?

pekkaar:
> how does that cause non-UTF8 to appear?

It doesn't cause it, that non-UTF8 was there before. The changes I made in media_cache.c appear to increase that call path and revealed this.

PR #1698, opened 2026-01-06
main/file.c: Video Voicemail Playback Fix

AryanM200:
main/file.c: Video Voicemail Playback Fix

This patch fixes uneven video playback in voicemails, where frames could appear too fast or too slow due to irregular RTP timestamps. It ensures smoother playback by clamping frame delays within a controlled range:

Minimum delay avoids rapid, burst-style frame playback.

Maximum delay prevents long pauses, keeping video timing smooth.

The result is more stable and consistent video playback during voicemail review.

User Note: Video voicemails now play naturally, without sudden speed changes.

Upgrade Note: Systems experiencing jittery or uneven voicemail video playback should apply this patch to stabilize playback timing.

### DIFF
--- a/main/file.c
+++ b/main/file.c
@@ -1055,7 +1055,7 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
                 int delta = whennext - s->lasttimeout;
                 int delay_ms;
 
-                /* Video RTP clock rate 90000Hz */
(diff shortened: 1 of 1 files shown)


AryanM200:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

PR #1699, opened 2026-01-06
file.c: asterisk-voicemail-video-playback-fix

AryanM200:
Convert RTP timestamp deltas to millisecond-based delays when scheduling
video frame playback.

This change updates the video playback scheduling logic to translate RTP
timestamp deltas into millisecond-based delays before frames are scheduled
for playback. Using millisecond-based timing ensures more consistent and
accurate pacing of video frames across different playback scenarios.

This behavior applies to video playback paths. The updated scheduling logic
helps prevent timing drift and improves overall video playback reliability
regardless of the source or context in which video frames are played.

### DIFF
--- a/funcs/func_talkdetect.c
+++ b/funcs/func_talkdetect.c
@@ -115,7 +115,6 @@
 			<example title="Enable and set talk threshold to 128">
 			same => n,Set(TALK_DETECT(set)=,128)
 			</example>
-			<para>This function will set the following variables:</para>
 			<note>
 				<para>The TALK_DETECT function uses an audiohook to inspect the
 				voice media frames on a channel. Other functions, such as JITTERBUFFER,


AryanM200:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

main/file.c
```diff
@@ -1033,7 +1033,7 @@ static int ast_fsread_video(const void *data);
 static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
 {
 	int whennext = 0;
-
+        static int scale_factor = 0;
```
jcolp: This doesn't work. This variable will be used by all streams. You're treating it as a per-stream state variable.

main/file.c
```diff
@@ -1052,7 +1052,35 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+                if (scale_factor == 0 && delta_ms > 0) {
+                     /* Assuming target frame interval ~60 ms  as standard */
+                      scale_factor = 60.0 / delta_ms;
+                }
+
+                /* Apply scale factor to all subsequent frames */
+                delay_ms = delta_ms * scale_factor;
+
+                /* Safety checks */
+                if (delay_ms < 20) delay_ms = 20;  /* avoid slow */
```
jcolp: This doesn't conform to our coding standards.

main/file.c
```diff
@@ -1052,7 +1052,35 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
 	}
 
 	if (whennext != s->lasttimeout) {
-		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), whennext / (ast_format_get_sample_rate(s->fmt->format) / 1000), ast_fsread_video, s));
+                int delta = whennext - s->lasttimeout;
+                int delay_ms;
+
+                /* Video RTP clock rate */
```
jcolp: I would suggest a comment about what the over all goal of this code actually is.

main/file.c
```diff
@@ -1052,7 +1052,35 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
 	}
 
 	if (whennext != s->lasttimeout) {
-		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), whennext / (ast_format_get_sample_rate(s->fmt->format) / 1000), ast_fsread_video, s));
+                int delta = whennext - s->lasttimeout;
```
jcolp: Is it possible for this to go negative?

main/file.c
AryanM200: Removed the static variable and replaced it with a per-instance int variable.

main/file.c
AryanM200: Updated the code to follow standard coding format.

main/file.c
AryanM200: We have added comments throughout the code to explain the goal.

main/file.c
AryanM200: yes , might be

AryanM200:
> The commit message also does not follow our format, or the special tags.
> 
> I'd also ask - was AI used to create this?

This was not AI-generated, it was developed by our team. The changes were made to fix with Asterisk formatting standards.

main/file.c
```diff
@@ -1052,7 +1051,34 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+                /*
+                * Normalize RTP-based video frame timing by scaling and clamping computed
+                * delays to ensure smooth playback despite irregular RTP timestamps.
+                */
+
+                int rtp_clock = 90000;
+
+                int delta_ms = (delta * 1000) / rtp_clock;
+
+                if (scale_factor == 0 && delta_ms > 0) {
```
jcolp: This will now always be 0, so what is the point of the check for 0?

main/file.c
```diff
@@ -1052,7 +1051,34 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
 	if (whennext != s->lasttimeout) {
-		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), whennext / (ast_format_get_sample_rate(s->fmt->format) / 1000), ast_fsread_video, s));
+		        int scale_factor = 0;
+                int delta = whennext - s->lasttimeout;
+                int delay_ms;
+
+                /*
+                * Normalize RTP-based video frame timing by scaling and clamping computed
+                * delays to ensure smooth playback despite irregular RTP timestamps.
+                */
```
jcolp: This comment needs to be expanded to explain the logic and why the values themselves were chosen.

main/file.c
AryanM200: We set it to zero on purpose so it starts fresh each time this code runs. The value is calculated again based on the current timing instead of reusing anything from a previous run. The check is there to make sure the value is only set when we have a valid timing difference.

jcolp: It'll always start 0 though, so why check if it is 0? It can never be non-zero there.

AryanM200: This check makes sure the system waits only when it actually needs to. If no adjustment is required and the calculated wait time is positive, the delay is applied. This avoids unnecessary waiting and prevents issues caused by invalid or negative timing values, helping everything run smoothly.

jcolp: scale_factor will always be 0 in this if statement, so checking if it is 0 is pointless

AryanM200: This check is kept as a safety step to keep video playback smooth. Even though scale_factor is 0 right now, the condition shows the purpose clearly and helps protect the code if changes are made in the future.

main/file.c
jcolp: It should be handled, then

main/file.c
```diff
@@ -1052,9 +1051,52 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
 	}
 
 	if (whennext != s->lasttimeout) {
-		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), whennext / (ast_format_get_sample_rate(s->fmt->format) / 1000), ast_fsread_video, s));
-		s->lasttimeout = whennext;
-		return FSREAD_SUCCESS_NOSCHED;
+                int scale_factor = 0;
+                int delta = whennext - s->lasttimeout;
+                int delay_ms;
```
jcolp: This is not following the indenting of everything.

main/file.c
```diff
@@ -1052,9 +1051,52 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+
+               /*
+                * Compute and schedule the delay for the next video frame in an RTP stream.
+                *
+                * `delta` calculates the difference between the expected next frame time (`whennext`)
+                *    and the last frame timestamp (`s->lasttimeout`).
+                *
+                * `delta_ms` converts this difference from RTP clock units (90 kHz) to milliseconds.
+                *
+                * `delay_ms` is then scaled by `scale_factor`. Currently `scale_factor` is 0,
```
jcolp: So you put in a scale factor and code that will not be used... why? What prompted it?

jcolp (changes requested):

main/file.c
```diff
@@ -1052,9 +1051,40 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+
+        delta_ms = (delta * 1000) / rtp_clock;
+
+        if (delta_ms < 60) {
+            delay_ms = 60;
+        } else {
+            delay_ms = 70;
+        }
+
+        ast_channel_vstreamid_set(s->owner,ast_sched_add(ast_channel_sched(s->owner),delay_ms,ast_fsread_video,s));
```
jcolp: Please use the previous spacing for arguments, I also don't have the ability to easily test this so if anyone else does that would be appreciated.

jcolp:
The indentation is messed up again, there's still two commits, and the commit message has not been fixed.

AryanM200:
@jcolp This PR has failing checks. Are there any issues in the code that need to be addressed?

jcolp:
The latest failure is because your change hasn't been rebased. It can be ignored.

AryanM200:
@jcolp Are there any issues with the code or commit messages? Is this patch still under review?

jcolp:
You still have two commits, there should be only one.

AryanM200:
@jcolp The commits have been updated.

jcolp:
Github says and shows you have 2 commits:

image

An older one, and a newer one. These should be squashed down into a single commit.

AryanM200:
@jcolp This PR was unexpectedly closed, so we have created a new PR. Please find it here:
https://github.com/asterisk/asterisk/pull/1753/commits

Issue #1700, opened 2026-01-06
[improvement]: Improve sorcery cache populate


mbradeen:
### Improvement Description

Reduce time locked when performing manual (AMI or CLI) "sorcery memory cache populate" commands to minimize the chances of other operations blocking on cache access.

This can be done by adding a specific lock for protecting population so it only occurs in one thread at a time and perform the following steps:

1. Lock repopulate lock
2. Retrieve all the backend objects
3. Turn the backend objects into sorcery_memory_cached_object objects, store in vector
4. Lock cache
5. Remove all existing objects from cache using remove_all_from_cache
6. Iterate through vector adding objects to cache using add_to_cache
7. Unlock cache
8. Cleanup any objects created for this task
9. Unlock repopulate lock

PR #1701, opened 2026-01-06
res_sorcery_memory_cache: Reduce cache lock time for sorcery memory cache populate command

mbradeen:
Reduce cache lock time for AMI and CLI sorcery memory cache populate
commands by adding a new populate_lock to the sorcery_memory_cache
struct which is locked separately from the existing cache lock so that
the cache lock can be maintained for a reduced time, locking only when
the cache objects are removed and re-populated.

Resolves: #1700

UserNote: The AMI command sorcery memory cache populate will now
return an error if there is an internal error performing the populate.
The CLI command will display an error in this case as well.

### DIFF
--- a/res/res_sorcery_memory_cache.c
+++ b/res/res_sorcery_memory_cache.c
@@ -164,6 +164,8 @@ struct sorcery_memory_cache {
 	const struct ast_sorcery *sorcery;
 	/*! \brief The type of object we are caching */
 	char *object_type;
+	/*! \brief Lock to serialize full cache population operations */
+	ast_mutex_t populate_lock;
 	/*! TRUE if trying to stop the oldest object expiration scheduler item. */
 	unsigned int del_expire:1;
 #ifdef TEST_FRAMEWORK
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

res/res_sorcery_memory_cache.c
```diff
@@ -1081,6 +1088,108 @@ static void memory_cache_populate(const struct ast_sorcery *sorcery, const char
 	ao2_ref(backend_objects, -1);
 }
 
+/*!
+ * \internal
+ * \brief Populate the cache with all objects from the backend (CLI/AMI version)
+ *
+ * This version is optimized for CLI and AMI calls. It performs all object allocation
+ * and conversion without holding locks, then quickly swaps everything into the cache.
+ * This minimizes lock contention and prevents blocking other operations.
```
jcolp: reduces blocking other operations.

Issue #1703, opened 2026-01-07
[bug]: res_pjsip_pubsub: ao2 reference leak of subscription tree in ast_sip_subscription


alex2grad:
### Severity

Major

### Versions

 20.17.0,  22.7.0,  23.1.0

### Components/Modules

res/res_pjsip_pubsub.c

### Operating Environment

Slackware 15.0

### Frequency of Occurrence

Constant

### Issue Description

ast_sip_subscription leaks an ao2 reference to its associated sip_subscription_tree.

In allocate_subscription(), the subscription takes a reference to the tree using ao2_bump(tree), but this reference was not consistently released during subscription destruction. As a result, destroyed subscriptions leave their subscription trees referenced, causing ao2 object leaks over time.

Some error paths attempted to manually clean up the tree, leading to inconsistent ownership handling and increased risk of refcount errors.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1704, opened 2026-01-07
res_pjsip_pubsub: Fix ao2 reference leak of subscription tree in ast_sip_subscription

alex2grad:
allocate_subscription() increments the ao2 reference count of the subscription tree,
but the reference was not consistently released during subscription destruction,
resulting in leaked sip_subscription_tree objects.

This patch makes destroy_subscription() responsible for releasing sub->tree,
removes ad-hoc cleanup in error paths,
and guards tree cleanup to ensure refcount symmetry and correct ownership.

Fixes: #1703

### DIFF
--- a/res/res_pjsip_pubsub.c
+++ b/res/res_pjsip_pubsub.c
@@ -1446,6 +1446,12 @@ static void destroy_subscription(struct ast_sip_subscription *sub)
 	ao2_cleanup(sub->datastores);
 	ast_json_unref(sub->persistence_data);
 	ast_free(sub->display_name);
+	if (sub->tree) {
+		/* Clear tree before cleanup to avoid re-entrant destruction */
+		struct sip_subscription_tree *tree = sub->tree;
+		sub->tree=NULL;
+		ao2_cleanup(tree);
+	}
 	ast_free(sub);
 }
 
@@ -1561,10 +1567,7 @@ static struct ast_sip_subscription *create_virtual_subscriptions(const struct as
 		if (AST_VECTOR_APPEND(&sub->children, child)) {
 			ast_debug(1, "Child subscription to resource %s could not be appended\n",
 					child_node->resource);
-			destroy_subscription(child);
-			/* Have to release tree here too because a ref was added
-			 * to child that destroy_subscription() doesn't release. */
-			ao2_cleanup(tree);
+			destroy_subscriptions(child);
 		}
 	}
 
@@ -1637,7 +1640,12 @@ void ast_sip_subscription_destroy(struct ast_sip_subscription *sub)
 {
 	ast_debug(3, "Removing subscription %p '%s->%s' reference to subscription tree %p\n",
 		sub, ast_sorcery_object_get_id(sub->tree->endpoint), sub->resource, sub->tree);
-	ao2_cleanup(sub->tree);
+	if (sub->tree) {
+		/* Clear tree before cleanup to avoid re-entrant destruction */
+		struct sip_subscription_tree *tree = sub->tree;
+		sub->tree = NULL;
+		ao2_cleanup(tree);
+	}
 }
 
 static void subscription_setup_dialog(struct sip_subscription_tree *sub_tree, pjsip_dialog *dlg)


alex2grad:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp:
It seems like this change causes a FRACK, resulting in the sub_term_and_removal test failure.

jcolp:
What were the cases/situations where the leak occurred?

alex2grad:
> What were the cases/situations where the leak occurred?

```
ERROR[25644] res_pjsip_pubsub.c: Excessive refcount 100000 reached on ao2 object 0x7f9d1f4614c0
ERROR[25644] res_pjsip_pubsub.c: FRACK!, Failed assertion Excessive refcount 100000 reached on ao2 object 0x7f9d1f4614c0 (0)
ERROR[25644] : Got 22 backtrace records
# 0: /usr/sbin/asterisk(__ast_assert_failed+0x68) [0x5aff98]
# 1: /usr/sbin/asterisk(__ao2_ref+0x2b7) [0x467597]
# 2: /usr/lib64/asterisk/modules/res_pjsip_pubsub.so(+0xa0a3) [0x7f9d23f140a3]
# 3: /usr/lib64/asterisk/modules/res_pjsip_pubsub.so(+0xa167) [0x7f9d23f14167]
# 4: /usr/lib64/asterisk/modules/res_pjsip_pubsub.so(+0xa1ff) [0x7f9d23f141ff]
# 5: /usr/lib64/asterisk/modules/res_pjsip_pubsub.so(+0x127d4) [0x7f9d23f1c7d4]
# 6: /usr/lib64/libasteriskpj.so.2(+0x8a2f7) [0x7f9d7568a2f7]
# 7: /usr/lib64/libasteriskpj.so.2(pjsip_dlg_on_tsx_state+0x5d) [0x7f9d756c33ad]
# 8: /usr/lib64/libasteriskpj.so.2(+0xbaf11) [0x7f9d756baf11]
# 9: /usr/lib64/libasteriskpj.so.2(+0xbbb45) [0x7f9d756bbb45]
#10: /usr/lib64/libasteriskpj.so.2(pjsip_tsx_recv_msg+0x4e) [0x7f9d756bfa9e]
#11: /usr/lib64/libasteriskpj.so.2(pjsip_dlg_on_rx_request+0xf5) [0x7f9d756c2f45]
#12: /usr/lib64/libasteriskpj.so.2(+0xc4a95) [0x7f9d756c4a95]
#13: /usr/lib64/libasteriskpj.so.2(pjsip_endpt_process_rx_data+0xf7) [0x7f9d756a1d77]
#14: /usr/lib64/asterisk/modules/res_pjsip.so(+0x37a4f) [0x7f9d70046a4f]
#15: /usr/sbin/asterisk(ast_taskprocessor_execute+0xd4) [0x59c4c4]
#16: /usr/sbin/asterisk() [0x5a33c8]
#17: /usr/sbin/asterisk(ast_taskprocessor_execute+0xd4) [0x59c4c4]
#18: /usr/sbin/asterisk() [0x5a3d9f]
#19: /usr/sbin/asterisk() [0x5ab89d]
#20: /lib64/libpthread.so.0(+0x7e45) [0x7f9d7585fe45]
#21: /lib64/libc.so.6(clone+0x3f) [0x7f9d7492c4af]
```


The leak occurred during RLS re-subscriptions, on SUBSCRIBE refreshes handled by pubsub_on_rx_refresh().

On every refresh:
1. pubsub_on_rx_refresh() calls create_virtual_subscriptions() to build a new subscription tree.
2. Each virtual ast_sip_subscription created by allocate_subscription() takes a reference to the tree:
    `sub->tree = ao2_bump(tree);`
3. Asterisk then compares the new tree with the existing one:
   If identical then destroy_subscriptions(new_root)
   If different then destroy_subscriptions(old_root)

This happens repeatedly for long-lived RLS dialogs.

destroy_subscriptions() ultimately destroys each ast_sip_subscription, but destroy_subscription() did not release sub->tree.
As a result:
- Every refresh added new ao2 references to the same sip_subscription_tree
- Those references were never released
- Over time, the tree’s refcount grew until hitting:
  `Excessive refcount 100000 reached on ao2 object`

The backtrace confirms this:
- Refcount growth happens at `allocate_subscription()`
`(sub->tree = ao2_bump(tree);)`
- Triggered repeatedly via:
`pubsub_on_rx_refresh() → create_virtual_subscriptions()`

This behavior was introduced by my commit: 
https://github.com/asterisk/asterisk/commit/edce85312389c8cd88428ff52eb0cd635e418308

That change increased usage of create_virtual_subscriptions() during refresh handling. At the time, I assumed destroy_subscriptions() fully cleaned up all subscription-owned resources. That assumption was incorrect: it destroyed the subscription objects, but did not release the ao2 reference taken on sub->tree.

So this is not a new refcount bug in allocate_subscription(), but a missing cleanup in the subscription destruction path, exposed by the increased refresh activity introduced in that commit.

res/res_pjsip_pubsub.c
```diff
@@ -1446,6 +1446,12 @@ static void destroy_subscription(struct ast_sip_subscription *sub)
 	ao2_cleanup(sub->datastores);
 	ast_json_unref(sub->persistence_data);
 	ast_free(sub->display_name);
+	if (sub->tree) {
+		/* Clear tree before cleanup to avoid re-entrant destruction */
+		struct sip_subscription_tree *tree = sub->tree;
+		sub->tree=NULL;
+		ao2_cleanup(tree);
+	}
```
gtjoseph, line 1454: Maybe I'm just being dense (it's early for me) but since ao2_cleanup is NULL tolerant, can't you just do..
```
ao2_cleanup(sub->tree);
sub->tree = NULL;
```

res/res_pjsip_pubsub.c
```diff
@@ -1637,7 +1640,12 @@ void ast_sip_subscription_destroy(struct ast_sip_subscription *sub)
 {
 	ast_debug(3, "Removing subscription %p '%s->%s' reference to subscription tree %p\n",
 		sub, ast_sorcery_object_get_id(sub->tree->endpoint), sub->resource, sub->tree);
-	ao2_cleanup(sub->tree);
+	if (sub->tree) {
+		/* Clear tree before cleanup to avoid re-entrant destruction */
+		struct sip_subscription_tree *tree = sub->tree;
+		sub->tree = NULL;
+		ao2_cleanup(tree);
+	}
```
gtjoseph, line 1648: Same here.  Just add the `sub->tree = NULL;` after the ao2_cleanup()?

gtjoseph (changes requested):

res/res_pjsip_pubsub.c
alex2grad: > Maybe I'm just being dense (it's early for me) but since ao2_cleanup is NULL tolerant, can't you just do..
> 
> ```
> ao2_cleanup(sub->tree);
> sub->tree = NULL;
> ```

The NULL tolerance itself isn’t the issue.
The problem is destruction ordering and re-entrancy, not whether ao2_cleanup() accepts NULL.

If we do:
```
ao2_cleanup(sub->tree);
sub->tree = NULL;
```
then ao2_cleanup() may drop the last reference to the tree and immediately trigger tree destruction. Tree destruction can cascade into subscription destructions, which can re-enter destroy_subscription() on the same sub instance.

At that point sub->tree is still non-NULL, so the re-entrant destroy path will call ao2_cleanup(sub->tree) again, resulting in a double release / refcount underflow / use-after-free.

By doing:
```
struct sip_subscription_tree *tree = sub->tree;
sub->tree = NULL;
ao2_cleanup(tree);
```
we break the ownership link before releasing the reference. This makes the destruction path re-entrancy-safe and ensures that any recursive cleanup sees a NULL tree pointer and does not attempt to release it again.

So the ordering is intentional to prevent re-entrant double cleanup, not to handle NULL input.

res/res_pjsip_pubsub.c
alex2grad: > Same here. Just add the `sub->tree = NULL;` after the ao2_cleanup()?

Same here. See above comment.

res/res_pjsip_pubsub.c
alex2grad: If you'd like, I can add a short code comment 
/* Clear tree before cleanup to avoid re-entrant destruction and double ao2 release */

gtjoseph: Ah, that makes sense now.

Issue #1707, opened 2026-01-08
[bug]: chan_iax2: Crash when processing video frames with negative length


InterLinked1:
### Severity

Critical

### Versions

Git

### Components/Modules

chan_iax2

### Operating Environment

N/A

### Frequency of Occurrence

Frequent

### Issue Description

When processing encrypted video frames (`AST_FRAME_VIDEO`), crashes sometimes occur due to frame length calculations that turn negative.

*Usually*, from the crashes that I've seen, this occurs when encryption is enabled and video is enabled on the calling side (with the crash occurring on the called side, at least). It may be possible in other cases as well.

Due to this bug, it is possible for an incoming call from a remote IAX2 system to crash the local Asterisk process, although I haven't traced the exact circumstances that lead to this.

In some parts of IAX2, there are checks for negative length; however, one code path is missing it, and when it encounters a negative length, this leads to memory corruption and subsequently a crash following memory operations using the invalid length calculation.

### Relevant log output

```shell
[2025-12-31 21:49:19.704] WARNING[1819857]: chan_iax2.c:10454 socket_process_helper: Packet Decrypt Failed!
[2025-12-31 21:49:19.784] ERROR[1819854]: iax2/parser.c:1224 iax_frame_wrap: Losing frame data because destination buffer size '4096' bytes not big enough for '-2' bytes in the frame
[2025-12-31 21:49:19.784] ERROR[1819854]: iax2/parser.c:1224 iax_frame_wrap: Losing frame data because destination buffer size '4096' bytes not big enough for '-2' bytes in the frame

#5  0x00007ff54c02686a in malloc_printerr (str=str@entry=0x7ff54c12f158 "double free or corruption (!prev)") at ./malloc/malloc.c:5660
#6  0x00007ff54c0288dc in _int_free (av=0x7ff53c000030, p=0x7ff53cada5a0, have_lock=<optimized out>, have_lock@entry=0) at ./malloc/malloc.c:4587
        size = 224
        fb = <optimized out>
        nextchunk = 0x7ff53cada680
        nextsize = <optimized out>
        nextinuse = <optimized out>
        prevsize = <optimized out>
        bck = <optimized out>
        fwd = <optimized out>
        __PRETTY_FUNCTION__ = "_int_free"
#7  0x00007ff54c02af4f in __GI___libc_free (mem=<optimized out>) at ./malloc/malloc.c:3385
        ar_ptr = <optimized out>
        p = <optimized out>
        err = 0
[...]

PR #1708, opened 2026-01-08
chan_iax2: Fix crash due to negative length frame lengths.

InterLinked1:
chan_iax2 has several code paths where a frame's data length is calculated by subtraction. On some paths, there is a check for negative length. One of these paths is missing this check, and on this path, it is possible for the result to be negative, leading to a crash as a result of memory operations using the bogus length.

Add a check to capture this off-nominal case. This will log the appropriate warnings as in other cases and prevent a crash. Also update the log messages to be clearer.

Resolves: #1707

### DIFF
--- a/channels/chan_iax2.c
+++ b/channels/chan_iax2.c
@@ -10145,7 +10145,7 @@ static int socket_process_meta(int packet_len, struct ast_iax2_meta_hdr *meta, s
 						iaxs[fr->callno]->last = fr->ts;
 				}
 			} else {
-				ast_log(LOG_WARNING, "Datalen < 0?\n");
+				ast_log(LOG_ERROR, "Dropping malformed frame (datalen %d?)\n", f.datalen);
 			}
 		}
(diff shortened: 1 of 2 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

channels/chan_iax2.c
```diff
@@ -12017,6 +12017,12 @@ static int socket_process_helper(struct iax2_thread *thread)
 			return 1;
 		}
 		f.datalen = res - sizeof(*vh);
+		if (f.datalen < 0) {
+			ast_log(LOG_WARNING, "Datalen < 0?\n");
```
seanbright: I know this is copy/paste from a similar error path but the message is not really helpful to the user. Maybe something about receiving a malformed frame and dropping it?

jcolp: Agreed. When it comes to additions of log messages we should strive to make the more useful with more information and as user friendly as possible.

asteriskteam:

Issue #1709, opened 2026-01-08
[improvement]: app_mf: Fix minor formatting issue.


InterLinked1:
### Improvement Description

Fix formatting to comply with the coding style.

PR #1710, opened 2026-01-08
app_mf: Fix minor formatting issue.

InterLinked1:
Resolves: #1709

### DIFF
--- a/apps/app_mf.c
+++ b/apps/app_mf.c
@@ -242,7 +242,8 @@ static const char sendmf_name[] = "SendMF";
  * \retval 0 if successful
  * \retval -1 if unsuccessful (including hangup).
  */
-static int read_mf_digits(struct ast_channel *chan, char *buf, int buflen, int timeout, int features, int laxkp, int override, int no_kp, int no_st, int maxdigits) {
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seanbright:
Can we just roll this in to the next functional change? An issue and a PR to add a newline character seems really excessive to me.

InterLinked1:
> Can we just roll this in to the next functional change? An issue and a PR to add a newline character seems really excessive to me.

I don't have any functional changes planned for `app_mf` currently, do you mean something else?

I _thought_ it was best practice to avoid doing formatting changes in other commits, and that issues were required, but happy to change practice here.

seanbright:
OK

jcolp:
My opinion is that I don't like code formatting change PRs. If there are fixes that occur as a result of functional changes then that's fine to include them, but PRs (particularly one line fixes) like this I don't like. It got past code review, we should have caught it, we didn't, it's not hurting anything, life goes on.

jcolp:
In fact, I'm closing this out.

PR #1713, opened 2026-01-11
res_pjsip_rfc7329: Add RFC7329 Session-ID support

mr-gmr:
Introduce a new res_pjsip_rfc7329 module that generates, stores, and propagates RFC7329 Session-ID values for PJSIP requests and responses. Session-IDs are reused across call legs via linkedid mapping and are ensured for in-dialog requests that bypass supplements. A new session supplement callback is added to hook channel creation so linkedid-based state can be attached when the channel is available.
Session-ID handling is now controlled by pjsip.conf. The module loads but remains inactive unless [global] rfc7329_enable=yes is set. rfc7329_enable=no, or an empty [global] leaves behavior disabled. Module reload res_pjsip_rfc7329.so re-reads the config and activates/deactivates the logic live.
The change also integrates Session-ID into REFER handling by persisting the inbound Refer-To Session-ID and adding a Session-ID parameter to outgoing Refer-To URIs.
When enabled, attempts to add, update, or remove Session-ID with PJSIP_HEADER are blocked to preserve RFC7329 immutability.

UserNote: Enable RFC7329 by setting [global] rfc7329_enable=yes in pjsip.conf and reloading the module. When enabled, Session-ID propagation is enforced and PJSIP_HEADER cannot modify Session-ID. REFER processing preserves Session-ID in Refer-To when available. https://www.rfc-editor.org/rfc/rfc7329

UpgradeNote: Realtime users must apply the Alembic migration that adds `rfc7329_enable` to `ps_globals` (config DB) to keep the schema in sync with the new global option.

### DIFF
--- a/include/asterisk/res_pjsip_session.h
+++ b/include/asterisk/res_pjsip_session.h
@@ -364,6 +364,11 @@ struct ast_sip_session_supplement {
 	 * Defaults to AST_SIP_SESSION_BEFORE_MEDIA
 	 */
 	enum ast_sip_session_response_priority response_priority;
+	/*!
+	 * \brief Notification that the session has an associated channel
+	 * This method will always be called from a SIP servant thread.
+	 */
+	void (*session_channel_created)(struct ast_sip_session *session);
 };
 
 enum ast_sip_session_sdp_stream_defer {
@@ -624,6 +629,12 @@ void ast_sip_session_register_supplement_with_module(struct ast_module *module, 
  * \param supplement The supplement to unregister
  */
 void ast_sip_session_unregister_supplement(struct ast_sip_session_supplement *supplement);
+
+/*!
+ * \brief Notify supplements that a session now has an associated channel
+ * \param session The session whose channel has been created
+ */
+void ast_sip_session_channel_created(struct ast_sip_session *session);
 
 /*!
  * \brief Add supplements to a SIP session
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -2756,6 +2756,9 @@ static struct ast_channel *chan_pjsip_request_with_stream_topology(const char *t
 		SCOPE_EXIT_RTN_VALUE(NULL, "Couldn't create channel\n");
 	}
 
+	/* Notify supplements that need channel-backed state (e.g., RFC7329 linkedid map). */
+	ast_sip_session_channel_created(session);
+
 	SCOPE_EXIT_RTN_VALUE(session->channel, "Channel: %s\n", ast_channel_name(session->channel));
 }
 
@@ -3060,6 +3063,9 @@ static int chan_pjsip_incoming_request(struct ast_sip_session *session, struct p
 		SCOPE_EXIT_LOG_RTN_VALUE(-1, LOG_ERROR, "%s: Failed to allocate new PJSIP channel on incoming SIP INVITE\n",
 			 ast_sip_session_get_name(session));
 	}
+
+	/* Notify supplements that need channel-backed state (e.g., RFC7329 linkedid map). */
+	ast_sip_session_channel_created(session);
 
 	set_sipdomain_variable(session);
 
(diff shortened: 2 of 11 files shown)


mr-gmr:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

seanbright:
> Loading res_pjsip_rfc7329 enables RFC7329 Session-ID propagation and prevents Session-ID modification via PJSIP_HEADER

If it is going to change behavior just by having the module loaded, I don't think it would be wise for it to go in to release branches.

gtjoseph:
A testsuite test should be created for this.

configs/samples/res_pjsip_rfc7329.conf.sample
```diff
@@ -0,0 +1,5 @@
+[general]
```
gtjoseph: Rather than adding a new config file just to enable/disable the module, you could add a parameter to the pjsip.conf `[global]` section.  Or...is this something that may need to be turned on or off on an endpoint-by-endpoint basis?

gtjoseph (changes requested):

configs/samples/res_pjsip_rfc7329.conf.sample
mr-gmr: I think this should be global. I can add a flag in asterisk.conf or pjsip.conf if it prefered

gtjoseph: The pjsip.conf [global] section would be the correct place.

mr-gmr: done

include/asterisk/res_pjsip_session.h
```diff
@@ -301,6 +301,11 @@ struct ast_sip_session_supplement {
 	 * This method will always be called from a SIP servant thread.
 	 */
 	void (*session_begin)(struct ast_sip_session *session);
+	/*!
+	 * \brief Notification that the session has an associated channel
+	 * This method will always be called from a SIP servant thread.
+	 */
+	void (*session_channel_created)(struct ast_sip_session *session);
```
gtjoseph: This callback needs to go at the end of the structure to preserve backwards compatibility.

gtjoseph (changes requested):
This isn't a complete review, just something that stood out initially.

include/asterisk/res_pjsip_session.h
mr-gmr: done

asteriskteam:

mr-gmr:
testsuite-test-pr: 127

mr-gmr:
I don't even understand what's wrong with the tests here. Locally, everything passes...

gtjoseph:
> I don't even understand what's wrong with the tests here. Locally, everything passes...

Let me investigate

gtjoseph:
@mr-gmr There's no documentation for the `rfc7329_enable` option in res/res_pjsip/pjsip_config.xml.  When Asterisk is configured with `--enable-dev-mode` (which all developers should be doing) a pjsip config option without XML documentation will cause the the containing object to not load and will generate error messages.  The effect is that rfc7329_enable is always "no" so your enable tests will all fail.

mr-gmr:
> @mr-gmr There's no documentation for the `rfc7329_enable` option in res/res_pjsip/pjsip_config.xml. When Asterisk is configured with `--enable-dev-mode` (which all developers should be doing) a pjsip config option without XML documentation will cause the the containing object to not load and will generate error messages. The effect is that rfc7329_enable is always "no" so your enable tests will all fail.


Thanks. Fixed it. I'll check the comments in testsuite tomorrow.

gtjoseph (dismissed):
`UserNote : Added tests` needs to be removed from the commit message.

mr-gmr:
> `UserNote : Added tests` needs to be removed from the commit message.

thanks. fixed

Issue #1711, opened 2026-01-09
[bug]: Missing Contact: header in 200 OK


popy2k14:
### Severity

Major

### Versions

20.6.0 (Starface 9.0.3.4 PBX - which is based on Asterisk)

### Components/Modules

chan_sip

### Operating Environment

Linux, Alma (Redhat), 8.10

### Frequency of Occurrence

Occasional

### Issue Description

In some cases asterisk answers a call (SIP 200 OK) without the {{Contact:}} header.
It results, for the receiving peer, to terminate the call (SIP BYE) because he can't answer.
For example, if the receiving peer is an linphone android, you get the following WARNING in the log:
`Missing contact header in resp [0xb40000796077c9a0] cannot set remote target for dialog [0xb4000079a0706d70]`

The issue doesn't occur every time with linphone 6.0.20 android on my test device auf in the following rate: ~4x/33x  calls
There is an archived issue which describes the issue in detail: https://issues-archive.asterisk.org/ASTERISK-24915

Here is the detail description of the old issue, which nails it down:
```
More precisely this happens when asterisk receives a SIP message with a bad CSeq (for example with CSeq-1). Then the next 200 OK doesn't contain the Contact: header.

*When does it happen in real life ?*

In real life I saw it with two asterisk interconnected with a SIP trunk via a WAN.
When there is some asymetric latency it results in retransmissions.
{code}
< INVITE            (CSeq 1)
> 401 Unauthorized  (CSeq 1)
> 401 Unauthorized  (CSeq 1) .... retransmission because ACK is not received on time
< ACK               (CSeq 1) .... ACK of the first 401 Unauthorized
< INVITE            (CSeq 2)
> 100 Trying        (CSeq 2)  
> 180 Ringing       (CSeq 2)
< ACK               (Cseq *1*) .. ACK of the retransmitted 401 Unauthorized
> 200 OK            (Cseq 2) .... /!\ which does not contain the Contact: header
{code}


*How to reproduce*
Below is a SIP exchange forged with SIPp which reproduce the problem :

{code}
< INVITE            (CSeq 1)
> 401 Unauthorized  (CSeq 1)
< ACK               (CSeq 1)
< INVITE            (CSeq 2)
> 100 Trying        (CSeq 2) ..... contains Contact: header
> 180 Ringing       (CSeq 2) ..... contains Contact: header
< ACK               (Cseq *1*)
> 200 OK            (Cseq 2) ..... /!\ does not contain Contact: header
{code}

[...]

PR #1714, opened 2026-01-11
chan_sip.c: Ensure Contact header is set on responses to INVITE.

seanbright:
From the original report* on ASTERISK-24915:

  > The problem occurs because the handle_incoming function updates
  p->method to req->method (p being a struct sip_pvt *) before
  checking if the CSeq makes sense, and if the CSeq is unexpected, it
  does not reset p->method to its old value before returning. Then,
  when asterisk sends the 200 OK response for the original INVITE,
  since p->method is now equal to SIP_ACK (instead of SIP_INVITE), the
  resp_need_contact function (called from respprep) says "its a SIP
  ACK, no need to add a Contact header for the response", which is
  wrong, since it's not a SIP ACK but a SIP INVITE dialog.

I have confirmed that the analysis is correct and that the patch fixes the behavior.

*: https://issues-archive.asterisk.org/ASTERISK-24915

Resolves: #1711

### DIFF
--- a/channels/chan_sip.c
+++ b/channels/chan_sip.c
@@ -29233,6 +29233,7 @@ static int handle_incoming(struct sip_pvt *p, struct sip_request *req, struct as
 			} else if (req->method != SIP_ACK) {
 				transmit_response(p, "500 Server error", req);	/* We must respond according to RFC 3261 sec 12.2 */
 			}
+			p->method = oldmethod;
 			return -1;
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: none

gtjoseph (approved):

jcolp (approved):

Issue #1716, opened 2026-01-14
[bug]: Ghost call when UAC didn't respond with 487 for a cancel request from server even after original call hangup.


hishamway:
### Severity

Major

### Versions

20,22,23

### Components/Modules

res_pjsip_session, chan_pjsip

### Operating Environment

Linux

### Frequency of Occurrence

Constant

### Issue Description

**Summary**

When a CANCEL is sent after caller hangup but the remote endpoint doesn't respond with 487 (due to network disconnection), Asterisk incorrectly attempts to failover and sends a new INVITE to the next address, even though the session is already DISCONNECTED.

**Problem**

In check_request_status() in res/res_pjsip_session.c, when a transaction times out (408) or receives a 503 error, the function attempts to failover to the next available address by calling pjsip_inv_uac_restart() and sending a new INVITE. However, the function does not check whether the inv_session is already in PJSIP_INV_STATE_DISCONNECTED state before attempting the failover.
This means:
* After the caller hangs up and CANCEL is sent, if the INVITE transaction times out (no 487 received), Asterisk sends a NEW outbound INVITE
* This violates RFC 3261 Section 9.1 which states that if no final response is received within 64T1 seconds, the transaction should be destroyed, not retried
* Unexpected outbound calls are placed after the original call was terminated
* The ghost call appears to ring devices that should no longer be involved in the call

**Scenario**

1. Inbound call arrives at Asterisk
2. Asterisk sends INVITE to multiple endpoints (e.g., ring group with 3 devices)
3. Caller hangs up before any device answers
4. Asterisk sends CANCEL to all devices
5. One device (e.g., mobile) loses network connectivity and doesn't respond with 487
6. INVITE transaction times out (PJSIP_EVENT_TIMER, status 408)
7. BUG: check_request_status() sees the 408 and multiple addresses available, calls pjsip_inv_uac_restart() and sends a new INVITE even though session is DISCONNECTED

### Relevant log output

```shell
[2026-01-13 08:00:01.115] DEBUG res_pjsip_session.c: sipcore: Destroying SIP session
[2026-01-13 08:00:01.157] DEBUG res_pjsip_session.c: sipcore: Source of transaction state change is TIMER
[2026-01-13 08:00:01.157] DEBUG res_pjsip_session.c: sipcore: Method is INVITE
[2026-01-13 08:00:01.157] DEBUG res_pjsip_session.c: The current inv state is DISCONNECTED
[...]

PR #1717, opened 2026-01-14
res_pjsip_session.c: Prevent INVITE failover when session is cancelled

hishamway:
When an outbound INVITE transaction times out (408) or receives a 503 error,
check_request_status() attempts to failover to the next available address by
restarting the INVITE session. However, the function did not check if the
inv_session was already cancelled before attempting the failover.

This caused unexpected behavior when a caller hung up during a ring group
scenario: after CANCEL was sent but the remote endpoint failed to respond
with 487 (e.g., due to network disconnection), the transaction timeout
would trigger a NEW outbound INVITE to the next address, even though the
session was already terminated.

This violates RFC 3261 Section 9.1 which states that if no final response
is received after CANCEL within 64*T1 seconds, the client should consider
the transaction cancelled and destroy it, not retry to another address.

The fix adds a check for both PJSIP_INV_STATE_DISCONNECTED and inv->cancelling
at the beginning of check_request_status(). This ensures that:
- Failover is blocked when the user explicitly cancelled the call (CANCEL sent)
- Failover is still allowed for legitimate timeout/503 scenarios where no
  CANCEL was initiated (e.g., SRV failover when first server is unreachable)

Resolves: #1716

### DIFF
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -4556,6 +4556,10 @@ static int check_request_status(pjsip_inv_session *inv, pjsip_event *e)
 	struct ast_sip_session *session = inv->mod_data[session_module.id];
 	pjsip_transaction *tsx = e->body.tsx_state.tsx;
 
+	if (inv->state == PJSIP_INV_STATE_DISCONNECTED && inv->cancelling) {
+		return 0;
+	}
+
 	if (tsx->status_code != 503 && tsx->status_code != 408) {
 		return 0;
 	}


hishamway:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

hishamway:
Updated Title and description, @jcolp could you please resume PR merge check runs.

hishamway:
This test failure is not related to the change in this PR. Correct me if I am wrong.
Also I changed the PR Title same as Commit title.

PR #1720, opened 2026-01-14
res_pjsip_session: Make sure NAT hook runs when packet is retransmitted for whatever reason.

arcivanov:
This hook may not be necessary when we do a retransmit, but when there are two
INVITEs, one *initial* and one with auth digest, the second INVITE contains wrong (unmodified) media address
due to the commented line below.
The NAT hook needs to run due to filters potentially reverting previously modified packets.

Fixes: #449

### DIFF
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -58,7 +58,6 @@
 #define SDP_HANDLER_BUCKETS 11
 
 #define MOD_DATA_ON_RESPONSE "on_response"
-#define MOD_DATA_NAT_HOOK "nat_hook"
 
 /* Most common case is one audio and one video stream */
 #define DEFAULT_NUM_SESSION_MEDIA 2
@@ -5494,8 +5493,6 @@ static pjsip_inv_callback inv_callback = {
 static void session_outgoing_nat_hook(pjsip_tx_data *tdata, struct ast_sip_transport *transport)
 {
 	RAII_VAR(struct ast_sip_transport_state *, transport_state, ast_sip_get_transport_state(ast_sorcery_object_get_id(transport)), ao2_cleanup);
-	struct ast_sip_nat_hook *hook = ast_sip_mod_data_get(
-		tdata->mod_data, session_module.id, MOD_DATA_NAT_HOOK);
 	pjsip_sdp_info *sdp_info;
 	pjmedia_sdp_session *sdp;
 	pjsip_dialog *dlg = pjsip_tdata_get_dlg(tdata);
@@ -5503,10 +5500,9 @@ static void session_outgoing_nat_hook(pjsip_tx_data *tdata, struct ast_sip_trans
 	int stream;
 
 	/*
-	 * If there's no transport_state or body, or the hook
-	 * has already been run, just return.
+	 * If there's no transport_state or body, just return.
 	 */
-	if (ast_strlen_zero(transport->external_media_address) || !transport_state || hook || !tdata->msg->body) {
+	if (ast_strlen_zero(transport->external_media_address) || !transport_state || !tdata->msg->body) {
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


arcivanov:
This is a draft, do not merge.

arcivanov:
This has been thoroughly tested. Suspicious behavior was identified to be due to host's routing table and unrelated to this change.

res/res_pjsip_session.c
```diff
@@ -5557,8 +5557,12 @@ static void session_outgoing_nat_hook(pjsip_tx_data *tdata, struct ast_sip_trans
...
 	}
 
-	/* We purposely do this so that the hook will not be invoked multiple times, ie: if a retransmit occurs */
-	ast_sip_mod_data_set(tdata->pool, tdata->mod_data, session_module.id, MOD_DATA_NAT_HOOK, nat_hook);
+	/* We purposely originally did this so that the hook would not be invoked multiple times, ie: if a retransmit occurs.
+	   However, as described in asterisk/asterisk/449: This hook may not be necessary when we do a retransmit, but when there are two
+	   INVITEs, one *initial* and one with auth digest, the second INVITE contains wrong (unmodified) media address
+	   due to the commented line below. The NAT hook needs to run due to filters potentially reverting previously modified packets.
+	*/
+	// ast_sip_mod_data_set(tdata->pool, tdata->mod_data, session_module.id, MOD_DATA_NAT_HOOK, nat_hook);
```
seanbright: All of this would be better in the commit message. Just delete everything here.

For future reference, we don't use `//` style comments.

seanbright (changes requested):

res/res_pjsip_session.c
arcivanov: Fixed

arcivanov:
Looks like test failure is spurious and unrelated.

seanbright:
I can't comment on the correctness of the change, so removing my review request.

arcivanov:
Please rerun the failed test?

seanbright:
[Please read and follow the instructions here](https://github.com/asterisk/asterisk/pull/1720#pullrequestreview-3662292914)

arcivanov:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

seanbright:
> [Please read and follow the instructions here](https://github.com/asterisk/asterisk/pull/1720#pullrequestreview-3662292914)

Nevermind. I took care of it. The only branches eligible for bug fixes are 20, 22, and 23. I think the automated message used to say that but doesn’t anymore for whatever reason.

res/res_pjsip_session.c
```diff
@@ -5557,8 +5557,6 @@ static void session_outgoing_nat_hook(pjsip_tx_data *tdata, struct ast_sip_trans
 		}
 	}
 
-	/* We purposely do this so that the hook will not be invoked multiple times, ie: if a retransmit occurs */
-	ast_sip_mod_data_set(tdata->pool, tdata->mod_data, session_module.id, MOD_DATA_NAT_HOOK, nat_hook);
```
gtjoseph, line 5565: If we blanket re-run the nat hook, does this mess up re-transmits?

arcivanov: The problem is the issue has nothing to do with retransmits - this suppresses the NAT hooks in the INVITE -> 401 -> INVITE w/auth sequence where no retransmits are involved. NAT hooks seem to be idempotent and when I was testing with retransmits they didn't seem to cause any problems.

gtjoseph: Yep, I understand your issue isn't retransmit related.  The existing comment just had me thinking.  I'm trying to remember if pjproject handles the retransmits without percolating it up to us.  I'm also thinking that if we don't set this, then there's no need to get or test it at the top of the function so that should be removed as well

asteriskteam:

arcivanov:
I will remove the get and test as well.

arcivanov:
Failures don't appear to have anything to do with this case here.

gtjoseph:
> Failures don't appear to have anything to do with this case here.

correct.  not your issue.

jcolp (changes requested):
The commit message needs to be updated to conform to our standard including subsystem at the front of the first line:

Otherwise I think this is fine.

PR #1732, opened 2026-01-20
CEL: Introduce Advanced CEL custom mapping for performance.

gtjoseph:
The legacy cel_custom and cel_sqlite3_custom channel event logging backends
are very heavy resource consumers because they use recursive string
interpolation and dialplan function evaluation. They're so resource intensive
that it actually takes more resources to write the event log records than it
does to actually process a call. Unfortunately, the existing configuration
file formats don't give us much leeway to capture the extra information
needed to shortcut the string interpolation so...

UpgradeNote: New "advanced" cel_custom and cel_sqlite3_custom configuration
formats are now available that allow significant reductions in resources
required to write channel event log entries.  The details are in the
cel_custom and cel_sqlite3_custom sample config files but by capturing
additional information at load time, we can eliminate the need for the string
interpolation and dialplan function calls which were the main contributors to
the resource over-utilization.  As an added benefit, the cel_custom module
can now also produce JSON formatted output.  The legacy configuration formats
are still supported but all users are encouraged to switch to the new format.

### DIFF
--- a/cel/cel_custom.c
+++ b/cel/cel_custom.c
@@ -42,6 +42,10 @@
 #include "asterisk/lock.h"
 #include "asterisk/threadstorage.h"
 #include "asterisk/strings.h"
+#include "asterisk/vector.h"
+#include "asterisk/json.h"
+
+#include "custom_common.h"
 
 #define CONFIG "cel_custom.conf"
 
@@ -49,14 +53,6 @@ AST_THREADSTORAGE(custom_buf);
 
 static const char name[] = "cel-custom";
 
-struct cel_config {
-	AST_DECLARE_STRING_FIELDS(
-		AST_STRING_FIELD(filename);
-		AST_STRING_FIELD(format);
-	);
-	ast_mutex_t lock;
-	AST_RWLIST_ENTRY(cel_config) list;
-};
 
 #define CUSTOM_BACKEND_NAME "CEL Custom CSV Logging"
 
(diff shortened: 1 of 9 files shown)


gtjoseph:
~~This is still very much a WORK IN PROGRESS.~~

FYI... cel_sqlite3_custom can also benefit from these changes.
The latest push includes it.

cel/cel_custom.c
```diff
@@ -62,22 +97,202 @@ struct cel_config {
...
+			strcpy(cel_field->name, field); /* Safe */
+		} else if (strcasecmp(field, "userdeftype") == 0) {
+			cel_field->ie_type = AST_EVENT_IE_CEL_USEREVENT_NAME;
+			cel_field->ie_pltype = AST_EVENT_IE_PLTYPE_STR;
+			strcpy(cel_field->name, field); /* Safe */
+		} else {
+			ast_debug(2, "%s: Field '%s' not found. Using as literal\n", sink->filename, field);
+			cel_field->ie_type = AST_EVENT_IE_CEL_LITERAL;
+			cel_field->ie_pltype = AST_EVENT_IE_PLTYPE_STR;
+			strcpy(cel_field->name, field); /* Safe */
```
seanbright: This same call appears in all 4 blocks. Could extract it.

main/event.c
```diff
@@ -226,6 +227,20 @@ enum ast_event_ie_pltype ast_event_get_ie_pltype(enum ast_event_ie_type ie_type)
 	return ie_maps[ie_type].ie_pltype;
 }
 
+int ast_event_str_to_ie_type(const char *str, enum ast_event_ie_type *ie_type)
+{
+	enum ast_event_ie_type ie;
+
+	for (ie = 1; ie < AST_EVENT_IE_TOTAL; ie++) {
```
seanbright: Use `AST_EVENT_IE_NEWMSGS` for the initial value of `ie` or throw in a comment explaining why we're starting at 1.

gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

cel/cel_custom.c
```diff
@@ -87,90 +302,300 @@ static int load_config(void)
...
+	} else {
+		ast_log(LOG_ERROR, "Unable to re-open master file %s : %s\n", config->filename, strerror(errno));
+	}
+
+	ast_mutex_unlock(&config->lock);
+}
+
+static char *quoter(const char *value, char quote, char quote_escape)
+{
+	size_t len = strlen(value) + 32;
```
phoneben: maybe ast_malloc(strlen(value) * 2 + 1)  to avoid buffer overflow.

worst-case you may need almost 2× the input size (if every character needs escaping)

cel/cel_custom.c
```diff
@@ -62,22 +97,202 @@ struct cel_config {
...
+
+static int load_basic_config(struct ast_category *category, int *mappings)
+{
+	struct ast_variable *var = NULL;
+	struct cel_config *sink = NULL;
+	int res = 0;
+
+	for (var = ast_category_first(category); var; var = var->next) {
+		if (ast_strlen_zero(var->name) || ast_strlen_zero(var->value)) {
+			ast_log(LOG_WARNING, "CEL Mapping must have both a filename and a format at line %d\n", var->lineno);
```
phoneben: not template ?
`ast_log(LOG_WARNING, "Legacy CEL Mapping must have both a filename and a template at line %d\n", var->lineno);`

gtjoseph: done.

cel/cel_custom.c
gtjoseph: done

gtjoseph:
Moving back to draft because cdr_custom and cdr_sqlite3_custom can also take advantage of what was done here.

cel/custom_common.c
```diff
@@ -0,0 +1,215 @@
...
+			*bufptr++ = quote_escape;
+		}
+		*bufptr++ = *dataptr++;
+	}
+	*bufptr='\0';
+	return ptr;
+}
+
+int csv_append_string(struct ast_str **str, int is_first, const char *value, struct cel_config *config)
+{
```
phoneben, line 52: `  value = S_OR(value, ""); `
should be added 

NULL safety: `csv_append_string(`) calls `strchr(value, …) `and formatting helpers that assume a non-NULL string.
`get_event_string() `may return NULL on allocation failure, so this can segfault.

cel/custom_common.c
```diff
@@ -0,0 +1,215 @@
...
+	return res;
+}
+
+int csv_append_uint(struct ast_str **str, int is_first, uint32_t value, struct cel_config *config)
+{
+	int res = 0;
+	char *sep_str = is_first ? "" : config->separator;
+
+	if (config->quoting_method == quoting_method_all) {
+		res = ast_str_append(str, 0, "%s%s%d%s", sep_str, config->quote, value, config->quote);
```
phoneben: `csv_append_uint()` prints unsigned as signed `(%d)`
You’re using` %d `with a `uint32_t`:

If value is > 2,147,483,647, it can print negative numbers. 
Fix: use `%u` (or PRIu32).

cel/custom_common.c
gtjoseph: done

cel/custom_common.c
gtjoseph: done

gtjoseph:
This latest push does a bunch of refactoring but doesn't include any CDR changes.  Those will be in a separate PR.

gtjoseph:
OK back to draft again because I thought of a few things while doing the CDRs.

cel/custom_common.c
```diff
@@ -0,0 +1,296 @@
...
+
+static int csv_append_uint(struct ast_str **str, int is_first, uint32_t value, struct cel_config *config)
+{
+	int res = 0;
+	char *sep_str = is_first ? "" : config->separator;
+
+	if (config->quoting_method == cel_quoting_method_all) {
+		res = ast_str_append(str, 0, "%s%s%u%s", sep_str, config->quote, value, config->quote);
+	} else {
+		res = ast_str_append(str, 0, "%s%d", sep_str, value);
```
phoneben, line 92: also should be `%u` vs `%d`

cel/custom_common.c
phoneben: did i miss something ?
i do not see  did you  take care of it in a other way ?
```
/* FIX 1: Guard against NULL immediately so strchr doesn't crash */
    value = S_OR(value, "");
```

cel/custom_common.c
```diff
@@ -0,0 +1,296 @@
...
+	csv_append_uint(str, is_first, ast_event_get_ie_uint(event, cel_field->ie_type), config);
+}
+
+/* JSON */
+
+static void append_json_event_string(struct ast_json *json, struct ast_event *event, struct cel_config *config,
+	struct cel_field *cel_field, int is_first)
+{
+	const char *value = ast_event_get_ie_str(event, cel_field->ie_type);
+	ast_json_object_set(json, cel_field->name, ast_json_string_create(value));
```
phoneben, line 166: suggestion 
Guard JSON string creation from NULL.

`ast_json_string_create(S_OR(value, "")) `
there are 4 like these

cel/custom_common.c
```diff
@@ -0,0 +1,296 @@
...
+		char *evalue = (char *)value;
+		if (strchr(value, config->quote[0])) {
+			evalue = quoter(value, config->quote[0], config->quote_escape[0]);
+		}
+		res = ast_str_append(str, 0, "%s%s%s%s", sep_str, config->quote, S_OR(evalue, ""), config->quote);
+		if (evalue != value) {
+			ast_free(evalue);
+		}
+	} else if (config->quoting_method == cel_quoting_method_minimal) {
+		if (strchr(value, config->separator[0])) {
```
phoneben, line 66: In` cel_quoting_method_minimal`, we only quote if the separator is present, but CSV also requires quoting if the value contains the quote character or newline/CR. Otherwise we can emit invalid CSV (e.g., values with `"` or `\n` will break parsing or split rows). Recommend extending the condition to quote when value contains separator or quote char or` \n`/`\r`.




```
} else if (config->quoting_method == cel_quoting_method_minimal) {

    if (strchr(value, config->separator[0]) ||
        strchr(value, config->quote[0]) ||
        strchr(value, '\n') ||
        strchr(value, '\r')) {

        char *evalue = (char *) value;

        /* Escape embedded quote characters */
        if (strchr(value, config->quote[0])) {
            evalue = quoter(value, config->quote[0], config->quote_escape[0]);
        }

        res = ast_str_append(str, 0, "%s%s%s%s",
            sep_str,
            config->quote,
            S_OR(evalue, ""),
            config->quote);

        if (evalue != value) {
            ast_free(evalue);
        }

    } else {
        /* Safe to emit unquoted */
        res = ast_str_append(str, 0, "%s%s", sep_str, value);
    }
```

cel/custom_common.c
```diff
@@ -0,0 +1,296 @@
...
+	*bufptr='\0';
+	return ptr;
+}
+
+static int csv_append_string(struct ast_str **str, int is_first, const char *value, struct cel_config *config)
+{
+	int res = 0;
+	char *sep_str = is_first ? "" : config->separator;
+
+	if (config->quoting_method == cel_quoting_method_all || config->quoting_method == cel_quoting_method_non_numeric) {
```
phoneben, line 56: `quoting_method_non_numeric` currently behaves identically to `all` — there is no numeric detection, so all string values are quoted.
This makes the option misleading. Either numeric detection should be added (and numeric fields left unquoted), or the option should be removed/renamed.

cel/custom_common.c
gtjoseph: Oops, missed that one.

cel/custom_common.c
gtjoseph: The JSON calls are all NULL tolerant but guarding them is a good idea anyway since jansson will leave the the key/value pair out altogether if the value is NULL.

seanbright: Because there is a semantic difference between `null` and `””`. I would stick with jansson’s default behavior.

cel/custom_common.c
gtjoseph: Whether something is numeric or not is based on the data type of the CEL event field, not whether there is only numeric data in the field.  Currently, only the amaflags field is defined as a uint and that IS left unquoted when non_numeric is used.

cel/custom_common.c
gtjoseph: In reality, they can't be NULL anyway.  These fields all come from STRING_FIELDS in a channel snapshot and STRING_FIELDs can't be NULL.

cel/custom_common.c
phoneben: Thanks for the clarification — that makes sense.
It might be helpful to add a brief comment noting that `quoting_method_non_numeric` is **type-based (CEL field type)** rather than content-based, since the name could otherwise be read as string numeric detection.

gtjoseph:
This PR is being closed in favor of an upcoming one that includes CDR.

PR #1734, opened 2026-01-21
chan_websocket.conf.sample: Fix category name.

gtjoseph:
UserNote: The category name in the chan_websocket.conf.sample file was
incorrect.  It should be "global" instead of "general".


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

PR #1741, opened 2026-01-23
app_queue: fix double increment of member->calls with shared_lastcall

### LINKED ISSUE #1736: [bug]: app_queue: update_queue() may double-increment member->calls with shared_lastcall=yes (regression observed after 20.17)
### Severity

Major

### Versions

20.17.0

### Components/Modules

app_queue.c

### Operating Environment

Ubuntu 24.04.3 LTS

### Frequency of Occurrence

Frequent

### Issue Description

Under high concurrency and with shared_lastcall=yes, app_queue.c:update_queue() may increment member->calls and queue-level counters more than once for a single completed call.

This results in observable counter inflation such as:

Agent has taken calls increasing by +2 instead of +1

callscompleted exceeding the actual number of handled calls

The issue becomes clearly reproducible in large-scale environments with many queues and concurrent calls.
Environment

Asterisk version: 20.17

Queues: ~2000

Concurrent calls: 100–200+

Configuration: shared_lastcall=yes

The same topology and traffic pattern did not exhibit this behavior on earlier 20.x deployments (e.g. 20.15.2). The issue became reproducible after upgrading to 20.17.

Observed Behavior

For a single call handled by an agent:

member->calls may be incremented more than once

Queue statistics reflect duplicated completion counts

This occurs when update_queue() is invoked multiple times for the same call (e.g. different state transitions reaching the function concurrently).

***Impact on Call Distribution (fewestcalls strategy)
This issue is not limited to statistics reporting.
When using queue strategies that depend on member->calls (such as fewestcalls), the double increment directly affects call distribution behavior.
Because member->calls may increase by +2 instead of +1 for a single completed call, the affected agent is incorrectly deprioritized in subsequent call selection. This leads to:
Skewed call distribution
Agents receiving fewer calls than intended
Violation of the expected fairness guarantees of the fewestcalls strategy
As a result, this race condition impacts functional call routing correctness, not just accounting metrics.

Root Cause Analysis:
update_queue() attempts to ensure a call is only counted once using:
if (!starttime || (member->starttime != starttime)) {
    return 0;
}

However, this check is not atomic.
Under concurrent execution, multiple threads may:
Observe the same member->starttime
Pass the check

Increment counters independently
With shared_lastcall=yes, the effect is amplified because the same agent entry is iterated across all queues.
The increased concurrency / task scheduling behavior in 20.17 appears to make this existing race condition much more visible, suggesting a regression in timing rather than logic.

Proposed Fix:
Treat member->starttime as a single-use token and consume it exactly once using an atomic compare-and-swap (CAS).
Example approach:

if (!__sync_bool_compare_and_swap(&member->starttime, starttime, 0)) {
    return 0;
}

This guarantees:
Only one execution path processes the call
Duplicate triggers safely exit
No additional locking is introduced
Existing semantics of shared_lastcall are preserved

Queue-level counters (member->calls) should only be updated if this gate is successfully passed.

I uploaded whole code part in update_queue_function_V1.c file.
[update_queue_function_V1.c](https://github.com/user-attachments/files/24789074/update_queue_function_V1.c)

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

smtcbn:
Under high concurrency, update_queue() may be invoked multiple times
for the same call, causing member->calls and queue-level counters to
be incremented more than once.

The existing starttime check is not atomic and allows concurrent
execution paths to pass. Treat member->starttime as a single-use token
and consume it via CAS to ensure the call is counted exactly once.

This also prevents incorrect call distribution when using strategies
such as fewestcalls.

Observed as a regression after upgrading to 20.17.

Resolves: #1736

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -6162,36 +6162,48 @@ static int wait_our_turn(struct queue_ent *qe, int ringing, enum queue_result *r
  * \brief update the queue status
  * \retval 0 always
 */
-static int update_queue(struct call_queue *q, struct member *member, int callcompletedinsl, time_t starttime)
+static int update_queue(struct call_queue *q, struct member *member,
+	int callcompletedinsl, time_t starttime)
 {
 	int oldtalktime;
-	int newtalktime = time(NULL) - starttime;
+	int newtalktime;
 	struct member *mem;
 	struct call_queue *qtmp;
 	struct ao2_iterator queue_iter;
+	int did_increment_any = 0;
+
+	if (!starttime) {
+		return 0;
+	}
+
+	newtalktime = (int)(time(NULL) - starttime);
 
 	/* It is possible for us to be called when a call has already been considered terminated
 	 * and data updated, so to ensure we only act on the call that the agent is currently in
 	 * we check when the call was bridged.
 	 */
-	if (!starttime || (member->starttime != starttime)) {
+	ao2_lock(q);
+	if (member->starttime != starttime) {
+		ao2_unlock(q);
 		return 0;
 	}
+	member->starttime = 0;
+	ao2_unlock(q);
 
 	if (shared_lastcall) {
 		queue_iter = ao2_iterator_init(queues, 0);
-		while ((qtmp = ao2_t_iterator_next(&queue_iter, "Iterate through queues"))) {
+		while ((qtmp = ao2_t_iterator_next(&queue_iter, "Iterate queues"))) {
 			ao2_lock(qtmp);
 			if ((mem = ao2_find(qtmp->members, member, OBJ_POINTER))) {
 				time(&mem->lastcall);
 				mem->calls++;
 				mem->callcompletedinsl = 0;
-				mem->starttime = 0;
 				mem->lastqueue = q;
+				did_increment_any = 1;
 				ao2_ref(mem, -1);
 			}
 			ao2_unlock(qtmp);
-			queue_t_unref(qtmp, "Done with iterator");
+			queue_t_unref(qtmp, "queue iteration done");
 		}
 		ao2_iterator_destroy(&queue_iter);
 	} else {
@@ -6199,8 +6211,8 @@ static int update_queue(struct call_queue *q, struct member *member, int callcom
 		time(&member->lastcall);
 		member->callcompletedinsl = 0;
 		member->calls++;
-		member->starttime = 0;
 		member->lastqueue = q;
+		did_increment_any = 1;
 		ao2_unlock(q);
 	}
 	/* Member might never experience any direct status change (local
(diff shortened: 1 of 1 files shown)


seanbright:
Please address [these issues](https://github.com/asterisk/asterisk/pull/1741#pullrequestreview-3697941864).

smtcbn:
> Please address [these issues](https://github.com/asterisk/asterisk/pull/1741#pullrequestreview-3697941864).

Hi @seanbright,

Thanks for the review.

I’ve addressed all the items mentioned in the review:
- Squashed the commits into a single commit
- Added the required cherry-pick-to headers
- The commit message now contains the full description of the change

Please let me know if anything else needs adjustment.

smtcbn:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

apps/app_queue.c
```diff
@@ -6127,64 +6127,85 @@ static int wait_our_turn(struct queue_ent *qe, int ringing, enum queue_result *r
...
-	/* It is possible for us to be called when a call has already been considered terminated
-	 * and data updated, so to ensure we only act on the call that the agent is currently in
-	 * we check when the call was bridged.
+	/*
+	 * GLOBAL GATE (minimal, reliable):
+	 * Consume the bridged starttime ONLY ONCE per call using the ORIGINAL member object.
+	 * This blocks duplicate triggers (status + hangup) from counting twice.
 	 */
-	if (!starttime || (member->starttime != starttime)) {
+	if (!__sync_bool_compare_and_swap(&member->starttime, starttime, 0)) {
```
jcolp: I believe it is possible that usage of this will break other existing working platforms, as we specifically have wrappers for other atomic functions to handle that case (while this is not implemented in those for reasons stated in the lock.h header file).

Is there no other way to implement this?

smtcbn: Thanks for the review and for pointing out the portability issues.

I’ve updated the patch to remove the direct use of GCC atomic builtins and instead
gate the update using the existing AO2 locking model.

The logic now checks and consumes member->starttime under ao2_lock(), ensuring that
each bridged call is only counted once while remaining portable across platforms.

Please let me know if this approach looks acceptable.

jcolp (commented):
Before I go through this - was this done using AI? Did you review it manually afterwards if so? The removal of still applicable comments and changing of ones arbitrarily is not good.

smtcbn:
> Before I go through this - was this done using AI? Did you review it manually afterwards if so? The removal of still applicable comments and changing of ones arbitrarily is not good.

No, this wasn’t blindly generated by AI.

I did use AI as a tool while exploring alternatives to atomics, but the final implementation and logic were reviewed and written by me with the existing AO2 locking model and portability constraints in mind.

You are absolutely right about the comments. Some still-applicable comments were removed or altered without sufficient justification, and that was a mistake on my side.

I’ll follow up with a cleanup commit that restores the relevant comments and keeps the change strictly focused on the behavioral fix, not cosmetic edits.

smtcbn:
rerun the PR because of the error meesage is not relavent the changes.

jcolp (approved):
Do you know the specific PR that introduced this originally?

gtjoseph:
@smtcbn You still have 3 commits in this PR than need to be squashed and why shouldn't this be cherry-picked to the 23, 22 and 20 branches?

smtcbn:
Thanks.

I have squashed the commits into a single logical change and force-pushed the branch.

The same code exists in branches 23, 22 and 20, so this fix should be applicable there as well. I can prepare backports if required.

gtjoseph:
> The same code exists in branches 23, 22 and 20, so this fix should be applicable there as well. I can prepare backports if required.

You need to edit the comment with `cherry-pick-to: none`, remove that line and add a new line for each branch.

You also haven't referenced the issue you opened for this so it won't get automatically closed or mentioned in the changelogs.

Once you've done that, you need to make the PR description match the commit message body (without the commit message title line).

See:

asteriskteam:

asteriskteam:

PR #1751, opened 2026-02-03
res_rtp_asterisk: fix audio clicks by using sample-based timestamps

### LINKED ISSUE #204: [bug]:
### Severity

Minor

### Versions

20.3, 20.1, 18.18

### Components/Modules

res_rtp_asterisk

### Operating Environment

CentOS 7.9

### Frequency of Occurrence

None

### Issue Description

We see that Asterisk is transmitting short RTP packets at the end of sound files being played.  These packets are shorter than 160ms in timestamp length and smaller payload.  These short packets manifest as clicking and popping of varying degrees in some remote call platforms that seem to be "filling in" the remainder of the short packet's coverage of 160ms.

From JIRA issue ASTERISK-30120 (https://issues.asterisk.org/jira/browse/ASTERISK-30120).
RTP Timestamp issue causes clicks & pops on calls routed the touch external call platforms.

This means that a short frame following a typical frame should have a timestamp that is a typical frame offset in time, it is the frame after the short frame that should have a shorter timestamp offset. However this does not seem to be the case currently.

With a shortened packet, the wrong value will be written.

### Relevant log output

```shell
From original report

RFC-3550
5.1 of the standard is direct:
The timestamp reflects the sampling instant of the first octet in the RTP data packet

This means that a short frame following a typical frame should have a timestamp that is a typical frame offset in time, it is the frame after the short frame that should have a shorter timestamp offset.2. In the discussion on jitter calculation, it explicitly refers to the need for the RTP timestamp to be like this:

Because the jitter calculation is based on the RTP timestamp which represents the instant when the first data in the packet was sampled, any variation in the delay between that sampling instant and the time the packet is transmitted will affect the resulting jitter that is calculated. Such a variation in delay would occur for audio packets of varying duration.

Appendix B changes since RFC-1889 refers to the fact this area has been reviewed and clarified:
The meaning of “sampling instant” was explained in more detail as part of the definition of the timestamp field of the RTP header in Section 5.1.

Further note though, the Henning’s Schulzrinne’s own FAQ (lead author) for RTP says it is OK to have short frames. He doesn’t expand on timestamp requirements there though.
```

### Asterisk Issue Guidelines

nrjchnd:
RTP timestamps were being calculated based on wall-clock time, causing clicks/pops when frames didn't align with the system clock. This patch uses sample-based increments for minor skews while forcing a clock resync and setting the marker bit for significant jumps (transfers/unholds) to ensure bridge stability.

Ref: ASTERISK-30120

Fixes: #204

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -5192,20 +5192,31 @@ static int rtp_raw_write(struct ast_rtp_instance *instance, struct ast_frame *fr
 #endif
 
 	if (frame->frametype == AST_FRAME_VOICE) {
+		/* Calculate what the next TS should be based strictly on samples */
 		pred = rtp->lastts + frame->samples;
 
-		/* Re-calculate last TS */
-		rtp->lastts = rtp->lastts + ms * rate;
 		if (ast_tvzero(frame->delivery)) {
-			/* If this isn't an absolute delivery time, Check if it is close to our prediction,
-			   and if so, go with our prediction */
-			if (abs((int)rtp->lastts - pred) < MAX_TIMESTAMP_SKEW) {
+			unsigned int time_based_ts = rtp->lastts + (ms * rate);
+			int delta = abs((int)time_based_ts - (int)pred);
+
+			if (delta < MAX_TIMESTAMP_SKEW) {
+				/* The difference is small; likely just jitter.
+				 * Use sample-based TS to prevent clicks. */
 				rtp->lastts = pred;
+			} else if (delta > (rate * 2)) {
+				/* The jump is huge (e.g., > 2 seconds).
+				 * This is likely a transfer or bridge change.
+				 * Reset to time-based TS and set the mark bit to signal a new stream. */
+				rtp->lastts = time_based_ts;
+				mark = 1;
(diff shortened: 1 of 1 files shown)


nrjchnd:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

nrjchnd:
Original report: https://issues-archive.asterisk.org/ASTERISK-30120

jcolp:
Please don't tag people. They will look if they wish to and when they have time.

jcolp (changes requested):
This is a fairly substantial change despite the size of the change and as a result I'm extremely hesitant on outright changing it, because I have a gut feeling (probably based on my past experience in timestamps and marker bits causing a mess) that by changing it and fixing it for the filed issue it's going to end up breaking other equipment. While I'm not a huge fan of adding an additional option I think having an option in rtp.conf to change timestamp behavior and not alter existing behavior would be acceptable.

nrjchnd:
🫡  
🚧 
will change approach to comply  and resubmit.

asteriskteam:

asteriskteam:

PR #1753, opened 2026-02-03
file.c: Convert RTP timestamp deltas to millisecond delays

AryanM200:
file.c: Convert RTP timestamp deltas to millisecond delays

Convert RTP timestamp deltas to millisecond-based delays when scheduling video frame playback.

This change updates the video playback scheduling logic to translate RTP timestamp deltas into millisecond-based delays before frames are scheduled for playback. Using millisecond-based timing ensures more consistent and accurate pacing of video frames across different playback scenarios.

This behavior applies to video playback paths. The updated scheduling logic helps prevent timing drift and improves overall video playback reliability regardless of the source or context in which video frames are played.

### DIFF
--- a/main/file.c
+++ b/main/file.c
@@ -1052,7 +1052,44 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
 	}
 
 	if (whennext != s->lasttimeout) {
-		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), whennext / (ast_format_get_sample_rate(s->fmt->format) / 1000), ast_fsread_video, s));
+		int delta;
+		int rtp_clock;
+		int delta_ms;
+		int delay_ms;
+
+		/*
+ 		 * Reason for this change:
+ 		 * After recording a video voicemail message, the playback was observed
+ 		 * to be very slow. This modification adjusts the frame scheduling logic
+		 * to correct the playback speed.
+ 		 *
+ 		 * Logic:
+ 		 * The code calculates the time difference between the next RTP video frame
+ 		 * timestamp and the previously scheduled timestamp using the RTP clock rate
+ 		 * (9 kHz). This difference is converted into milliseconds to determine the
+ 		 * appropriate delay before reading the next video frame.
+ 		 *
+ 		 * If the calculated delay is less than 60 ms, a minimum delay of 60 ms is
+ 		 * enforced. Otherwise, a delay of 70 ms is applied. The next video frame
+ 		 * read is then scheduled using the channel scheduler. The video stream ID
+ 		 * for the channel is updated, and the last scheduled RTP timestamp is stored
+ 		 * for use in the next iteration.
+ 		 *
+ 		 * The function returns success, indicating that no additional scheduling
+ 		 * action is required at this stage.
+ 		 */
+
+		delta = whennext - s->lasttimeout;
+		rtp_clock = 9000;
+		delta_ms = (delta * 1000) / rtp_clock;
+
+		if (delta_ms < 60) {
+			delay_ms = 60;
+		} else {
+			delay_ms = 70;
+		}
+
+		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), delay_ms, ast_fsread_video, s));
 		s->lasttimeout = whennext;
 		return FSREAD_SUCCESS_NOSCHED;
 	}


jcolp:
The first line no longer complies to commit message guidelines.

AryanM200:
@jcolp is it correct commit message?

AryanM200:
@jcolp Is there any issue in the PR?

jcolp:
Please stop asking and tagging me. When it's reviewed then you'll see. Asking me just pushes it towards the top of my queue, and I generally work oldest to newest updated.

jcolp (approved):

main/file.c
```diff
@@ -1052,11 +1052,38 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+		rtp_clock = 90000;
+		delta_ms = (delta * 1000) / rtp_clock;
+
+		if (delta_ms < 60) {
+			delay_ms = 60;
+		} else {
+			delay_ms = 70;
+		}
+
+		ast_channel_vstreamid_set(s->owner,ast_sched_add(ast_channel_sched(s->owner),delay_ms,ast_fsread_video,s));
```
seanbright: Add a space after each `,`

main/file.c
```diff
@@ -1052,11 +1052,38 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
 		s->lasttimeout = whennext;
 		return FSREAD_SUCCESS_NOSCHED;
 	}
-
```
seanbright, line 1059: This whitespace change is unnecessary, please revert it.

main/file.c
```diff
@@ -1052,11 +1052,38 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+		 * This code calculates the time difference between the next RTP video frame
+		 * and the previously scheduled frame using the RTP clock rate (90 kHz),
+		 * converts that difference into milliseconds, and determines an appropriate
+		 * delay for scheduling the next video frame read. If the calculated delay is
+		 * less than 60 ms, a minimum delay of 60 ms is enforced; otherwise, a delay
+		 * of 70 ms is used. The next video frame read is then scheduled using the
+		 * channel scheduler, the video stream ID is updated for the channel, and
+		 * the last scheduled RTP timestamp is stored for the next iteration. The
+		 * function returns a success status indicating that no additional scheduling
+		 * is required at this point.
```
seanbright: This comment describe what this code is doing but not _why_ it's doing it. Please update the comment to explain the rationale behind the numbers/calculations being used. Alternatively, just delete the comment altogether as it adds no value.

seanbright (changes requested):

main/file.c
```diff
@@ -1052,7 +1052,44 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+		/*
+ 		 * Reason for this change:
+ 		 * After recording a video voicemail message, the playback was observed
+ 		 * to be very slow. This modification adjusts the frame scheduling logic
+		 * to correct the playback speed.
+ 		 *
+ 		 * Logic:
+ 		 * The code calculates the time difference between the next RTP video frame
+ 		 * timestamp and the previously scheduled timestamp using the RTP clock rate
+ 		 * (90 kHz). This difference is converted into milliseconds to determine the
```
seanbright: Do all video formats use a sample rate of 90kHz?

main/file.c
```diff
@@ -1052,7 +1052,44 @@ static enum fsread_res ast_readvideo_callback(struct ast_filestream *s)
...
+		rtp_clock = 90000;
+		delta_ms = (delta * 1000) / rtp_clock;
+
+		if (delta_ms < 60) {
+			delay_ms = 60;
+		} else {
+			delay_ms = 70;
+		}
+
+		ast_channel_vstreamid_set(s->owner, ast_sched_add(ast_channel_sched(s->owner), delay_ms,ast_fsread_video, s));
```
seanbright: There needs to be a space after `delay_ms,` and before `ast_fsread_video`

seanbright:
FWIW - I've tested this patch with an h263 file I recorded with Bria 4 & Asterisk and the playback, while faster than without this patch, is still too slow.

tmittelstaedt:
Couple of things on this to keep in mind when testing.  First, this patch was originally created due to problems playing back video files that were recorded to the voicemail.  When you call an extension with a video enabled phone, like your Bria 4 softphone, and you have video turned on, and there is no answer and you get a voicemail prompt for the recipient extension, then you record a message, it will record both the audio and the video from your phone.  Playback of that message is the problem - without this patch, Asterisk takes so long to seek to the end of the video file, it is many minutes after the audio file portion of the voicemail has been played, as a result, the listener never hears the "press 3 for advanced options, press 5 to repeat press 7 to delete, etc. etc." and it sounds to them like the system has hung.  With this patch, immediately after the audio completes playing, the usual audio menu is then presented to the user because Asterisk has finished sending the video frames to the phone listening in.  That way the audio voicemail can be properly dealt with.  Incidentally this was first noted on h.264 not h.263.  It is good to hear that it also helps with h.263 but possibly you never noticed voicemailbox breakage because Bria handles h.263 differently.  I do not know, I am just guessing.

Now we get to the video playback side of things you are noticing.  For starters, you can adjust the speed frames are played back by playing with the 60, changing it to 30 and so on.  However, there is no codec negotiation with this patch - there isn't intended to be since the voicemail system does not do that.

When the voicemail system records a video VM, it does it using the codec settings on that phone - not negotiated settings with that phone and the recipient phone.  For example, you can establish a call between Linphone, with the Cisco H.264 codec installed, and a Cisco CP-8845 phone which uses the H.264 codec, and video will work fine because the codec properly negotiates since the negotiation is not happening inside of Asterisk.  However, if you save a video voicemail with linphone to an extension it will be saved as H.264 and then retrieve that voicemail with the same CP-8845, the video voicemail will NOT be played back you will just get a black screen, because the CP-8845 cannot negotiate H.264 codec settings since there's nothing to negotiate to.  Asterisk will not transcode to the phones video codec it just dumps the file as-is.   The only time the video WILL be played back is if the video voicemail is saved with the same make and model phone.  So, you saved a video voicemail with your Bria 4 running H.263 codec and you were able to play it back, yes, even though it was slow.  But if you tried retrieving that video voicemail with, say, Linphone which does not support that codec - it would NOT play because Asterisk does not transcode video voicemails to a codec that Linphone DOES support.  In fact, if you SCP the saved .h263 or h.264 file off the Asterisk system to your PC and ran a video identifier program like MediaInfo on it, it wouldn't even show at all as a video file because it's not encapsulated as one, it's just a raw video file. 

For you to properly test you must be able to record an h.263 file from your Bria phone, then SCP the file to your PC and then play it back normally.  Or at least, examine it to even see if it was recorded at 60fps.  It may not have been.  Since when recording the video voicemail no codec negotiation happened we don't know what fps was used by your softphone it may have been a very low fps as a default.

Third, whether or not you could even display 30fps or 60fps video at all from a saved voicemail is dependent on the disk speed in the Asterisk PBX - and the bandwidth setting (for example did you even think to raise the video bandwidth to 2500 kbps which is needed for 1080p HD in chan_pjsip since the default is 384 kbps) as well as speed of the underlying network.  In testing this on my local net, playback is much faster, in testing this on my remote net that's 80 miles away from the PBX, tied to the PBX via VPN, playback of video is far slower to the point of being unusable.  Audio playback is fine with both setups with only a few ms dropouts on the audio noticed here and there over the VPN.  My PBXes are virtualized so disk speed is terrible but audio works fine since it's so much smaller than video.

In summary, videophone to videophone calls already work on an Asterisk system because the phones themselves are negotiating the bandwidth and frame rate and the rest of it.  But recording voicemail greetings from a videophone breaks the voicemailbox playback, and recording voicemails from another videophone to an extension also breaks voicemailbox playback and has since this scheduling code was broken in Asterisk 13.  This patch corrects that so voicemail now works, but does NOT attempt to address all issues with videophones and voicemail.

main/file.c
AryanM200: yes, it works with all video formats

AryanM200:
cherry-pick-to: none

main/file.c
AryanM200: > Do all video formats use a sample rate of 90kHz?

based on source code  all video codecs already supported 90khz

AryanM200:
https://github.com/asterisk/asterisk/pull/1753#:~:text=PRCheck%20/%20run%2Dcheck%20/%20Check%20/%20Gate%20(pjs5)%20/%20master%2Dpjs5%20(pull_request_target)

Since this PR check has failed, could it cause any issues during the approval process? or we can ignore it

jcolp:
The failure is unrelated.

jcolp:
Are there any other individuals who can test this change and provide feedback?

gareth-palmer:
I don't think this change should be merged.

Instead, I think the issue is being caused because the `sample_rate` for the video codec is incorrectly specified in `codec_builtin.c`, it should be 90000 not 1000. That causes the sched time calculation to be wrong.

Changing that made the voicemail options start playing right after the message had finished.

Branch commit showing the changes: https://github.com/gareth-palmer/asterisk/commit/271ee8c3c05cc37c6d2eca617cb35c1c72e2c2f3

jcolp:
I believe it originally was 90000 and I changed it to fix another issue, so changing it back is just going to result in another problem coming back:

    rtp: Preserve timestamps on video frames.

    Currently when receiving video over RTP we store only
    a calculated samples on the frame. When starting the video
    it can take some time for this calculation to actually yield
    a value as it requires constant changing timestamps. As well
    if a video frame passes over multiple RTP packets this calculation
    will fail as the timestamp is the same as the previous RTP
    packet and the number of samples calculated will be 0.

    This change preserves the timestamp on the frame and allows
    it to pass through the core. When sending the video this timestamp
    is used instead of a new one being calculated.

Specifically on large packets you'd encounter the "multiple RTP packets" scenario so video would not work during that. It's changing one problem for another.

tmittelstaedt:
Joshua I sent an email on the 18th to @alcanzar asking him to test, he made a post on this issue 10 years ago but never followed through with submitting a patch, however he has apparently gone silent, now.  (the last change he made on Github was 5 years ago on https://github.com/alcanzar/logstash-filter-augment)

Posting on the Asterisk mailing list got no takers for testing, either.  I can try asking on the FreePBX mailing list but I think we won't find any testers there, either.  Absent a complete rewrite of video handling in the voicemail system in Asterisk, the only other option to this change would be to put in another option to Asterisk that disabled video recording completely, and it seems pointless to put in a user-selectable option disabling a section of code that isn't working right in the first place.

Can I ask you, if someone had tested this and identified the problem with the voicemail recording from video phones immediately after you had made the change in codec_builtin.c to fix the "multiple RTP packets" issue, would you have committed that change anyway, or rewritten it, or committed that change plus a change here in file.c?

Nowadays it seems most people running desk videophones are doing so in conjunction with door camera phones, video through the PBX is a border case setup now.  Few to no sites are running 100% or close to 100% desk videophones, and sites running softphones would be also using them in conjunction with Zoom or Teams videoconferencing or Facetime video calls.  This miscalculation of the scheduled time is a usability fix for Asterisk voicemail, it's exactly the kind of bug that hits niche setups and those new to Asterisk who would really struggle with even identifying it's a bug in the PBX in the first place.  It's also most likely to be visible and embarrassing when it does hit, since orgs that do experiment with buying videophones that are likely to be used to record voicemails are likely going to only buy them for the top brass in the org.

jcolp:
If an issue was brought up at the time it most likely would have gotten looked at then.

tmittelstaedt:
I'm also sure it would have - but it wasn't brought up then.  Thus, a fix went in that fixed the "multiple RTP packets" problem but created the "broken voicemail if greeting recorded with a videophone" problem.

What about modifying this patch to change 2 files, main/file.c and codec_builtin.c?

We could possibly increase sample_rate from 1000 up to maybe 10000 in codec_builtin.c and maybe change
delta_ms = (delta * 10000 / rtp_clock); in main/file.c or do some other playing around with the variables?

The playback speed issue of video voicemails is a cosmetic issue.  The broken voicemail problem is structural.  Let's fix the structure, since any voicemail left from an incoming call from the PSTN is not going to have a video component thus playback speed of video does not apply.  However, modifying a greeting from a videophone definitely causes problems in voicemail playback.

What do you think?

jcolp:
I don't know. I can only answer surface level things, as this is not something I'm working on and not something I'm digging into to investigate at this time. Such a change may work, or may still suffer from other problems.

jcolp:
Also you keep saying voicemail. This isn't voicemail specific. It's file recording/playback. You just see it in voicemail because that's what you're doing.

jcolp:
From a higher level perspective altering codec_builtin can have an effect on timestamps of video packets passed through Asterisk, which is done in WebRTC. Such a change could cause video to fail/stutter/freeze with WebRTC, which I would expect is the vast usage of video in Asterisk these days.

tmittelstaedt:
Then we are back to the issue which is the change that you made to fix the timestamps of video packets passed through Asterisk, and fix possible multiple RTP packets in video streams, and possible stuttering with WebRTC, has caused a problem in the voicemail portion of Asterisk when videophones are leaving messages and greetings and regular phones are getting messages and greetings.

If the only thing that is stopping you from approving this commit into asterisk:master is the fear it would harm WebRTC, then would you be satisfied with a test of this patch and a WebRTC client such as https://github.com/InnovateAsterisk/Browser-Phone or https://www.mizu-voip.com/Software/WebPhone.aspx ?

jcolp:
My comment was not regarding the commit on this PR, I was referring to comment on the codec_builtin change. I have not included the commit in this PR because Sean did testing and showed that it did not completely solve it for him, so I wanted further data points as to how much this does actually solve the situation on a wider group of people.

tmittelstaedt:
After Sean tested, I asked him to test and comment on the voicemail problem he did not do so, his comment was only regarding playback speed of video recorded with voicemails - which is frankly a cosmetic issue and does not affect usability of voicemails stored under Asterisk.  I would guess Sean isn't interested in the voicemail aspect and so lost interest when the playback speed was not fixed.

The playback speed is dependent on what is going on inside of codec_builtin.  When testing changes to sample_rate inside of codec_builtin, we have observed playback speeds too fast.

It may be that accurate playback of recorded video voicemail messages is not possible by merely making changes to delta_ms but instead by changing both, make sample_rate also to a value that is higher but not so high as to cause problems with WebRTC then modify delta_ms.

Do we even have a wide group of beta testers for new Asterisk versions?

jcolp:
There isn't an explicit beta test group. We make release candidates available and encourage people to test them and report issues.

tmittelstaedt:
Then if we expanded the patch to modify all files necessary to get the mis-synchronizing on the voicemail fixed and the slow playback on video voicemails fixed and it tested out fine with WebRTC would it be committed into master?

jcolp:
Assuming it passed code review, there is a high chance, yes.

seanbright:
> After Sean tested, I asked him to test and comment on the voicemail problem he did not do so

I apologize - you sent a wall of text which I didn’t read to see that there was something more you were asking me to do.

[I tested already and reported my results.](https://github.com/asterisk/asterisk/pull/1753#issuecomment-3899007593)

My test results were not intended as a rejection of the change, they were just a data point.

tmittelstaedt:
Thanks, all. I'll add a webRTC client into my phone test lab and see what we can do.

gareth-palmer:
I will have a look at the zero-samples issue caused by getting the same timestamp.

When calculating the number of samples for audio codecs the code in `res/res_rtp_asterisk.c` uses a `get_samples` callback.

Perhaps an equivalent can be added for the video codecs, falling back to using the timestamp if no callback is present.

AryanM200:
If the sample rate is modified in codec_builtin.c, is that the correct approach, or should another location be considered for this change? Also, shouldn’t this change avoid affecting normal video calls?”

tmittelstaedt:
> I don't think this change should be merged.
> 
> Instead, I think the issue is being caused because the `sample_rate` for the video codec is incorrectly specified in `codec_builtin.c`, it should be 90000 not 1000. That causes the sched time calculation to be wrong.
> 
> Changing that made the voicemail options start playing right after the message had finished.
> 

Gareth, I can't get video to work at all in the voicemail system when using your modified chan_sip that is part of the USECALLMANAGER patch, at least not under FreePBX with that patch applied to Asterisk.  In testing with Cisco videophones configured on that driver, even with video set to yes in the extension definition, while the phones can make video calls to each other, any messages left in the voicemail system by videophones is only left as an audio message.  It seems only that when phones are used with chan_pjsip that they can leave messages and greeting messages with both audio and video in the voicemail system.  I'm not sure if there's a configuration option that FreePBX sets that is global for video that applies to that driver, though. perhaps there is?

gareth-palmer:
@tmittelstaedt If you can make video calls between two phones in chan_sip then video is enabled in the driver.

I do not know why the video stream is not being recorded as well. Try creating a an extension that executes the `Record()` application and see if that saves the video stream.

gareth-palmer:
I have a branch with working video playback now:
https://github.com/gareth-palmer/asterisk/tree/video-playback

The important changes to note are that frames without the marker bit are written immediately with only the end-frame causing `ast_readvideo_callback` to be rescheduled and RTP timestamps are now preserved for each video frame that spans multiple packets.

Currently the playback speed is hard-coded to 30FPS, it may be possible to have `h264_write` to calculate the actual FPS based on the timestamp. **Note** the on-disk format has changed so existing recordings won't work.

I ran into a few issues when testing with Cisco phones using the Enterprise firmware where if it doesn't receive any video then the video it sends is invalid so by the time `Voicemail()` starts recording the actual video being sent is useless. Also, when `transmit_silence` is enable it also causes video to be just a blank.

So testing was done with the following dial-plan:

```
exten => 100,1,Record(${ASTSPOOLDIR}/recording/video-test.wav,,,kq)
same => next,Hangup()

exten => 200,1,Playback(${ASTSPOOLDIR}/recording/video-test)
same => next,Hangup()
```

tmittelstaedt:
Thanks, Gareth, I'll test

AryanM200:
> I have a branch with working video playback now: https://github.com/gareth-palmer/asterisk/tree/video-playback
> 
> The important changes to note are that frames without the marker bit are written immediately with only the end-frame causing `ast_readvideo_callback` to be rescheduled and RTP timestamps are now preserved for each video frame that spans multiple packets.
> 
> Currently the playback speed is hard-coded to 30FPS, it may be possible to have `h264_write` to calculate the actual FPS based on the timestamp. **Note** the on-disk format has changed so existing recordings won't work.
> 
> I ran into a few issues when testing with Cisco phones using the Enterprise firmware where if it doesn't receive any video then the video it sends is invalid so by the time `Voicemail()` starts recording the actual video being sent is useless. Also, when `transmit_silence` is enable it also causes video to be just a blank.
> 
> So testing was done with the following dial-plan:
> 
> ```
> exten => 100,1,Record(${ASTSPOOLDIR}/recording/video-test.wav,,,kq)
> same => next,Hangup()
> 
> exten => 200,1,Playback(${ASTSPOOLDIR}/recording/video-test)
> same => next,Hangup()
> ```

@gareth-palmer , 

I tested the code, and during video playback it seems like the video speed is faster than normal, while the audio is working properly.

gareth-palmer:
Link to the PR for my changes is below. Video should playback at the correct rate now.
https://github.com/asterisk/asterisk/pull/1909

AryanM200:
Hi @gareth-palmer ,

I tested your code and it is working. The video speed issue is fixed but During testing, in random scenarios the video playback is not working.

My code changes are also working. You have made changes in 5 files, while I made changes in 1 file.

Please let us know which solution would be better suited for the PR merging process.

gareth-palmer:
@AryanM200 Thank you for testing my changes.

You are correct that I have made changes to 5 files as opposed to only 1 for you change, though I am not sure of the relevance of this.

However, it is incorrect to state that your changes "are also working". I have tested your change and the playback of the stream does not finish at the correct time and video playback does not work at all.

You didn't state which scenarios video playback is not working, as noted above the only one likely to work is a simple `Playback()` extension that plays the stream immediately after answering.

This is likely due to Asterisk not being able to send an empty video stream along with the audio causing the phone to give-up on playback. That is outside the current scope of this change.

AryanM200:
> @AryanM200 Thank you for testing my changes.
> 
> You are correct that I have made changes to 5 files as opposed to only 1 for you change, though I am not sure of the relevance of this.
> 
> However, it is incorrect to state that your changes "are also working". I have tested your change and the playback of the stream does not finish at the correct time and video playback does not work at all.
> 
> You didn't state which scenarios video playback is not working, as noted above the only one likely to work is a simple `Playback()` extension that plays the stream immediately after answering.
> 
> This is likely due to Asterisk not being able to send an empty video stream along with the audio causing the phone to give-up on playback. That is outside the current scope of this change.

@gareth-palmer , I have checked my changes in the code and also tested the implementation in different scenarios on my side, and it seems to be working properly during testing. Playback is completing correctly, and video voicemail playback also seems to be working in my environment.

I tested multiple call flows and playback scenarios, including video voicemail testing, and did not observe the issue consistently.

gareth-palmer:
@AryanM200 
To play video at 30 frames per second in sync with the audio stream the delay between full frames needs to be 33ms (`1s / 30fps = 33ms`). Your change uses either 60ms or 70ms for some reason.

[This comment](https://github.com/asterisk/asterisk/pull/1753#issuecomment-3899007593) also notes that playback was too slow when testing your change.

What audio codec, video codec and device are you using to test? I have tested it with G722 and H264 on a Cisco 8845 using firmware version 14.4.1.

AryanM200:
> > @AryanM200 Thank you for testing my changes.
> > You are correct that I have made changes to 5 files as opposed to only 1 for you change, though I am not sure of the relevance of this.
> > However, it is incorrect to state that your changes "are also working". I have tested your change and the playback of the stream does not finish at the correct time and video playback does not work at all.
> > You didn't state which scenarios video playback is not working, as noted above the only one likely to work is a simple `Playback()` extension that plays the stream immediately after answering.
> > This is likely due to Asterisk not being able to send an empty video stream along with the audio causing the phone to give-up on playback. That is outside the current scope of this change.
> 
> @gareth-palmer , I have checked my changes in the code and also tested the implementation in different scenarios on my side, and it seems to be working properly during testing. Playback is completing correctly, and video voicemail playback also seems to be working in my environment.
> 
> I tested multiple call flows and playback scenarios, including video voicemail testing, and did not observe the issue consistently.

@gareth-palmer , 

We have used audio codecs such as PCMA, PCMU, and G.722, and video codecs including H.264 and VP8. Testing was performed on Linphone, MicroSIP, and WebRTC call. We also changed the RTP clock rate from 90 kHz to 9 kHz, and during testing, the playback was completed within the correct timeframe.

gareth-palmer:
@AryanM200 You had video playback working with VP8?

AryanM200:
@gareth-palmer ,
Does vp8 supported file playback?

gareth-palmer:
@AryanM200 Asterisk doesn't support vp8 playback, which is why it was odd that you mentioned testing with it.

@tmittelstaedt Have you had a chance to test both this and my changes?

AryanM200:
@gareth-palmer ,We are already aware that VP8 does not support playback. We have check and testing this during video call testing using the VP8 codec.

asteriskteam:

asteriskteam:

Issue #1755, opened 2026-02-04
[bug]: app_dial, utils.h: Compilation failure with -Wold-style-declaration and -Wdiscarded-qualifiers


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

app_dial, utils.h

### Operating Environment

Debian 13, gcc 14.2.0

### Frequency of Occurrence

Constant

### Issue Description

See log output:

### Relevant log output

```shell
In file included from /usr/src/asterisk/asterisk/include/asterisk/alertpipe.h:22,
                 from /usr/src/asterisk/asterisk/include/asterisk/channel.h:125,
                 from app_dial.c:47:
/usr/src/asterisk/asterisk/include/asterisk/utils.h:772:1: warning: ‘inline’ is not at beginning of declaration [-Wold-style-declaration]
  772 | static void force_inline _ast_assert(int condition, const char *condition_str, const char *file, int line, const char *function)
      | ^~~~~~

app_dial.c:976:22: warning: assignment discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  976 |                 tech = "Local";
```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1756, opened 2026-02-04
app_dial, utils.h: Avoid old style declaration and discarded qualifier.

InterLinked1:
* app_dial: Use const char* for fixed strings.
* utils.h: inline should come before return type.

Resolves: #1755


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

PR #1759, opened 2026-02-05
GHSA 2026 1

gtjoseph:
- **xml.c: Replace XML_PARSE_NOENT with XML_PARSE_NONET for xmlReadFile.**
  The xmlReadFile XML_PARSE_NOENT flag, which allows parsing of external
  entities, could allow a potential XXE injection attack.  Replacing it with
  XML_PARSE_NONET, which prevents network access, is safer.

- **ast_coredumper: check ast_debug_tools.conf permissions**
  Prevent ast_coredumper from using ast_debug_tools.conf files that are
  not owned by root or are writable by other users or groups.
  
  Prevent ast_logescalator and ast_loggrabber from doing the same if
  they are run as root.
  
  UserNote: ast_debug_tools.conf must be owned by root and not be
  writable by other users or groups to be used by ast_coredumper or
  by ast_logescalator or ast_loggrabber when run as root.
  
- **http.c: Change httpstatus to default disabled and sanitize output.**
  To address potential security issues, the httpstatus page is now disabled
  by default and the echoed query string and cookie output is html-escaped.
  
  UpgradeNote: To prevent possible security issues, the `/httpstatus` page
  served by the internal web server is now disabled by default.  To explicitly
  enable it, set `enable_status=yes` in http.conf.
  
- **ast_coredumper: create gdbinit file with restrictive permissions**
  Modify gdbinit to use the install command with explicit permissions (-m 600)
  when creating the .ast_coredumper.gdbinit file. This ensures the file is
  created with restricted permissions (readable/writable only by the owner)
  to avoid potential privilege escalation.
  
Resolves: #GHSA-rvch-3jmx-3jf3
Resolves: #GHSA-85x7-54wr-vh42
Resolves: #GHSA-v6hp-wh3r-cwxh
Resolves: #GHSA-xpc6-x892-v83c

### DIFF
--- a/contrib/scripts/ast_coredumper
+++ b/contrib/scripts/ast_coredumper
@@ -115,6 +115,7 @@ if $RUNNING ; then
 		if ${DRY_RUN} ; then
 			echo "Would run: ${GDB} -p $MAIN_PID -q --batch --ex gcore $cf"
 		else
+			install -m 0600 /dev/null "$cf"
 			${GDB} -p "$MAIN_PID" -q --batch --ex "gcore $cf" >/dev/null 2>&1
 		fi
 		echo "$(S_COR "${DRY_RUN}" 'Simulated dump' 'Dump') is complete."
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23
multiple-commits: standalone

PR #1765, opened 2026-02-10
fix: resolve compilation warnings in utils.h and app_dial.c

dmzoneill:
REDACTED


silentindark:
@dmzoneill same https://github.com/asterisk/asterisk/pull/1756

jcolp:
Marking as closed as there is already a PR up.

PR #1766, opened 2026-02-10
endpoints: Allow access to latest snapshot directly.

jcolp:
This change adds an API call to allow direct access to the latest snapshot of an ast_endpoint. This is then used by chan_pjsip when calculating device state, eliminating the need to access the cache which would incur a container find and access.


jcolp:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

mbradeen (approved):

PR #1767, opened 2026-02-10
fix: correctly track queue member pause time in realtime mode

dmzoneill:
REDACTED

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -3826,8 +3826,12 @@ static void rt_handle_member_record(struct call_queue *q, char *category, struct
 			ast_copy_string(m->rt_uniqueid, rt_uniqueid, sizeof(m->rt_uniqueid));
 			if (paused_str) {
 				m->paused = paused;
-				if (paused && m->lastpause == 0) {
-					time(&m->lastpause); /* XXX: Should this come from realtime? */
+				if (paused) {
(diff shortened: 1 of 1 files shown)


jcolp:
We will begin enforcing this AI policy soon, you will be required to follow it:

https://github.com/asterisk/documentation/pull/176/files#diff-9e6ba8b594f901c1bb1ec1e1a9ee73765120f057493f9faa9c5877cb23d9c828

jcolp:
> Testing
> 
> The fix follows the logic suggested in the issue and maintains backward compatibility while correctly handling the realtime member state transitions.

There has been no statement that any testing has actually been done. The original reporter also stated they believed this was incomplete and there were other time related issues, so was anything done about that?

What are you exactly doing with your AI issue?

jcolp:
> @jcolp hi, whats the issue? i use auto generated commit messages all the time.. I've never had anyone complain about it. infact people appreciate it a lot as it helps establish good context. thanks

1. We have specific commit message guidelines
2. If testing is stated in the message then there is an expectation that it talks about testing, which from what I can tell there was none in this - if there was none then it should be stated explicitly
3. It does not appear as though a complete investigation was done to cover time related things, merely the possible fix from the reporter was turned into a PR which could very well be incomplete

PR #1768, opened 2026-02-10
chan_websocket: Add media direction.

bkford:
Currently, WebSockets both accept and send media without the option to
disable one or the other. This commit adds the ability to set the media
direction for a WebSocket, making it unidirectional or bidirectional
(the default). Direction is done from the point of view of the
application, NOT Asterisk. The allowed values are 'both', 'in', and
'out'. If media direction is 'both' (the default), Asterisk accepts and
sends media to the application. If it is 'in', Asterisk will drop any
media received from the application. If it is 'out', Asterisk will not
write any media frames to the application.

UserNote: WebSocket now supports media direction, allowing for
unidirectional media. This is done from the perspective of the
application and can be set via channel origination, external media, or
commands sent from the application. Check out
https://docs.asterisk.org/Configuration/Channel-Drivers/WebSocket/ for
more.

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -70,6 +70,19 @@ struct webchan_conf_global {
 	enum webchan_control_msg_format control_msg_format;
 };
 
+/* This is from the perspective of the app, NOT Asterisk */
+enum webchan_media_direction {
+	WEBCHAN_MEDIA_DIRECTION_BOTH,
+	WEBCHAN_MEDIA_DIRECTION_OUT,
+	WEBCHAN_MEDIA_DIRECTION_IN,
+};
+
+static const char *websocket_media_direction_map[] = {
+	[WEBCHAN_MEDIA_DIRECTION_BOTH] = "both",
+	[WEBCHAN_MEDIA_DIRECTION_OUT] = "out",
+	[WEBCHAN_MEDIA_DIRECTION_IN] = "in",
+};
+
 static struct ast_websocket_server *ast_ws_server;
 
 static struct ao2_container *instances = NULL;
@@ -102,6 +115,7 @@ struct websocket_pvt {
 	int frame_queue_length;
 	int queue_full;
 	int queue_paused;
+	int media_direction;
 	char connection_id[0];
 };
 
@@ -126,6 +140,7 @@ struct websocket_pvt {
 #define REPORT_QUEUE_DRAINED "REPORT_QUEUE_DRAINED"
 #define PAUSE_MEDIA "PAUSE_MEDIA"
 #define CONTINUE_MEDIA "CONTINUE_MEDIA"
+#define SET_MEDIA_DIRECTION "SET_MEDIA_DIRECTION"
 
 #define QUEUE_LENGTH_MAX 1000
 #define QUEUE_LENGTH_XOFF_LEVEL 900
@@ -142,6 +157,7 @@ static struct ast_frame *webchan_read(struct ast_channel *ast);
 static int webchan_write(struct ast_channel *ast, struct ast_frame *f);
 static int webchan_hangup(struct ast_channel *ast);
 static int webchan_send_dtmf_text(struct ast_channel *ast, char digit, unsigned int duration);
+static int set_channel_timer(struct websocket_pvt *instance);
 
 #define websocket_request_hangup(_instance, _cause, _tech) \
 	_websocket_request_hangup(_instance, _cause, _tech, __LINE__, __FUNCTION__)
@@ -746,6 +762,27 @@ static int queue_option_frame(struct websocket_pvt *instance,
 	return 0;
 }
 
+#define ERROR_ON_PASSTHROUGH_MODE_RTN(instance, command) \
+({ \
+	if (instance->passthrough) { \
+		send_event(instance, ERROR, "%s not supported in passthrough mode", command); \
+		ast_debug(4, "%s: WebSocket in passthrough mode. Ignoring %s command.\n", \
+			ast_channel_name(instance->channel), command); \
+		return 0; \
+	} \
+})
+
+#define ERROR_ON_INVALID_MEDIA_DIRECTION_RTN(instance, command, direction) \
+({ \
+	if (instance->media_direction == direction) { \
+		send_event(instance, ERROR, "%s not supported while media direction " \
+			"is '%s'", command, websocket_media_direction_map[direction]); \
+		ast_debug(4, "%s: WebSocket media direction is '%s'. Ignoring %s command.\n", \
+			ast_channel_name(instance->channel), websocket_media_direction_map[direction], command); \
+		return 0; \
+	} \
+})
+
 /*!
  * \internal
  * \brief Handle commands from the websocket
@@ -787,12 +824,8 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
 		websocket_request_hangup(instance, AST_CAUSE_NORMAL, AST_WEBSOCKET_STATUS_NORMAL);
 
 	} else if (ast_strings_equal(command, START_MEDIA_BUFFERING)) {
-		if (instance->passthrough) {
-			send_event(instance, ERROR, "%s not supported in passthrough mode", command);
-			ast_debug(4, "%s: WebSocket in passthrough mode. Ignoring %s command.\n",
-				ast_channel_name(instance->channel), command);
-			return 0;
-		}
+		ERROR_ON_PASSTHROUGH_MODE_RTN(instance, command);
+		ERROR_ON_INVALID_MEDIA_DIRECTION_RTN(instance, command, WEBCHAN_MEDIA_DIRECTION_IN);
 		AST_LIST_LOCK(&instance->frame_queue);
 		instance->bulk_media_in_progress = 1;
 		AST_LIST_UNLOCK(&instance->frame_queue);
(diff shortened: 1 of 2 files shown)


bkford:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

mbradeen:
```
For example, let's say that we wanted to change
the media direction to 'out' after the session has already been started.
The following command will change it:

{ "command": "SET_MEDIA_DIRECTION", "direction": "in" }
```
The text example says 'out' but the json says 'in'

channels/chan_websocket.c
```diff
@@ -916,10 +966,93 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
...
+				instance->timer = NULL;
+				ast_channel_internal_fd_clear(instance->channel, WS_TIMER_FDNO);
+				ast_queue_frame(instance->channel, &ast_null_frame);
+			}
+
+			AST_LIST_LOCK(&instance->frame_queue);
+			while ((frame = AST_LIST_REMOVE_HEAD(&instance->frame_queue, frame_list))) {
+				ast_frfree(frame);
+			}
+			instance->frame_queue_length = 0;
```
mbradeen: I have mixed feelings about flushing the queue implicitly when this is set to 'in' via the application.  I would consider setting the mode to 'in' without waiting for the queue to empty or be flushed explicitly to be the application acting inappropriately and should generate a warning.

channels/chan_websocket.c
```diff
@@ -916,10 +966,93 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
...
+
+		} else if (!strcmp("in", direction)) {
+			struct ast_frame *frame = NULL;
+
+			if (instance->media_direction == WEBCHAN_MEDIA_DIRECTION_IN) {
+				return 0;
+			}
+
+			if (instance->timer) {
+				ast_timer_close(instance->timer);
```
gtjoseph, line 977: Closing the timer should probably be done after clearing it from the channel.

channels/chan_websocket.c
gtjoseph: Thinking about it, yeah.  Setting "in" shouldn't automatically flush the "out" direction.  Thinking further, we should probably remove the checks for FLUSH and REPORT_QUEUE_DRAINED.  They don't hurt anything if called when media direction is "in".

channels/chan_websocket.c
```diff
@@ -1597,6 +1744,7 @@ AST_APP_OPTIONS(websocket_options, BEGIN_OPTIONS
 	AST_APP_OPTION_ARG('v', OPT_WS_URI_PARAM, OPT_ARG_WS_URI_PARAM),
 	AST_APP_OPTION('p', OPT_WS_PASSTHROUGH),
 	AST_APP_OPTION_ARG('f', OPT_WS_MSG_FORMAT, OPT_ARG_WS_MSG_FORMAT),
+	AST_APP_OPTION_ARG('m', OPT_WS_MEDIA_DIRECTION, OPT_ARG_WS_MEDIA_DIRECTION),
```
gtjoseph: How about "d" for direction instead?

gtjoseph (dismissed):

channels/chan_websocket.c
bkford: So if I'm understanding correctly, we want:
1. Setting the media direction to "in" (Asterisk to the application) has no effect on what's in the queue from the application already.
2. If the user wants the queue to be flushed, they should do so themselves.
3. By having it set up this way, removing the checks for FLUSH and REPORT_QUEUE_DRAINED makes sense so the user can control it on their end.

Am I on the same page?

channels/chan_websocket.c
bkford: I'm fine with changing it if other things already use 'd' for direction or if it is more intuitive. I was looking at it from the perspective of an m() line, but I'm not sure what would be more natural from a user perspective. Thoughts?

jcolp: "d" is better.

channels/chan_websocket.c
```diff
@@ -916,10 +966,93 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
...
+		}
+
+		direction = (char *)ast_json_object_string_get(json, "direction");
+		if (!direction) {
+			send_event(instance, ERROR, "%s requires a 'direction' parameter.\n", command);
+			return 0;
+		}
+
+		direction = ast_str_to_lower(direction);
+		if (!strcmp("both", direction)) {
```
seanbright: It feels unsafe to me to cast away the `const` and then change the casing of the string that belongs to the json object. Can we just use strcasecmp instead?

Or just ignore it if it's not the correct casing (my preference).

res/ari/resource_channels.c
```diff
@@ -2212,11 +2212,21 @@ static int external_media_websocket(struct ast_ari_channels_external_media_args
...
 	struct varshead *vars;
+	RAII_VAR(char *, direction, NULL, ast_free);
 
-	if (ast_asprintf(&endpoint, "WebSocket/%s%s%s",
+	/* If direction is set here, it WILL override any m() line in transport data
+	 * since it is appended to the end of the string.
+	 */
+	if (args->direction && ast_asprintf(&direction, "m(%s)", args->direction) == -1) {
+		return 1;
+	}
```
seanbright: I would just merge this in to the `ast_asprintf` call on line 2224.

channels/chan_websocket.c
```diff
@@ -916,10 +966,93 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
 				ast_channel_name(instance->channel), command);
 			return 0;
 		}
+		if (instance->media_direction == WEBCHAN_MEDIA_DIRECTION_IN) {
+			send_event(instance, ERROR, "%s not supported while media direction is 'in'", command);
+			ast_debug(4, "%s: WebSocket media direction is 'in'. Ignoring %s command.\n",
+				ast_channel_name(instance->channel), command);
+			return 0;
+		}
```
seanbright: This feels like something we could extract in to a macro or function

channels/chan_websocket.c
bkford: When I discussed case sensitivity for this with the team, I was told to make it case insensitive. If anyone has changed their mind, feel free to chime in. Otherwise I'll change it to a strcasecmp!

gtjoseph: I'm good with case insensitivity and using strcasecmp().  If this were handling hundreds of frames a second, converting it to lower case once using ast_str_to_lower() would probably be better but these are just control messages.

channels/chan_websocket.c
gtjoseph: > Am I on the same page?

Yep.

res/ari/resource_channels.c
gtjoseph: I'd leave it as is.  That asprintf on 2224 is already hard to read.  Extending would mean having to test and add the "m(" (actually, "d(" now) then the value, then the ")".

You could save the memory allocation and free though by ...
```
char direction[strlen(S_OR(args->direction, "")) + 4];
```
then just use sprintf.

seanbright: Or just direction[16] and use snprintf?

gtjoseph: that's fine as well.

gtjoseph (approved):

channels/chan_websocket.c
seanbright: I am not good with case insensitivity.

res/ari/resource_channels.c
```diff
@@ -2212,11 +2212,23 @@ static int external_media_websocket(struct ast_ari_channels_external_media_args
 	char *endpoint;
 	struct ast_channel *chan;
 	struct varshead *vars;
+	char direction[16];
```
seanbright: Initialize this to `""`

res/ari/resource_channels.c
```diff
@@ -2212,11 +2212,23 @@ static int external_media_websocket(struct ast_ari_channels_external_media_args
...
 
-	if (ast_asprintf(&endpoint, "WebSocket/%s%s%s",
+	/* If direction is set here, it WILL override any m() line in transport data
+	 * since it is appended to the end of the string.
+	 */
+	if (args->direction) {
+		snprintf(direction, sizeof(direction), "d(%s)", args->direction);
+	} else {
+		direction[0] = '\0';
```
seanbright: Remove this

res/ari/resource_channels.c
```diff
@@ -2212,11 +2212,23 @@ static int external_media_websocket(struct ast_ari_channels_external_media_args
...
+	}
+
+	if (ast_asprintf(&endpoint, "WebSocket/%s%s%s%s%s",
 			args->external_host,
 			S_COR(args->transport_data, "/", ""),
-			S_OR(args->transport_data, "")) == -1) {
+			S_OR(args->transport_data, ""),
+			S_COR(!args->transport_data && args->direction, "/", ""),
+			S_OR(direction, "")) == -1) {
```
seanbright: This does not need an `S_OR(...)`, it can just be `direction`

channels/chan_websocket.c
```diff
@@ -899,27 +914,79 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
...
+
+		direction = ast_json_object_string_get(json, "direction");
+		if (!direction) {
+			send_event(instance, ERROR, "%s requires a 'direction' parameter.\n", command);
+			return 0;
+		}
+
+		if (!strcasecmp("both", direction)) {
```
seanbright: This should be `strcmp`

channels/chan_websocket.c
```diff
@@ -899,27 +914,79 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
...
+			}
+
+			if (!instance->timer) {
+				set_channel_timer(instance);
+				ast_queue_frame(instance->channel, &ast_null_frame);
+			}
+
+			instance->media_direction = WEBCHAN_MEDIA_DIRECTION_BOTH;
+
+		} else if (!strcasecmp("out", direction)) {
```
seanbright: This should be `strcmp`

channels/chan_websocket.c
```diff
@@ -899,27 +914,79 @@ static int handle_command(struct websocket_pvt *instance, char *buffer)
...
+			}
+
+			if (!instance->timer) {
+				set_channel_timer(instance);
+				ast_queue_frame(instance->channel, &ast_null_frame);
+			}
+
+			instance->media_direction = WEBCHAN_MEDIA_DIRECTION_OUT;
+
+		} else if (!strcasecmp("in", direction)) {
```
seanbright: This should be `strcmp`

channels/chan_websocket.c
jcolp: @seanbright brought up on IRC that if we are defining this why can't we make it strict. I looked at the existing usage of "direction" and they just use strcmp so let's be consistent with strictness, so change from strcasecmp to strcmp for "direction".

jcolp (changes requested):

jcolp (approved):

PR #1772, opened 2026-02-12
build: Fix GCC discarded-qualifiers const errors.

jcolp:
GCC 15.2.1 pays attention to the discarding of the const qualifier when strchr, strrchr, memchr, or memrchr are now used. This change fixes numerous errors with this throughout the tree. The fixes can be broken down into the following:

1. The return value should be considered const.
2. The value passed to strchr or strrchr can be cast as it is expected and allowed to be modified.
3. The pointer passed to strchr or strrchr is not meant to be modified and so the contents must be duplicated.
4. It was declared const and never should have been.


jcolp:
cherry-pick-to: 20
cherry-pick-to: 21
cherry-pick-to: 22
cherry-pick-to: 23

jcolp:
I haven't yet fixed up extconf.

jcolp:
Alrighty, that should cover the latest batch.

mbradeen (approved):

PR #1777, opened 2026-02-17
CDR/CEL Custom Performance Improvements

gtjoseph:
There is a LOT of work in this commit but the TL;DR is that it takes
CEL processing from using 38% of the CPU instructions used by a call,
which is more than that used by the call processing itself, down to less
than 10% of the instructions.

So here's the deal...  cdr_custom, cdr_sqlite3_custom, cel_custom
and cel_sqlite3_custom all shared one ugly trait...they all used
ast_str_substitute_variables() or pbx_substitute_variables_helper()
to resolve the dialplan functions used in their config files.  Not only
are they both extraordinarily expensive, they both require a dummy
channel to be allocated and destroyed for each record written.  For CDRs,
that's not too bad because we only write one CDR per call.  For CELs however,
it's a disaster.

As far as source code goes, the modules basically all did the same thing.
Unfortunately, they did it badly.  The config files simply contained long
opaque strings which were intepreted by ast_str_substitute_variables() or
pbx_substitute_variables_helper(), the very functions that ate all the
instructions.  This meant introducing a new "advanced" config format much
like the advanced manager event filtering added to manager.conf in 2024.
Fortunately however, if the legacy config was recognizable, we were able to
parse it as an advanced config and gain the benefit.  If not, then it
goes the legacy, and very expensive, route.

Given the commonality among the modules, instead of making the same
improvements to 4 modules then trying to maintain them over time, a single
module "res_cdrel_custom" was created that contains all of the common code.
A few bonuses became possible in the process...
* The cdr_custom and cel_custom modules now support JSON formatted output.
* The cdr_sqlite_custom and cel_sqlite3_custom modules no longer have
  to share an Sqlite3 database.

Summary of changes:

A new module "res/res_cdrel_custom.c" has been created and the existing
cdr_custom, cdr_sqlite3_custom, cel_custom and cel_sqlite3_custom modules
are now just stubs that call the code in res_cdrel_custom.

res_cdrel_custom contains:
* A common configuration facility.
* Getters for both CDR and CEL fields that share the same abstraction.
* Formatters for all data types found in the ast_cdr and ast_event
  structures that share the same abstraction.
* Common writers for the text file and database backends that, you guessed it,
  share the same abstraction.

The result is that while there is certainly a net increase in the number of
lines in the code base, most of it is in the configuration handling at
load-time.  The run-time instruction path length is now significanty shorter.

```
Scenario                   Instructions     Latency
=====================================================
CEL pre changes                  38.49%     37.51%
CEL Advanced                      9.68%      6.06%
CEL Legacy (auto-conv to adv)     9.95%      6.13%

CEL Sqlite3 pre changes          39.41%     39.90%
CEL Sqlite3 Advanced             25.68%     24.24%
CEL Sqlite3 Legacy (auto conv)   25.88%     24.53%

CDR pre changes                   4.79%      2.95%
CDR Advanced                      0.79%      0.47%
CDR Legacy (auto conv to adv)     0.86%      0.51%

CDR Sqlite3 pre changes           4.47%      2.89%
CEL Sqlite3 Advanced              2.16%      1.29%
CEL Sqlite3 Legacy (auto conv)    2.19%      1.30%
```

Notes:
* We only write one CDR per call but every little bit helps.
* Sqlite3 still takes a fair amount of resources but the new config
  makes a decent improvement.
* Legacy configs that we can't auto convert will still take the
  "pre changes" path.

If you're interested in more implementation details, see the comments
at the top of the res_cdrel_custom.c file.

One minor fix to CEL is also included...Although TenantID was added to the
ast_event structure, it was always rendered as an empty string.  It's now
properly rendered.

UserNote: Significant performance improvements have been made to the
cdr_custom, cdr_sqlite3_custom, cel_custom and cel_sqlite3_custom modules.
See the new sample config files for those modules to see how to benefit
from them.

### DIFF
--- a/cdr/cdr_custom.c
+++ b/cdr/cdr_custom.c
@@ -28,6 +28,9 @@
  *
  * Logs in LOG_DIR/cdr_custom
  * \ingroup cdr_drivers
+ *
+ * The logic for this module now resides in res/res_cdrel_custom.c.
+ *
  */
 
 /*! \li \ref cdr_custom.c uses the configuration file \ref cdr_custom.conf
@@ -40,192 +43,73 @@
  */
 
 /*** MODULEINFO
+	<depend>res_cdrel_custom</depend>
 	<support_level>core</support_level>
  ***/
 
 #include "asterisk.h"
 
-#include <time.h>
-
-#include "asterisk/paths.h"	/* use ast_config_AST_LOG_DIR */
-#include "asterisk/channel.h"
 #include "asterisk/cdr.h"
 #include "asterisk/module.h"
-#include "asterisk/config.h"
-#include "asterisk/pbx.h"
-#include "asterisk/utils.h"
-#include "asterisk/lock.h"
-#include "asterisk/threadstorage.h"
-#include "asterisk/strings.h"
+#include "asterisk/res_cdrel_custom.h"
 
 #define CONFIG "cdr_custom.conf"
 
-AST_THREADSTORAGE(custom_buf);
+#define CUSTOM_BACKEND_NAME "CDR File custom backend"
 
-static const char name[] = "cdr-custom";
+static struct cdrel_configs *configs;
 
-struct cdr_custom_config {
-	AST_DECLARE_STRING_FIELDS(
-		AST_STRING_FIELD(filename);
-		AST_STRING_FIELD(format);
-		);
-	ast_mutex_t lock;
-	AST_RWLIST_ENTRY(cdr_custom_config) list;
-};
+/*!
+ * Protects in-flight log transactions from reloads.
+ */
+static ast_rwlock_t configs_lock;
 
-static AST_RWLIST_HEAD_STATIC(sinks, cdr_custom_config);
+#define CDREL_RECORD_TYPE cdrel_record_cdr
+#define CDREL_BACKEND_TYPE cdrel_backend_text
 
-static void free_config(void)
+static int custom_log(struct ast_cdr *cdr)
 {
-	struct cdr_custom_config *sink;
-
-	while ((sink = AST_RWLIST_REMOVE_HEAD(&sinks, list))) {
-		ast_mutex_destroy(&sink->lock);
-		ast_string_field_free_memory(sink);
-		ast_free(sink);
-	}
-}
-
-static int load_config(void)
-{
-	struct ast_config *cfg;
-	struct ast_variable *var;
-	struct ast_flags config_flags = { 0 };
 	int res = 0;
 
-	cfg = ast_config_load(CONFIG, config_flags);
-	if (!cfg || cfg == CONFIG_STATUS_FILEINVALID) {
-		ast_log(LOG_ERROR, "Unable to load " CONFIG ". Not logging custom CSV CDRs.\n");
-		return -1;
-	}
-
-	var = ast_variable_browse(cfg, "mappings");
-	while (var) {
-		if (!ast_strlen_zero(var->name) && !ast_strlen_zero(var->value)) {
-			struct cdr_custom_config *sink = ast_calloc_with_stringfields(1, struct cdr_custom_config, 1024);
-
-			if (!sink) {
-				ast_log(LOG_ERROR, "Unable to allocate memory for configuration settings.\n");
-				res = -2;
-				break;
-			}
-
-			ast_string_field_build(sink, format, "%s\n", var->value);
-			if (var->name[0] == '/') {
-				ast_string_field_build(sink, filename, "%s", var->name);
-			} else {
-				ast_string_field_build(sink, filename, "%s/%s/%s", ast_config_AST_LOG_DIR, name, var->name);
-			}
-			ast_mutex_init(&sink->lock);
-
-			AST_RWLIST_INSERT_TAIL(&sinks, sink, list);
-		} else {
-			ast_log(LOG_NOTICE, "Mapping must have both a filename and a format at line %d\n", var->lineno);
-		}
-		var = var->next;
-	}
-	ast_config_destroy(cfg);
+	ast_rwlock_rdlock(&configs_lock);
+	res = cdrel_logger(configs, cdr);
+	ast_rwlock_unlock(&configs_lock);
 
 	return res;
 }
 
-static int custom_log(struct ast_cdr *cdr)
-{
-	struct ast_channel *dummy;
-	struct ast_str *str;
-	struct cdr_custom_config *config;
-
-	/* Batching saves memory management here.  Otherwise, it's the same as doing an allocation and free each time. */
-	if (!(str = ast_str_thread_get(&custom_buf, 16))) {
-		return -1;
-	}
-
-	dummy = ast_dummy_channel_alloc();
-	if (!dummy) {
-		ast_log(LOG_ERROR, "Unable to allocate channel for variable substitution.\n");
-		return -1;
-	}
-
-	/* We need to dup here since the cdr actually belongs to the other channel,
-	   so when we release this channel we don't want the CDR getting cleaned
-	   up prematurely. */
-	ast_channel_cdr_set(dummy, ast_cdr_dup(cdr));
-
-	AST_RWLIST_RDLOCK(&sinks);
-
-	AST_LIST_TRAVERSE(&sinks, config, list) {
-		FILE *out;
-
-		ast_str_substitute_variables(&str, 0, dummy, config->format);
-
-		/* Even though we have a lock on the list, we could be being chased by
-		   another thread and this lock ensures that we won't step on anyone's
-		   toes.  Once each CDR backend gets it's own thread, this lock can be
-		   removed. */
-		ast_mutex_lock(&config->lock);
-
-		/* Because of the absolutely unconditional need for the
-		   highest reliability possible in writing billing records,
-		   we open write and close the log file each time */
-		if ((out = fopen(config->filename, "a"))) {
-			fputs(ast_str_buffer(str), out);
-			fflush(out); /* be particularly anal here */
-			fclose(out);
-		} else {
-			ast_log(LOG_ERROR, "Unable to re-open master file %s : %s\n", config->filename, strerror(errno));
-		}
-
-		ast_mutex_unlock(&config->lock);
-	}
-
-	AST_RWLIST_UNLOCK(&sinks);
-
-	ast_channel_unref(dummy);
-
-	return 0;
-}
-
 static int unload_module(void)
 {
-	if (ast_cdr_unregister(name)) {
-		return -1;
+	int res = 0;
+
+	ast_rwlock_wrlock(&configs_lock);
+	res = cdrel_unload_module(CDREL_BACKEND_TYPE, CDREL_RECORD_TYPE, configs, CUSTOM_BACKEND_NAME);
+	ast_rwlock_unlock(&configs_lock);
+	if (res == 0) {
+		ast_rwlock_destroy(&configs_lock);
 	}
 
-	if (AST_RWLIST_WRLOCK(&sinks)) {
-		ast_cdr_register(name, ast_module_info->description, custom_log);
-		ast_log(LOG_ERROR, "Unable to lock sink list.  Unload failed.\n");
-		return -1;
-	}
-
-	free_config();
-	AST_RWLIST_UNLOCK(&sinks);
-	return 0;
+	return res;
 }
 
 static enum ast_module_load_result load_module(void)
 {
-	if (AST_RWLIST_WRLOCK(&sinks)) {
-		ast_log(LOG_ERROR, "Unable to lock sink list.  Load failed.\n");
+	if (ast_rwlock_init(&configs_lock) != 0) {
 		return AST_MODULE_LOAD_DECLINE;
 	}
 
-	load_config();
-	AST_RWLIST_UNLOCK(&sinks);
-	ast_cdr_register(name, ast_module_info->description, custom_log);
-	return AST_MODULE_LOAD_SUCCESS;
+	configs = cdrel_load_module(CDREL_BACKEND_TYPE, CDREL_RECORD_TYPE, CONFIG, CUSTOM_BACKEND_NAME, custom_log);
+
+	return configs ? AST_MODULE_LOAD_SUCCESS : AST_MODULE_LOAD_DECLINE;
 }
 
 static int reload(void)
 {
-	if (AST_RWLIST_WRLOCK(&sinks)) {
-		ast_log(LOG_ERROR, "Unable to lock sink list.  Load failed.\n");
-		return AST_MODULE_LOAD_DECLINE;
-	}
-
-	free_config();
-	load_config();
-	AST_RWLIST_UNLOCK(&sinks);
-	return AST_MODULE_LOAD_SUCCESS;
+	int res = 0;
+	ast_rwlock_wrlock(&configs_lock);
+	res = cdrel_reload_module(CDREL_BACKEND_TYPE, CDREL_RECORD_TYPE, &configs, CONFIG);
+	ast_rwlock_unlock(&configs_lock);
+	return res;
 }
 
 AST_MODULE_INFO(ASTERISK_GPL_KEY, AST_MODFLAG_LOAD_ORDER, "Customizable Comma Separated Values CDR Backend",
@@ -234,5 +118,5 @@ AST_MODULE_INFO(ASTERISK_GPL_KEY, AST_MODFLAG_LOAD_ORDER, "Customizable Comma Se
 	.unload = unload_module,
 	.reload = reload,
 	.load_pri = AST_MODPRI_CDR_DRIVER,
-	.requires = "cdr",
+	.requires = "cdr,res_cdrel_custom",
 );
--- a/cel/cel_custom.c
+++ b/cel/cel_custom.c
@@ -24,197 +24,75 @@
  * \author Steve Murphy <murf@digium.com>
  * Logs in LOG_DIR/cel_custom
  * \ingroup cel_drivers
+ *
+ * The logic for this module now resides in res/res_cdrel_custom.c.
+ *
  */
 
 /*** MODULEINFO
+	<depend>res_cdrel_custom</depend>
 	<support_level>core</support_level>
  ***/
 
 #include "asterisk.h"
 
-#include "asterisk/paths.h"
-#include "asterisk/channel.h"
 #include "asterisk/cel.h"
 #include "asterisk/module.h"
-#include "asterisk/config.h"
-#include "asterisk/pbx.h"
-#include "asterisk/utils.h"
-#include "asterisk/lock.h"
-#include "asterisk/threadstorage.h"
-#include "asterisk/strings.h"
+#include "asterisk/res_cdrel_custom.h"
 
 #define CONFIG "cel_custom.conf"
 
-AST_THREADSTORAGE(custom_buf);
-
-static const char name[] = "cel-custom";
-
-struct cel_config {
-	AST_DECLARE_STRING_FIELDS(
-		AST_STRING_FIELD(filename);
-		AST_STRING_FIELD(format);
-	);
-	ast_mutex_t lock;
-	AST_RWLIST_ENTRY(cel_config) list;
-};
-
 #define CUSTOM_BACKEND_NAME "CEL Custom CSV Logging"
 
-static AST_RWLIST_HEAD_STATIC(sinks, cel_config);
+static struct cdrel_configs *configs;
 
-static void free_config(void)
-{
-	struct cel_config *sink;
+/*!
+ * Protects in-flight log transactions from reloads.
+ */
+static ast_rwlock_t configs_lock;
 
-	while ((sink = AST_RWLIST_REMOVE_HEAD(&sinks, list))) {
-		ast_mutex_destroy(&sink->lock);
-		ast_string_field_free_memory(sink);
-		ast_free(sink);
-	}
-}
-
-static int load_config(void)
-{
-	struct ast_config *cfg;
-	struct ast_variable *var;
-	struct ast_flags config_flags = { 0 };
-	int mappings = 0;
-	int res = 0;
-
-	cfg = ast_config_load(CONFIG, config_flags);
-	if (!cfg || cfg == CONFIG_STATUS_FILEINVALID) {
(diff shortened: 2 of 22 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
It looks like I've got some typos in the commit message and the comments in res_cdrel_custom.c.  I'll fix them in due course.

res/cdrel_custom/config.c
```diff
@@ -0,0 +1,1408 @@
...
+	*column_count = 0;
+
+	/* We need to trim and remove any single or double quotes from each column name. */
+	while((col = ast_strsep(&cols, ',', AST_STRSEP_TRIM))) {
+		col = ast_strip(ast_strip_quoted(col, "'\"", "'\""));
+		if (ast_str_append(&column_string, 0, "%s%s", is_first ? "" : ",", col) <= 0) {
+			return -1;
+		}
+		is_first = 0;
+		(*column_count)++;
```
seanbright: Huge optimization opportunity here. You don't need is_first, you can just use column_count.

gtjoseph: Done.  I can't wait to see what else you find. :smile:

configs/samples/cel_custom.conf.sample
```diff
@@ -36,5 +35,218 @@
...
 ;                  CHANNEL(peer) could also be used.
 ;
-[mappings]
+; Legacy Mappings
+;
+; Within a legacy mapping, use the CALLERID(), CHANNEL() and CSV_QUOTE()
+; functions to retrieve values from the CEL event (and pay the price).
+;
+; NOTE: If your legacy mapping uses commas as field separators and only the
+; CSV_QUOTE, CALLERID and CHANNEL dialplan functions or one of teh special
```
InterLinked1: the

gtjoseph: done

phoneben:
just curiosity regular cell does not benefit from this pr correct ?

jcolp:
> just curiosity regular cell does not benefit from this pr correct ?

Channels that produce CDRs or CEL events benefit when the respective modules are in use.

gtjoseph:
> just curiosity regular cell does not benefit from this pr correct ?

@phoneben What do you mean by "regular cell"?
The *custom* modules were the only ones that used the dialplan functions and ast_str_substitute_variables() which is where the resources were going.

res/res_cdrel_custom.c
```diff
@@ -0,0 +1,229 @@
...
+ * registry.c.
+ *
+ * "loggers.c": Contains the common "cdrel_logger" entrypoint that the
+ * individual modules call to log a record.  It takes the module's
+ * cdrel_configs vector and the record to log it got from the core
+ * cel.c and cdr.c.  It then looks up and runs the logger implementation
+ * based on the backend type (text file or database) and config type
+ * (legacy or advanced).
+ *
+ * "getters_cdr.c", "getters_cdr.c": Contain the getters that retrieve
```
phoneben: the second one should be `getters_cel.c`?

gtjoseph: oops, yep.

phoneben:
thank you just using regular cel not the custom 

> > just curiosity regular cell does not benefit from this pr correct ?
> 
> @phoneben What do you mean by "regular cell"? The _custom_ modules were the only ones that used the dialplan functions and ast_str_substitute_variables() which is where the resources were going.

i got excited i thought i had another improvement  like stasis!
 but i do not use the custom models  just the plain cel !
 
 thank you 

i have some more typos should i comment?

include/asterisk/cel.h
```diff
@@ -117,6 +118,18 @@ const char *ast_cel_get_type_name(enum ast_cel_event_type type);
  */
 enum ast_cel_event_type ast_cel_str_to_event_type(const char *name);
 
+/*!
+ * \brief Format an event timeval using dateformat from cel.conf
+ *
+ * \param eventtime The timeval to format
+ * \param timebuf A buffer of at least 30 chracters to place the result in
```
phoneben: characters

res/cdrel_custom/cdrel.h
```diff
@@ -0,0 +1,343 @@
...
+		AST_STRING_FIELD(db_columns);              /*!< List of columns for database backends */
+		AST_STRING_FIELD(db_table);                /*!< Table name for database backends */
+	);
+	sqlite3 *db;                                   /*!< sqlite3 database handle */
+	int busy_timeout;                              /*!< sqlite3 query timeout value */
+	cdrel_dummy_channel_alloc dummy_channel_alloc; /*!< Legacy config types need a dummy channel */
+	enum cdrel_backend_type backend_type;          /*!< Text file or database */
+	enum cdrel_config_type config_type;            /*!< Legacy or advanced */
+	enum cdrel_text_format_type format_type;       /*!< For text files, CSV or JSON */
+	enum cdrel_quoting_method quoting_method;      /*!< When to qquote */
```
phoneben: quote

configs/samples/cdr_custom.conf.sample
```diff
@@ -1,19 +1,143 @@
...
+
+;format = dsv                 ; Advanced mappings can actually have two types
+                              ; "dsv" or "json".  This example uses "dsv", which
+                              ; is the default, but see the example below for more
+                              ; info on the json format.
+
+;separator_character = ,      ; The character to use for the field separator,
+                              ; It defaults to a comma but you can use any
+                              ; character you want.  For instance, specify
+                              ; \t to seprate the fields with a tab.
```
phoneben: separate

configs/samples/cel_custom.conf.sample
```diff
@@ -36,5 +35,218 @@
...
+
+;format = dsv                 ; Advanced mappings can actually have two types
+                              ; "dsv" or "json".  This example uses "dsv", which
+                              ; is the default, but see the example below for more
+                              ; info on the json format.
+
+;separator_character = ,      ; The character to use for the field separator,
+                              ; It defaults to a comma but you can use any
+                              ; character you want.  For instance, specify
+                              ; \t to seprate the fields with a tab.
```
phoneben: separate

configs/samples/cel_sqlite3_custom.conf.sample
```diff
@@ -1,11 +1,154 @@
...
+;
+; NOTE: If your legacy mapping uses only those two functions and the special
+; variables, the module will attempt to strip the functions and create a much
+; faster advanced mapping for it. However, we urge you to create a real advanced
+; mapping and not rely on this process. If the mapping contains something not
+; recognized it will go the slower legacy route.
+;
+;
+;[cel_master]
+;table = cel             ; The name of the table in the datasbase into which records
```
phoneben: database

configs/samples/cel_sqlite3_custom.conf.sample
```diff
@@ -1,11 +1,154 @@
...
+    ; number of columns or the config will fail to load.
+
+;busy_timeout = 1000    ; The number of milliseconds to wait for a database operation
+                        ; to complete before an error is returned.
+
+;
+; Advanced Mappings
+;
+;[cdr_advanced]
+;table = cel             ; The name of the table in the datasbase into which records
```
phoneben: database

configs/samples/cdr_sqlite3_custom.conf.sample
```diff
@@ -1,11 +1,105 @@
...
+;
+; NOTE: If your legacy mapping uses only the CDR dialplan function, the module
+; will attempt to strip the functions and create a much faster advanced mapping
+; for it. However, we urge you to create a real advanced mapping and not rely
+; on this process. If the mapping contains something not recognized it will go
+; the slower legacy route.
+;
+;
+;[cdr_master]
+;table = cdr             ; The name of the table in the datasbase into which records
```
phoneben: database

configs/samples/cdr_sqlite3_custom.conf.sample
```diff
@@ -1,11 +1,105 @@
...
+    ; number of columns or the config will fail to load.
+
+;busy_timeout = 1000    ; The number of milliseconds to wait for a database operation
+                        ; to complete before an error is returned.
+
+;
+; Advanced Mappings
+;
+;[cdr_advanced]
+;table = cdr             ; The name of the table in the datasbase into which records
```
phoneben: database

configs/samples/cel_sqlite3_custom.conf.sample
```diff
@@ -1,11 +1,154 @@
...
 ; Asterisk Channel Event Logging (CEL) - SQLite 3 Backend
 ;
 
 ;
 ; Mappings for sqlite3 config file
 ;
-; Within a mapping, use the CALLERID() and CHANNEL() functions to retrieve
+; Legacy vs Advanced Mappings
+;
+; Legacy mappings are very resource intentsive because you have to use
```
phoneben: intensive

res/cdrel_custom/config.c
```diff
@@ -0,0 +1,1406 @@
...
+	tmp_field = ast_strip(ast_strdupa(input_field_template));
+
+	if (tmp_field[0] == '\'') {
+		result.csv_quote = 1;
+	}
+
+	/*
+	 * I really hate the fact that ast_strip really trims whitespace
+	 * and ast_strip_quoted will strip anything in pairs.
+	 * Anyway, get rid of any remaining encosing quotes.
```
phoneben: enclosing

res/cdrel_custom/config.c
```diff
@@ -0,0 +1,1406 @@
...
+	 */
+	tmp_field = ast_strip(ast_strip_quoted(tmp_field, "'\"'", "'\"'"));
```
phoneben: ast_strip_quoted(tmp_field, "'\"'", "'\"'") looks like it has an extra quote compared to the pattern used elsewhere ("'\"" / "'\"").

seanbright:
@phoneben each of your one-off comments generates a separate notification. Can you submit multiple comments as a review instead so they are consolidated?

cdr/cdr_custom.c
```diff
@@ -40,192 +43,71 @@
  */
 
 /*** MODULEINFO
+	<depend>res_cdrel_custom</depend>
```
InterLinked1: I haven't tested this, but if `autoload=no`, does this now manually need to be loaded for CDR backends to load properly, or will it get automatically loaded?

configs/samples/cel_custom.conf.sample
```diff
@@ -36,5 +35,218 @@
...
+;    The quote escape character is always the backslash.
+;    The quoting method is always "non_numeric".
+;
+;  Since JSON requires both a name and value, the name is always
+;  the field name.  For literals however you must specify the literal
+;  as a "name: value" pair as demonstrated above.  The output record
+;  would then look something like:
+; {"eventtype":"HANGUP","eventenum":"HANGUP","userdeftype":"","My":"Literal","eventtime":1771359872.0, ...}
+
+
```
InterLinked1: Why so many blank lines here?

configs/samples/cdr_custom.conf.sample
```diff
@@ -1,19 +1,143 @@
...
+                              ; Additionally, the following options are available:
+                              ; "all": Quote all fields. The default.
+                              ; "non_numeric": Quote all non numeric fields.
+                              ; "none": Don't quote any field (unless it contains
+                              ; a separator character).
+
+;fields = clid,"Some Literal",src, dst, dcontext, channel, dstchannel, lastapp, lastdata, start, answer, end, duration, billsec, disposition, amaflags, accountcode, uniqueid(noquote), userfield, sequence,ds_type(uservar)
+    ; This is the list of fields to include in the record. The field names are the
+    ; same as in the legacy mapping but without any enclosing dialplan functions.
+    ; You can specify literals to be placed in the in output record by double-quoting
```
mbradeen: "in the in" - also in 4 other places with the same blurb.

res/cdrel_custom/config.c
```diff
@@ -0,0 +1,1406 @@
...
+	}
+
+	return 0;
+}
+
+/*!
+ * \internal
+ * \brief Load a database config from a config file category.
+ *
+ * \param record_type CDRF or CEL.
```
mbradeen: CDRF?

res/cdrel_custom/cdrel.h
```diff
@@ -0,0 +1,343 @@
...
+	char separator[2];                             /*!< For text files, the field separator */
```
mbradeen, line 297: Is it worth having this as a 2 char array and always referencing [0] over copying the first char from the value string?

cdr/cdr_custom.c
```diff
@@ -40,192 +43,71 @@
...
 
-	if (AST_RWLIST_WRLOCK(&sinks)) {
-		ast_cdr_register(name, ast_module_info->description, custom_log);
-		ast_log(LOG_ERROR, "Unable to lock sink list.  Unload failed.\n");
-		return -1;
-	}
+	ast_rwlock_wrlock(&configs_lock);
+	res = cdrel_unload_module(CDREL_BACKEND_TYPE, CDREL_RECORD_TYPE, configs, CUSTOM_BACKEND_NAME);
+	ast_rwlock_unlock(&configs_lock);
+	ast_rwlock_destroy(&configs_lock);
```
phoneben: ai 
In unload_module() you always call `ast_rwlock_destroy(&configs_lock);` (line 88), even when `cdrel_unload_module(...)` returns an error (line 86).

That’s risky because in Asterisk, if u`nload_module() `returns non‑zero, the module does not unload. In that failure case, your module can continue running, and `custom_log()` can be called again — but now it will try to lock a destroyed lock → crash / UB

maybe something like 

```
	if (res == 0) {
		ast_rwlock_destroy(&configs_lock);
		configs = NULL;
	}
```

cel/cel_custom.c
```diff
@@ -24,197 +24,75 @@
...
-#include "asterisk/threadstorage.h"
-#include "asterisk/strings.h"
-
-#define CONFIG "cel_custom.conf"
-
-AST_THREADSTORAGE(custom_buf);
+#include "asterisk/res_cdrel_custom.h"
 
-static const char name[] = "cel-custom";
+#include "asterisk/res_cdrel_custom.h"
```
phoneben: #include "`asterisk/res_cdrel_custom.h`" twice

include/asterisk/res_cdrel_custom.h
```diff
@@ -0,0 +1,117 @@
...
+ * \param backend_type One of \ref cdrel_backend_type.
+ * \param record_type One of \ref cdrel_record_type.
+ * \param config_filename The config file name.
+ * \param backend_name The name to register the backend as.
+ * \param logging_cb The logging callback to register with CDR or CEL.
+ * \returns A pointer to a VECTOR or config objects read from the config file.
+ */
+struct cdrel_configs *cdrel_load_module(enum cdrel_backend_type backend_type,
+	enum cdrel_record_type record_type, const char *config_filename,
+	const char *backend_name, void *logging_cb);
```
phoneben, line 66: a type mismatch: you’re passing an int (*)(struct ast_cdr *) where a void * is expected.

`configs = cdrel_load_module(CDREL_BACKEND_TYPE, CDREL_RECORD_TYPE, CONFIG, CUSTOM_BACKEND_NAME, custom_log);`

`static int custom_log(struct ast_cdr *cdr)
`

res/cdrel_custom/cdrel.h
gtjoseph: It's a string because this allows me to optionally use it in printf format strings as a `%s`  I.E. `"%s%s", is_first ? "" : separator, value`

include/asterisk/res_cdrel_custom.h
gtjoseph: This is intentional because the CDR and CEL log functions have different signatures.

gtjoseph:
The last push includes...
* Pre-preparing the SQL INSERT statement once per config for the database backend instead of preparing it for every record written.  Contributed by @seanbright 
* Addressed all review feedback
* Change cdrel_field->flags from an enum to an ast_flags structure.
* Fixed a few bugs.

mbradeen (approved):

include/asterisk/res_cdrel_custom.h
```diff
@@ -0,0 +1,117 @@
...
+ * \author George Joseph <gjoseph@sangoma.com>
+ *
+ * \brief Protected header for the CDR and CEL Custom Backends
+ *
+ * \warning This file should be included only by CDR and CEL backends.
+ *
+ */
+
+#ifndef _RES_CDREL_CUSTOM_H
+#define _RES_CDREL_CUSTOM_H
```
seanbright, line 30: From C99:

> All identifiers that begin with an underscore and either an uppercase letter or another
underscore are always reserved for any use.

So maybe `AST_RES_...` instead?

main/cel.c
```diff
@@ -673,6 +673,34 @@ static void check_retire_linkedid(struct ast_channel_snapshot *snapshot, const s
...
+		ast_localtime(&eventtime, &tm, NULL);
+		ast_strftime(timebuf, len, cfg->general->date_format, &tm);
+	}
+
+	return 0;
+}
+
+int ast_cel_format_eventtime(struct timeval eventtime, char *timebuf, size_t len)
+{
+	RAII_VAR(struct cel_config *, cfg, ao2_global_obj_ref(cel_configs), ao2_cleanup);
```
seanbright: Seems a _little_ excessive given there's only 1 failure path, but not the end of the world.

main/cel.c
```diff
@@ -759,6 +780,7 @@ struct ast_channel *ast_cel_fabricate_channel_from_event(const struct ast_event
 	ast_channel_accountcode_set(tchan, record.account_code);
 	ast_channel_peeraccount_set(tchan, record.peer_account);
 	ast_channel_userfield_set(tchan, record.user_field);
+	ast_channel_tenantid_set(tchan, record.tenant_id);
```
seanbright, line 788: Make sure this is noted in the commit message. This could potentially be a separate fix?

res/cdrel_custom/cdrel.h
```diff
@@ -0,0 +1,363 @@
...
+/*!
+ * \file
+ * \author George Joseph <gjoseph@sangoma.com>
+ *
+ * \brief Private header for res_cdrel_custom.
+ *
+ */
+
+#ifndef _CDREL_H
+#define _CDREL_H
```
seanbright, line 28: Previous comment about reserved identifiers

res/cdrel_custom/cdrel.h
```diff
@@ -0,0 +1,363 @@
...
+
+/*! Represents a field definition */
+struct cdrel_field {
+	enum cdrel_record_type record_type;     /*!< CDR or CEL */
+	int field_id;                           /*!< May be an AST_EVENT_IE_CEL or a cdr_field_id */
+	char *name;                             /*!< The name of the field */
+	enum cdrel_data_type input_data_type;   /*!< The data type of the field in the source record */
+	enum cdrel_data_type output_data_type;
+	struct ast_flags flags;                 /*!< Flags used during config parsing */
+	char data[1];                           /*!< Could be literal data, a user variable name, etc */
```
seanbright, line 276: I think we're safe to make this `char data[];` instead (we do it elsewhere in core code at least).

res/cdrel_custom/config.c
```diff
@@ -0,0 +1,1464 @@
...
+		[cdrel_record_cel] = "cel-custom",
+	},
+	[cdrel_backend_db] = {
+		[cdrel_record_cdr] = NULL,
+		[cdrel_record_cel] = NULL,
+	}
+};
+
+static void field_free(struct cdrel_field *f) {
+	if (f) {
```
seanbright: `ast_free(...)` is `NULL` safe.

res/cdrel_custom/writers.c
```diff
@@ -0,0 +1,231 @@
...
+{
+	FILE *out;
+	int res = 0;
+
+	ast_mutex_lock(&config->lock);
+	if ((out = fopen(config->output_filename, "a"))) {
+		res = fputs(ast_str_buffer(record), out);
+		fputs("\n", out);
+		fflush(out);
+		fclose(out);
```
seanbright: From [`fclose()`](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fclose.html):

> The `fclose()` function shall cause the stream pointed to by stream to be flushed and the associated file to be closed.

include/asterisk/res_cdrel_custom.h
gtjoseph: Not following.  We use the _RES pattern in many other include files.

main/cel.c
gtjoseph: Ah yeah, that snuck in when this was just a CEL PR. I'll add it to the commit msg.

include/asterisk/res_cdrel_custom.h
seanbright: Not following what? Just because it’s wrong in other places doesn’t mean we keep doing it.

res/cdrel_custom/cdrel.h
gtjoseph: `data[]` and `data[0]` causes sizeof to NOT include space for it.  Try this and see what happens...

```
struct test_struct {
	int dummy;
	char data[];
};

int main(int argc, char **argv) {
	struct test_struct *ts = calloc(2, sizeof(*ts));
	ts->dummy = 0xFF;
	struct test_struct *ts2 = ts + 1;
	ts2->dummy = 66;
	printf("%p %d %d %d '%s' '%hhu'\n", ts, (int)sizeof(struct test_struct), (int)sizeof(*ts), ts->dummy, ts->data, ts->data[0]);
	printf("%p %d %d %d '%s' '%hhu'\n", ts2, (int)sizeof(struct test_struct), (int)sizeof(*ts2), ts2->dummy, ts2->data, ts2->data[0]);
	return 0;
}
```

res/cdrel_custom/writers.c
gtjoseph: Copy and paste from the old code.  Removed.

gtjoseph: Hmmm.  I wonder how much is spent opening the file for append for every record vs just keeping the handle open and flushing after each record.  I'll have to test.

PR #1778, opened 2026-02-18
astconfigparser.py: Fix regex pattern error by properly escaping string

juliandunn:
"SyntaxWarning: invalid escape sequence '\s'" occurs when using the pjsip
migration script because '\' is an escape character in Python. Instead,
use a raw string for the regex.


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1783, opened 2026-02-18
[bug]: Several unused-but-set-variable warnings with gcc 16


InterLinked1:
### Severity

Minor

### Versions

Git

### Components/Modules

N/A

### Operating Environment

gcc 16.0.1 20260218

### Frequency of Occurrence

Constant

### Issue Description

A few unused variable warnings with gcc 16

### Relevant log output

```shell
[CCi] app_externalivr.i -> app_externalivr.o
app_externalivr.c: In function ‘ast_eivr_getvariable’:
app_externalivr.c:314:13: error: variable ‘j’ set but not used [-Werror=unused-but-set-variable=]
  314 |         int j;
      |             ^
   [CCi] app_festival.i -> app_festival.o
cc1: all warnings being treated as errors


app_minivm.c: In function ‘message_template_parse_filebody’:
app_minivm.c:2622:13: error: variable ‘lines’ set but not used [-Werror=unused-but-set-variable=]
 2622 |         int lines = 0;
      |             ^~~~~

    inlined from ‘acf_curl_helper’ at func_curl.c:776:30:
func_curl.c:723:31: error: call to ‘_curl_easy_setopt_err_long’ declared with attribute warning: curl_easy_setopt expects a long argument [-Werror=attribute-warning]
  723 |         curl_easy_setopt(*curl, CURLOPT_NOSIGNAL, 1);
      |                               ^~~~~~~~~~~~~~~~~~~~~~~
func_curl.c:724:31: error: call to ‘_curl_easy_setopt_err_long’ declared with attribute warning: curl_easy_setopt expects a long argument [-Werror=attribute-warning]
  724 |         curl_easy_setopt(*curl, CURLOPT_TIMEOUT, 180);
      |                               ^~~~~~~~~~~~~~~~~~~~~~~~
In function ‘curl_instance_init’,
    inlined from ‘ast_threadstorage_get’ at /usr/src/asterisk-22.8.2/include/asterisk/threadstorage.h:180:28,
    inlined from ‘acf_curl_helper’ at func_curl.c:794:15:
func_curl.c:723:31: error: call to ‘_curl_easy_setopt_err_long’ declared with attribute warning: curl_easy_setopt expects a long argument [-Werror=attribute-warning]
  723 |         curl_easy_setopt(*curl, CURLOPT_NOSIGNAL, 1);
      |                               ^~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1784, opened 2026-02-18
build: Fix unused-but-set-variable warnings with gcc 16.

InterLinked1:
Fix or remove a few variables that were being set but not actually used anywhere, causing warnings with gcc 16.

Resolves: #1783


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1785, opened 2026-02-18
[bug]: chan_websocket doesn’t work with genericplc and transcoding


daninmadison:
### Severity

Minor

### Versions

20.18.2 and also replicated with master branch

### Components/Modules

chan_websocket

### Operating Environment

Ubuntu 22.04.5

### Frequency of Occurrence

Constant

### Issue Description

Place a G.722 call into Asterisk and answer it.  Answer the call.  If you Dial a chan_websocket with a ulaw codec, the audio will be ulaw if you perform this through the dial plan (see 999 below).  However, if you use AGI/AMI, when you perform an Exec Dial WebSocket with the same ulaw codec, the inbound audio stream doesn't realize it needs to translate ulaw.

The following is the relevant Transcode information on the WebSocket channel.  It sends ulaw@8000.  However, it doesn't realize it is receiving ulaw@8000 so it's transcoding the data received incorrectly.  This causes the audio to be garbled.

WriteTranscode: Yes (slin@16000)->(slin@8000)->(ulaw@8000)
ReadTranscode: Yes (slin@8000)->(slin@16000)

exten => 999,1(newchannel),Answer()
same => n(test),Dial(WebSocket/1/c(ulaw),20)
same => n(exit),Verbose(Exit)

exten => 998,1(agitest),AGI(agi:async)
same => n(exit),Verbose(Exit)

Worked with gjoseph to diagnose this and determined if 

asterisk.conf has 
transcode_via_sln = no

and 
codecs.conf has
genericplc = false

The problem goes away.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
I can reproduce.

gtjoseph:
@daninmadison I just pushed PR #1788 up which should fix this issue.  Give it a try and let me know.

daninmadison:
This worked for me without the additional configuration file changes.

seanbright:
@gtjoseph I assume you want this open?

daninmadison:
Not sure what the policy is.  George's pull request does fix things for me.
The only thing remaining for me is the merge into master and then the actual version with this fix.

gtjoseph:
Usually I just note in the PR that the reporter confirmed the fix and let the issue auto-close  when the PR merges.  No biggie either way.

PR #1788, opened 2026-02-19
chan_websocket: Remove silence generation and frame padding.

gtjoseph:
The original chan_websocket implementation attempted to improve the
call quality experience by generating silence frames to send to the core
when no media was being read from the websocket and padding frames with
silence when short frames were read from the websocket.  Both of these
required switching the formats on the channel to slin for short periods
of time then switching them back to whatever format the websocket channel
was configured for.  Unfortunately, the format switching caused issues
when transcoding was required and the transcode_via_sln option was enabled
in asterisk.conf (which it is by default).  The switch would cause the
transcoding path to be incorrectly set resulting in malformed RTP packets
being sent back out to the caller on the other end which the caller heard
as loud noise.

After looking through the code and performing multiple listening tests,
the best solution to this problem was to just remove the code that was
attempting to generate the silence.  There was no decrease in call quality
whatsoever and the code is a bit simpler.  None of the other channel drivers
nor res_rtp_asterisk generate silence frames or pad short frames which
backed up decision.

Resolves: #1785

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -93,12 +93,8 @@ struct websocket_pvt {
 	struct ast_websocket *websocket;
 	struct ast_format *native_format;
 	struct ast_codec *native_codec;
-	struct ast_format *slin_format;
-	struct ast_codec *slin_codec;
 	struct ast_channel *channel;
 	struct ast_timer *timer;
-	struct ast_frame silence;
-	struct ast_trans_pvt *translator;
 	AST_LIST_HEAD(, ast_frame) frame_queue;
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
The issue reporter has confirmed the fix.

gtjoseph:
NOTE TO SELF:  @bkford 's PR #1768 needs to be merged first, then this PR will need to be rebased on top of it.

PR #1789, opened 2026-02-20
Add rt_order attrb for members

bulatov-an:
In realtime queues, you cannot specify the order of iteration, for example, in a linear iteration strategy. This patch adds a new order attribute, which allows you to set the sorting order of members in such a queue. Th

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -3775,7 +3775,7 @@ static void rt_handle_member_record(struct call_queue *q, char *category, struct
 	const char *paused_str = ast_variable_retrieve(member_config, category, "paused");
 	const char *wrapuptime_str = ast_variable_retrieve(member_config, category, "wrapuptime");
 	const char *reason_paused = ast_variable_retrieve(member_config, category, "reason_paused");
-	const char *rt_order_str = ast_variable_retrieve(member_config, category, "order");
+	const char *rt_order_str = ast_variable_retrieve(member_config, category, "rt_order");
 
 	if (ast_strlen_zero(rt_uniqueid)) {
 		ast_log(LOG_WARNING, "Realtime field 'uniqueid' is empty for member %s\n",


phoneben:
how can you set it ?
are you adding a column in `queue_member` table ?

bulatov-an:
> how can you set it ? are you adding a column in `queue_member` table ?

Yes. This is an optional column for queue_member. If it is specified other than 0, it is used as a position in the queue. It doesn't have to be a column in the database. I have this attribute coming from a web server in my configuration. The order of the members is determined in the WEB UI, for example, by dragging the mouse on the fly. Another interesting case is to change the order of calling members for each individual call, taking into account the loading of the member in chats (not voice channels). If a member has processed the chat, we put him at the end of the queue.

phoneben:
should be adding this optional column  to database upgrade so it could be used 
a user note of the changes 

> . It doesn't have to be a column in the database.

realtime or not ?

have you tested it ?

bulatov-an:
> should be adding this optional column to database upgrade so it could be used a user note of the changes
> 
> > . It doesn't have to be a column in the database.
> 
> realtime or not ?
> 
> have you tested it ?

Yes, real-time, for example, through the res_config_curl module. Yes, tested in a production environment. The members call according to the order parameter in real time. Each new call gets its own order of members. If the parameter is missing or equal to 0, then the old queue logic is used.

> should be adding this optional column to database upgrade

Can you tell me which file describes the structure of the database, where should I add this information?

phoneben:
contrib/ast-db-manage/

phoneben:
also Note: order is a reserved SQL keyword in most databases. You may need to quote it — e.g., order in MySQL or "order" in PostgreSQL. Worth considering renaming the column to rt_order in the DB as well and adjusting the ast_variable_retrieve call accordingly.?

phoneben:
you have to squash them !

> `There is more than 1 commit in this PR and no PR comment with a multiple-commits: special header. Please squash the commits down into 1`

PR #1790, opened 2026-02-20
app_record: Fix hangup handling before record loop

silverjoe:
When a hangup is received while app_record is playing the
initial beep (or immediately after it), but before entering
the main recording loop, the application does not detect
the hangup and continues running until timeout.

This change adds an early hangup check before the
ast_waitfor() loop to ensure the application exits
immediately when the channel is already in a hung-up state.

### DIFF
--- a/apps/app_record.c
+++ b/apps/app_record.c
@@ -511,6 +511,12 @@ static int record_exec(struct ast_channel *chan, const char *data)
 	if (maxduration <= 0)
 		maxduration = -1;
 
+	if (ast_test_flag(ast_channel_flags(chan), AST_FLAG_ZOMBIE) || ast_check_hangup(chan)) {
+		ast_debug(1, "Channel hangup detected before ast_waitfor().\n");
+		status_response = "HANGUP";
+		res = -1;
+		goto out;
+	}
 	start = ast_tvnow();
 	while ((ms = ast_remaining_ms(start, maxduration))) {
 		ms = ast_waitfor(chan, ms);


jcolp:
Per the AI policy you must disclose usage of it:

https://github.com/asterisk/documentation/blob/main/docs/Development/Policies-and-Procedures/AI-Policy.md

silverjoe:
> Per the AI policy you must disclose usage of it:
> 
> https://github.com/asterisk/documentation/blob/main/docs/Development/Policies-and-Procedures/AI-Policy.md

I used AI to help improve the wording of the PR description, since English is not my native language.

The issue analysis, debugging, and the actual code change were done by me and tested in my environment.

silverjoe:
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: master

InterLinked1 (changes requested):
We should probably be handling the failure of `ast_streamfile` directly. Right now the code assumes the audio file didn't exist, but if a hangup occurred, that should probably be dealt with there.

Also, BYE is SIP terminology, so replace "BYE" with "hangup" in your commit message.

phoneben:
> cherry-pick-to: 21 cherry-pick-to: 22
> cherry-pick-to: 21 cherry-pick-to: 22

why not 20?

jcolp:
The current supported branches:

20, 22, 23, and master

This change will not go into 21.

silverjoe:
> We should probably be handling the failure of `ast_streamfile` directly. Right now the code assumes the audio file didn't exist, but if a hangup occurred, that should probably be dealt with there.

The audio file is created correctly. But no media frames arrive, actually no frames at all. So the loop just waits until ast_waitfor() finishes the waiting time.


> Also, BYE is SIP terminology, so replace "BYE" with "hangup" in your commit message.

Fixed.

silverjoe:
> The current supported branches:
> 
> 20, 22, 23, and master
> 
> This change will not go into 21.

I updated my comment with "cherry-pick-to".

InterLinked1:
> > We should probably be handling the failure of `ast_streamfile` directly. Right now the code assumes the audio file didn't exist, but if a hangup occurred, that should probably be dealt with there.
> 
> The audio file is created correctly. But no media frames arrive, actually no frames at all. So the loop just waits until ast_waitfor() finishes the waiting time.

So what does `ast_waitstream` return in this case?

silverjoe:
> > cherry-pick-to: 21 cherry-pick-to: 22
> > cherry-pick-to: 21 cherry-pick-to: 22
> 
> why not 20?

This PR is for branch 20.

phoneben:
> cherry-pick-to: 22
> cherry-pick-to: 23
> cherry-pick-to: master

it should look like this  because the pr is against master, and we also cherry-picking to other branches 

```
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 20
```

silverjoe:
> > > We should probably be handling the failure of `ast_streamfile` directly. Right now the code assumes the audio file didn't exist, but if a hangup occurred, that should probably be dealt with there.
> > 
> > 
> > The audio file is created correctly. But no media frames arrive, actually no frames at all. So the loop just waits until ast_waitfor() finishes the waiting time.
> 
> So what does `ast_waitstream` return in this case?

ast_waitstream returns the timeout value in milliseconds that we set in the maxduration argument.
You can try to reproduce it - it is really simple. The key point is that the hangup from your side must reach Asterisk at the 'beep' moment.

silverjoe:
> it should look like this because the pr is against master, and we also cherry-picking to other branches

O, thank you!

InterLinked1:
> > > > We should probably be handling the failure of `ast_streamfile` directly. Right now the code assumes the audio file didn't exist, but if a hangup occurred, that should probably be dealt with there.
> > > 
> > > 
> > > The audio file is created correctly. But no media frames arrive, actually no frames at all. So the loop just waits until ast_waitfor() finishes the waiting time.
> > 
> > 
> > So what does `ast_waitstream` return in this case?
> 
> ast_waitstream returns the timeout value in milliseconds that we set in the maxduration argument

Um, no, that's not what it's supposed to return: https://github.com/asterisk/asterisk/blob/c2ee154ed2897a8461fd2f39733a4b4f4c627bd1/include/asterisk/file.h#L211-L215

Did you even read the documentation?

My hunch is that some function that is already being called is returning -1 and we can simply catch that and abort, rather than doing an explicit check for hangup, which usually isn't needed.

silverjoe:
> > > > > We should probably be handling the failure of `ast_streamfile` directly. Right now the code assumes the audio file didn't exist, but if a hangup occurred, that should probably be dealt with there.
> > > > 
> > > > 
> > > > The audio file is created correctly. But no media frames arrive, actually no frames at all. So the loop just waits until ast_waitfor() finishes the waiting time.
> > > 
> > > 
> > > So what does `ast_waitstream` return in this case?
> > 
> > 
> > ast_waitstream returns the timeout value in milliseconds that we set in the maxduration argument
> 
> Um, no, that's not what it's supposed to return:
> 
> https://github.com/asterisk/asterisk/blob/c2ee154ed2897a8461fd2f39733a4b4f4c627bd1/include/asterisk/file.h#L211-L215
> 
> Did you even read the documentation?

I are talking about the ast_waitfor() function.
```
 * \brief Wait for input on a channel
 * \param chan channel to wait on
 * \param ms length of time to wait on the channel
 * \details
 * Wait for input on a channel for a given # of milliseconds (<0 for indefinite).
 * \retval < 0 on failure
 * \retval 0 if nothing ever arrived
 * \retval the # of ms remaining otherwise
 */
int ast_waitfor(struct ast_channel *chan, int ms);
```

> My hunch is that some function that is already being called is returning -1 and we can simply catch that and abort, rather than doing an explicit check for hangup, which usually isn't needed.

I did not see any function returning -1 by the time the loop starts and the operation ms = ast_waitfor(chan, ms); is called.

asteriskteam:

silverjoe:
Guys, what do you need from me to proceed with this pull request?
@jcolp @InterLinked1

InterLinked1:
> Guys, what do you need from me to proceed with this pull request? @jcolp @InterLinked1

So I just tested this, I'm able to reproduce the problem (I swapped out "beep" for a longer file to make it easier).

`ast_streamfile` does return 0, interestingly, upon user hangup during the audio file, which seems odd to me. I had assumed it would return -1 if hangup occurred while playing the file.

However, I also added the check you made in the `if (!ast_test_flag(&flags, OPTION_QUIET)) {` block, right after `ast_stopstream`, and indeed that condition is also true there. So while your current solution works and is probably fine, I still wonder if handling failure / hangup of the audio file specifically is a more targeted place to be handling this. I'm assuming you don't ever encounter this issue with the `q` option?

silverjoe:
> I'm assuming you don't ever encounter this issue with the `q` option?

I think it’s also possible to run into this scenario with the q option, but it’s less likely.

gtjoseph:
> Guys, what do you need from me to proceed with this pull request?

For a start, per the PR Checklist, you need to make the PR description match the commit message.  You should probably create a new issue with the details from the PR description then update the commit message with a "Resolves" trailer then copy the commit message to the PR description.

See:

asteriskteam:

asteriskteam:

jcolp:
There are now two independent reviews up. As the other one is more active I am closing this one out, though it can be reopened if needed.

Issue #1657, opened 2025-12-17
[bug]: Wrong dtmf payload is used when inbound invite contains 8K and 16K, and outgoing leg is using G722 and SRTP


pnlarsson:
### Severity

Minor

### Versions

20.17.0, 22.7.0

### Components/Modules

res_pjsip_sdp_rtp, res_rtp_asterisk

### Operating Environment

Debian 12, asterisk compiled from source without and patches

### Frequency of Occurrence

Constant

### Issue Description

This is an intresting one. Inbound Invite (snipped sdp):

```
a=rtpmap:104 AMR-WB/16000/1
a=fmtp:104 mode-change-capability=2; max-red=0
a=rtpmap:110 EVS/16000
a=fmtp:110 max-red=0; br=5.9-24.4
a=rtpmap:9 G722/8000
a=rtpmap:102 AMR/8000/1
a=fmtp:102 mode-change-capability=2; max-red=0
a=rtpmap:8 PCMA/8000
a=rtpmap:96 telephone-event/16000
a=fmtp:96 0-15
a=rtpmap:97 telephone-event/8000
a=fmtp:97 0-15
```

When outgoing leg is using G722 and SRTP (enough with media_encryption_optimistic=yes, so no SRTP is used), rtpevent has payload 96 even though the 200OK has 97. If we change to media_encryption=no or used alaw the payload is correct (97).

### Relevant log output

```shell
pjsip.conf:

[global]
type=global
endpoint_identifier_order=header,ip,username,anonymous

[transport-udp]
type=transport
protocol=udp
bind=0.0.0.0:5060

[siptrunk-basic](!)
type=endpoint
transport=transport-udp
context=from-siptrunk
disallow=all
allow=g722,alaw
aors=siptrunk
direct_media=no
send_diversion=no
t38_udptl=yes
t38_udptl_maxdatagram=400
t38_udptl_ec=redundancy
allow_transfer=no
send_rpid=no
send_pai=no
language=se
dtmf_mode=auto
allow_transfer=no
allow_subscribe=no
ice_support=no
use_avpf=no
identify_by=ip
moh_passthrough=yes
preferred_codec_only=yes

[siptrunk](siptrunk-basic)
type=endpoint
aors=siptrunk

[siptrunk]
type=aor
contact=sip:X.X.X.X:5060

[siptrunk]
type=identify
endpoint=siptrunk
match=X.X.X.X

[aor-single-reg](!)
type=aor
max_contacts=1
[...]

PR #1791, opened 2026-02-23
res_rtp_asterisk: use correct sample rate lookup to account for g722

mbradeen:
Swap out ast_rtp_get_rate for ast_format_get_sample_rate when looking
at the paired audio codec rate to account for g722 oddness.

Resolves: #1657

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -4391,15 +4391,19 @@ static int ast_rtp_dtmf_begin(struct ast_rtp_instance *instance, char digit)
 		return -1;
 	}
 
+
+	/* g722 is a 16K codec that masquerades as an 8K codec within RTP. ast_rtp_get_rate was written specifically to
+	   handle this. If we use the actual sample rate of g722 in this scenario and there is a 16K telephone-event on
+	   offer, we will end up using that instead of the 8K rate telephone-event that is expected with g722. */
 	if (rtp->lasttxformat == ast_format_none) {
 		/* No audio frames have been written yet so we have to lookup both the preferred payload type and bitrate. */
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

seanbright:
I’d love a bit more detail on how/why this fixes things.

res/res_rtp_asterisk.c
```diff
@@ -4399,7 +4399,7 @@ static int ast_rtp_dtmf_begin(struct ast_rtp_instance *instance, char digit)
 			sample_rate = ast_format_get_sample_rate(payload_format);
 		}
 	} else {
-		sample_rate = ast_format_get_sample_rate(rtp->lasttxformat);
+		sample_rate = ast_rtp_get_rate(rtp->lasttxformat);
```
jcolp, line 4406: I think a comment should be added here to explain the mess of G.722 sample rate and why ast_rtp_get_rate is used instead, so that it answers both @seanbright and any future readers.

mbradeen:
> I’d love a bit more detail on how/why this fixes things.

I've added a comment to the code that hopefully explains why we have to use ast_rtp_get_rate

res/res_rtp_asterisk.c
```diff
@@ -4391,15 +4391,19 @@ static int ast_rtp_dtmf_begin(struct ast_rtp_instance *instance, char digit)
 		return -1;
 	}
 
+
+	/* g722 is a 16K codec that masquerades as an 8K codec. ast_rtp_get_rate was written specifically to handle this.
```
jcolp: as an 8K codec within RTP.

Issue #1799, opened 2026-02-27
[new-feature]: Add path based ACLs for http requests and user based ACLs for ARI requests


mbradeen:
### Feature Description

Add the ability to configure an acl per uri prefix to http.conf and an acl per user in ari.conf

PR #1800, opened 2026-02-27
acl: Add ACL support to http and ari

mbradeen:
Add uri prefix based acl support to the built in http server.
This allows an acl to be added per uri prefix (ie '/metrics'
or '/ws') to restrict access.

Add user based acl support for ARI. This adds new acl options
to the user section of ari.conf to restrict access on a per
user basis.

resolves: #1799

UserNote: A new section, type=restriction has been added to http.conf
to allow an uri prefix based acl to be configured. See
http.conf.sample for examples and more information.
The user section of ari.conf can now contain an acl configuration
to restrict users access. See ari.conf.sample for examples and more
information

### DIFF
--- a/configs/samples/ari.conf.sample
+++ b/configs/samples/ari.conf.sample
@@ -35,6 +35,22 @@ enabled = yes       ; When set to no, ARI support is disabled.
 ; When set to plain, the password is in plaintext.
 ;
 ;password_format = plain
+;
+; The following three options (permit, deny, acl) allow for a per-user acl to be
+; configured.  The format follows the rules as documented in acl.conf.sample
+;
+; If no restriction is defined for a given user, no IPs will be blocked by
+; Asterisk (legacy behavior).
+;
+;deny =             ; Deny acces from the subnet for the given user
+;permit =           ; Permit access from the subnet(s) for the given user
+;acl =              ; Optional name for the acl.
+;
+;Example:
+;deny = 0.0.0.0/0
+;permit = 127.0.0.1,10.0.0.0/24
+;acl = localasteriskuser
+;
 
 ; Outbound Websocket Connections
 ;
--- a/configs/samples/http.conf.sample
+++ b/configs/samples/http.conf.sample
@@ -130,3 +130,41 @@ bindaddr=127.0.0.1
 ; POST URL: /asterisk/uploads will put files in /var/lib/asterisk/uploads/.
 ;uploads = /var/lib/asterisk/uploads/
 ;
+
+;[uripath]
+;
+;type = restriction ; Specifies acl configuration
+;
+; The following options (permit, deny, acl) allow for an acl to be configured
+; on a per uri prefix basis. The first character should be an '/'
+;
+; The format follows the rules as documented in acl.conf.sample
+;
+; If no restriction is defined for a given prefix, legacy behavior will apply.
+;
+; If multiple restrictions apply, any restriction that denies will supersede
+; any another restrictions that permit.  For example if an /ari restriction
+; results in a deny, but an /ari/channels restriction would permit, the
+; attempt would still be denied.
+;
+;deny =             ; Deny the subnet access for the given user
+;permit =           ; Permit the subnet(s) access for the given user
+;acl =              ; Optional name for the acl.
+;
+;Examples:
+;
+; Only allow ari connections from localhost:
+;
+;[/ari]
+;type = restriction
+;deny = 0.0.0.0/0
+;permit = 127.0.0.1
+;acl = localarionly
+;
+; Only allow metrics to be gathered by 10.0.0.23
+;
+;[/metrics]
+;type = restriction
+;deny = 0.0.0.0/0
+;permit = 10.0.0.23
+;
--- a/main/http.c
+++ b/main/http.c
@@ -51,6 +51,7 @@
 #include <fcntl.h>
 
 #include "asterisk/paths.h"	/* use ast_config_AST_DATA_DIR */
+#include "asterisk/acl.h"
 #include "asterisk/cli.h"
 #include "asterisk/tcptls.h"
 #include "asterisk/http.h"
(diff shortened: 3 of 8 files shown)


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
testsuite-test-pr: 126

main/http.c
```diff
@@ -177,6 +178,17 @@ struct http_uri_redirect {
 
 static AST_RWLIST_HEAD_STATIC(uri_redirects, http_uri_redirect);
 
+/*! \brief Per-path ACL restriction */
+struct http_restriction {
+	AST_LIST_ENTRY(http_restriction) entry;
+	struct ast_acl_list *acl;
+	char path[0];
```
InterLinked1: `char path[];`

main/http.c
```diff
@@ -2484,10 +2533,12 @@ static int __ast_http_load(int reload)
 	char newprefix[MAX_PREFIX] = "";
 	char server_name[MAX_SERVER_NAME_LENGTH];
 	struct http_uri_redirect *redirect;
+	struct http_restriction *restriction;
 	struct ast_flags config_flags = { reload ? CONFIG_FLAG_FILEUNCHANGED : 0 };
 	uint32_t bindport = DEFAULT_PORT;
 	int http_tls_was_enabled = 0;
 	char *bindaddr = NULL;
+	char *cat = NULL;
```
InterLinked1: Can this be a const char*?
I'm thinking it will cause compilation failure on gcc 15.2.1 otherwise

main/http.c
```diff
@@ -2602,6 +2662,50 @@ static int __ast_http_load(int reload)
...
+				}
+			}
+		}
+
+		new_restriction->acl = acl;
+
+		AST_RWLIST_WRLOCK(&restrictions);
+		AST_RWLIST_INSERT_TAIL(&restrictions, new_restriction, entry);
+		AST_RWLIST_UNLOCK(&restrictions);
+		ast_verb(3, "HTTP: Added restriction for path '%s'\n", cat);
```
InterLinked1: Verbose 3 seems too high, I would make it a debug or at least higher verbose

InterLinked1 (changes requested):

main/http.c
```diff
@@ -2602,6 +2662,50 @@ static int __ast_http_load(int reload)
...
+		}
+
+		type = ast_variable_retrieve(cfg, cat, "type");
+		if (!type || strcasecmp(type, "restriction") != 0) {
+			continue;
+		}
+
+		new_restriction = ast_calloc(1, sizeof(*new_restriction) + strlen(cat) + 1);
+		if (!new_restriction) {
+			ast_log(LOG_ERROR, "Failed to allocate restriction for path '%s'\n", cat);
```
InterLinked1: Logging an error for allocation failures is duplicative so this can be removed

InterLinked1: Logging an error for allocation failures is duplicative so this can be removed

InterLinked1 (changes requested):

configs/samples/http.conf.sample
```diff
@@ -130,3 +130,37 @@ bindaddr=127.0.0.1
...
+;type = restriction ; Specifies acl configuration
+;
+; The following options (permit, deny, acl) allow for an acl to be configured
+; on a per uri prefix basis. The first character should be an '/'
+;
+; The format follows the rules as documented in acl.conf.sample
+;
+; If no restriction is defined for a given prefix, no IPs will be blocked by
+; Asterisk (legacy behavior).
+;
```
jcolp, line 144: Document the behavior if subpaths are configured as well. For example a rule for /ari and one for /ari/channels

main/http.c
```diff
@@ -2527,6 +2578,15 @@ static int __ast_http_load(int reload)
 	}
 	AST_RWLIST_UNLOCK(&uri_redirects);
 
+	AST_RWLIST_WRLOCK(&restrictions);
```
jcolp: I believe this logic would temporarily provide a short period of time where ACLs do not exist, and thus are not enforced allowing access to potentially things that shouldn't be. You probably instead want to build up a temporary linked list of the current configuration, and once loading is done do the swap over with the lock on restrictions held.

main/http.c
```diff
@@ -2602,6 +2662,49 @@ static int __ast_http_load(int reload)
...
+		struct ast_acl_list *acl = NULL;
+		int acl_error = 0;
+		int acl_subscription_flag = 0;
+
+		if (strcasecmp(cat, "general") == 0) {
+			continue;
+		}
+
+		type = ast_variable_retrieve(cfg, cat, "type");
+		if (!type || strcasecmp(type, "restriction") != 0) {
```
jcolp, line 2672: We use !strcasecmp

main/http.c
```diff
@@ -2602,6 +2662,49 @@ static int __ast_http_load(int reload)
...
+			continue;
+		}
+
+		new_restriction = ast_calloc(1, sizeof(*new_restriction) + strlen(cat) + 1);
+		if (!new_restriction) {
+			continue;
+		}
+
+		strcpy(new_restriction->path, cat);
```
jcolp, line 2682: Mark as /* Safe */

configs/samples/http.conf.sample
mbradeen: Currently, if _any_ acl results in a deny, that supersedes any other rule that might permit.  Is this the way we want it to work, or should we look at the overall path length and use the 'best' match as the deciding rule?

jcolp: I think any is best, it just needs to be explicitly documented

configs/samples/http.conf.sample
```diff
@@ -130,3 +130,42 @@ bindaddr=127.0.0.1
...
+; If multiple restrictions apply, any restriction that denies will supersede
+; any another restrictions that permit.  For example if an /ari retstriction
```
jcolp: restriction

jcolp (changes requested):

configs/samples/ari.conf.sample
```diff
@@ -35,6 +35,22 @@ enabled = yes       ; When set to no, ARI support is disabled.
 ; When set to plain, the password is in plaintext.
 ;
 ;password_format = plain
+;
+; The following three options (permit, deny, acl) allow for a per-user acl to be
+; configured.  The format follows the rules as documented in acl.conf.sample
+;
+; If no restriction is defined for a given user, no IPs will be blocked by
+; Asterisk (legacy behavior).
```
InterLinked1, line 43: I feel like this could be worded a little more clearly.
Maybe "If no restriction is defined for a given user, the user will be allowed to connect from any IP"?

("No IPs will be blocked" could be misinterpreted to have a global meaning when it doesn't)

configs/samples/ari.conf.sample
```diff
@@ -35,6 +35,22 @@ enabled = yes       ; When set to no, ARI support is disabled.
...
+;
+;deny =             ; Deny the subnet access for the given user
+;permit =           ; Permit the subnet(s) access for the given user
```
InterLinked1: These phrasings seem backwards to me, the subnet isn't taking any action, the user is... e.g. "Permit user access from specified subnet(s)" would be clearer

Issue #1797, opened 2026-02-27
[bug]: Potential logic issue in translated frame write loop (main/file.c)


Tinet-mucw:
### Severity

Trivial

### Versions

centos8

### Components/Modules

main

### Operating Environment

centos8

### Frequency of Occurrence

None

### Issue Description

```c

int ast_writestream(struct ast_filestream *fs, struct ast_frame *f)
{
....
		if (!fs->trans) {
			ast_log(LOG_WARNING, "Unable to translate to format %s, source format %s\n",
				fs->fmt->name, ast_format_get_name(f->subclass.format));
		} else {
			struct ast_frame *trf;
			ao2_replace(fs->lastwriteformat, f->subclass.format);
			/* Get the translated frame but don't consume the original in case they're using it on another stream */
			if ((trf = ast_translate(fs->trans, f, 0))) {
				struct ast_frame *cur;

				/* the translator may have returned multiple frames, so process them */
				for (cur = trf; cur; cur = AST_LIST_NEXT(cur, frame_list)) {
					if ((res = fs->fmt->write(fs, trf))) {
						ast_log(LOG_WARNING, "Translated frame write failed\n");
						break;
					}
				}
				ast_frfree(trf);
			} else {
				res = 0;
			}
		}

````

In main/file.c, there appears to be a likely logic issue in the translated-frame write loop.
Inside the loop, the code iterates with cur but writes trf; it may be more correct to write cur for each iteration.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
@Tinet-mucw Will you be submitting a pull request to fix this?

PR #1801, opened 2026-02-28
main/file: fix translated-frame write loop to use current frame

Tinet-mucw:
write each translated frame from translator output.

Resolves: #1797


Tinet-mucw:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1804, opened 2026-03-01
[new-feature]: dsp.c: Add support for R2 signaling


InterLinked1:
### Feature Description

Add ability to detect R2 signaling tones in Asterisk

PR #1805, opened 2026-03-01
dsp.c: Add support for detecting R2 signaling tones.

InterLinked1:
Extend the existing DTMF/MF tone detection support by adding support for R2 tones, another variant of MF (R1) signaling. Both forward and backward signaling are supported.

Resolves: #1804


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1795, opened 2026-02-26
[bug]: DTLS packets dropped when TURN configured in rtp.conf due to loopback re-injection occurring before ICE candidate source check


robertwilson88:
### Severity

Minor

### Versions

20.15.2

### Components/Modules

res_rtp_asterisk.so

### Operating Environment

CentOS Stream 9 / RHEL x86_64

### Frequency of Occurrence

Constant

### Issue Description

When TURN is configured within Asterisk via rtp.conf (turnaddr, turnusername, turnpassword), all DTLS packets are dropped with the following warning:
WARNING: res_rtp_asterisk.c:3315 __rtp_recvfrom: DTLS packet from 127.0.0.1:port dropped. Source not in ICE active candidate list.

A tcpdump on the Asterisk loopback interface confirms that pjproject receives incoming TURN relay packets on eth0, decapsulates them, and re-injects the payload internally via the loopback interface (127.0.0.1). The socket is confirmed as owned by the Asterisk process via ss -ulnp.
In __rtp_recvfrom() within res/res_rtp_asterisk.c, the DTLS candidate source check at line 3315 fires before the loopback address substitution code runs. By the time Asterisk checks whether the source address is in the ICE active candidate list it sees 127.0.0.1 rather than the real remote address, fails the check, and drops the packet.
Patch:
The fix is to add a loopback substitution block immediately after the ast_debug_dtls line and before the #ifdef HAVE_PJPROJECT ICE candidate check:
`ast_debug_dtls(3, "(%p) DTLS - __rtp_recvfrom rtp=%p - Got SSL packet '%d'\n", instance, rtp, *in);

#ifdef HAVE_PJPROJECT
		/* If this packet arrived via TURN/ICE loopback re-injection,
		 * substitute the real remote address before the candidate check
		 * otherwise the DTLS check will see 127.0.0.1 and drop the packet.
		 */
		if (!ast_sockaddr_isnull(&rtp->rtp_loop) && !ast_sockaddr_cmp(&rtp->rtp_loop, sa)) {
			ast_rtp_instance_get_remote_address(instance, sa);
		} else if (rtcp && !ast_sockaddr_isnull(&rtp->rtcp_loop) && !ast_sockaddr_cmp(&rtp->rtcp_loop, sa)) {
			ast_sockaddr_copy(sa, &rtp->rtcp->them);
		}
#endif

		/*
		 * If ICE is in use, we can prevent a possible DOS attack...
		 */
`
[...]

PR #1809, opened 2026-03-03
res_rtp_asterisk.c: Fix DTLS packet drop when TURN loopback re-injection occurs before ICE candidate check

robertwilson88:
When TURN is configured in rtp.conf, pjproject re-injects TURN packets
via 127.0.0.1 (the loopback address). The DTLS packet handler checks the
source address against the ICE active candidate list before the loopback
address substitution runs, causing the packet to be silently dropped as
the source 127.0.0.1 is not in the candidate list.

Fix by performing the loopback address substitution before the ICE
candidate source check in the DTLS path, mirroring the logic already
present in the non-DTLS RTP path.

Fixes: #1795

UserNote: WebRTC calls using TURN configured in rtp.conf (turnaddr,
turnusername, turnpassword) will now correctly complete DTLS/SRTP
negotiation. Previously all DTLS packets were silently dropped due to
the loopback re-injection address not being in the ICE active candidate
list.

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -3267,6 +3267,18 @@ static int __rtp_recvfrom(struct ast_rtp_instance *instance, void *buf, size_t s
 
 		ast_debug_dtls(3, "(%p) DTLS - __rtp_recvfrom rtp=%p - Got SSL packet '%d'\n", instance, rtp, *in);
 
+#ifdef HAVE_PJPROJECT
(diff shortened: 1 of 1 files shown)


robertwilson88:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
Asterisk 21 is security fix only. It has been removed from the cherry pick list.

robertwilson88:
Description updated to match commit

Issue #1814, opened 2026-03-06
[bug]: A pjsip transport with an invalid config can cause issues with other transports


gtjoseph:
### Severity

Minor

### Versions

All

### Components/Modules

res_pjsip, res_pjsip_nat, res_pjsip_refer

### Operating Environment

Any

### Frequency of Occurrence

Frequent

### Issue Description

If a pjsip transport (A) fails to load, its temporary state gets left behind causing the next transport (B) to load to pick up some of its parameters, including its name.  This can cause B to have the correct name (B) in its transport object but the wrong name (A) in its internal state object.  When a transport state is searched for later on, transport state B is returned but a retrieval of the actual transport object will fail because B's transport state _id_ is actually "A".

Currently, searching for transport state is done by res_pjsip_nat and res_pjsip_refer so operations they perform could be affected.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

seanbright:
I remember @Kobaz bringing this up on IRC a few years ago.

gtjoseph:
> I remember [@kobaz](https://github.com/kobaz) bringing this up on IRC a few years ago.

Hmmm.  I wonder why we didn't pursue it.  Was it just an anecdotal "nat doesn't work sometimes" or more than that?

jcolp:
It was:

https://github.com/asterisk/asterisk/issues/293

Which looks like got auto closed as there wasn't a follow-up after I said I couldn't reproduce it with the details given

gtjoseph:
Well, what do you know.  I could probably look at #436 also while I'm here.

seanbright:
And #867? And and and and…

gtjoseph:
possibly

gtjoseph:
I haven't been able to reproduce either #436 or #867 in current codebase so maybe they got fixed along the way.  I'll ask the OPs to re-confirm.

PR #1815, opened 2026-03-06
res_pjsip: Remove temp transport state when a transport fails to load.

gtjoseph:
If a pjsip transport (A) fails to load, its temporary state gets left behind
causing the next transport to load (B) to pick up some of its parameters,
including its name. This can cause B to have the correct name (B) in its
transport object but the wrong name (A) in its internal state object. When a
transport state is searched for later on, transport state B is returned but a
retrieval of the actual transport object will fail because B's transport
state id is actually "A" and transport "A" doesn't exist because it failed
to load.

remove_temporary_state() is now being called in all error paths in
config_transport.c functions that call find_or_create_temporary_state().

A bit of extra debugging was also added to res_pjsip_nat.c.

Resolves: #1814


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
Well, that was embarrassing.

Issue #1816, opened 2026-03-06
[bug]: ARI: RTPAUDIO channel vars aren't set if call hung up by ARI.


gtjoseph:
### Severity

Minor

### Versions

All

### Components/Modules

res_ari

### Operating Environment

All

### Frequency of Occurrence

Frequent

### Issue Description

If a channel is hung up by ARI, the channel leaves Stasis before the RTPAUDIO channel variables are set so the app doesn't get the VarSet events.  If the channel is hung up by the channel driver, then the variables ARE set before the channel leaves Statis and the app _will_ get the VarSet messages.

Subscribing to all events solves the issue but the app is then flooded with every event on the system which is far from ideal.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1817, opened 2026-03-06
rtp: Set RTPAUDIOQOS variables when ast_softhangup is called.

gtjoseph:
If a channel in Stasis/ARI is hung up by the channel driver, the RTPAUDIOQOS
variables are set before the channel leaves Stasis and are therefore
available to the ARI app via ChannelVarset events.  If the channel is hung up
by ARI however, the channel leaves Stasis before the RTPAUDIOQOS variables
are set so the app may not get the ChannelVarset events.

We now set the RTPAUDIOQOS variables when ast_softhangup() is called as well
as when the channel driver hangs up a channel.  Since ARI hangups call
ast_softhangup(), the variables will be set before the channel leaves Stasis
and the app should get the ChannelVarset events.
ast_rtp_instance_set_stats_vars(), which actually sets the variables, now
checks to see if the variables are already set before attempting to set them.
This prevents double messages from being generated.

Resolves: #1816

### DIFF
--- a/main/channel.c
+++ b/main/channel.c
@@ -75,6 +75,7 @@
 #include "asterisk/max_forwards.h"
 #include "asterisk/stream.h"
 #include "asterisk/message.h"
+#include "asterisk/rtp_engine.h"
 
 #include "channelstorage.h"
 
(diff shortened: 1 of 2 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

PR #1820, opened 2026-03-10
res_pjsip_maintenance: Add PJSIP endpoint maintenance mode

danieldonoghue:
Introduces res_pjsip_maintenance, a loadable module that allows
operators to place individual PJSIP endpoints into maintenance mode
at runtime without unregistering or disabling them.

While an endpoint is in maintenance mode:
 * New inbound INVITE and SUBSCRIBE dialogs are rejected with
   503 Service Unavailable and a Retry-After: 300 header.
 * In-progress dialogs (re-INVITE, UPDATE, BYE, etc.) are
   unaffected and complete normally.
 * Outbound originations via Dial() or ARI originate are refused
   before any SIP session is created.

State is held in-memory only and is cleared on module unload
or Asterisk restart.

This module was developed with AI assistance (Claude).  All code
has been reviewed and tested by the author, who takes full
responsibility for the submission.

CLI interface:
  pjsip set maintenance <on|off> <endpoint|all>
  pjsip show maintenance [endpoint]

AMI interface:
  Action: PJSIPSetMaintenance
  Endpoint: <name>|all
  State: on|off

  Action: PJSIPShowMaintenance
  Endpoint: <name>  (optional; omit to list all)

  Emits PJSIPMaintenanceStatus events per result, followed by
  PJSIPMaintenanceStatusComplete. State changes also emit an
  unsolicited PJSIPMaintenanceStatus event.

To support outbound blocking, a new session_create callback is
added to ast_sip_session_supplement. Supplements that set this
callback are invoked at the start of ast_sip_session_create_outgoing()
in res_pjsip_session, before any dialog or invite session resources
are allocated. res_pjsip_maintenance registers itself as a session
supplement and uses this callback to gate outbound session creation
on a per-endpoint basis.

MODULEINFO:
  <depend>pjproject</depend>
  <depend>res_pjsip</depend>
  <depend>res_pjsip_session</depend>

UserNote: New module res_pjsip_maintenance adds runtime maintenance
mode for PJSIP endpoints. Use "pjsip set maintenance <on|off>
<endpoint|all>" to enable or disable, and "pjsip show maintenance"
to list affected endpoints. AMI actions PJSIPSetMaintenance and
PJSIPShowMaintenance provide programmatic access. No configuration
file changes required.

DeveloperNote: ast_sip_session_supplement gains a new optional
callback - int (*session_create)(struct ast_sip_endpoint *endpoint,
const char *destination). It is called from the global supplement
list (not per-session) at the start of ast_sip_session_create_outgoing()
via ast_sip_session_check_supplement_create(). Returning non-zero
blocks the outgoing session. Modules that need to gate outbound
SIP session creation should register a supplement with this callback
set rather than hooking into chan_pjsip directly.

### DIFF
--- a/include/asterisk/res_pjsip_session.h
+++ b/include/asterisk/res_pjsip_session.h
@@ -364,6 +364,31 @@ struct ast_sip_session_supplement {
 	 * Defaults to AST_SIP_SESSION_BEFORE_MEDIA
 	 */
 	enum ast_sip_session_response_priority response_priority;
+	/*!
+	 * \brief Called before an outgoing session is created
+	 *
+	 * This is called before the session dialog is created and can be used to
+	 * block the creation of the session entirely.  A non-zero return value
+	 * prevents the session from being created.  The callback is called from
+	 * the global supplement list, not per-session, so the session does not
+	 * yet exist when this is called.
+	 *
+	 * \since 20.20.0
+	 * \since 22.10.0
+	 * \since 23.4.0
+	 *
+	 * \param endpoint The endpoint the outgoing session would be created for
+	 * \param contact The contact to use for the outgoing session, or NULL
+	 * \param location Name of the location to call, be it named location or explicit URI, or NULL
+	 * \param request_user Optional request user to place in the request URI, or NULL
+	 * \param req_topology The requested stream capabilities
+	 *
+	 * \retval non-zero Block session creation
+	 * \retval 0 Allow session creation
+	 */
+	int (*session_create)(struct ast_sip_endpoint *endpoint,
+		struct ast_sip_contact *contact, const char *location,
+		const char *request_user, struct ast_stream_topology *req_topology);
 };
 
 enum ast_sip_session_sdp_stream_defer {
@@ -624,6 +649,33 @@ void ast_sip_session_register_supplement_with_module(struct ast_module *module, 
  * \param supplement The supplement to unregister
  */
 void ast_sip_session_unregister_supplement(struct ast_sip_session_supplement *supplement);
+
+/*!
+ * \brief Check registered supplements for permission to create an outgoing session
+ *
+ * Iterates the global supplement list and calls any registered \c session_create
+ * callbacks.  The first callback to return a non-zero value stops the iteration
+ * and causes this function to return -1, blocking the session creation.
+ *
+ * This is called at the beginning of ast_sip_session_create_outgoing() before
+ * any dialog or invite session resources are allocated.
+ *
+ * \since 20.20.0
+ * \since 22.10.0
+ * \since 23.4.0
+ *
+ * \param endpoint The endpoint the outgoing session would be created for
+ * \param contact The contact to use for the outgoing session, or NULL
+ * \param location Name of the location to call, be it named location or explicit URI, or NULL
+ * \param request_user Optional request user to place in the request URI, or NULL
+ * \param req_topology The requested stream capabilities
+ *
+ * \retval 0 Session creation is allowed
+ * \retval -1 Session creation is blocked by a supplement
+ */
+int ast_sip_session_check_supplement_create(struct ast_sip_endpoint *endpoint,
+	struct ast_sip_contact *contact, const char *location,
+	const char *request_user, struct ast_stream_topology *req_topology);
 
 /*!
  * \brief Add supplements to a SIP session
--- a/res/res_pjsip/pjsip_session.c
+++ b/res/res_pjsip/pjsip_session.c
@@ -108,6 +108,23 @@ int ast_sip_session_add_supplements(struct ast_sip_session *session)
 	return 0;
 }
 
+int ast_sip_session_check_supplement_create(struct ast_sip_endpoint *endpoint,
+	struct ast_sip_contact *contact, const char *location,
+	const char *request_user, struct ast_stream_topology *req_topology)
+{
+	struct ast_sip_session_supplement *iter;
+	SCOPED_LOCK(lock, &session_supplements, AST_RWLIST_RDLOCK, AST_RWLIST_UNLOCK);
+
+	AST_RWLIST_TRAVERSE(&session_supplements, iter, next) {
+		if (iter->session_create && iter->session_create(endpoint, contact, location,
+				request_user, req_topology)) {
+			return -1;
+		}
+	}
+
+	return 0;
+}
+
 void ast_sip_session_remove_supplements(struct ast_sip_session *session)
 {
 	struct ast_sip_session_supplement *iter;
--- a/res/res_pjsip_maintenance.c
+++ b/res/res_pjsip_maintenance.c
@@ -1,0 +1,754 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2026, Aurora Innovation
+ *
+ * Daniel Donoghue
+ *
+ * See http://www.asterisk.org for more information about
+ * the Asterisk project. Please do not directly contact
+ * any of the maintainers of this project for assistance;
+ * the project provides a web site, mailing lists and IRC
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
+ * at the top of the source tree.
+ */
+
+/*!
+ * \file
+ * \brief PJSIP Endpoint Maintenance Mode
+ *
+ * Provides a runtime toggle to place individual PJSIP endpoints into
+ * maintenance mode.  While an endpoint is in maintenance mode:
+ *
+ * - New \b inbound out-of-dialog requests are rejected with
+ *   "503 Service Unavailable" and a Retry-After: 300 header
+ *   (except SUBSCRIBE/REGISTER with Expires: 0).
+ * - \b Outbound originations (Dial, ARI originate) are refused before
+ *   any SIP session or Asterisk channel is created.
+ * - Active in-progress dialogs (BYE, re-INVITE, UPDATE, etc.) are
+ *   completely unaffected.
+ * - Existing presence/BLF subscriptions are left to expire naturally.
+ *
+ * CLI:
+ *   pjsip set maintenance <on|off> <endpoint|all>
+ *   pjsip show maintenance [endpoint]
+ *
+ * AMI actions: PJSIPSetMaintenance, PJSIPShowMaintenance
+ *
+ * \ingroup res_pjsip
+ */
+
+/*** MODULEINFO
+	<depend>pjproject</depend>
+	<depend>res_pjsip</depend>
+	<depend>res_pjsip_session</depend>
+	<support_level>extended</support_level>
+ ***/
+
+/*** DOCUMENTATION
+	<manager name="PJSIPSetMaintenance" language="en_US">
+		<since>
+			<version>20.20.0</version>
+			<version>22.10.0</version>
+			<version>23.4.0</version>
+		</since>
+		<synopsis>
+			Enable or disable maintenance mode for a PJSIP endpoint.
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="Endpoint" required="true">
+				<para>The PJSIP endpoint name, or <literal>all</literal> to
+				toggle maintenance mode for every configured endpoint.</para>
+			</parameter>
+			<parameter name="State" required="true">
+				<para>Desired maintenance state.</para>
+				<enumlist>
+					<enum name="on" />
+					<enum name="off" />
+				</enumlist>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Enables or disables maintenance mode for the specified PJSIP
+			endpoint.  While in maintenance mode, new inbound out-of-dialog
+			requests are rejected with 503 Service Unavailable (except
+			SUBSCRIBE/REGISTER with Expires: 0), and outbound originations via
+			Dial() or ARI are refused before any SIP session or channel is
+			created.  In-progress dialogs are unaffected.</para>
+			<para>A <literal>PJSIPMaintenanceStatus</literal> event is emitted
+			when the state changes.</para>
+		</description>
+	</manager>
+	<manager name="PJSIPShowMaintenance" language="en_US">
+		<since>
+			<version>20.20.0</version>
+			<version>22.10.0</version>
+			<version>23.4.0</version>
+		</since>
+		<synopsis>
+			Show maintenance mode status for PJSIP endpoints.
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="Endpoint" required="false">
+				<para>If specified, show the status for this endpoint only.
+				If omitted, list all endpoints currently in maintenance
+				mode.</para>
+			</parameter>
+		</syntax>
+		<description>
+			<para>Emits one <literal>PJSIPMaintenanceStatus</literal> event
+			per result, followed by a
+			<literal>PJSIPMaintenanceStatusComplete</literal> event.</para>
+		</description>
+	</manager>
+	<managerEvent language="en_US" name="PJSIPMaintenanceStatus">
+		<managerEventInstance class="EVENT_FLAG_SYSTEM">
+			<since>
+				<version>20.20.0</version>
+				<version>22.10.0</version>
+				<version>23.4.0</version>
+			</since>
+			<synopsis>
+				Reports the maintenance mode state of a PJSIP endpoint.
+			</synopsis>
+			<syntax>
+				<parameter name="Endpoint">
+					<para>The PJSIP endpoint name.</para>
+				</parameter>
+				<parameter name="Status">
+					<para>Current maintenance state.</para>
+					<enumlist>
+						<enum name="enabled" />
+						<enum name="disabled" />
+					</enumlist>
+				</parameter>
+			</syntax>
+			<description>
+				<para>Emitted when an endpoint enters or leaves maintenance
+				mode, and as a list entry in response to
+				<literal>PJSIPShowMaintenance</literal>.</para>
+			</description>
+		</managerEventInstance>
+	</managerEvent>
+ ***/
+
+#include "asterisk.h"
+
+#include <pjsip.h>
+
+#include "asterisk/res_pjsip.h"
+#include "asterisk/res_pjsip_session.h"
+#include "asterisk/manager.h"
+#include "asterisk/module.h"
+#include "asterisk/logger.h"
+#include "asterisk/cli.h"
+#include "asterisk/sorcery.h"
+#include "asterisk/astobj2.h"
+#include "asterisk/strings.h"
+
+enum {
+	MAINT_HASH_BUCKETS = 53,
+};
+
+/*! Endpoints currently in maintenance mode.
+ *  Protected by the container's own internal RWLOCK.
+ *  No other locks are ever held simultaneously with this container.
+ */
+static struct ao2_container *maintenance_set;
+
+/*!
+ * \internal
+ * \brief Add an endpoint to the maintenance set.
+ * \retval  1  Added successfully.
+ * \retval  0  Already in maintenance (no-op).
+ * \retval -1  Allocation failure.
+ */
+static int maint_set_add(const char *endpoint_name)
+{
+	char *entry;
+
+	entry = ao2_find(maintenance_set, endpoint_name, OBJ_SEARCH_KEY);
+	if (entry) {
+		ao2_ref(entry, -1);
+		return 0; /* already in maintenance */
+	}
+	return ast_str_container_add(maintenance_set, endpoint_name) ? -1 : 1;
+}
+
+/*!
+ * \internal
+ * \brief Remove an endpoint from the maintenance set.
+ * \retval 1  Removed successfully.
+ * \retval 0  Was not in maintenance (no-op).
+ */
+static int maint_set_remove(const char *endpoint_name)
+{
+	char *entry;
+
+	entry = ao2_find(maintenance_set, endpoint_name, OBJ_SEARCH_KEY | OBJ_UNLINK);
+	if (!entry) {
+		return 0;
+	}
+	ao2_ref(entry, -1);
+	return 1;
+}
+
+/*!
+ * \internal
+ * \brief Apply a maintenance state change to the maintenance set.
+ *
+ * Does not validate endpoint existence; callers are responsible for that.
+ * Callers are also responsible for emitting log messages and AMI events.
+ *
+ * \retval  1  State changed.
+ * \retval  0  Already in requested state (no-op).
+ * \retval -1  Allocation failure (enable path only).
+ */
+static int apply_maintenance_state(const char *endpoint_name, int enable)
+{
+	return enable ? maint_set_add(endpoint_name) : maint_set_remove(endpoint_name);
+}
+
+/* Session supplement: block outgoing session creation when endpoint is in maintenance. */
+
+/*!
+ * \internal
+ * \brief Session supplement session_create callback: block outgoing sessions to
+ *        endpoints currently in maintenance mode.
+ * \retval  1  Endpoint is in maintenance; session creation blocked.
+ * \retval  0  Endpoint is not in maintenance; session creation allowed.
+ */
+static int maint_session_create(struct ast_sip_endpoint *endpoint,
+	struct ast_sip_contact *contact, const char *location,
+	const char *request_user, struct ast_stream_topology *req_topology)
+{
+	const char *endpoint_name = ast_sorcery_object_get_id(endpoint);
+	char *entry = ao2_find(maintenance_set, endpoint_name, OBJ_SEARCH_KEY);
+
+	if (entry) {
+		ao2_ref(entry, -1);
+		ast_log(LOG_NOTICE, "PJSIP: Refusing outbound call to endpoint '%s': maintenance mode active\n",
+			endpoint_name);
+		return 1;
+	}
+	return 0;
+}
+
+static struct ast_sip_session_supplement maintenance_session_supplement = {
+	.session_create = maint_session_create,
+	.priority = AST_SIP_SUPPLEMENT_PRIORITY_FIRST,
+};
+
+/* Inbound request hook for maintenance_pjsip_mod.
+ *
+ * For endpoints in maintenance mode, blocks new out-of-dialog requests
+ * with 503 + Retry-After: 300. Any in-dialog request is passed through
+ * unmodified. SUBSCRIBE and REGISTER with Expires: 0 are also passed
+ * through, allowing un-subscribe and de-registration. */
+
+static pj_bool_t maintenance_on_rx_request(pjsip_rx_data *rdata)
+{
+	pjsip_msg *msg = rdata->msg_info.msg;
+	const pjsip_method *method = &msg->line.req.method;
+	pjsip_to_hdr *to;
+	pjsip_expires_hdr *expires_hdr;
+	struct ast_sip_endpoint *endpoint;
+	char *entry;
+	pjsip_hdr hdr_list;
+	pjsip_generic_int_hdr *retry_after;
+	static const pj_str_t str_retry_after = { "Retry-After", 11 };
+	int is_subscribe;
+	int is_register;
+
+	is_subscribe = pjsip_method_cmp(method, pjsip_get_subscribe_method()) == 0;
+	is_register = pjsip_method_cmp(method, pjsip_get_register_method()) == 0;
+
+	/* Any in-dialog request is always allowed through. */
+	to = rdata->msg_info.to;
+	if (to->tag.slen > 0) {
+		return PJ_FALSE;
+	}
+
+	/* SUBSCRIBE or REGISTER with Expires: 0: allow un-subscribe / de-register. */
+	if (is_subscribe || is_register) {
+		expires_hdr = pjsip_msg_find_hdr(msg, PJSIP_H_EXPIRES, NULL);
+		if (expires_hdr && expires_hdr->ivalue == 0) {
+			return PJ_FALSE;
+		}
+	}
+
+	endpoint = ast_pjsip_rdata_get_endpoint(rdata);
+	if (!endpoint) {
+		return PJ_FALSE;
+	}
+
+	entry = ao2_find(maintenance_set, ast_sorcery_object_get_id(endpoint), OBJ_SEARCH_KEY);
+	if (!entry) {
+		ao2_ref(endpoint, -1);
+		return PJ_FALSE;
+	}
+	ao2_ref(entry, -1);
+
+	ast_log(LOG_NOTICE, "PJSIP: Endpoint '%s' is in maintenance mode; rejecting new %.*s from %s\n",
+		ast_sorcery_object_get_id(endpoint),
+		(int)method->name.slen, method->name.ptr,
+		rdata->pkt_info.src_name);
+
+	ao2_ref(endpoint, -1);
+
+	pj_list_init(&hdr_list);
+	retry_after = pjsip_generic_int_hdr_create(rdata->tp_info.pool,
+		&str_retry_after, 300);
+	if (retry_after) {
+		pj_list_push_back(&hdr_list, retry_after);
+	}
+
+	pjsip_endpt_respond_stateless(ast_sip_get_pjsip_endpoint(), rdata, 503, NULL,
+		retry_after ? &hdr_list : NULL, NULL);
+
+	return PJ_TRUE;
+}
+
+static struct pjsip_module maintenance_pjsip_mod = {
+	.name     = { "Maintenance Module", 18 },
+	/*
+	 * Run after endpoint identification (endpoint_mod,
+	 * PJSIP_MOD_PRIORITY_TSX_LAYER - 3) so that
+	 * ast_pjsip_rdata_get_endpoint() returns the identified endpoint,
+	 * but before the request authenticator
+	 * (PJSIP_MOD_PRIORITY_APPLICATION - 2) so that a maintenance
+	 * endpoint receives 503 rather than a 401 challenge.
+	 */
+	.priority = PJSIP_MOD_PRIORITY_APPLICATION - 3,
+	.on_rx_request = maintenance_on_rx_request,
+};
+
+/* Sorcery observer - clean up stale entries when an endpoint is deleted. */
+
+static void maint_endpoint_deleted(const void *obj)
+{
+	maint_set_remove(ast_sorcery_object_get_id(obj));
+}
+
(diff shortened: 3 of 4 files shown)


danieldonoghue:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
When doing stuff like this an explanation of... why... and scenarios is nice to have, as well as why any existing (even loosely capable) functionality isn't suitable because ultimately this is "I've done this work and I want the project to include it and have the burden of dealing with further supporting it and dealing with consequences".

I would also fully expect people to rarely do CLI for this and instead do AMI programmatically.

As well if AI was used in the creation, see our policy: https://docs.asterisk.org/Development/Policies-and-Procedures/AI-Policy/?h=ai+policy

danieldonoghue:
Wow, thanks for the quick and detailed feedback. I will address the points you raised :) 

> When doing stuff like this an explanation of... why... and scenarios is nice to have, as well as why any existing (even loosely capable) ..

should I add this to the commit message/pr description or is a separate comment sufficient?

> I would also fully expect people to rarely do CLI for this and instead do AMI programmatically.

good point, I hadn't considered AMI since im using ansible and the CLI for my deployment management; will add that,

> As well if AI was used in the creation, see our policy: https://docs.asterisk.org/Development/Policies-and-Procedures/AI-Policy/?h=ai+policy

will read and do what is necessary.

jcolp:
A separate comment for the explanation is fine.

danieldonoghue:
> Workflow Check failed [master-ari1](https://github.com/asterisk/asterisk/actions/runs/22943683342/job/66591547283): FAILED TEST: rest_api/bridges/softmix_unhold

I think this is perhaps a pre-existing flaky test. I've run the specific ARI tests locally and they pass.

jcolp:
It is.

danieldonoghue:
The core use case for this module is graceful endpoint draining during maintenance windows - particularly in clustered or carrier-connected deployments where you might need to stop new calls arriving on a specific endpoint without disrupting calls already in progress.

**Why existing mechanisms don't cover this:**

- **Disabling/removing the endpoint from config + reload** - requires touching config files and triggering a reload, gives no indication to the peer when to retry, and sends no SIP-layer response to callers in flight. A well-behaved SIP peer receiving 503 with `Retry-After` will back off gracefully; a config reload gives it nothing.
- **PJSIP ACLs** - send a `403 Forbidden` response which is not correct for a maintenance scenario. ACLs are also IP-based rather than endpoint-based and require a config change and reload to toggle.
- **Unregistering an endpoint** - only applicable to endpoints using outbound registration, and not all do. Even where registration is used, unregistering relies on the remote party to stop routing calls once it detects the de-registration or the registration expires - there's no guarantee of immediacy, and no active rejection of calls that arrive in the window before the remote party catches up.
- **Dialplan flags (AstDB/global variables)** - only intercept calls that pass through dialplan; outbound originations via `Dial()` or ARI that target an endpoint directly and bypass dialplan entirely have no dialplan context in which a flag could be checked.
- **`max_calls` (global)** - a system-wide cap with no per-endpoint granularity; setting it to zero also doesnt actually block calls but sets the system to unlimited; it also requires a reload to change.
- **`core stop gracefully` / `core pause`** - system-wide; no per-endpoint granularity.
- **`pjsip set qualify`** - controls OPTIONS probing, not call acceptance.

**Scenarios this addresses:**

1. **Rolling upgrades in a cluster** - drain one node by putting its trunks into maintenance mode, wait for in-progress calls to complete naturally, then upgrade and re-enable. System-wide methods such as `core stop gracefully` can achieve a full-node drain, but at the cost of taking every endpoint on that node offline simultaneously.
2. **Carrier trunk issues** - a PSTN trunk is degraded; block new attempts with a proper 503+Retry-After while keeping existing sessions alive and letting the carrier know when to retry.
3. **Programmatic call control** - an ARI/AMI-driven orchestration system needs to gracefully take an endpoint offline without touching config files or triggering reloads.

The common thread is: **per-endpoint, runtime-changeable, graceful rejection with correct SIP signalling, without touching existing calls or requiring a reload or config file changes.**

Beyond the technical gaps, none of the above can be driven cleanly from an external orchestration system without config file manipulation or reload cycles. This module exposes the operation as a single AMI action or CLI command, making it straightforward to automate, for example, as a pre-step in a deployment pipeline or a health-check-triggered drain.

**Real-world context:** We operate multiple ARI applications against shared Asterisk servers, making outbound calls across dedicated SIP trunks, 24/7. The need for this module arose from three concrete operational requirements:

- **Trunk maintenance and migration** - when a trunk needs repair, upgrade, or replacement, we need to stop new calls using it while allowing calls already in progress to complete naturally. Waiting for the entire server to drain before touching a single trunk is impractical - in a busy system this can take hours.
- **Granular infrastructure upgrades** - being able to drain a specific endpoint means individual components of our infrastructure can be upgraded independently without a full service window affecting everything.
- **Testing** - maintenance mode provides a controlled, repeatable mechanism for regression and failure testing of the ARI applications themselves, simulating trunk unavailability without requiring actual infrastructure changes.

jcolp:
I would suggest trying to get feedback from others on the community forum[1] as to whether this actually meets the needs of others and would be something others would use.

[1] https://community.asterisk.org/

danieldonoghue:
added a post to the forum: https://community.asterisk.org/t/rfc-res-pjsip-maintenance-per-endpoint-maintenance-mode-module/112254

channels/chan_pjsip.c
```diff
@@ -2717,6 +2718,18 @@ static int request(void *obj)
 		}
 	}
 
+	/* Outbound endpoint policy check */
+	if (ast_sip_invoke_outbound_endpoint_checks(endpoint,
+		(struct ast_channel *)req_data->requestor)) {
+		ast_log(LOG_NOTICE,
+			"Outbound call to endpoint '%s' rejected by endpoint policy check\n",
+			ast_sorcery_object_get_id(endpoint));
```
gtjoseph: `endpoint_name` is already populated so you can use that instead of the call to `ast_sorcery_object_get_id()`.

channels/chan_pjsip.c
```diff
@@ -2717,6 +2718,18 @@ static int request(void *obj)
...
+		req_data->cause = AST_CAUSE_SWITCH_CONGESTION;
+		ao2_ref(endpoint, -1);
+		SCOPE_EXIT_RTN_VALUE(-1, "Endpoint '%s' blocked by outbound policy\n",
+			ast_sorcery_object_get_id(endpoint));
```
gtjoseph: Same as above.

include/asterisk/res_pjsip.h
```diff
@@ -1472,6 +1472,51 @@ int ast_sip_register_service(pjsip_module *module);
...
 
+/*!
+ * \brief Register a chainable outbound endpoint policy check callback.
+ *
+ * Multiple modules may register independent callbacks.  When an outbound
+ * call is initiated via chan_pjsip, all registered callbacks are invoked in
+ * registration order.  If any callback returns non-zero the call is refused
+ * before any SIP session or Asterisk channel is created.
+ *
+ * \since 22.0
```
gtjoseph: The next releases will be 20.19.0, 22.9.0 and 23.3.0 so...
```
\since 20.19.0
\since 22.9.0
\since 23.3.0
```

include/asterisk/res_pjsip.h
```diff
@@ -1472,6 +1472,51 @@ int ast_sip_register_service(pjsip_module *module);
...
+ *
+ * \param fn Callback to register.  Receives the resolved endpoint (borrowed
+ *           reference - do not unref) and the requesting channel (may be NULL
+ *           for ARI originations without a bridge channel).  Return non-zero
+ *           to block the call.
+ */
+void ast_sip_register_outbound_endpoint_check(
```
gtjoseph: Since these affect both inbound and outbound, I think better names are needed.

EDIT:  After going through the rest of the PR I guess these calls really are for outbound only.  The implementation seems a bit lopsided though.  Let me think more about it.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,664 @@
...
+static int maint_outbound_check(struct ast_sip_endpoint *endpoint,
+	struct ast_channel *requester)
+{
+	const char *endpoint_name = ast_sorcery_object_get_id(endpoint);
+	struct maint_ep *entry = ao2_find(maintenance_set, endpoint_name, OBJ_SEARCH_KEY);
+	if (entry) {
+		ao2_ref(entry, -1);
+		ast_log(LOG_NOTICE,
+			"PJSIP: Refusing outbound call to endpoint '%s': maintenance mode active\n",
+			endpoint_name);
```
gtjoseph: This results in double log messages because chan_pjsip is also logging it.
I'd leave it to whoever calls `ast_sip_invoke_outbound_endpoint_checks` to emit the message.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,664 @@
...
+	}
+
+	entry = ao2_find(maintenance_set, ast_sorcery_object_get_id(endpoint), OBJ_SEARCH_KEY);
+	if (!entry) {
+		ao2_ref(endpoint, -1);
+		return PJ_FALSE;
+	}
+	ao2_ref(entry, -1);
+
+	ast_verb(3, "PJSIP: Endpoint '%s' is in maintenance mode; rejecting new %.*s from %s\n",
```
gtjoseph: I'd make this one a NOTICE as well.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,664 @@
...
+	struct maint_ep *entry;
+	int rc;
+
+	entry = ao2_alloc_options(sizeof(*entry) + strlen(endpoint_name) + 1,
+		NULL, AO2_ALLOC_OPT_LOCK_NOLOCK);
+	if (!entry) {
+		return -1;
+	}
+	strcpy(entry->name, endpoint_name); /* safe: allocation is exactly the right size */
+	rc = ao2_link(maintenance_set, entry);
```
gtjoseph: This isn't failing on duplicate objects.  See comment in `load_module()`.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,664 @@
...
+	return 0;
+}
+
+/* Module load / unload */
+
+static int load_module(void)
+{
+	maintenance_set = ao2_container_alloc_hash(AO2_ALLOC_OPT_LOCK_RWLOCK,
+		AO2_CONTAINER_ALLOC_OPT_DUPS_OBJ_REJECT,
+		MAINT_HASH_BUCKETS, maint_ep_hash_fn, NULL, maint_ep_cmp_fn);
```
gtjoseph: You can't use `AO2_CONTAINER_ALLOC_OPT_DUPS_OBJ_REJECT` on an unsorted container.  See the documentation in astobj2.h.  You'll need to add a sort_fn and you'll want to use `AO2_CONTAINER_ALLOC_OPT_DUPS_REJECT` instead of `AO2_CONTAINER_ALLOC_OPT_DUPS_OBJ_REJECT`.

Having said all of that, since you're only storing the endpoint name, you'd be better using `ast_str_container_alloc_options(AO2_ALLOC_OPT_LOCK_RWLOCK, MAINT_HASH_BUCKETS)` then you wouldn't need the maint_ep structure or the `maint_ep_hash_fn` and `maint_ep_cmp_fn` functions.  Then in `maint_set_add()` just do an `ao2_find()` at the beginning of the function to check if the entry already exists.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,664 @@
...
+	int rc;
+
+	rc = enable ? maint_set_add(endpoint_name) : maint_set_remove(endpoint_name);
+	if (rc > 0) {
+		manager_event(EVENT_FLAG_SYSTEM, "PJSIPMaintenanceStatus",
+			"Endpoint: %s\r\n"
+			"Status: %s\r\n",
+			endpoint_name,
+			enable ? "enabled" : "disabled");
+		ast_verb(3, "PJSIP: Maintenance mode %s for endpoint '%s'\n",
```
gtjoseph: Log levels are a bit inconsistent. Some are NOTICEs and some are Verbose. I'd make this one a NOTICE.

gtjoseph:
Oh, one other thing, you should add an "all" option so you can pause all endpoints with one command: `pjsip set maintenance on all`, `pjsip set maintenance off all`

InterLinked1:
> I would suggest trying to get feedback from others on the community forum[1] as to whether this actually meets the needs of others and would be something others would use.
> 
> [1] https://community.asterisk.org/

Personally, if this is going to go forward, I would prefer to see a technology-neutral implementation of this concept, not something PJSIP-specific, since others may want to do something similar with other channel types.

The minimum callback logic required could be put into the core, and then each channel driver / external module could implement the protocol-specific support for different technologies. The AMI actions and CLI commands would be named generically and could be put into another module.

channels/chan_pjsip.c
danieldonoghue: fair point; will address

include/asterisk/res_pjsip.h
danieldonoghue: great; will make sure to update those

res/res_pjsip_maintenance.c
danieldonoghue: sounds much easier; thanks

res/res_pjsip_maintenance.c
danieldonoghue: will do

danieldonoghue:
> Oh, one other thing, you should add an "all" option so you can pause all endpoints with one command: `pjsip set maintenance on all`, `pjsip set maintenance off all`

I think this is a great idea. will do that :)

danieldonoghue:
> Personally, if this is going to go forward, I would prefer to see a technology-neutral implementation of this concept, not something PJSIP-specific, since others may want to do something similar with other channel types.
> 
> The minimum callback logic required could be put into the core, and then each channel driver / external module could implement the protocol-specific support for different technologies. The AMI actions and CLI commands would be named generically and could be put into another module.

I like this idea in principle but im not really sure how I'd go about doing it. could you point me in the right direction?

res/res_pjsip_maintenance.c
danieldonoghue: the messages are subtly different though; the one in chan_pjsip is about the outbound call being blocked due to policy whilst this one is specifically stating that the policy failure is maintenance mode. would it be beneficial to do this as a debug message instead?

res/res_pjsip_maintenance.c
danieldonoghue: fair enough

InterLinked1:
> > Personally, if this is going to go forward, I would prefer to see a technology-neutral implementation of this concept, not something PJSIP-specific, since others may want to do something similar with other channel types.
> > The minimum callback logic required could be put into the core, and then each channel driver / external module could implement the protocol-specific support for different technologies. The AMI actions and CLI commands would be named generically and could be put into another module.
> 
> I like this idea in principle but im not really sure how I'd go about doing it. could you point me in the right direction?

For example, `ast_call` or `ast_request` would return failure if a channel technology / specific devices/endpoints/etc are in maintenance mode (what it looks like is implementation dependent for each channel driver).

The PJSIP stuff would mostly be the same, I'm just suggesting moving the callback registration into the core in a way that could accommodate other channel drivers, not saying you need to implement the other ones at this time.

include/asterisk/res_pjsip.h
danieldonoghue: @gtjoseph have you had any further thoughts on this? what did you mean by lopsided? im happy to discuss the shape of the API :)

res/res_pjsip_maintenance.c
danieldonoghue: @gtjoseph would downgrading this message to ast_debug be acceptable? that way the NOTICE from chan_pjsip covers operational visibility and the debug line gives the specific reason without duplicating at the same log level.

include/asterisk/res_pjsip.h
jcolp: Asking him for thoughts less than 24 hours after his original comment is a bit... much. He'll respond when he's had time to think, ponder, and wishes to.

danieldonoghue: my apologies; wasn't meaning to push; just eager to do what I can :)

danieldonoghue:
> For example, `ast_call` or `ast_request` would return failure if a channel technology / specific devices/endpoints/etc are in maintenance mode (what it looks like is implementation dependent for each channel driver).
> 
> The PJSIP stuff would mostly be the same, I'm just suggesting moving the callback registration into the core in a way that could accommodate other channel drivers, not saying you need to implement the other ones at this time.

I think there's real merit to this suggestion but I'd like to understand the shape of it better before committing to any sort of refactor.

As I see it, there are 2 parts to this module that have different requirements when making it more generalisable:

**Outbound blocking** is pretty straightforward. The hook mechanism I've added could reasonably live in the core instead, hooked into `ast_request`/`ast_call` as you suggested, with some signature changes, so that any channel driver could participate.

**Inbound rejection** is inherently protocol-specific. while other technologies like IAX2 and ISDN have their own call rejection mechanisms, the exact behaviour differs in each case and would need a separate per-driver implementation. 

**CLI/AMI naming** raises its own question. The current `pjsip set maintenance` and `PJSIPSetMaintenance` names are clear and well-scoped. A generic rename such as `channel set maintenance` and `SetMaintenance` implies cross-technology support that is only partially implemented, which might be confusing to operators.

**The `all` question** is where I could most use more input. Right now `pjsip set maintenance on all` iterates all PJSIP endpoints - a well-defined, bounded set. In a technology-agnostic framework, what does `all` mean? If it means all devices across all channel technologies that have a maintenance handler registered, does that imply a need for a registry of per-technology maintenance implementations? That would likely require significantly more work.

With that in mind, I have a few questions:

1. What would the hook signature look like when `struct ast_sip_endpoint` is no longer available? Something based on channel tech + device string, or a `void *` with per-driver interpretation?
2. Should the CLI/AMI names change now, or could that wait until a second technology actually implements the interface?
3. How should `all` behave generically? scoped per-technology (e.g. `pjsip/*`), or truly global across all technologies with a registered handler?
4. For the inbound side, what would the trigger point be? for outbound it's the `ast_request/ast_call` functions that provide a natural hook in the core but inbound calls are pushed into the system by each driver independently so my feeling is that the generalisation likely only makes sense for outbound calls or did you have some thoughts on this?

Happy to look at going in this direction but really want to understand the scope before starting.

include/asterisk/res_pjsip.h
gtjoseph: > @gtjoseph have you had any further thoughts on this? what did you mean by lopsided? im happy to discuss the shape of the API :)

I just meant that the implementation for outbound is via the new callback registration method but the process for inbound is this module registering a sip service.  It just seemed asymmetrical. For outbound, could you just add a new "session_create" callback to the ast_sip_session_supplement structure and check them at the beginning of res_pjsip_session.c:ast_sip_session_create_outgoing()?  This module then could register itself with ast_sip_session_register_supplement().

res/res_pjsip_maintenance.c
gtjoseph: Yep, that's fine.

gtjoseph:
Making this a global capability is a good idea but let's leave that for a separate effort. As @danieldonoghue points out, it's not as straightforward as it seems.

include/asterisk/res_pjsip.h
danieldonoghue: will get started :)

InterLinked1:
> > For example, `ast_call` or `ast_request` would return failure if a channel technology / specific devices/endpoints/etc are in maintenance mode (what it looks like is implementation dependent for each channel driver).
> > The PJSIP stuff would mostly be the same, I'm just suggesting moving the callback registration into the core in a way that could accommodate other channel drivers, not saying you need to implement the other ones at this time.
> 
> I think there's real merit to this suggestion but I'd like to understand the shape of it better before committing to any sort of refactor.
> 
> As I see it, there are 2 parts to this module that have different requirements when making it more generalisable:
> 
> **Outbound blocking** is pretty straightforward. The hook mechanism I've added could reasonably live in the core instead, hooked into `ast_request`/`ast_call` as you suggested, with some signature changes, so that any channel driver could participate.
> 
> **Inbound rejection** is inherently protocol-specific. while other technologies like IAX2 and ISDN have their own call rejection mechanisms, the exact behaviour differs in each case and would need a separate per-driver implementation.
> 
> **CLI/AMI naming** raises its own question. The current `pjsip set maintenance` and `PJSIPSetMaintenance` names are clear and well-scoped. A generic rename such as `channel set maintenance` and `SetMaintenance` implies cross-technology support that is only partially implemented, which might be confusing to operators.

I don't think this is an issue, really. There are a lot of things that are standardized in the core as implemented by channel drivers, but are not actually supported by channel drivers.

The CLI commands or AMI actions would simply return a clear error if the channel tech isn't supported for that operation. The for the AMI command could include a current list of technologies supported that could be updated in the future.

> **The `all` question** is where I could most use more input. Right now `pjsip set maintenance on all` iterates all PJSIP endpoints - a well-defined, bounded set. In a technology-agnostic framework, what does `all` mean? If it means all devices across all channel technologies that have a maintenance handler registered, does that imply a need for a registry of per-technology maintenance implementations? That would likely require significantly more work.
>
> With that in mind, I have a few questions:
> 
> 1. What would the hook signature look like when `struct ast_sip_endpoint` is no longer available? Something based on channel tech + device string, or a `void *` with per-driver interpretation?

I would imagine something like `channel set maintenance on all` which might involve a callback per tech to do tech-specific things. But I'm not suggesting you abstract everything into the core so that you can't look at `ast_sip_endpoint` anymore - mainly just the core callbacks and CLI/AMI, so that other techs could do something similar without having to duplicate all the plumbing required.

> 2. Should the CLI/AMI names change now, or could that wait until a second technology actually implements the interface?

If making it generic now, it should be made generic now. Changing things later is always a pain since you need to take backwards compatibility into account. The other implementations, of course, can always be added later without breaking backwards compatibility.

> 3. How should `all` behave generically? scoped per-technology (e.g. `pjsip/*`), or truly global across all technologies with a registered handler?

I'm not sure I understand the question exactly but it might be desirable to do both, i.e. truly put everything into maintenance more, or just those of a specific technology, e.g. PJSIP (or even further subcategories within each as appropriate, as defined by templates or some other criteria, etc)

> 4. For the inbound side, what would the trigger point be? for outbound it's the `ast_request/ast_call` functions that provide a natural hook in the core but inbound calls are pushed into the system by each driver independently so my feeling is that the generalisation likely only makes sense for outbound calls or did you have some thoughts on this?

Since each technology will involve some kind of tech-specific implementation anyways, I suppose that could be left up to each tech implementation.

danieldonoghue:
since I've now shifted away from the custom outbound hook in favour of using a session supplement, I think the outbound generalisation suggestion isn't so relevant any more - and since broader cross-technology capabilities have been earmarked for a separate effort anyway, I'll park the questions raised above for now but im happy to revisit should that effort get traction :)

include/asterisk/res_pjsip.h
danieldonoghue: Ive added a new session_create callback to ast_sip_session_supplement and check it at the very top of ast_sip_session_create_outgoing() via a new helper, ast_sip_session_check_supplement_create(), in pjsip_session.c. res_pjsip_maintenance now registers itself as a session supplement via ast_sip_session_register_supplement() rather than using a custom callback chain in res_pjsip. The hook in chan_pjsip and the three ast_sip_*_outbound_endpoint_check* functions are gone. Is this more in line with what you had in mind?

danieldonoghue:
testsuite-test-pr: 128

channels/chan_pjsip.c
```diff
@@ -2621,6 +2621,7 @@ struct request_data {
 	struct ast_sip_session *session;
 	struct ast_stream_topology *topology;
 	const char *dest;
+	const struct ast_channel *requestor;
```
gtjoseph: You don't need this anymore.

channels/chan_pjsip.c
```diff
@@ -2741,6 +2742,7 @@ static struct ast_channel *chan_pjsip_request_with_stream_topology(const char *t
 
 	req_data.topology = topology;
 	req_data.dest = data;
+	req_data.requestor = requestor;
```
gtjoseph: Or this.

include/asterisk/res_pjsip_session.h
```diff
@@ -313,6 +313,26 @@ struct ast_sip_session_supplement {
...
+	 * \since 22.9.0
+	 * \since 23.3.0
+	 *
+	 * \param endpoint The endpoint the outgoing session would be created for
+	 * \param destination The request user (user part of the Request-URI), or NULL
+	 *
+	 * \retval non-zero Block session creation
+	 * \retval 0 Allow session creation
+	 */
+	int (*session_create)(struct ast_sip_endpoint *endpoint, const char *destination);
```
gtjoseph: This needs to go at the end of the structure or it will break ABI compatibility with earlier releases.

You might as well pass all of the arguments to this function that are passed to ast_sip_session_create_outgoing().  They're not strictly needed now but better to add them for the future.

include/asterisk/res_pjsip_session.h
```diff
@@ -625,6 +645,29 @@ void ast_sip_session_register_supplement_with_module(struct ast_module *module,
...
+ *
+ * \param endpoint The endpoint the outgoing session would be created for
+ * \param destination The request user (user part of the Request-URI), or NULL
+ *
+ * \retval 0 Session creation is allowed
+ * \retval -1 Session creation is blocked by a supplement
+ */
+int ast_sip_session_check_supplement_create(struct ast_sip_endpoint *endpoint,
+	const char *destination);
+
```
gtjoseph, line 679: See previous comment about passing more arguments.

res/res_pjsip_session.c
```diff
@@ -3248,6 +3248,10 @@ struct ast_sip_session *ast_sip_session_create_outgoing(struct ast_sip_endpoint
 	SCOPE_ENTER(1, "%s %s Topology: %s\n", ast_sorcery_object_get_id(endpoint), request_user,
 		ast_str_tmp(256, ast_stream_topology_to_str(req_topology, &STR_TMP)));
 
+	if (ast_sip_session_check_supplement_create(endpoint, request_user)) {
+		SCOPE_EXIT_RTN_VALUE(NULL, "Session creation blocked by supplement\n");
+	}
+
```
gtjoseph: See comment above about passing more arguments.

Also, add the endpoint name to the message:
```
SCOPE_EXIT_RTN_VALUE(NULL, "%s: Session creation blocked by supplement\n", ast_sorcery_object_get_id(endpoint));
```

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,713 @@
...
+	rc = enable ? maint_set_add(endpoint_name) : maint_set_remove(endpoint_name);
+	if (rc > 0) {
+		manager_event(EVENT_FLAG_SYSTEM, "PJSIPMaintenanceStatus",
+			"Endpoint: %s\r\n"
+			"Status: %s\r\n",
+			endpoint_name,
+			enable ? "enabled" : "disabled");
+		ast_log(LOG_NOTICE, "PJSIP: Maintenance mode %s for endpoint '%s'\n",
+			enable ? "enabled" : "disabled", endpoint_name);
+	}
```
gtjoseph: If a system has thousands of endpoints, using "all" will generate thousands of log messages and ami events.  I think the logging should be done in handle_cli_pjsip_set_maintenance() and ami_set_maintenance() so that only one log message or AMI event is issued when "all" is used.

res/res_pjsip/pjsip_session.c
```diff
@@ -108,6 +108,21 @@ int ast_sip_session_add_supplements(struct ast_sip_session *session)
 	return 0;
 }
 
+int ast_sip_session_check_supplement_create(struct ast_sip_endpoint *endpoint,
+	const char *destination)
```
gtjoseph: See previous comment about passing more arguments.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,713 @@
...
+			if (rc > 0) {
+				count++;
+			} else if (rc < 0) {
+				failed++;
+			}
+			ao2_ref(endpoint, -1);
+		}
+		ao2_iterator_destroy(&it);
+		ao2_ref(all_endpoints, -1);
+		ast_cli(a->fd, "Maintenance mode %s for %d endpoint%s%s\n",
```
gtjoseph, line 456: See comment above about doing only one log message/AMI event if "all" is specified.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,713 @@
...
+			return 0;
+		}
+		it = ao2_iterator_init(all_endpoints, 0);
+		while ((endpoint = ao2_iterator_next(&it))) {
+			apply_maintenance_state(ast_sorcery_object_get_id(endpoint), enable);
+			ao2_ref(endpoint, -1);
+		}
+		ao2_iterator_destroy(&it);
+		ao2_ref(all_endpoints, -1);
+		astman_send_ack(s, m,
```
gtjoseph, line 620: See comment above about doing only one log message/ami event if "all" is used.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,713 @@
...
+{
+	pjsip_msg *msg = rdata->msg_info.msg;
+	pjsip_to_hdr *to;
+	struct ast_sip_endpoint *endpoint;
+	char *entry;
+	pjsip_hdr hdr_list;
+	pjsip_generic_int_hdr *retry_after;
+	static const pj_str_t str_retry_after = { "Retry-After", 11 };
+
+	if (pjsip_method_cmp(&msg->line.req.method, &pjsip_invite_method) != 0 &&
```
gtjoseph: Use pjsip_get_invite_method(), pjsip_get_register_method(), pjsip_get_options_method().

danieldonoghue:
> I just realized that while SUBSCRIBEs are blocked, REGISTERs aren't. Was there a reason for that?
> 
> Also for SUBSCRIBE and REGISTER, I think the Expires header should be checked and if "0" should be allowed. This allows unsubscribe and unregister to work but not new subscriptions or registers.
> 
> EDIT: Add OPTIONS checking for incoming requests as well.

REGISTER was an oversight; will address that and also allow through with `Expires: 0`. I wasn't sure whether I should do OPTIONS or not, on account that it might trigger external monitoring alarms, etc but, on reflection - those will likely happen anyway if a call is rejected so I will go back and add that too.

res/res_pjsip_maintenance.c
danieldonoghue: @gtjoseph, when "all" is used ive moved to a single summary log as requested, but im unsure about the AMI event. 

Firing PJSIPMaintenanceStatus with `Endpoint: all` feels inconsistent since "all" is really just a command keyword, not a real endpoint name and  any AMI client tracking per-endpoint state cant act on it directly. Other bulk operations like `pjsip qualify all` don't emit a synthetic summary event either - clients are expected to query state separately if needed. My instinct is to suppress the AMI event entirely for the bulk path and let clients use PJSIPShowMaintenance to get the current picture, but wanted your guidance before proceeding..

gtjoseph:
> I wasn't sure whether I should do OPTIONS or not, on account that it might trigger external monitoring alarms, etc but, on reflection - those will likely happen anyway if a call is rejected so I will go back and add that too.

OPTIONS came up as we were talking about this PR internally.  Rejecting them would allow a load-balancer or proxy to mark the Asterisk instance as unavailable and (hopefully) use another instance in the same group.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,762 @@
...
+		return PJ_FALSE;
+	}
+
+	/* In-dialog INVITE, SUBSCRIBE, or OPTIONS; pass through. */
+	if (is_invite || is_subscribe || is_options) {
+		to = rdata->msg_info.to;
+		if (to->tag.slen > 0) {
+			return PJ_FALSE;
+		}
+	}
```
gtjoseph: After talking about it internally, we think any in-dialog request should be allowed otherwise we'd also have to explicitly allow REFER, UPDATE, etc.  Better to just remove the outer `if` statement that checks the method.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,762 @@
...
+	struct ast_sip_contact *contact, const char *location,
+	const char *request_user, struct ast_stream_topology *req_topology)
+{
+	const char *endpoint_name = ast_sorcery_object_get_id(endpoint);
+	char *entry = ao2_find(maintenance_set, endpoint_name, OBJ_SEARCH_KEY);
+
+	if (entry) {
+		ao2_ref(entry, -1);
+		ast_debug(3, "PJSIP: Refusing outbound call to endpoint '%s': maintenance mode active\n",
+			endpoint_name);
```
gtjoseph: I know I had you change this to a debug but after talking about it, making it a NOTICE would be better because otherwise you just get the following without any reason...
```
ERROR[732388]: ... Failed to create outgoing session to endpoint 'alice'
NOTICE[735530][C-00000005]: ...: Unable to create channel of type 'PJSIP' (cause 3 - No route to destination)
```

gtjoseph (changes requested):
Sorry, we've been really slammed the past 2 weeks so I'm just getting back to this.  I think with these 2 minor changes we'll be good to go.

include/asterisk/res_pjsip_session.h
```diff
@@ -625,6 +650,33 @@ void ast_sip_session_register_supplement_with_module(struct ast_module *module,
...
+ * Iterates the global supplement list and calls any registered \c session_create
+ * callbacks.  The first callback to return a non-zero value stops the iteration
+ * and causes this function to return -1, blocking the session creation.
+ *
+ * This is called at the beginning of ast_sip_session_create_outgoing() before
+ * any dialog or invite session resources are allocated.
+ *
+ * \since 20.19.0
+ * \since 22.9.0
+ * \since 23.3.0
```
gtjoseph: Oops, one more sorry.  These need to be bumped to 20.20.0, 22.10.0 and 23.4.0.

res/res_pjsip_maintenance.c
```diff
@@ -0,0 +1,762 @@
...
+	<depend>res_pjsip_session</depend>
+	<support_level>extended</support_level>
+ ***/
+
+/*** DOCUMENTATION
+	<manager name="PJSIPSetMaintenance" language="en_US">
+		<since>
+			<version>20.19.0</version>
+			<version>22.9.0</version>
+			<version>23.3.0</version>
```
gtjoseph: These also need to be bumped to 20.20.0, 22.10.0 and 23.4.0.

gtjoseph (changes requested):

res/res_pjsip_maintenance.c
danieldonoghue: that makes sense. will address

res/res_pjsip_maintenance.c
danieldonoghue: not a problem :)

gtjoseph (approved):

Issue #1819, opened 2026-03-10
[bug]: When a 302 is received from a UAS, the cause and tech_cause codes set on the channel are incorrect.


gtjoseph:
### Severity

Minor

### Versions

All

### Components/Modules

chan_pjsip

### Operating Environment

Any

### Frequency of Occurrence

Constant

### Issue Description

If a UAS returns a "302 Moved permanently", the channel hangup cause is set to "127 INTERWORKING" and the channel tech_cause is set to "487 Request Terminated".  This gives no indication of what really happened.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1821, opened 2026-03-10
chan_pjsip: Set correct cause codes for non-2XX responses.

gtjoseph:
Redirects initiated by 302 response codes weren't handled correctly
when setting the hangup cause code and tech cause code on the responding
channel.  They're now set to 23 (REDIRECTED_TO_NEW_DESTINATION) and
302 (Moved permanently).  Other non-2XX response codes also had issues.

A new API ast_channel_dialed_causes_iterator() was added to retrieve
the hangup cause codes for a channel.

chan_pjsip_session_end() in chan_pjsip has been refactored to set the
correct cause codes on a channel based on the cause codes added by
chan_pjsip_incoming_response_update_cause().  Copious amounts of
debugging and comments were also added.

Resolves: #1819

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -2453,6 +2453,8 @@ static int chan_pjsip_call(struct ast_channel *ast, const char *dest, int timeou
 static int hangup_cause2sip(int cause)
 {
 	switch (cause) {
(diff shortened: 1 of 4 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph:
This PR will stay in draft mode for a bit.

PR #1822, opened 2026-03-11
manager.c : add early disabledevents fast-path

phoneben:
manager.c:  Add early `disabledevents` fast-path

- Add early `disabledevents` check in `manager_default_msg_cb()` and `manager_generic_msg_cb()` before `append_event()`, the `all_events` write lock, and per-session wakeup loop; existing check in `__manager_event_sessions_va()` preserved as fallback for all other callers

### DIFF
--- a/main/manager.c
+++ b/main/manager.c
@@ -589,6 +589,18 @@ static void manager_default_msg_cb(void *data, struct stasis_subscription *sub,
 		ao2_cleanup(sessions);
 		return;
 	}
+	/* Early drop before append_event(), the all_events write lock, and the
+	 * per-session wakeup loop. stasis_message_to_ami() cost above is already
+	 * paid at this point. The check in __manager_event_sessions_va() is
+	 * intentionally kept as fallback for all other callers. */
+	if (!ast_strlen_zero(manager_disabledevents)
+		&& !ast_strlen_zero(ev->manager_event)
+		&& ast_in_delimited_string(ev->manager_event,
+			manager_disabledevents, ',')) {
+		ao2_ref(ev, -1);
+		ao2_cleanup(sessions);
+		return;
+	}
 
 	manager_event_sessions(sessions, ev->event_flags, ev->manager_event,
 		"%s", ev->extra_fields);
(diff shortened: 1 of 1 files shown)


phoneben:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

phoneben:
want to test locally to see if i get a gain in performance

main/manager.c
```diff
@@ -589,6 +589,18 @@ static void manager_default_msg_cb(void *data, struct stasis_subscription *sub,
...
 		return;
 	}
+	/* Early drop before append_event(), the all_events write lock, and the
+	 * per-session wakeup loop. stasis_message_to_ami() cost above is already
+	 * paid at this point. The check in __manager_event_sessions_va() is
+	 * intentionally kept as fallback for all other callers. */
+	if (!ast_strlen_zero(manager_disabledevents)
+		&& !ast_strlen_zero(ev->manager_event)
+		&& ast_in_delimited_string(ev->manager_event,
+			manager_disabledevents, ',')) {
```
seanbright, line 599: This same check is now in this file 3 separate times. Time to extract in to a function or macro.

seanbright:
> multiple modules registering the same event name

Which modules are doing this? What event?

phoneben:
> > multiple modules registering the same event name
> 
> Which modules are doing this? What event?

VarSet  *3
VoicemailPasswordChange  *3

VarSet appears three times in manager show events because exactly three files embed a <managerEvent name="VarSet"> XML documentation block:

- apps/app_stack.c line 230
- main/stasis_channels.c line 47
- funcs/func_global.c line 134

and VoicemailPasswordChange   managerEvent is  the only managerEvent  in app_voicemail.c
so it gets populated for all backend 

i think the correct way to fix it is 
in` xmldoc.c` with `AO2_CONTAINER_ALLOC_OPT_DUPS_REJECT` — because that fixes the general case for any file that has this pattern now or in the future,

jcolp:
Please don't include multiple fixes or changes in a single commit. Unrelated things should be separate PRs.

phoneben:
are you  referring to  `disabledevents  `vs the duplicate and the length? will make a new pr for this after i finish testing this ....
`disabledevents ` was the original reason i opened it up still working on it!  (github has the .patch feature i use  lot )
it did not improve  still have  high water trying to find where is the bottle-neck  is 
any pointers to it really appreciated 
```
Taskprocessor: stasis/m:manager:core-00000012
===========================================
Subsystem:            stasis
Tasks processed:      1081294294
Current queue size:   0
Max queue depth:      60592
Low water mark:       2700
High water mark:      3000
High water alert:     No
Suspended:            No
Currently executing:  No
Highest time (us):    172395
  Highest task origin: stasis.c:1455 (dispatch_message)
Lowest time (us):     0
```

at this point i think the queue does work even if the slow receiving endpoint it does not stall the whole queue from accepting new event and processing it

jcolp:
I don't have any pointers. If I knew with any certainty then it would most likely already be something I would have improved further or done.

phoneben:
closing as duplicate 
https://github.com/asterisk/asterisk/pull/1984

PR #1824, opened 2026-03-17
res_audiosocket: Fix header read loop to use correct buffer offset

SvenKube:
The PR #1522 introduced the header read loop for audiosocket packets which does not handle partial header reads correctly. This commit adds the missing buffer offsets.


SvenKube:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1831, opened 2026-03-20
[bug]:raise_exception() and EXCEPTION() read use channel datastores without holding ast_channel_lock


Tinet-mucw:
### Severity

Trivial

### Versions

18.10

### Components/Modules

main

### Operating Environment

centos8

### Frequency of Occurrence

Occasional

### Issue Description

We’re sorry, but we have not yet fully completed the Asterisk version migration.

Problem Summary
Symptom: Process abort with glibc double free or corruption (fasttop) inside ast_datastore_free(), called from ast_channel_destructor() while tearing down channel datastores. The backtrace went through chan_pjsip hangup → ast_channel_unref → destructor.

Observed state (GDB): chan->datastores had first == NULL but last non-NULL—invalid for Asterisk’s AST_LIST_HEAD list. A datastore on that tail pointed at max_forwards_info.

Interpretation: The datastore linked list on the channel was inconsistent (broken first/last/entry.next chain) and/or the heap was already corrupted. double free can be either a true second free of the same struct ast_datastore *, or glibc reporting corruption at the next free().

Analysis Process
API contract
channel.h states that ast_channel_datastore_add / find / remove do not lock the channel; the caller must hold ast_channel_lock(chan).

Destructor path
ast_channel_destructor drains the list with AST_LIST_REMOVE_HEAD while holding the channel lock. If the list is already corrupted (e.g. first NULL but last set), REMOVE_HEAD does not fix last, and behavior becomes inconsistent (skipped frees, leaks, or worse if the heap is already damaged).

Root-cause classes (not mutually exclusive)

Concurrent access: One thread modifying the datastore list without the channel lock while another holds the lock or destroys the channel.
Double free: Same datastore freed twice (e.g. removed from list and freed, then freed again in destructor).
Other memory corruption unrelated to datastores, with failure first appearing at ast_free(datastore).
raise_exception review
raise_exception() in pbx.c used ast_channel_datastore_find / add without locking. Dialplan execution runs pbx_exec() with the channel unlocked during app execution, and several __ast_pbx_run branches call raise_exception without holding the channel lock. That violates the datastore API and is a plausible contributor to list corruption under concurrency (not necessarily the only one).

### Relevant log output

```shell
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x0000ffffa9cb8a0c in __GI_abort () at abort.c:79
[...]

PR #1832, opened 2026-03-20
pbx: Hold channel lock for exception datastore access

Tinet-mucw:
ast_channel_datastore_find() and ast_channel_datastore_add() must only be called while the channel is locked (see channel.h). raise_exception() and the EXCEPTION dialplan function read path accessed the exception datastore without holding ast_channel_lock, which could corrupt the per-channel datastore list under concurrency and lead to crashes during teardown (e.g. double free in ast_datastore_free).

Resolves: #1831


Tinet-mucw:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

Issue #1840, opened 2026-03-27
[bug]: Asterisk fails to compile with --enable-dev-mode=yes due to INIT_RETURN undeclared in bundled pjproject Python bindings


alex2grad:
### Severity

Minor

### Versions

20.19.0-rc1

### Components/Modules

third-party/pjproject

### Operating Environment

Slackware 15.0 x86_64
Python 2.7.18
Python 3.9.25

### Frequency of Occurrence

Constant

### Issue Description

Building Asterisk 20.19.0 with `--enable-dev-mode=yes` fails to compile the bundled pjproject Python bindings with the following error:

```
[pjproject]  Compiling python bindings
source/pjsip-apps/src/python/_pjsua.c: In function 'init_pjsua':
source/pjsip-apps/src/python/_pjsua.c:4471:16: error: 'INIT_RETURN' undeclared (first use in this function)
 4471 |         return INIT_RETURN;
      |                ^~~~~~~~~~~
```

`_pjsua.c` (pjproject 2.16) uses the macro `INIT_RETURN` unconditionally throughout `init_pjsua()`. The macro is defined in `python3_compat.h`, but only inside the `#if PY_MAJOR_VERSION > 2` block:

```c
// python3_compat.h
#if PY_MAJOR_VERSION > 2
    // ...
    #define INIT_RETURN  (m)   // or empty for Python 3.0-3.9
#endif
```

When Asterisk's `configure` detects `python2.7` (which is exactly what `configure.ac` searches for in dev mode - Python 3 support is intentionally disabled per a comment in that file), the pjproject Python bindings are compiled against Python 2.7. With `PY_MAJOR_VERSION == 2`, the entire `python3_compat.h` block is skipped and `INIT_RETURN` is never defined, causing the build to fail.

A one-line fix adding an `#else` branch to `python3_compat.h`:

```c
#else /* PY_MAJOR_VERSION <= 2 */

#define INIT_RETURN

#endif // PY_MAJOR_VERSION > 2
```

This has been submitted upstream to pjproject:
**https://github.com/pjsip/pjproject/pull/4892**

### Relevant log output

```shell
[pjproject]  Compiling python bindings
source/pjsip-apps/src/python/_pjsua.c: In function 'init_pjsua':
source/pjsip-apps/src/python/_pjsua.c:4471:16: error: 'INIT_RETURN' undeclared (first use in this function)
 4471 |         return INIT_RETURN;
      |                ^~~~~~~~~~~
[...]

PR #1841, opened 2026-03-27
pjproject: fix INIT_RETURN undeclared when building Python bindings with Python 2.7

alex2grad:
This bug has been reported and a fix submitted upstream to pjproject:
https://github.com/pjsip/pjproject/pull/4892

Fixes: #1840

### DIFF
--- a/third-party/pjproject/patches/0060-python2-fix-INIT_RETURN-undeclared.patch
+++ b/third-party/pjproject/patches/0060-python2-fix-INIT_RETURN-undeclared.patch
@@ -1,0 +1,25 @@
+From 40f7a3766cda838f77ea84066d4374b4ef9d61e0 Mon Sep 17 00:00:00 2001
+From: Alexei Gradinari <alex2grad@gmail.com>
+Date: Thu, 26 Mar 2026 19:15:06 -0400
+Subject: [PATCH] pjsip-apps/python: fix INIT_RETURN undeclared with Python 2
+
+---
+ pjsip-apps/src/python/python3_compat.h | 4 ++++
+ 1 file changed, 4 insertions(+)
+
+diff --git a/pjsip-apps/src/python/python3_compat.h b/pjsip-apps/src/python/python3_compat.h
+index 392d60ce1f..9d903ffa58 100644
+--- a/pjsip-apps/src/python/python3_compat.h
++++ b/pjsip-apps/src/python/python3_compat.h
+@@ -47,6 +47,10 @@
+ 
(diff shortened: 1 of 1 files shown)


alex2grad:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

silentindark:
@alex2grad Your PR https://github.com/pjsip/pjproject/pull/4892 was abandoned by the pjproject team because this module is now deprecated: https://github.com/pjsip/pjproject/blob/master/pjsip-apps/src/python/DEPRECATED.txt

jcolp:
I believe we substantially reduced or eliminated our pjsua usage in the testsuite, so the fix may be to just remove the building altogether.

seanbright:
> was abandoned by the pjproject team

A bit silly on their part for a simple bug fix

alex2grad:
@jcolp, do you mean there's no need to compile pjproject's pjsua, pjsystest and python bindings at all, even if dev-mode=yes? And then we can completely get rid of Python dependency in Asterisk?

jcolp:
I believe so, yes. What say you @mbradeen?

alex2grad:
> > was abandoned by the pjproject team
> 
> A bit silly on their part for a simple bug fix

This is despite the fact that they introduced https://github.com/pjsip/pjproject/pull/4577
this bug after the Python pjsua module was deprecated.

mbradeen:
> I believe so, yes. What say you @mbradeen?

Yes, we had to get rid of pjsua when we went to Python3 because it would hang due to a conflict between async calls within pjsua and twisted.  There are still some old references to tests we couldn't fully convert to sipp, but those are skipped.

alex2grad:
> > I believe so, yes. What say you @mbradeen?
> 
> Yes, we had to get rid of pjsua when we went to Python3 because it would hang due to a conflict between async calls within pjsua and twisted. There are still some old references to tests we couldn't fully convert to sipp, but those are skipped.

I can prepare PR.
To do this, I need to modify "configure.ac." Will I also need to generate "configure" and add it to the commit?

jcolp:
Yes.

alex2grad:
Closing this PR in favor of [#1854](https://github.com/asterisk/asterisk/pull/1854).

After discussion with @jcolp and @mbradeen, the correct fix is not to patch the broken Python bindings but to remove the entire pjsua/pjsystest/Python bindings build from dev mode altogether:

- The pjsua Python module is officially deprecated upstream (pjproject added `pjsip-apps/src/python/DEPRECATED.txt`)
- The Asterisk Test Suite has already migrated away from pjsua to SIPp

PR #1843, opened 2026-03-31
pbx_functions/func_odbc: fix use-after-free crash during func_odbc reload

### LINKED ISSUE #1818: [bug]: func_odbc: possible use-after-free crash during reload with active calls
### Severity

Minor

### Versions

20.18.2

### Components/Modules

fun_odbc

### Operating Environment

centos 7

### Frequency of Occurrence

One Time

### Issue Description

Asterisk 20.18.2 crashes with SIGSEGV when `func_odbc.so` is reloaded while a call is actively executing an `ODBC_*` custom function. In the observed case, the log shows:
`[2026-03-09 13:01:07] VERBOSE[29784] loader.c: Reloading module 'func_odbc.so' (ODBC lookups)`
and the crashing call was expanding:
`dynfeature=${ODBC_GET_FROMOUTSIDE_DYNFEATURES(${TENANTID})}`
The backtrace shows a stale/invalid module pointer during cleanup:
```
#2  __ast_module_user_remove(mod=0x21, ...)
#3  ast_func_read() at pbx_functions.c:618
#4  pbx_substitute_variables_helper_full_location()

```
`mod=0x21` strongly indicates a freed or never-valid pointer.
In reload(), `func_odbc `tears down all existing queries under the write lock:

```
while (!AST_RWLIST_EMPTY(&queries)) {
    oldquery = AST_RWLIST_REMOVE_HEAD(&queries, list);
    ast_custom_function_unregister(oldquery->acf);
    free_acf_query(oldquery);  // frees query->acf and the query struct
}

```
However, `acf_odbc_read()` acquires `AST_RWLIST_RDLOCK(&queries)` only to locate `query`, then drops the lock and continues using `query` afterward — accessing `readhandle`, `flags`, `rowlimit`, and running SQL execution. A concurrent reload can therefore free the query/function state while the read path is still executing through it.
This appears to be a general reload-time use-after-free. It may manifest in at least two places:

- Inside `acf_odbc_read()` via direct access to a freed query
- Later in `ast_func_read()` when releasing the module user reference, as in the observed crash

The source also contains an existing note suggesting the lifetime/use-count handling is known to be incomplete:
`/* XXX need to revise usecount - set if query_lock is set */
`
`unload_module()`also has a`usleep(1)` workaround that does not appear sufficient to guarantee in-flight users have drained before the free.

Expected behavior:
Reload should not free query or function objects while active threads may still be executing through them. Query/function state should remain alive until in-flight users drain, or reload should defer freeing until it is safe to do so.

ai was used to describe the issue 

[core-asterisk-2026-03-09T11-01-12Z-locks.txt](https://github.com/user-attachments/files/25841623/core-asterisk-2026-03-09T11-01-12Z-locks.txt)
[core-asterisk-2026-03-09T11-01-12Z-thread1.txt](https://github.com/user-attachments/files/25841654/core-asterisk-2026-03-09T11-01-12Z-thread1.txt)
[core-asterisk-2026-03-09T11-01-12Z-inforedac.txt](https://github.com/user-attachments/files/25842195/core-asterisk-2026-03-09T11-01-12Z-inforedac.txt)
### Relevant log output

```shell

```

### Asterisk Issue Guidelines

phoneben:
pbx_functions/func_odbc: fix use-after-free crash during func_odbc reload with active calls

Asterisk crashes when `reload func_odbc.so` races with an active call executing an `ODBC_*` function:

```
[2026-03-09 13:01:07] loader.c: Reloading module 'func_odbc.so'

#2  __ast_module_user_remove(mod=0x21, ...)
#3  ast_func_read() at pbx_functions.c:618
```

`reload()` frees `acf_odbc_query` objects immediately while `ast_func_read()` is still executing through them. Two windows exist: between `find()` and `__ast_module_user_add()`, and after `acf_odbc_read()` drops `queries` but continues using `query->readhandle[]`.

Fix by holding `acf_root` across `find()` and `__ast_module_user_add()` in `pbx_functions.c`, making them atomic with respect to a concurrent unregister. In `func_odbc`, convert `acf_odbc_query` to ao2, take a ref in `acf_odbc_read()` before dropping `queries`, defer freeing old query objects in `reload()`, and in `unload_module()` keep backing storage alive until `__ast_module_user_hangup_all()` has forced in-flight callers to drain. This replaces the existing `usleep(1)` hack, which was an insufficient attempt at the same problem.

Resolves: #1818

### DIFF
--- a/funcs/func_odbc.c
+++ b/funcs/func_odbc.c
@@ -144,6 +144,12 @@ enum odbc_option_flags {
 
 struct acf_odbc_query {
 	AST_RWLIST_ENTRY(acf_odbc_query) list;
+	/*
+	 * Separate linkage for the deferred-free list in reload/unload.
+	 * Must not share 'list' to avoid macro mismatches and to make
+	 * ownership transfer between the two lists explicit.
+	 */
+	AST_LIST_ENTRY(acf_odbc_query) pending;
 	char readhandle[5][30];
 	char writehandle[5][30];
 	char *sql_read;
@@ -445,6 +451,10 @@ static inline void release_obj_or_dsn(struct odbc_obj **obj, struct dsn **dsn)
 }
 
 static AST_RWLIST_HEAD_STATIC(queries, acf_odbc_query);
+
+AST_MUTEX_DEFINE_STATIC(retired_queries_lock);
+/* Retired queries: unregistered but kept alive for in-flight calls */
+static AST_LIST_HEAD_NOLOCK(retired_query_list, acf_odbc_query) retired_queries = AST_LIST_HEAD_NOLOCK_INIT_VALUE;
 
 static int resultcount = 0;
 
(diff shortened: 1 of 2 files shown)


phoneben:
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 20

phoneben:
i have a issue 
 reload() accumulates retired queries indefinitely with no reclaim until unload_module(). The fix requires either a comment documenting this as intentional (safest with current APIs), or a proper drain primitive added to the Asterisk core

Issue #1845, opened 2026-03-31
[bug]:res_cdrel_custom produces wrong float timestamps


nappsoft:
### Severity

Minor

### Versions

22.9.0-rc1, 20.19.0-rc1, 23.3.0-rc1

### Components/Modules

res_cdrel_custom

### Operating Environment

Linux

### Frequency of Occurrence

None

### Issue Description

Double can not hold enough significant digits to hold a unixtime timestamp with microseconds (not even with seconds). So the code for the floater formatter should be changed as follows: 

https://github.com/asterisk/asterisk/pull/1842/changes/cd64d6c8459230436c69620adf37518925183665

Another problem that should be fixed as well in the final release: cel used to produce floating point timestamps. This can still be done with the right configuration, what's fine. However: if one is not aware of the changes (we talk here about a minor release, not a major one...), res_cdrel_custom can confuse existing code as the legacy config-converter by default is using a formatted timestamp in cel. So without rewriting the custom_cel configuration one will observe a behavior change.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

gtjoseph:
The issue with the timestamp format change is covered in #1844

nappsoft:
Just tested the patch, looks good (both aspects: timestamp accuracy and default format in cel)

Issue #1844, opened 2026-03-31
[bug]: cdrel_custom isn't respecting the default time format for CEL records


gtjoseph:
### Severity

Blocker

### Versions

23.3.0-rc1, et. al.

### Components/Modules

res_cdrel_custom

### Operating Environment

All

### Frequency of Occurrence

None

### Issue Description

The default time format for CEL records has historically been `<seconds>.<microseconds>` since epoch but it's now `%F %T`.  This can break existing log analyzers that expect the original format.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

PR #1846, opened 2026-03-31
res_cdrel_custom: Resolve several formatting issues.

gtjoseph:
Several issues are resolved:

* Internally, floats were used for timestamp values but this could result
in wrapping so they've been changed to doubles.

* Historically, the default CEL eventtime format is `<seconds>.<microseconds>`
with `<microseconds>` always being 6 digits.  This should have continued to be
the case but res_cdrel_custom wasn't checking the `dateformat` setting in
cel.conf and was defaulting to `%F %T`.  res_cdrel_custom now gets the default
date format from cel.conf, which will be whatever the `dateformat` parameter
is set to or `<seconds>.<microseconds>` if not set.

* The timeval field formatter for both CDR and CEL wasn't handling custom
strftime format strings correctly.  This is now fixed so you should be able
to specifiy custom strftime format strings for the CEL `eventtime` and CDR
`start`, `answer` and `end` fields.  For example: `eventtime(%FT%T%z)`.

Resolves: #1844
Resolves: #1845

### DIFF
--- a/configs/samples/cdr_custom.conf.sample
+++ b/configs/samples/cdr_custom.conf.sample
@@ -106,7 +106,8 @@
     ;
     ; The default output format for the "start", "answer" and "end" timestamp fields
     ; is the "%Y-%m-%d %T" strftime string format however you can also format those
-    ; fields as an int64 or a float: `start(int64),answer(float),end`.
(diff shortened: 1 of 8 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1847, opened 2026-03-31
res_pjsip_sdp_rtp: RTP socket binds to [::] when endpoint transport is auto-selected


arcivanov:
## Summary

When a PJSIP endpoint does not explicitly set `transport = <name>`, the RTP socket binds to `[::]` (or `0.0.0.0`) instead of the transport's bind address. PJSIP correctly auto-selects the transport for signaling, but `create_rtp()` in `res_pjsip_sdp_rtp.c` only resolves the transport bind address via a sorcery lookup on `session->endpoint->transport`, which is empty when the transport is auto-selected.

This causes a one-way audio problem when Asterisk runs directly on a NAT gateway using iptables DNAT + MASQUERADE, because the wildcard-bound RTP socket allows the kernel to select the WAN IP as source, creating a conntrack tuple collision with inbound DNAT entries. MASQUERADE silently remaps the RTP source port to resolve the collision, causing the remote end to receive RTP from an unexpected port.

## Steps to Reproduce

1. Configure a PJSIP transport bound to a specific LAN address:
   ```ini
   [transport-udp]
   type = transport
   protocol = udp
   bind = 192.168.x.1:5060
   external_media_address = my.public.host
   external_signaling_address = my.public.host
   local_net = 192.168.x.0/24
   ```

2. Configure an endpoint **without** explicit `transport`:
   ```ini
   [my-trunk]
   type = endpoint
   ; transport = transport-udp   ← omitted, relying on auto-selection
   ```

3. Asterisk runs on the NAT gateway with:
   ```
   -A PREROUTING -i $WAN -p udp --dport 63000:63199 -j DNAT --to-destination 192.168.x.1:63000-63199
   -A POSTROUTING -o $WAN -j MASQUERADE
   ```

4. Make an outbound call through the trunk.

**Result:** One-way audio. The remote party cannot hear the local caller. RTP from Asterisk to the provider has its source port silently remapped by MASQUERADE due to a conntrack collision. No errors appear in Asterisk logs even at debug level 99.

**Workaround:** Explicitly set `transport = transport-udp` on the endpoint.

## Root Cause

In `res/res_pjsip_sdp_rtp.c`, `create_rtp()` (line 224) determines the RTP bind address:

```c
struct ast_sockaddr *media_address = &address_rtp; /* default: [::] */

if (session->endpoint->media.bind_rtp_to_media_address && ...) {
    /* ... */
} else {
[...]

PR #1848, opened 2026-04-01
res_pjsip_sdp_rtp: Use auto-selected transport bind address for RTP

arcivanov:
When a PJSIP endpoint does not explicitly set `transport`, the RTP
socket binds to `[::]` (or `0.0.0.0`) instead of the transport's bind
address. This is because `create_rtp()` only resolves the transport
via a sorcery lookup on `session->endpoint->transport`, which is empty
when the transport is auto-selected.

On a multi-homed host or NAT gateway whose transport binds to a
specific interface, the wildcard-bound RTP socket causes the kernel to
select a source IP that may conflict with conntrack DNAT entries for
inbound RTP, causing MASQUERADE to silently remap the RTP source port.
The remote end receives RTP from an unexpected port and drops it,
resulting in one-way audio with no errors logged by Asterisk.

When no explicit transport is configured and ICE is not in use,
determine the transport type from the dialog's target URI and acquire
the transport via `pjsip_tpmgr_acquire_transport()`. If the
transport's bind address is a specific (non-wildcard) address, use it
for RTP. When the transport binds to a wildcard address (0.0.0.0 or
[::]), the RTP socket is left on the wildcard as well, matching the
behaviour of the explicit-transport code path and preserving
dual-stack operation.

When ICE is enabled the wildcard bind is left as-is since ICE needs
to discover candidates on all interfaces.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

UserNote: When a PJSIP endpoint does not have an explicit `transport`
configured, ICE is disabled, and the auto-selected transport binds to
a specific (non-wildcard) address, the RTP socket now binds to that
address instead of the wildcard. This fixes one-way audio on
multi-homed hosts and NAT gateways where the wildcard bind caused the
kernel to select an incorrect source address for outbound RTP.

Fixes: #1847

### DIFF
--- a/res/res_pjsip_sdp_rtp.c
+++ b/res/res_pjsip_sdp_rtp.c
@@ -264,6 +264,82 @@ static int create_rtp(struct ast_sip_session *session, struct ast_sip_session_me
 				ao2_ref(trans_state, -1);
 			}
 			ao2_ref(transport, -1);
+		} else if (!session->endpoint->media.rtp.ice_support
+				&& session->inv_session && session->inv_session->dlg
+				&& (PJSIP_URI_SCHEME_IS_SIP(session->inv_session->dlg->target)
+					|| PJSIP_URI_SCHEME_IS_SIPS(session->inv_session->dlg->target))) {
+			/*
+			 * The endpoint does not have an explicit transport configured
+			 * and ICE is not in use. Acquire the transport that would be
+			 * used for the dialog's target URI and, when it is bound to a
+			 * specific (non-wildcard) address, use that address for RTP.
+			 *
+			 * On a multi-homed or NAT-gateway host whose transport binds
+			 * to a specific interface, a wildcard-bound RTP socket causes
+			 * the kernel to choose a source IP that may conflict with
+			 * conntrack DNAT entries for inbound RTP.
+			 *
+			 * When the transport itself binds to a wildcard address
+			 * (0.0.0.0 or [::]), the RTP socket is left on the wildcard
+			 * as well — this matches the behaviour of the explicit
+			 * transport code path above and preserves dual-stack
+			 * operation.
+			 *
+			 * When ICE is enabled the wildcard bind is intentional so
+			 * that ICE can discover candidates on all interfaces.
+			 */
+			pjsip_sip_uri *sip_uri;
+			pjsip_transport_type_e type;
+
+			sip_uri = pjsip_uri_get_uri(session->inv_session->dlg->target);
+			type = pjsip_transport_get_type_from_name(&sip_uri->transport_param);
+			if (PJSIP_URI_SCHEME_IS_SIPS(sip_uri)) {
+				if (type == PJSIP_TRANSPORT_UNSPECIFIED
+						|| !(pjsip_transport_get_flag_from_type(type) & PJSIP_TRANSPORT_SECURE)) {
+					type = PJSIP_TRANSPORT_TLS;
+				}
+			} else if (!sip_uri->transport_param.slen) {
+				type = PJSIP_TRANSPORT_UDP;
+			}
+
+			if (pj_strchr(&sip_uri->host, ':')) {
+				type |= PJSIP_TRANSPORT_IPV6;
+			}
+
+			if (type != PJSIP_TRANSPORT_UNSPECIFIED
+					&& (pjsip_transport_get_flag_from_type(type) & PJSIP_TRANSPORT_DATAGRAM)) {
+				pjsip_transport *tp = NULL;
+				pj_sockaddr remote;
+				int addr_len;
+
+				pj_bzero(&remote, sizeof(remote));
+				if (type & PJSIP_TRANSPORT_IPV6) {
+					addr_len = sizeof(pj_sockaddr_in6);
+					remote.addr.sa_family = pj_AF_INET6();
+				} else {
+					addr_len = sizeof(pj_sockaddr_in);
+					remote.addr.sa_family = pj_AF_INET();
+				}
+
+				if (pjsip_tpmgr_acquire_transport(
+						pjsip_endpt_get_tpmgr(ast_sip_get_pjsip_endpoint()),
+						type, &remote, addr_len,
+						&session->inv_session->dlg->tp_sel,
+						&tp) == PJ_SUCCESS) {
+					if (pj_sockaddr_has_addr(&tp->local_addr)) {
+						char hoststr[PJ_INET6_ADDRSTRLEN];
+
+						pj_sockaddr_print(&tp->local_addr, hoststr,
+							sizeof(hoststr), 0);
+						if (ast_sockaddr_parse(&temp_media_address, hoststr, 0)) {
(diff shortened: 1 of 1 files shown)


arcivanov:
This has been verified to work with the patch on top of 22.8.2.

res/res_pjsip_sdp_rtp.c
```diff
@@ -264,6 +264,52 @@ static int create_rtp(struct ast_sip_session *session, struct ast_sip_session_me
 				ao2_ref(trans_state, -1);
 			}
 			ao2_ref(transport, -1);
+		} else if (session->inv_session && session->inv_session->dlg
+				&& (PJSIP_URI_SCHEME_IS_SIP(session->inv_session->dlg->target)
+					|| PJSIP_URI_SCHEME_IS_SIPS(session->inv_session->dlg->target))) {
+			/*
+			 * The endpoint does not have an explicit transport configured.
```
jcolp: This change will not be accepted as-is. The binding is done on purpose against wildcard so that ICE negotiation can use all available candidates on the system and allow media to flow using any of them. That behavior will not change. If you want new behavior, then that's a new option you have to implement.

jcolp (changes requested):

res/res_pjsip_sdp_rtp.c
jcolp: Additionally, you must disclose use of AI.

arcivanov: > This change will not be accepted as-is. The binding is done on purpose against wildcard so that ICE negotiation can use all available candidates on the system and allow media to flow using any of them. That behavior will not change. If you want new behavior, then that's a new option you have to implement.

It's not that I want a new behavior is that the there is a broken outcome with misrouted packets.

If a transport is specified explicitly the IP binding is correct. If the transport is auto-selected the wildcard produces one-way audio in a multi-homed gateway system. I will recheck the configuration but ICE is explicitly turned off on the endpoint. 

Would you accept this if ICE is checked to be off?

> Additionally, you must disclose use of AI.

It's done with Claude Code assistance. What other type of disclosure would you like?

jcolp: If ICE is turned off, then yes, I think that could be accepted. As for AI, our policy is here:

https://docs.asterisk.org/Development/Policies-and-Procedures/AI-Policy/

arcivanov: I'll submit the patch with a check for ICE being off and an AI signoff notice.

arcivanov: Should be done now.

arcivanov: Verified on top of the same 22.8.2 LTS

jcolp (changes requested):
The commit message needs to have a UserNote[1] informing users of this behavior change.

[1] https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/?h=usernote#special-trailers-for-commit-messages

jcolp:
Okay, you're merging to 22 and not master. That's why those are failing because your 22 is out of date that you're based on. I was  confused.

You also need to take care of the cherry-pick-to lines that the bot stated.

In the future base things off of master and ensure you're up to date, or else you'll end up with old test failures.

jcolp (dismissed):
Code seems okay, but process side needs to be taken care of.

arcivanov:
Would you like me to rebase on the latest 22 or submit a different PR against the master?
I was under impression cherry-pick is something you guys decide, no?

jcolp:
When putting up a PR you add cherry-pick-to lines in a comment to specify where the change goes. If something is amiss with that, then that is brought up for discussion. I don't see why this couldn't go in the applicable branches which would be 20, 23, and master alongside 22. This does not need to be put up as a separate PR against master, my comment was for the future.

https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/#cherry-picking

arcivanov:
Understood in re cherry-picking. What would you like to do with PR? Rebase on latest or close and create PR against the master with cherry-picking to 20, 22, 23?

jcolp:
Unless @gtjoseph says otherwise, this can be rebased on latest 22 and the cherry-pick-to lines done here.

arcivanov:
cherry-pick-to: 20
cherry-pick-to: 23
cherry-pick-to: 24
cherry-pick-to: master

arcivanov:
Hopefully everything's fixed now!

gtjoseph:
> Hopefully everything's fixed now!

Sorry, not yet.

You can ignore the caller_local_blonde test failure but I tested the other 3 locally and they all fail with your PR but pass with the current 22 branch.  You'll need to investigate and determine if your PR needs to be adjusted or whether the testsuite tests need to be adjusted.  You should fork and clone the testsuite repository and run those tests yourself.

If you determine that the tests need to be adjusted, you can create a PR in the testsuite repo, then in this PR, add a `testsuite-test-pr: xxxx` line to the existing cherry-pick comment which will allow this PR to be tested with your testsuite PR.

See:
https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/#test-against-a-testsuite-pr

gtjoseph (dismissed):
Previously mentioned test failures need investigation.

arcivanov:
will do

arcivanov:
@gtjoseph why would IAX2 fail with SIP changes is beyond me

gtjoseph:
> @gtjoseph why would IAX2 fail with SIP changes is beyond me

`local_iax2_mgr` is just the name of the group of tests.  In this case, it contains tests for channels/local, channels/iax2 and manager.

arcivanov:
Reconfirmed working on top of LTS 22.2.8.

arcivanov:
FYI: Tested the patch with 22.9.0, no problems, no one-way audio: https://copr.fedorainfracloud.org/coprs/karellen/asterisk/

res/res_pjsip_sdp_rtp.c
```diff
@@ -264,6 +264,82 @@ static int create_rtp(struct ast_sip_session *session, struct ast_sip_session_me
...
+			 * transport code path above and preserves dual-stack
+			 * operation.
+			 *
+			 * When ICE is enabled the wildcard bind is intentional so
+			 * that ICE can discover candidates on all interfaces.
+			 */
+			pjsip_sip_uri *sip_uri;
+			pjsip_transport_type_e type;
+
+			sip_uri = pjsip_uri_get_uri(session->inv_session->dlg->target);
```
jcolp, line 294: Has the case been tested with this where a SIP INVITE is sent to a hostname that does SRV/NAPTR and resolves down to multiple targets, and then goes from IPv6 to IPv4 as a result of failure/failover? Would that result in the SDP containing IPv6 despite being an IPv4 target? Or worse - do we update the SDP elsewhere under the assumption that it is wildcard, but it's not so audio would fail?

arcivanov: I definitely didn't test that scenario. Let me see what I can do.

PR #1851, opened 2026-04-02
res_pjsip: Add per-endpoint RTP port range configuration

### LINKED ISSUE #71: [new-feature]: Per-endpoint RTP port range configuration for PJSIP
### Feature Description

**1. The Problem**
Currently, Asterisk manages RTP ports globally through rtp.conf. While this is suitable for most deployments, it lacks granular control. In complex environments, especially those involving multiple NATs or specific firewall rules, it would be highly beneficial to define a dedicated RTP port range for each PJSIP endpoint.

For instance, a proxy or a session border controller (SBC) might require specific endpoints to communicate on a predefined, restricted set of ports. The current global rtp.conf setting forces all endpoints to use the same large port range, which is inefficient and difficult to manage in such scenarios.

**2. The Proposed Solution**
I propose adding a new configuration parameter to the PJSIP endpoint configuration in pjsip.conf. This parameter would allow developers to specify a dedicated RTP port range for each endpoint, overriding the global rtp.conf settings for that specific endpoint.

**Proposed Parameters:**
**rtp_start**: Defines the starting port number for the dedicated RTP range.
**rtp_end**: Defines the ending port number for the dedicated RTP range.

Example in pjsip.conf:

```
[webrtc_endpoint_1]
type=endpoint
context=webrtc-context
aors=webrtc_endpoint_1
transport=transport-ws
rtp_start=15000
rtp_end=15005

[webrtc_endpoint_2]
type=endpoint
context=webrtc-context
aors=webrtc_endpoint_2
transport=transport-ws
rtp_start=15006
rtp_end=15011

```
**3. Benefits of this Feature**
Improved NAT Traversal: Provides better control for devices behind complex NAT configurations, allowing specific port forwarding rules.

Enhanced Security: Allows administrators to restrict RTP traffic to a minimal, endpoint-specific port range, reducing the attack surface.

Greater Flexibility for Integrators: Enables third-party applications, like media proxies or SBCs, to integrate with Asterisk more seamlessly by using predictable and non-conflicting port ranges.

Efficient Port Management: Prevents the exhaustion of the global port range in large-scale deployments by assigning smaller, dedicated ranges to each endpoint.

**4. Technical Considerations**
Conflict Resolution: The implementation must handle potential conflicts if multiple endpoints are configured with overlapping or identical port ranges. The system should log an error and fall back to the global rtp.conf setting in case of a conflict.

Dynamic vs. Static: This feature should be an override for the dynamic port allocation. If rtp_start and rtp_end are not defined for an endpoint, Asterisk should continue to use the global rtp.conf range as it does today.

**5. Call to Action**
I believe this feature would be a valuable addition to Asterisk, addressing real-world challenges faced by system integrators and administrators. I am open to discussing the implementation details and contributing to the development of this feature.

Thank you for your consideration.

wlp2s0:
Add rtp_port_start and rtp_port_end options to PJSIP endpoint
configuration, allowing each endpoint to use a dedicated RTP port
range instead of the global rtp.conf setting.

This is useful for scenarios where different endpoints need isolated
port ranges, such as firewall rules per trunk, multi-tenant systems,
or network QoS policies tied to port ranges.

The implementation adds ast_rtp_instance_new_with_port_range() to the
RTP engine API, which sets the port range on the instance before the
engine allocates the transport. The default RTP engine
(res_rtp_asterisk) checks for per-instance overrides in
rtp_allocate_transport() and falls back to the global range when
none is set.

Both options must be set together, with values >= 1024 and
rtp_port_end > rtp_port_start. Setting both to 0 (the default)
preserves existing behavior.

Resolves: https://github.com/asterisk/asterisk-feature-requests/issues/71

UserNote: PJSIP endpoints now support rtp_port_start and
rtp_port_end options to configure a dedicated RTP port range per
endpoint, overriding the global rtp.conf setting.

UpgradeNote: An alembic database migration has been added to add
the rtp_port_start and rtp_port_end columns to the ps_endpoints
table. Run "alembic upgrade head" to apply the schema change.

DeveloperNote: New public API: ast_rtp_instance_new_with_port_range()
creates an RTP instance with a per-instance port range.
ast_rtp_instance_get_port_start() and ast_rtp_instance_get_port_end()
allow RTP engines to query the override. Third-party RTP engines can
use these getters to support per-instance port ranges.

### DIFF
--- a/include/asterisk/rtp_engine.h
+++ b/include/asterisk/rtp_engine.h
@@ -981,6 +981,60 @@ struct ast_rtp_instance *ast_rtp_instance_new(const char *engine_name,
                 void *data);
 
 /*!
+ * \brief Options for creating a new RTP instance
+ *
+ * This structure allows passing additional options when creating an
+ * RTP instance via \ref ast_rtp_instance_new_with_options. New fields
+ * can be added in the future without changing the function signature.
+ *
+ * \since 20.20.0
+ * \since 22.10.0
+ * \since 23.4.0
+ */
+struct ast_rtp_instance_options {
+	/*! Starting port number for this instance (0 to use global) */
+	unsigned int port_start;
+	/*! Ending port number for this instance (0 to use global) */
+	unsigned int port_end;
+};
+
+/*!
+ * \brief Create a new RTP instance with additional options
+ *
+ * \param engine_name Name of the engine to use for the RTP instance
+ * \param sched Scheduler context that the RTP engine may want to use
+ * \param sa Address we want to bind to
+ * \param data Unique data for the engine
+ * \param options Pointer to options struct, or NULL to use defaults
+ *
+ * \retval non-NULL success
+ * \retval NULL failure
+ *
+ * Example usage:
+ *
+ * \code
+ * struct ast_rtp_instance_options options = { .port_start = 15000, .port_end = 15010 };
+ * struct ast_rtp_instance *instance = NULL;
+ * instance = ast_rtp_instance_new_with_options(NULL, sched, &sin, NULL, &options);
+ * \endcode
+ *
+ * This creates a new RTP instance using the specified options. When a
+ * per-instance port range is provided the global rtp.conf range is ignored.
+ * If options is NULL or port values are both 0 the global range is used.
+ *
+ * \note The per-instance port range overrides the global rtp.conf settings
+ *       for this specific RTP instance only. It does not need to be a
+ *       subset of the global range.
+ *
+ * \since 20.20.0
+ * \since 22.10.0
+ * \since 23.4.0
+ */
+struct ast_rtp_instance *ast_rtp_instance_new_with_options(const char *engine_name,
+                struct ast_sched_context *sched, const struct ast_sockaddr *sa,
+                void *data, const struct ast_rtp_instance_options *options);
+
+/*!
  * \brief Destroy an RTP instance
  *
  * \param instance The RTP instance to destroy
@@ -2667,6 +2721,32 @@ int ast_rtp_instance_get_hold_timeout(struct ast_rtp_instance *instance);
  * \since 1.8
  */
 int ast_rtp_instance_get_keepalive(struct ast_rtp_instance *instance);
+
+/*!
+ * \brief Get the per-instance RTP port range start
+ *
+ * \param instance The RTP instance
+ *
+ * \return port start value (0 means use global setting)
+ *
+ * \since 20.20.0
+ * \since 22.10.0
+ * \since 23.4.0
+ */
+unsigned int ast_rtp_instance_get_port_start(struct ast_rtp_instance *instance);
+
+/*!
+ * \brief Get the per-instance RTP port range end
+ *
+ * \param instance The RTP instance
+ *
+ * \return port end value (0 means use global setting)
+ *
+ * \since 20.20.0
+ * \since 22.10.0
+ * \since 23.4.0
+ */
+unsigned int ast_rtp_instance_get_port_end(struct ast_rtp_instance *instance);
 
 /*!
  * \brief Get the RTP engine in use on an RTP instance
--- a/res/res_pjsip/pjsip_config.xml
+++ b/res/res_pjsip/pjsip_config.xml
@@ -1659,6 +1659,35 @@
 						channel is hung up. By default this option is set to 0, which means do not check.
 					</para></description>
 				</configOption>
+				<configOption name="rtp_port_start" default="0">
+					<since>
+						<version>20.20.0</version>
+						<version>22.10.0</version>
+						<version>23.4.0</version>
+					</since>
+					<synopsis>Per-endpoint starting RTP port number.</synopsis>
+					<description><para>
+						Defines the starting port number for a dedicated per-endpoint RTP port
+						range, overriding the global rtp.conf setting. Must be used together
+						with <literal>rtp_port_end</literal>. When set to 0, the global
+						rtp.conf range is used. The value must be between 1024 and 65535.
+					</para></description>
+				</configOption>
+				<configOption name="rtp_port_end" default="0">
+					<since>
+						<version>20.20.0</version>
+						<version>22.10.0</version>
+						<version>23.4.0</version>
+					</since>
+					<synopsis>Per-endpoint ending RTP port number.</synopsis>
+					<description><para>
+						Defines the ending port number for a dedicated per-endpoint RTP port
+						range, overriding the global rtp.conf setting. Must be used together
+						with <literal>rtp_port_start</literal>. When set to 0, the global
+						rtp.conf range is used. The value must be greater than
+						<literal>rtp_port_start</literal> and between 1024 and 65535.
+					</para></description>
+				</configOption>
 				<configOption name="acl">
 					<since>
 						<version>13.10.0</version>
--- a/res/res_pjsip/pjsip_configuration.c
+++ b/res/res_pjsip/pjsip_configuration.c
@@ -1666,6 +1666,30 @@ static int sip_endpoint_apply_handler(const struct ast_sorcery *sorcery, void *o
 		return -1;
 	}
 
+	if (endpoint->media.rtp.port_start || endpoint->media.rtp.port_end) {
+		if (!endpoint->media.rtp.port_start || !endpoint->media.rtp.port_end) {
+			ast_log(LOG_ERROR, "Endpoint '%s': Both rtp_port_start and rtp_port_end must be set together\n",
+				ast_sorcery_object_get_id(endpoint));
+			return -1;
+		}
+		if (endpoint->media.rtp.port_start < 1024 || endpoint->media.rtp.port_end < 1024) {
+			ast_log(LOG_ERROR, "Endpoint '%s': rtp_port_start and rtp_port_end must be at least 1024\n",
+				ast_sorcery_object_get_id(endpoint));
+			return -1;
+		}
+		if (endpoint->media.rtp.port_end <= endpoint->media.rtp.port_start) {
+			ast_log(LOG_ERROR, "Endpoint '%s': rtp_port_end (%u) must be greater than rtp_port_start (%u)\n",
+				ast_sorcery_object_get_id(endpoint),
+				endpoint->media.rtp.port_end,
+				endpoint->media.rtp.port_start);
+			return -1;
+		}
+		ast_debug(1, "Endpoint '%s': Using per-endpoint RTP port range %u-%u\n",
+			ast_sorcery_object_get_id(endpoint),
+			endpoint->media.rtp.port_start,
+			endpoint->media.rtp.port_end);
+	}
+
 	if (endpoint->preferred_codec_only) {
 		if (endpoint->media.incoming_call_offer_pref.flags != (AST_SIP_CALL_CODEC_PREF_LOCAL | AST_SIP_CALL_CODEC_PREF_INTERSECT | AST_SIP_CALL_CODEC_PREF_ALL)) {
 			ast_log(LOG_ERROR, "Setting both preferred_codec_only and incoming_call_offer_pref is not supported on endpoint '%s'\n",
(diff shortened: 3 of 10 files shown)


wlp2s0:
**AI Disclosure**

This implementation was developed with the assistance of AI tools. AI was primarily used for code style normalization, ensuring consistency with Asterisk's coding guidelines regarding indentation, brace style, line length, and Doxygen formatting. As a non-native English speaker, I also relied on AI to refine the English in code comments, documentation, and this PR description for maximum clarity.

All code was written with full understanding of the Asterisk RTP engine architecture, reviewed line-by-line, compiled locally, and functionally tested in a Docker-based deployment with multiple endpoints verifying correct port allocation. I'm happy to discuss any aspect of the implementation.

wlp2s0:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

include/asterisk/rtp_engine.h
```diff
@@ -980,6 +980,39 @@ struct ast_rtp_instance *ast_rtp_instance_new(const char *engine_name,
...
+ * \endcode
+ *
+ * This creates a new RTP instance using the specified port range instead of
+ * the global rtp.conf range. If port_start and port_end are both 0, the
+ * global range from rtp.conf will be used as a fallback.
+ *
+ * \note The per-instance port range overrides the global rtp.conf settings
+ *       for this specific RTP instance only.
+ *
+ * \since master
```
jcolp: We don't use \since master as master is not a fixed version, if this is not going to go into current branches then it would be the next major version number instead - 24

include/asterisk/rtp_engine.h
```diff
@@ -980,6 +980,39 @@ struct ast_rtp_instance *ast_rtp_instance_new(const char *engine_name,
...
+ * \code
+ * struct ast_rtp_instance *instance = NULL;
+ * instance = ast_rtp_instance_new_with_port_range(NULL, sched, &sin, NULL, 15000, 15010);
+ * \endcode
+ *
+ * This creates a new RTP instance using the specified port range instead of
+ * the global rtp.conf range. If port_start and port_end are both 0, the
+ * global range from rtp.conf will be used as a fallback.
+ *
+ * \note The per-instance port range overrides the global rtp.conf settings
```
jcolp, line 1025: Clarify that this means it does not need to be a subset

include/asterisk/rtp_engine.h
```diff
@@ -2668,6 +2701,28 @@ int ast_rtp_instance_get_hold_timeout(struct ast_rtp_instance *instance);
...
 int ast_rtp_instance_get_keepalive(struct ast_rtp_instance *instance);
 
+/*!
+ * \brief Get the per-instance RTP port range start
+ *
+ * \param instance The RTP instance
+ *
+ * \return port start value (0 means use global setting)
+ *
+ * \since master
```
jcolp: See previous comment

include/asterisk/rtp_engine.h
```diff
@@ -2668,6 +2701,28 @@ int ast_rtp_instance_get_hold_timeout(struct ast_rtp_instance *instance);
...
+int ast_rtp_instance_get_port_start(struct ast_rtp_instance *instance);
+
+/*!
+ * \brief Get the per-instance RTP port range end
+ *
+ * \param instance The RTP instance
+ *
+ * \return port end value (0 means use global setting)
+ *
+ * \since master
```
jcolp: See previous comment

res/res_pjsip/pjsip_config.xml
```diff
@@ -1659,6 +1659,31 @@
 						channel is hung up. By default this option is set to 0, which means do not check.
 					</para></description>
 				</configOption>
+				<configOption name="rtp_port_start" default="0">
+					<since>
+						<version>master</version>
```
jcolp: See previous comment

res/res_pjsip/pjsip_config.xml
```diff
@@ -1659,6 +1659,31 @@
...
+					<description><para>
+						Defines the starting port number for a dedicated per-endpoint RTP port
+						range, overriding the global rtp.conf setting. Must be used together
+						with <literal>rtp_port_end</literal>. When set to 0, the global
+						rtp.conf range is used. The value must be between 1024 and 65535.
+					</para></description>
+				</configOption>
+				<configOption name="rtp_port_end" default="0">
+					<since>
+						<version>master</version>
```
jcolp: See previous comment

res/res_pjsip/pjsip_configuration.c
```diff
@@ -1666,6 +1666,35 @@ static int sip_endpoint_apply_handler(const struct ast_sorcery *sorcery, void *o
...
+			return -1;
+		}
+		if (endpoint->media.rtp.port_end <= endpoint->media.rtp.port_start) {
+			ast_log(LOG_ERROR, "Endpoint '%s': rtp_port_end (%u) must be greater than rtp_port_start (%u)\n",
+				ast_sorcery_object_get_id(endpoint),
+				endpoint->media.rtp.port_end,
+				endpoint->media.rtp.port_start);
+			return -1;
+		}
+		ast_log(LOG_NOTICE, "Endpoint '%s': Using per-endpoint RTP port range %u-%u\n",
```
jcolp: We don't do notice messages for things like this, if it's useful for debugging then it should be a debug message instead

jcolp (changes requested):

include/asterisk/rtp_engine.h
wlp2s0: Fixed, changed to \since 24 in all three declarations

include/asterisk/rtp_engine.h
wlp2s0: Added "It does not need to be a subset of the global range." :+1:

res/res_pjsip/pjsip_configuration.c
wlp2s0: Changed to ast_debug

res/res_pjsip/pjsip_config.xml
wlp2s0: Fixed

res/res_pjsip/pjsip_config.xml
wlp2s0: Fixed

include/asterisk/rtp_engine.h
wlp2s0: Fixed

include/asterisk/rtp_engine.h
wlp2s0: Fixed

include/asterisk/rtp_engine.h
```diff
@@ -980,6 +980,40 @@ struct ast_rtp_instance *ast_rtp_instance_new(const char *engine_name,
...
+ * the global rtp.conf range. If port_start and port_end are both 0, the
+ * global range from rtp.conf will be used as a fallback.
+ *
+ * \note The per-instance port range overrides the global rtp.conf settings
+ *       for this specific RTP instance only. It does not need to be a
+ *       subset of the global range.
+ *
+ * \since 24
+ */
+struct ast_rtp_instance *ast_rtp_instance_new_with_port_range(const char *engine_name,
```
seanbright: Even though it’s unlikely to be needed I think it would be better for this to be `ast_rtp_instance_new_options(…)` or `ast_rtp_instance_new_with_options(…)` and take a pointer to a new options struct.

This would allow future customization without needing additional initialization functions.

wlp2s0: Makes sense and it wasn't a big change.. Definitely more future proof. Already tested and working, updated in the latest push. Thanks!

include/asterisk/rtp_engine.h
```diff
@@ -980,6 +980,56 @@ struct ast_rtp_instance *ast_rtp_instance_new(const char *engine_name,
...
+ * RTP instance via \ref ast_rtp_instance_new_with_options. New fields
+ * can be added in the future without changing the function signature.
+ *
+ * \since 24
+ */
+struct ast_rtp_instance_options {
+	/*! Starting port number for this instance (0 to use global) */
+	int port_start;
+	/*! Ending port number for this instance (0 to use global) */
+	int port_end;
```
seanbright: Make these unsigned to be consistent with the `ast_sip_media_rtp_configuration` changes.

res/res_pjsip/pjsip_configuration.c
```diff
@@ -2291,6 +2320,8 @@ int ast_res_pjsip_initialize_configuration(void)
 	ast_sorcery_object_field_register(sip_sorcery, "endpoint", "rtp_keepalive", "0", OPT_UINT_T, 0, FLDSET(struct ast_sip_endpoint, media.rtp.keepalive));
 	ast_sorcery_object_field_register(sip_sorcery, "endpoint", "rtp_timeout", "0", OPT_UINT_T, 0, FLDSET(struct ast_sip_endpoint, media.rtp.timeout));
 	ast_sorcery_object_field_register(sip_sorcery, "endpoint", "rtp_timeout_hold", "0", OPT_UINT_T, 0, FLDSET(struct ast_sip_endpoint, media.rtp.timeout_hold));
+	ast_sorcery_object_field_register(sip_sorcery, "endpoint", "rtp_port_start", "0", OPT_UINT_T, 0, FLDSET(struct ast_sip_endpoint, media.rtp.port_start));
+	ast_sorcery_object_field_register(sip_sorcery, "endpoint", "rtp_port_end", "0", OPT_UINT_T, 0, FLDSET(struct ast_sip_endpoint, media.rtp.port_end));
```
seanbright: Use `PARSE_IN_RANGE` instead - see how `external_signaling_port` is registered in `res/res_pjsip/config_transport.c`

include/asterisk/rtp_engine.h
wlp2s0: Updated the whole chain to unsigned int. Thanks!

res/res_pjsip/pjsip_configuration.c
wlp2s0: Done, much cleaner :+1:

res/res_pjsip/pjsip_config.xml
```diff
@@ -1659,6 +1659,31 @@
...
+				<configOption name="rtp_port_start" default="0">
+					<since>
+						<version>24</version>
+					</since>
+					<synopsis>Per-endpoint starting RTP port number.</synopsis>
+					<description><para>
+						Defines the starting port number for a dedicated per-endpoint RTP port
+						range, overriding the global rtp.conf setting. Must be used together
+						with <literal>rtp_port_end</literal>. When set to 0, the global
+						rtp.conf range is used. The value must be between 1024 and 65535.
```
seanbright, line 1671: > The value must be between 1024 and 65535.

Is that true?

Edit: Or I guess I should ask - why is that? Why limit the lower bound to 1024?

seanbright: Ah. Because that is what we do in res_rtp_asterisk. Sorry for the noise.

res/res_pjsip/pjsip_config.xml
```diff
@@ -1659,6 +1659,31 @@
 						channel is hung up. By default this option is set to 0, which means do not check.
 					</para></description>
 				</configOption>
+				<configOption name="rtp_port_start" default="0">
+					<since>
+						<version>24</version>
```
jcolp: 24.0.0

res/res_pjsip/pjsip_config.xml
```diff
@@ -1659,6 +1659,31 @@
...
+					<description><para>
+						Defines the starting port number for a dedicated per-endpoint RTP port
+						range, overriding the global rtp.conf setting. Must be used together
+						with <literal>rtp_port_end</literal>. When set to 0, the global
+						rtp.conf range is used. The value must be between 1024 and 65535.
+					</para></description>
+				</configOption>
+				<configOption name="rtp_port_end" default="0">
+					<since>
+						<version>24</version>
```
jcolp: 24.0.0

jcolp (changes requested):

res/res_pjsip/pjsip_config.xml
wlp2s0: Fixed, updated to `24.0.0`.

res/res_pjsip/pjsip_config.xml
wlp2s0: Fixed, updated to `24.0.0`.

jcolp (approved):

gtjoseph (dismissed):
After talking it over with @jcolp,  we'd like this PR to go into the 20, 22 and 23 branches as well.  Since you touch some common files like the alembic scripts and documentation, leaving this as master-only can cause problems for future pull requests because, especially for alembic, automatic cherry-picking will fail.  Sorry for the change in direction.

The change should be fairly simple... just replace the single `<version>24.0.0</version>` elements in the XML files with...

```xml
<version>20.20.0</version>
<version>22.10.0</version>
<version>23.4.0</version>
```

...and the single `\since 24` tags in the header files with...

```
 * \since 20.20.0
 * \since 22.10.0
 * \since 23.4.0
```

Thanks!

Oh, and update the existing PR comment containing `cherry-pick-to: none` to add a line for each branch.

wlp2s0:
Done :)

mahirgul:
Thanks guys

Issue #1852, opened 2026-04-02
[bug]: res_cdrel_custom: connection to the sqlite3 database closes from time to time


nappsoft:
### Severity

Major

### Versions

22.9.0-rc1

### Components/Modules

res_cdrel_custom

### Operating Environment

Linux 6.6.130 with musl libc

### Frequency of Occurrence

None

### Issue Description

After some time (sometimes many hours) logging into the cel or cdr database over SQL stops working. Checking the list of open files with lsof shows that asterisk is not accessing the master.db file anymore in this case.

This never happened before the refactoring in https://github.com/asterisk/asterisk/commit/fd781d966a9be7cce946ca1ba27ea3447bfdfdeb

I will for the moment revert this change on our productive system but will let the new code run on some testsystems. Unfortunately I do not have any logs indicating reasons so far. However I thought it would be important for the core devs to know that this change might need to be excluded from the 22.9.0 release till the reason for this behavior is found.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

- [x] Yes, I have read the Asterisk Issue Guidelines

nappsoft:
Might be faster in analyzing than I thought: a core reload is enough to let the connection stop:

> schulung02*CLI> core restart now
schulung02*CLI> 
Disconnected from Asterisk server
Asterisk cleanly ending (0).
Executing last minute cleanups
~ # lsof | grep master.db
1875    /usr/sbin/asterisk      19      /var/log/asterisk/master.db
1875    /usr/sbin/asterisk      20      /var/log/asterisk/master.db-wal
1875    /usr/sbin/asterisk      21      /var/log/asterisk/master.db-shm
1875    /usr/sbin/asterisk      22      /var/log/asterisk/master.db
1875    /usr/sbin/asterisk      23      /var/log/asterisk/master.db-wal
~ # asterisk -r
'alwaysfork' is not compatible with console or remote console mode; ignored
Asterisk 22.9.0-rc1, Copyright (C) 1999 - 2025, Sangoma Technologies Corporation and others.
Created by Mark Spencer <markster@digium.com>
Asterisk comes with ABSOLUTELY NO WARRANTY; type 'core show warranty' for details.
This is free software, with components licensed under the GNU General Public
License version 2 and other licenses; you are welcome to redistribute it under
certain conditions. Type 'core show license' for details.
=========================================================================
Connected to Asterisk 22.9.0-rc1 currently running on schulung02 (pid = 1875)
[...]

PR #1853, opened 2026-04-02
res_cdrel_custom: do not free config when no new config was loaded

nappsoft:
When the res_cdrel_custom modules is reloaded and the config has not been changed asterisk should not free the old config. Otherwise the connection to the database will be closed and no new connection will be opened.

Resolves: #1852


nappsoft:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

PR #1854, opened 2026-04-06
build: remove pjsua, pjsystest, Python bindings and asterisk_malloc_debug stubs from pjproject dev build

alex2grad:
The pjsua Python module and the pjsua/pjsystest apps were used by the
Asterisk Test Suite for SIP simulation in dev mode builds. They are now
fully obsolete for three independent reasons:

1. **pjsua Python bindings officially deprecated upstream.** The pjproject
   maintainers added `pjsip-apps/src/python/DEPRECATED.txt` directing
   users to the PJSUA2 SWIG binding instead. A build-fix PR
   (https://github.com/pjsip/pjproject/pull/4892) was closed by the
   maintainer explicitly citing this deprecation.

2. **Removed from the Asterisk Test Suite.** As confirmed by @mbradeen:
   > *"We had to get rid of pjsua when we went to Python3 because it would
   > hang due to a conflict between async calls within pjsua and twisted.
   > There are still some old references to tests we couldn't fully convert
   > to sipp, but those are skipped."*

3. **Broken and unmaintained.** Building with Python 2.7 (the only version
   `configure.ac` searched for) fails with:
   ```
   _pjsua.c: error: 'INIT_RETURN' undeclared (first use in this function)
   ```
   due to a bug in pjproject 2.16's `python3_compat.h` that upstream
   declined to fix.

This PR removes all pjsua-related build artifacts from Asterisk's bundled
pjproject build: the pjsua and pjsystest application binaries, the deprecated
Python (`_pjsua.so`) bindings, the `asterisk_malloc_debug.c` stubs, and the
`PYTHONDEV` detection from `configure.ac`. Also removes `libpjsua` from
Asterisk's main linker flags.

DeveloperNote: The pjsua and pjsystest application binaries, the deprecated
Python pjsua bindings (`_pjsua.so`), and the `asterisk_malloc_debug.c` stub
implementations are no longer built or installed as part of the bundled
pjproject dev mode build. The `PYTHONDEV` (python2.7-dev) build dependency
is also removed. Developers who relied on the pjsua binary for Test Suite
SIP simulation should use SIPp instead, which is the current Asterisk Test
Suite standard.

Fixes: #1840


alex2grad:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

Issue #1855, opened 2026-04-07
[bug]: core reload deadlocks Asterisk (pjsip, CLI, etc.)


Innosoft1:
### Severity

Major

### Versions

20.18.2

### Components/Modules

res_pjsip, res_pjsip_config_wizard

### Operating Environment

Debian 12.13

### Frequency of Occurrence

Occasional

### Issue Description

After upgrading to 20.18.2, we observed the Asterisk CLI freezing and pjsip stopping to process SIP packets while invoking "core reload".

This happened on two different Asterisk instances with different network setups within a week. 

core reload reloads res_pjsip_config_wizard which refreshes the transport states and gets stuck waiting for a lock. I see that other threads are waiting for core reload to finish but I don't know why or for which lock core reload is waiting for.

I've read several possibly, similiar issues but there are no network/DNS issues on both VMs which run on different hardware. Our transport config is minimalistic and does not use any invalid syntax. allow_reload defaults to false.

```
[transport-udp]
type = transport
protocol = udp
bind = 1.2.3.4:5160

[transport-tcp]
type = transport
protocol = tcp
bind = 1.2.3.4:5160

[transport-tls]
type = transport
protocol = tls
bind = 1.2.3.4:5161
cert_file = /etc/letsencrypt/live/domain.com/fullchain.pem
priv_key_file = /etc/letsencrypt/live/domain.com/privkey.pem
method = tlsv1_2

```
We had issues with pjsip freezing (not processing any packets until Asterisk was restarted) on earlier version of Asterisk 20 but those happened very sporadic like every 2-3 months. We mitigated those by restarting Asterisk every 2-3 weeks and implementing a monitoring task which kills Asterisk if it stops processing OPTIONS packets and does not process CLI commands anymore.

I saw there is a patch which improves the handling of temp transport states (Commit 7226577) in Asterisk 20.19 but I'm not sure if our issue has anything to do with it.

I attached a redacted coredumper output of both coredumps:

[core-asterisk-2026-04-02T10-10-30Z-brief.txt](https://github.com/user-attachments/files/26532467/core-asterisk-2026-04-02T10-10-30Z-brief.txt)
[core-asterisk-2026-04-02T10-10-30Z-full.txt](https://github.com/user-attachments/files/26532468/core-asterisk-2026-04-02T10-10-30Z-full.txt)
[...]

PR #1856, opened 2026-04-07
res_pjsip_config_wizard: Trigger reloads from a pjsip servant thread

gtjoseph:
When res_pjsip is reloaded directly, it does the sorcery reload in a pjsip
servant thread as it's supposed to.  res_pjsip_config_wizard however
was not which was leading to occasional deadlocks.  It now does the reload
in a servant thread just like res_pjsip.

Resolves: #1855


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

PR #1857, opened 2026-04-08
func_talkdetect.c: Clarify dsp_talking_threshold documentation.

### LINKED ISSUE #1761: func_talkdetect.c: TALK_DETECT docs wording mistake
[>TALK_DETECT docs ](https://docs.asterisk.org/Asterisk_22_Documentation/API_Documentation/Dialplan_Functions/TALK_DETECT/#description)

> dsp_talking_threshold is the time in milliseconds of sound above what the dsp has established as base line silence for a user before a user is considered to be talking. By default, the value of silencethreshold from dsp.conf is used. If this value is set too tight events may be falsely triggered by variants in room noise.

Here, the dsp_talking_threshold is described as a "time in milliseconds", but below on the same page it is described as "The minimum average magnitude per sample in a frame for the DSP to consider talking/noise present".

So to me, this means the parameter is not a "time in milliseconds" value, but rather some unit of energy instead.

I think the wording needs to be fixed, or at least made less confusing.

seanbright:
Fixes: #1761

### DIFF
--- a/funcs/func_talkdetect.c
+++ b/funcs/func_talkdetect.c
@@ -86,13 +86,14 @@
 			<para>The function has two parameters that can optionally be passed
 			when <literal>set</literal> on a channel: <replaceable>dsp_talking_threshold</replaceable>
 			and <replaceable>dsp_silence_threshold</replaceable>.</para>
-			<para><replaceable>dsp_talking_threshold</replaceable> is the time in milliseconds of sound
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph (approved):

InterLinked1 (approved):

PR #1860, opened 2026-04-08
res_pjsip_outbound_registration: only update the Expires header if the value has changed

### LINKED ISSUE #1859: [bug]: res_pjsip_outbound_registration: No expires header set when triggered via CLI
### Severity

Trivial

### Versions

20, 22, 23

### Components/Modules

res_pjsip_outbound_registration

### Operating Environment

all

### Frequency of Occurrence

Frequent

### Issue Description

If you trigger a registration with pjsip send register <name> from the CLI, we send a REGISTER with an Expires: 0  to remove the current binding but then Asterisk sends a REGISTER with no Expires header at all which causes the UAS to respond with whatever their default expiry time is.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

mbradeen:
The PJSIP outbound registration API has undocumented behavior when reconfiguring
the outbound registration if the expires value being set is the same as what was
previously set.

In this case PJSIP will remove the Expires header entirely from subsequent
outbound REGISTER requests. To eliminate this as an issue we now check the current
expires value against the configured expires value and only apply it if it differs.

This ensures that outbound REGISTER requests always contain an Expires header.

Resolves: #1859

### DIFF
--- a/res/res_pjsip_outbound_registration.c
+++ b/res/res_pjsip_outbound_registration.c
@@ -2105,6 +2105,7 @@ static int sip_outbound_registration_perform(void *data)
 	struct sip_outbound_registration *registration = ao2_bump(state->registration);
 	size_t i;
 	int max_delay;
+	pjsip_regc_info info;
 
 	/* Just in case the client state is being reused for this registration, free the auth information */
 	ast_sip_auth_vector_destroy(&state->client_state->outbound_auths);
@@ -2132,7 +2133,14 @@ static int sip_outbound_registration_perform(void *data)
 	state->client_state->auth_rejection_permanent = registration->auth_rejection_permanent;
 	max_delay = registration->max_random_initial_delay;
 
-	pjsip_regc_update_expires(state->client_state->client, registration->expiration);
+	/*
+	 * pjsip_regc_update_expires will remove the Expires header from the REGISTER request if the
+	 * expiration interval is re-set to the same value as the current interval.  We want to avoid
+	 * this so we only call it if the interval has changed.
+	 */
+	if (pjsip_regc_get_info(state->client_state->client, &info) == PJ_SUCCESS && info.interval != (unsigned) registration->expiration) {
+		pjsip_regc_update_expires(state->client_state->client, registration->expiration);
+	}
 
 	/* n mod 0 is undefined, so don't let that happen */
 	schedule_registration(state->client_state, (max_delay ? ast_random() % max_delay : 0) + 1);


mbradeen:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
I don't really understand the commit message and its meaning or the scenario.

mbradeen:
> I don't really understand the commit message and its meaning or the scenario.

sip_outbound_registration_perform which calls pjsip_regc_get_info is only called when the config is loaded, applied or via queue_register which is done via AMI or CLI command.  In the first two scenarios it does not cause an issue because the regc structure is 'fresh'.  When done via queue_register by AMI or CLI, the regc structure already exists so we are calling pjsip_regc_update_expires on an existing struct with the same value, leading to the header being deleted.

jcolp:
I would say:

"The PJSIP outbound registration API has undocumented behavior where when reconfiguring the outbound registration if the expires value we are setting is the same as what was previously set it will remove the Expires header entirely from subsequent outbound REGISTER requests. To eliminate this as an issue we check the current expires value against the configured expires value and only change it if it differs ensuring that outbound REGISTER requests always contain an Expires."

res/res_pjsip_outbound_registration.c
```diff
@@ -2132,7 +2133,14 @@ static int sip_outbound_registration_perform(void *data)
 	state->client_state->auth_rejection_permanent = registration->auth_rejection_permanent;
 	max_delay = registration->max_random_initial_delay;
 
-	pjsip_regc_update_expires(state->client_state->client, registration->expiration);
+	/*
+	 * pjsip_regc_update_expires will remove the expires header from the REGISTER request if the
+	 * expiration interval is re-set to the same value as the current interval.  We want to avoid
+	 * this so we only call it if the interval has changed.
+	 */
+	if (pjsip_regc_get_info(state->client_state->client, &info) == PJ_SUCCESS && info.interval != (unsigned) registration->expiration) {
```
jcolp, line 2141: What failure cases are there for pjsip_regc_get_info? Is there a scenario where it would fail and we'd want to update the expires anyway?

res/res_pjsip_outbound_registration.c
```diff
@@ -2132,7 +2133,14 @@ static int sip_outbound_registration_perform(void *data)
 	state->client_state->auth_rejection_permanent = registration->auth_rejection_permanent;
 	max_delay = registration->max_random_initial_delay;
 
-	pjsip_regc_update_expires(state->client_state->client, registration->expiration);
+	/*
+	 * pjsip_regc_update_expires will remove the expires header from the REGISTER request if the
```
jcolp: Expires header

jcolp (changes requested):

res/res_pjsip_outbound_registration.c
mbradeen: In theory pjsip_regc_get_info should return PJ_SUCCESS unless the info or regc is invalid.  If the Info is bad, we have bigger problems.  If the regc is bad, pjsip_regc_update_expires will fail anyway.

jcolp (approved):

PR #1869, opened 2026-04-15
ARI: Added paths to get and set multiple channel variables.

bkford:
Two new paths exist for ARI to get and set multiple channel variables at
the same time. This is done via GET and POST like the single get and set
variable equivalents. Leading and trailing whitespace will be stripped
from the variable names for both paths. When setting variables, the
values will be read as-is, whitespace included. GET takes in a single
string with comma-separated values, while POST takes in a dictionary of
key value pairs. The code follows the same paths as when setting
multiple variables when originating a channel via ARI.

UserNote: Added new ARI paths for getting and setting multiple channel
variables at a time. For GET, this takes in a single string of
comma-separated variable names, while POST takes in a dictionary of key
value pairs. The behavior is the same as passing in variables when
originating a channel.

### DIFF
--- a/res/ari/resource_channels.c
+++ b/res/ari/resource_channels.c
@@ -1559,6 +1559,98 @@ void ast_ari_channels_get_channel_var(struct ast_variable *headers,
 	ast_ari_response_ok(response, ast_json_ref(json));
 }
 
+void ast_ari_channels_get_channel_vars(struct ast_variable *headers,
+	struct ast_ari_channels_get_channel_vars_args *args,
+	struct ast_ari_response *response)
+{
+	int res;
+	RAII_VAR(struct ast_json *, json, ast_json_object_create(), ast_json_unref);
+	RAII_VAR(struct ast_json *, inner_json, ast_json_object_create(), ast_json_unref);
+	RAII_VAR(struct ast_str *, value, ast_str_create(32), ast_free);
+	RAII_VAR(struct ast_channel *, channel, NULL, ast_channel_cleanup);
+
+	ast_assert(response != NULL);
+
+	if (!json || !inner_json || !value) {
+		ast_ari_response_alloc_failed(response);
+		return;
+	}
+
+	if (args->variables_count == 0) {
+		ast_ari_response_error(
+			response, 400, "Bad Request",
+			"At least one variable name is required");
+		return;
+	}
+
+	if (ast_strlen_zero(args->channel_id)) {
+		ast_ari_response_error(
+			response, 400, "Bad Request",
+			"Channel ID is required");
+		return;
+	}
+
+	channel = ast_channel_get_by_name(args->channel_id);
+	if (!channel) {
+		ast_ari_response_error(
+			response, 404, "Channel Not Found",
+			"Provided channel was not found");
+		return;
+	}
+
+	for (int i = 0; i < args->variables_count; i++) {
+		struct ast_json *json_str;
+		char buf[strlen(args->variables[i]) + 1];
+		char *variable;
+
+		strcpy(buf, args->variables[i]);
+		variable = ast_strip(buf);
+		if (ast_strlen_zero(variable)) {
+			ast_ari_response_error(
+				response, 400, "Bad Request",
+				"Variable names are required");
+			return;
+		}
+
+		if (variable[strlen(variable) - 1] == ')') {
+			if (ast_func_read2(channel, variable, &value, 0)) {
+				ast_ari_response_error(
+					response, 500, "Error With Function",
+					"Unable to read provided function");
+				return;
+			}
+		} else {
+			if (!ast_str_retrieve_variable(&value, 0, channel, NULL, variable)) {
+				ast_ari_response_error(
+					response, 404, "Variable Not Found",
+					"Provided variable was not found");
+				return;
+			}
+		}
+
+		json_str = ast_json_string_create(ast_str_buffer(value));
+		if (!json_str) {
+			ast_ari_response_alloc_failed(response);
+			return;
+		}
+
+		res = ast_json_object_set(inner_json, variable, json_str);
+		if (res) {
+			ast_ari_response_alloc_failed(response);
+			ast_json_unref(json_str);
+			return;
+		}
+	}
+
+	res = ast_json_object_set(json, "variables", ast_json_ref(inner_json));
+	if (res) {
+		ast_ari_response_alloc_failed(response);
+		return;
+	}
+
+	ast_ari_response_ok(response, ast_json_ref(json));
+}
+
 void ast_ari_channels_set_channel_var(struct ast_variable *headers,
 	struct ast_ari_channels_set_channel_var_args *args,
 	struct ast_ari_response *response)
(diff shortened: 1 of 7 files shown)


bkford:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

rest-api/api-docs/channels.json
```diff
@@ -1620,6 +1620,103 @@
...
+							"dataType": "string"
+						},
+						{
+							"name": "values",
+							"description": "The values to set the variables to",
+							"paramType": "query",
+							"required": false,
+							"allowMultiple": true,
+							"dataType": "string"
+						}
```
gtjoseph, line 1692: This should look like the `variables` entries in channels create and originate.

gtjoseph (changes requested):

gtjoseph:
In the commit message, I don't understand what the following means...

> You can also use getChannelVars and setChannelVars in your applications.

bkford:
**NOTE**
One concern I have is the locking, so extra eyes on that would be greatly appreciated. The main concerns are:
1. When setting multiple variables, ast_set_variables is called and locks the channel. It eventually gets to pbx_builtin_setvar_helper, which ALSO locks the channel. Originating a channel also does this.
2. When setting a single variable (the original, existing path), it calls a different function: stasis_app_control_set_channel_var. This does NOT lock the channel before calling, but it ALSO eventually gets to pbx_builtin_setvar_helper.
3. In the original, existing get variable function, it makes a note of not locking when reading from the channel due to potential locking issues, especially with autoservice dialplan functions. It may not have anything to do with a write, but it's a concern I had.

rest-api/api-docs/channels.json
```diff
@@ -1620,6 +1620,95 @@
...
+							"name": "channelId",
+							"description": "Channel's id",
+							"paramType": "path",
+							"required": true,
+							"allowMultiple": false,
+							"dataType": "string"
+						},
+						{
+							"name": "variables",
+							"description": "The \"variables\" key in the body object holds variable key/value pairs to set on the channel. Other keys in the body object are interpreted as query parameters. Ex. { \"variables\": { \"CALLERID(name)\": \"Alice\" } }",
```
gtjoseph: >  Other keys in the body object are interpreted as query parameters

There aren't any other keys in the body for this endpoint so this doesn't apply.

res/ari/resource_channels.c
```diff
@@ -1590,6 +1672,84 @@ void ast_ari_channels_set_channel_var(struct ast_variable *headers,
...
+	if (!variables) {
+		ast_ari_response_error(
+			response, 409, "Conflict",
+			"Could not create channel variables from JSON");
+		return;
+	}
+
+	ast_channel_lock(channel);
+	ast_set_variables(channel, variables);
+	ast_channel_unlock(channel);
```
gtjoseph: I don't think you want to be calling ast_set_variables() directly here because of the locking issues you mentioned.  The reason stasis_app_control_set_channel_var() is used when setting a single variable is because the actual locking and setting is done in another thread.  You should use stasis_app_control_set_channel_var() in the loop above for that same reason.  This also allows you to get rid of the intermediate ast_variable list.

gtjoseph (changes requested):

res/ari/resource_channels.c
```diff
@@ -1559,6 +1559,88 @@ void ast_ari_channels_get_channel_var(struct ast_variable *headers,
...
+		ast_ari_response_error(
+			response, 404, "Channel Not Found",
+			"Provided channel was not found");
+		return;
+	}
+
+	for (int i = 0; i < args->variables_count; i++) {
+		int res;
+		struct ast_json *json_str;
+		char *variable = ast_strip(ast_strdupa(args->variables[i]));
```
gtjoseph: ast_strdupa() can be problematic in a loop because the memory doesn't get released until the function exits.  While it's probably not an issue in this particular case you could argue that someone could exhaust the stack by requesting a lot of variables with very long names.

This would be better:
```
		char buf[strlen(args->variables[i]) + 1];
		char *variable = ast_strip(buf);
```

res/ari/resource_channels.c
```diff
@@ -1590,6 +1672,107 @@ void ast_ari_channels_set_channel_var(struct ast_variable *headers,
...
+	if (control == NULL) {
+		/* response filled in by find_control */
+		return;
+	}
+
+	json_variables = ast_json_object_get(args->variables, "variables");
+	for (it_json_var = ast_json_object_iter(json_variables); it_json_var;
+		it_json_var = ast_json_object_iter_next(json_variables, it_json_var)) {
+		const char *key = ast_json_object_iter_key(it_json_var);
+		char *stripped_key = ast_strip(ast_strdupa(key));
```
gtjoseph: Same issue here with ast_strdupa().

res/ari/resource_channels.c
```diff
@@ -1590,6 +1672,107 @@ void ast_ari_channels_set_channel_var(struct ast_variable *headers,
...
+
+		json_value = ast_json_object_iter_value(it_json_var);
+		if (ast_json_typeof(json_value) != AST_JSON_STRING) {
+			ast_ari_response_error(
+				response, 400, "Bad Request",
+				"Variable values must be strings");
+			return;
+		}
+
+		res = ast_json_object_set(stripped_json, stripped_key, json_value);
```
gtjoseph: JSON is pretty expensive to use as a temporary container.  I'd use an ast_variables list then iterate over that below. Unfortunately, ast_json_to_ast_variables() doesn't strip the keys or return an error if the key is empty so you still need the loop.

gtjoseph (dismissed):

gtjoseph (dismissed):
LGTM pending the addition of the testsuite test.
When that's ready, don't forget to update this PR with a `testsuite-test-pr: nnnn` comment so the two PRs can be tested together.

rest-api/api-docs/channels.json
```diff
@@ -1620,6 +1620,95 @@
...
 			]
 		},
+		{
+			"path": "/channels/{channelId}/variables",
+			"description": "Multiple variables on a channel",
+			"operations": [
+				{
+					"httpMethod": "GET",
+					"since": [
+						"20.20.0"
```
mbradeen: since should show 22.10.0 and 23.4.0 as well

rest-api/api-docs/channels.json
```diff
@@ -1620,6 +1620,95 @@
...
+						{
+							"code": 409,
+							"reason": "Channel not in a Stasis application"
+						}
+					]
+				},
+				{
+					"httpMethod": "POST",
+					"since": [
+						"20.20.0"
```
mbradeen: since should show 22.10.0 and 23.4.0 as well

mbradeen (dismissed):

bkford:
testsuite-test-pr: 131

mbradeen (approved):

gtjoseph (dismissed):
Need failing get_set testsuite test to pass.

gtjoseph (approved):

gtjoseph:
Moving to DRAFT because @mbradeen has a related PR coming and we want to make sure they're in sync before merging.

jcolp (approved):

bkford:
Included in #1917

PR #1871, opened 2026-04-16
extension_state: Add new extension state API.

jcolp:
Extension state to this point has been an API implemented inside the PBX core resulting in its state being intermingled with that of the dialplan. This increased the complexity of the PBX core and made it difficult to enact improvements.

This change adds a new separate extension state API which receives updates from the PBX core as configuration changes but maintains its own separate state. The API is also written to fully take advantage of modern APIs in a more selective manner by subscribing each extension state to only the devices it is interested in, ultimately reducing resource consumption during updates. Presence state updates being infrequently done use a single shared subscription that goes through the extension states to find and update ones that the update is applicable to.

Legacy API support is provided by reimplementing the API as wrappers over the new extension state API. This improves the legacy API by making it multithreaded, with each callback being individually subscribed.

Autohints support is maintained but has been separated out into a self contained new implementation.

Synchronous subscription support has also been added to Stasis to remove the overhead of asynchronous publishing when the handling of published messages is small and fast.

### DIFF
--- a/include/asterisk/_private.h
+++ b/include/asterisk/_private.h
@@ -59,6 +59,9 @@ void ast_msg_shutdown(void);        /*!< Provided by message.c */
 int aco_init(void);             /*!< Provided by config_options.c */
 int dns_core_init(void);        /*!< Provided by dns_core.c */
 int ast_refer_init(void);             /*!< Provided by refer.c */
+int ast_extension_state_init(void);     /*!< Provided by extension_state.c */
+int ast_extension_state_legacy_init(void); /*!< Provided by extension_state_legacy.c */
+int ast_extension_state_autohints_init(void); /*!< Provided by extension_state_autohints.c */
 
 /*!
  * \brief Initialize malloc debug phase 1.
--- a/include/asterisk/extension_state.h
+++ b/include/asterisk/extension_state.h
@@ -1,0 +1,210 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2026, Sangoma Technologies Corporation
+ *
+ * Joshua C. Colp <jcolp@sangoma.com>
+ *
+ * See http://www.asterisk.org for more information about
+ * the Asterisk project. Please do not directly contact
+ * any of the maintainers of this project for assistance;
+ * the project provides a web site, mailing lists and IRC
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
+ * at the top of the source tree.
+ */
+
+/*! \file
+ * \ref ExtensionState
+ *
+ * \page ExtensionState API providing extension state management.
+
+Before we talk about extension state let's talk about the fundamentals that
+drive it. Extension state is driven based on three things: hints, device state,
+and presence state. Hints are the stateless configuration that map a dialplan location,
+context and extension, to zero or more devices and/or zero or more presence state
+providers. Device state provides information about the associated device(s).
+Presence state provides information about the associated presence state provider(s).
+
+The extension state API itself acts as an aggregator of the device state and presence
+state information using the hint configuration to determine both the individual
+identifier as well as the aggregation sources. The API provides the ability to query
+the current state of an extension, as well as subscribe to be notified when the state
+of an extension changes.
+
+When hint configuration changes this is reconciled and the extension state is updated
+accordingly. If a hint has been added the corresponding extension state is created. If
+a hint has been removed the corresponding extension state is removed. For cases where
+the hint has been added or updated the sources of information are updated on the
+extension state and its state is recalculated. This may involve subscribing to new
+device state topics or unsubscribing from old ones.
+
+Extension state uses synchronous per-device subscriptions to receive device state
+updates. Synchronous is used as the overhead of asynchronous delivery is not worth
+the added overhead and CPU for the small amount of work done when device states change.
+On receipt of a device state update the extension device state is recalculated and if
+the state has changed the extension device state is updated and any subscribers are
+notified.
+
+Presence state uses a single global subscription to receive presence state updates as
+no per-presence state provider topic is available and also due to the extremely small
+number of presence state updates that occur on a system. Just like device state
+updates the extension presence state is recalculated and if it has changed it is
+updated and any subscribers are notified.
+
+To minimize querying of other APIs in Asterisk extension state keeps an internal
+cache of device states and presence state on each extension state. This cache is
+updated when device state or presence state changes and is used to determine the
+aggregated state of an extension. The aggregated state is also cached on the
+extension state for quick access by API users who do not subscribe to receive
+updates.
+
+*/
+
+#ifndef _ASTERISK_EXTENSION_STATE_H
+#define _ASTERISK_EXTENSION_STATE_H
+
+#include "asterisk/pbx.h"
+
+/*! \brief Individual device states that contributed to snapshot */
+struct ast_extension_state_device_state_info {
+	/*! \brief The state of the device */
+	enum ast_device_state state;
+	/*! \brief The name of the device */
+	char device[0];
+};
+
+/*! \brief Device snapshot for an extension state*/
+struct ast_extension_state_device_snapshot {
+	/*! \brief The state of the extension */
+	enum ast_extension_states state;
+	/*! \brief The device that caused this update */
+	struct ast_extension_state_device_state_info *causing_device;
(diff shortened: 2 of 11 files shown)


jcolp:
This is in draft because it's not yet ready for inclusion. Tests will break due to ordering and other issues, but I wanted to get this up so people could start giving it a try and providing input if they want.

include/asterisk/_private.h
```diff
@@ -59,6 +59,9 @@ void ast_msg_shutdown(void);        /*!< Provided by message.c */
 int aco_init(void);             /*!< Provided by config_options.c */
 int dns_core_init(void);        /*!< Provided by dns_core.c */
 int ast_refer_init(void);             /*!< Provided by refer.c */
+int ast_extension_state_init(void);     /*!< Provided by extensionstate.c */
```
phoneben: extensionstate.c should be `extension_state.c.`

main/extension_state.c
```diff
@@ -0,0 +1,1500 @@
...
+		return NULL;
+	}
+
+	/*
+	 * We do a synchronous subscription to the device state topic, as our callback is extremely
+	 * short lived and the added overhead of queueing to a taskprocessor for another thread to handle
+	 * it is just not worth it.
+	 */
+	source->device_state_subscription = stasis_subscribe_synchronous(topic, extension_state_device_state_cb, state);
+	if (!source->device_state_subscription) {
```
phoneben, line 487: If `extension_state_device_state_info_alloc()` succeeds but` stasis_subscribe_synchronous()` fails, the code frees `source `and returns without releasing` source->info`. Since` source->info` is AO2-managed, that leaks the object.
add 
`  ao2_ref(source->info, -1);`

main/extension_state.c
```diff
@@ -0,0 +1,1500 @@
...
+		ao2_ref(state, -1);
+		return NULL;
+	}
+
+	if (ast_asprintf(&extension_state_topic_name, "extension_state:extension/%s", extension) < 0) {
+		ao2_ref(state, -1);
+		return NULL;
+	}
+
+	state->extension_state_topic = stasis_topic_create(extension_state_topic_name);
```
phoneben, line 979: String buffer never freed.

string is built with `ast_asprintf() `to name a topic, then handed to` stasis_topic_create()`. After that, the string just sits in memory forever — whether the topic creation succeeded or failed. The fix is simply to call `ast_free(extension_state_topic_name)`

jcolp:
Update: This has now been completed.

I originally broke up device and presence updates separately, to allow more specific filtering for subscribers. In practice, though, it's kind of a 50/50 where a subscriber wants the update to contain both device and presence or doesn't care. Since presence updates are infrequent I'll probably go back to a single update message that contains both, as well as the old and new versions of each.

include/asterisk/extension_state.h
```diff
@@ -0,0 +1,210 @@
...
+
+/*! \brief Stasis message for extension state update message */
+struct ast_extension_state_update_message {
+	/*! \brief The old device snapshot */
+	struct ast_extension_state_device_snapshot *old_device_snapshot;
+	/*! \brief The new device snapshot - will be pointer equivalent to old if unchanged */
+	struct ast_extension_state_device_snapshot *new_device_snapshot;
+	/*! \brief The old presence snapshot */
+	struct ast_extension_state_presence_snapshot *old_presence_snapshot;
+	/*! \brief The new presence snapshot - will be pointed equivalent to old if unchanged */
```
mbradeen: pointer

main/extension_state.c
```diff
@@ -0,0 +1,1598 @@
...
+	}
+
+	/* Initialize extension state subsystem */
+	extension_states = ao2_container_alloc_hash(AO2_ALLOC_OPT_LOCK_MUTEX, 0,
+		EXTENSION_STATE_BUCKETS, extension_state_hash_fn, NULL,
+		extension_state_cmp_fn);
+	if (!extension_states) {
+		ast_log(LOG_ERROR, "Failed to allocate new states container\n");
+		return -1;
+	}
```
gtjoseph: In the current implementation, although hints aren't sorted when you do a `core show hints` they always seem to be somewhat grouped.  In the new implementation, they're all over the place which makes it hard to navigate.  If this is changed to an rbtree, "core show hints" will be naturally sorted.  Various other CLI commands use an intermediate rbtree container to sort when the CLI command runs but most of those are displaying configuration data which is fairly static so locking the base hash container while running ao2_container_dup() isn't a big deal.  They also tend to be displaying data from sorcery which is hash container based so there's no opportunity to use an rbtree as the base.

main/extension_state_autohints.c
```diff
@@ -0,0 +1,183 @@
...
+#include "asterisk/pbx.h"
+#include "asterisk/stasis.h"
+#include "asterisk/vector.h"
+#include "pbx_private.h"
+
+/*! \brief Lock to protect the autohints vector */
+AST_MUTEX_DEFINE_STATIC(extension_state_autohints_lock);
+
+/*! \brief Contexts which have autohints enabled */
+static AST_VECTOR(, struct ast_context *) extension_state_autohints;
```
gtjoseph, line 34: Every DPMA-managed phone will have an autohint.  Will a vector perform well enough with a thousand or more phones?

gtjoseph (dismissed):
Some initial stuff...

main/extension_state_autohints.c
jcolp: During normal operation it's iterating through the vector, and the vector only stores contexts, so it should be fine. The autohints code isn't storing the extension state or hints itself, just what contexts have autohints enabled so things can be created there as needed.

gtjoseph: Ah yeah, I didn't notice that it was a vector of contexts.

jcolp:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
Once more review is done and we're close to merging I'll look at additional PRs for elsewhere

asteriskteam:

gtjoseph (approved):

jcolp:
This has been rebased onto latest master and I've tweaked structure positioning such that it cherry picks cleanly to 20 now as well.

main/extension_state_legacy.c
```diff
@@ -0,0 +1,472 @@
...
+				return 0;
+			}
+		}
+		ast_mutex_unlock(&extension_state_legacy_callbacks_lock);
+
+		return -1;
+	}
+
+	ast_mutex_lock(&extension_state_legacy_callbacks_lock);
+	if (id > 0 && id <= AST_VECTOR_SIZE(&extension_state_legacy_callbacks)) {
```
mbradeen: if id <0, can we handle that at the beginning of the func w/o having to lock/unlock, etc?

main/extension_state.c
```diff
@@ -0,0 +1,1597 @@
...
+	device_snapshot = ao2_alloc_options(sizeof(*device_snapshot),
+		extension_state_device_snapshot_destroy, AO2_ALLOC_OPT_LOCK_NOLOCK);
+	if (!device_snapshot) {
+		return NULL;
+	}
+
+	device_snapshot->state = device_state;
+	device_snapshot->causing_device = ao2_bump(causing_device);
+	if (AST_VECTOR_INIT(&device_snapshot->additional_devices, AST_VECTOR_SIZE(device_state_sources))) {
+		ao2_ref(device_snapshot, -1);
```
mbradeen, line 306: Do we need to decrement the count on the causing_device here or is that handled in the device_snapshot cleanup?

main/extension_state.c
```diff
@@ -0,0 +1,1597 @@
...
+		ast_channel_lock(chan);
+		/* this channel's state doesn't match */
+		if (search_state != ast_channel_state(chan)) {
+			ast_channel_unlock(chan);
+			continue;
+		}
+		/* any non-ringing channel will fit */
+		if (search_state != AST_STATE_RINGING) {
+			ast_channel_unlock(chan);
+			channel = chan;
```
mbradeen, line 1115: why don't we need to ref the channel here as we do below?

mbradeen (commented):
Just a couple of questions.

main/extension_state.c
jcolp: It's handled in the device_snapshot destructor.

main/extension_state.c
jcolp: By breaking out the ast_channel_unref of the for loop won't execute so the reference remains valid. In the logic below the loop keeps going, so an unref would happen thus to keep it valid an explicit ref has to be done.

mbradeen (approved):

PR #1876, opened 2026-04-17
res_stasis/resource_bridges: Split bridge playback control and wrapper cleanup

### LINKED ISSUE #1861: [bug]: Possible heap corruption in audiohook/translate write path during bridged media
### Severity

Major

### Versions

20.18.2

### Components/Modules

`main/audiohook.c`, `main/channel.c`, `main/translate.c`, `main/frame.c`

### Operating Environment

- Asterisk 20.18.2
- Release tag/commit analyzed locally: `d49c5d454f42fc11da7631bea496c66c6473798e` (`Update for 20.18.2`, dated February 5, 2026)
- Ubuntu 24 on Azure VM
- Source build
- Debug symbols present in the binary
- We use ARI/Stasis heavily and also stream media externally via AudioSocket
- No traditional dialplan usage except routing all calls into the Stasis app
- PJSIP endpoints/trunks primarily use alaw
- The system is high-concurrency, with many worker threads active at any given time (as well during the crash)

### Frequency of Occurrence

One Time

### Issue Description

We hit a production crash with a glibc abort:

```text
malloc(): unsorted double linked list corrupted
```

Which looks like a delayed detection of heap corruption rather than a direct allocator bug. 

The backtrace points to the bridged media write path through audiohook processing, frame duplication, and translation:

```text
#6  malloc_printerr("malloc(): unsorted double linked list corrupted")
#7  _int_malloc
#8  __libc_calloc
#9  __ast_repl_calloc
#10 __ast_calloc
#11 __ast_frdup
#12 ast_translate
#13 audiohook_list_translate_to_slin
#14 audio_audiohook_write_list
#15 ast_audiohook_write_list
#16 ast_write_stream
#17 bridge_channel_handle_write
#18 bridge_channel_wait
#19 bridge_channel_internal_join
#20 bridge_channel_depart_thread
```

The crashing thread is a bridge worker in the bridged media write path.

### What we found from the core

- Crashing channel was a PJSIP channel running Stasis
- Two active audiohooks were attached, both with source Snoop
- One Snoop hook was read-direction, the other write-direction
- No whisper/manipulate hooks were active
- No channel monitor was active
- The immediate RTP alaw frame and the translated alawtolin SLIN frame both looked structurally valid in GDB
- Heap corruption was detected while duplicating the translated frame in `__ast_frdup()`

So the current working theory is that memory was corrupted earlier, but the failure surfaced in the bridged write/audiohook/translation path with active ARI snoop audiohooks.

The allocator wrapper arguments at the abort site were already corrupted which reinforces that this is prior heap corruption rather than a simple bad allocation request:

- `__libc_calloc` elem_size = 101507530449710
- `__ast_repl_calloc` size = 101507530449710
- but `__ast_calloc` size = 530, which feels like the correct value
- file = " Redirects:\n" is clearly not a C source filename
- func = "handle_parking_bridge_enter_message" which is not in line with the actual call path

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

pekkaar:
Modified the bridge playback teardown so the worker thread removes only the
playback control, while the after-bridge callback removes the playback
wrapper once the announcer has actually left the bridge.

This avoids a stale window where a new playback request could create a
replacement announcer before the old announcer had fully exited the holding
bridge.

Also replaced the flexible trailing bridge_id storage in the shared worker
thread data with an optional bridge_id pointer, since recording paths use the
same structure without a bridge id.

Fixes: #1861

### DIFF
--- a/include/asterisk/stasis_app.h
+++ b/include/asterisk/stasis_app.h
@@ -796,6 +796,15 @@ int stasis_app_bridge_playback_channel_add(struct ast_bridge *bridge,
 	struct stasis_app_control *control);
 
 /*!
+ * \brief Remove a bridge playback channel's control from the app controls list.
+ *
+ * \param bridge_id The unique ID of the bridge the playback channel is in.
+ * \param control The app control structure for the playback channel
+ */
+void stasis_app_bridge_playback_channel_control_remove(const char *bridge_id,
+	struct stasis_app_control *control);
+
+/*!
  * \brief remove channel from list of ARI playback channels for bridges.
  *
  * \param bridge_id The unique ID of the bridge the playback channel is in.
--- a/res/ari/resource_bridges.c
+++ b/res/ari/resource_bridges.c
@@ -282,7 +282,7 @@ struct bridge_channel_control_thread_data {
 	struct ast_channel *bridge_channel;
 	struct stasis_app_control *control;
 	struct stasis_forward *forward;
-	char bridge_id[0];
+	char *bridge_id;
 };
 
 static void *bridge_channel_control_thread(void *data)
@@ -292,7 +292,7 @@ static void *bridge_channel_control_thread(void *data)
 	struct stasis_app_control *control = thread_data->control;
 	struct stasis_forward *forward = thread_data->forward;
 	ast_callid callid = ast_channel_callid(bridge_channel);
-	char *bridge_id = ast_strdupa(thread_data->bridge_id);
+	char *bridge_id = thread_data->bridge_id;
 
 	if (callid) {
 		ast_callid_threadassoc_add(callid);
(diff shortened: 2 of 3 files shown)


pekkaar:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

pkrall-buzzeasy:
Reviewed both CPCheck artifacts. The 20-ari1 failure is a SIPp referee.xml abort in an attended-transfer test, while the direct bridge playback tests that more closely relate to my commit still pass. The 22-pjs5 failure is an RLS/MWI validation failure, which is in a different subsystem from this ARI/stasis bridge playback change.

gtjoseph:
Yeah, the transfer and mwi tests are timing sensitive and can have issues if GitHub's runners are slow.

PR #1890, opened 2026-04-22
res_audiosocket: Tolerate non-audio frame types

SvenKube:
This commit implements the handling of non-voice or DTMF frames like the chan_websocket handling added in #1588. Rather than treating unsupported frames as fatal errors, silently ignore CNG frames and log a warning for other unsupported types.


SvenKube:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

PR #1902, opened 2026-04-26
app_queue: Fix raise_respect_min lost in copy_rules() breaking rN queue rules

### LINKED ISSUE #1901: [bug]: QUEUE_RAISE_PENALTY=rN ignored when set via queue rules
### Severity

Trivial

### Versions

20.19.0

### Components/Modules

app queue rules 

### Operating Environment

centos 9

### Frequency of Occurrence

Constant

### Issue Description

When using `QUEUE_RAISE_PENALTY=rN` via a timed penaltychange rule, the `r` flag is silently dropped and members below `min_penalty` that should be excluded receive calls.

**Root cause:** `copy_rules()` never copies `raise_respect_min` into the per-call rule list, so the flag is always 0 when the rule fires.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

phoneben:
app_queue: Fix raise_respect_min not copied in copy_rules() causing rN rules to be ignored.

`copy_rules()` never copied `raise_respect_min` into the per-call rule list, so the flag was always 0 when a timed penaltychange rule fired, making `rN` behave like plain `N` and raising members below `min_penalty` that should have been excluded.

Also fixes `update_qe_rule()` not propagating the flag from `qe->pr` to `qe`, and dropping the `r` prefix when saving back to `QUEUE_RAISE_PENALTY`.

Resolves: #1901

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -6049,7 +6049,12 @@ static void update_qe_rule(struct queue_ent *qe)
 			raise_penalty = max_penalty;
 		}
 
-		snprintf(raise_penalty_str, sizeof(raise_penalty_str), "%d", raise_penalty);
+		qe->raise_respect_min = qe->pr->raise_respect_min;
+		if (qe->raise_respect_min) {
+			snprintf(raise_penalty_str, sizeof(raise_penalty_str), "r%d", raise_penalty);
+		} else {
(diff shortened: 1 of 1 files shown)


phoneben:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

PR #1908, opened 2026-05-04
manager: Eliminate unnecessary code, simplify sessions in stasis callbacks

jcolp:
Due to stasis filtering the stasis callback for AMI type messages is guaranteed to only receive messages that can be turned into AMI events, so remove the check done in the callback.

The sessions container usage for the stasis callbacks has also been simplified by having a reference on the message router subscription instead of having to acquire the sessions from the global object each time.

### DIFF
--- a/main/manager.c
+++ b/main/manager.c
@@ -568,32 +568,27 @@ struct ast_str *ast_manager_str_from_json_object(struct ast_json *blob, key_excl
 static void manager_default_msg_cb(void *data, struct stasis_subscription *sub,
 				    struct stasis_message *message)
 {
-	struct ao2_container *sessions;
+	struct ao2_container *sessions = data;
 	struct ast_manager_event_blob *ev;
 
-	if (!stasis_message_can_be_ami(message)) {
-		/* Not an AMI message; disregard */
-		return;
-	}
-
-	sessions = ao2_global_obj_ref(mgr_sessions);
+	/*
+	 * This callback only receives messages that can be turned into AMI events, so
+	 * no need to check that the message can be turned into an event before checking for listeners.
+	 */
 	if (!any_manager_listeners(sessions)) {
 		/* Nobody is listening */
-		ao2_cleanup(sessions);
 		return;
 	}
 
 	ev = stasis_message_to_ami(message);
 	if (!ev) {
 		/* Conversion failure */
-		ao2_cleanup(sessions);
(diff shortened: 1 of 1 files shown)


jcolp:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

mbradeen (approved):

main/manager.c
```diff
@@ -621,14 +614,20 @@ static void manager_generic_msg_cb(void *data, struct stasis_subscription *sub,
...
 		return;
 	}
 
 	manager_event_sessions(sessions, class_type, type,
 		"%s", ast_str_buffer(event_buffer));
 	ast_free(event_buffer);
-	ao2_cleanup(sessions);
+}
+
+static void manager_subscription_change_msg_cb(void *userdata, struct stasis_subscription *sub,
```
gtjoseph, line 631: I know it's sort of documented down in manager_subscriptions_init() but it'd be good to document here what `userdata` actually is, what triggers this function and why the cleanup on final message.

main/manager.c
```diff
@@ -9598,11 +9598,21 @@ static int manager_subscriptions_init(void)
...
+	 * sessions container so it has to exist.
+	 */
+	sessions = ao2_global_obj_ref(mgr_sessions);
+
 	stasis_message_router_set_formatters_default(stasis_router,
-		manager_default_msg_cb, NULL, STASIS_SUBSCRIPTION_FORMATTER_AMI);
+		manager_default_msg_cb, sessions, STASIS_SUBSCRIPTION_FORMATTER_AMI);
+
+	res |= stasis_message_router_add(stasis_router,
+		ast_manager_get_generic_type(), manager_generic_msg_cb, sessions);
```
gtjoseph, line 9626: I'd be explicit here about the fact that manager_subscription_change_msg_cb's purpose is to clean up sessions.  

There's no possibility that manager_subscription_change_msg_cb could get invoked with a final message and trigger cleanup before any other messages have finished processing right? :)

gtjoseph (changes requested):

main/manager.c
jcolp: Stasis guarantees it doesn't. This is a common pattern across the tree for subscriptions.

gtjoseph (approved):

PR #1909, opened 2026-05-05
rtp: Use timestamp difference for video playback rate.

gareth-palmer:
When playing an audio file that has an associated video file that stream will not finish at the correct time due to incorrect scheduling of ast_readvideo_callback.

Non-zero scheduling between frames now only occurs when the frame timestamp changes and when-next is now
calculated using the  difference in timestamps between frames.

Video frames were also sent out with an incorrectly calcuated timestamp because the flag AST_FRFLAG_HAS_TIMING_INFO was not set on frames read from a file.

UserNote: The on-disk format for H263 and H264 video streams has changed so any previous recordings are not be playable.

### DIFF
--- a/formats/format_h263.c
+++ b/formats/format_h263.c
@@ -45,89 +45,121 @@
  * theoretical limit is not much larger (32k = 15bits), we'll go for that
  * size to ensure we don't corrupt frames sent to us (unless they're
  * ridiculously large). */
-#define	BUF_SIZE	32768	/* Four real h.263 Frames */
-
-#define FRAME_ENDED 0x8000
+#define BUF_SIZE	32768	/* Four real h.263 Frames */
+#define FRAME_ENDED	0x8000
 
 struct h263_desc {
 	unsigned int lastts;
 };
 
-
-static int h263_open(struct ast_filestream *s)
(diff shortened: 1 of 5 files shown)


gareth-palmer:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

AryanM200:
Hi @gareth-palmer ,

I tested your code and it is working. The video speed issue is fixed but During testing, in random scenarios the video playback is not working.

My code changes are also working. You have made changes in 5 files, while I made changes in 1 file.
#1753 

Can you explain which calculation or logic you used for the video playback handling?

gareth-palmer:
The calculation is that `fr->ts` contains the frame timestamp in milliseconds. When the timestamp of the current frame differs from that of the next frame the difference is used to calculate the delay (`*whennext = ts - desc->lastts`). There is no delay for frames that have the same timestamp.

Because `*whennext` is used as a timeout value for `ast_sched_add` the delay is decremented by 1ms so that the callback runs on the 33rd millisecond rather that just after the 33rd millisecond, or whatever the frame rate is guessed to be.

PR #1913, opened 2026-05-05
third-party: add CVE-APPLICABILITY.md for bundled deps

jelmerdehen:
third-party: add CVE-APPLICABILITY.md for bundled deps

Per-CVE applicability decisions for each bundled dep listed in
third-party/versions.mak: pjproject, jansson, libjwt.

third-party/CVE-APPLICABILITY-FORMAT.md describes the shared format
(buckets, OpenVEX justification mapping, suffix markers, scoping rule,
row schema). Each per-dep file is per-dep content only and references
the format doc.

pjproject: 14 in-scope CVEs classified, 24 out-of-scope listed in an
appendix. Each row's bucket is backed by a build flag in
Makefile.rules, a #if guard in the affected file, a verified
Asterisk-source callgraph, or a patch filename in
third-party/pjproject/patches/.

jansson: zero published upstream GHSAs.

libjwt: two upstream GHSAs target the 3.x branch and do not affect
the pinned major.

Bucket vocabulary maps to OpenVEX justification codes
(vulnerable_code_not_present,
vulnerable_code_cannot_be_controlled_by_adversary,
vulnerable_code_not_in_execute_path, affected).

In-tree pattern follows what Erlang/OTP commits for vendored OpenSSL at
vex/otp-N.openvex.json.

The version pinned in third-party/versions.mak is referenced rather
than restated in each file, so the maintainer does not have to update
two places when bumping the pin. Classifications are valid for the
pin at the same commit as each file; on bump, in-scope rows whose
upstream fix is now in the new pin move to out-of-scope, and new
upstream CVEs since the previous pin need rows added.

### DIFF
--- a/third-party/CVE-APPLICABILITY-FORMAT.md
+++ b/third-party/CVE-APPLICABILITY-FORMAT.md
@@ -1,0 +1,49 @@
+# CVE applicability log format
+
+Per-bundled-dep `CVE-APPLICABILITY.md` files in `third-party/<dep>/` use the format described here. Each file is per-dep, but the bucket vocabulary, OpenVEX justification mapping, and column conventions are shared across all bundled deps.
+
+## When to add a row
+
+* Open one row when an upstream GHSA is published against a bundled dep listed in `third-party/versions.mak`.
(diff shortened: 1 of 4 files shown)


jcolp:
I've placed this into draft for now until I decide whether this is something the project will undertake.

PR #1916, opened 2026-05-06
app_dial: Properly handle callee hangup while sending digits.

### LINKED ISSUE #1915: [bug]: app_dial: Channel not handled properly if callee disconnects while caller is sending it digits prior to answer
### Severity

Minor

### Versions

Git

### Components/Modules

app_dial

### Operating Environment

N/A

### Frequency of Occurrence

None

### Issue Description

Dial is supposed to continue dialplan execution unless the call is answered.

Simple example where all works as intended:

```
; channel originate Local/s@prog-test application Wait 30
[prog-test]
exten => s,1,Dial(Local/ss@prog-test/n)
	same => n,NoOp(got here) ; <---- get here afterwards
	same => n,Hangup()
exten => ss,1,Progress()
	same => n,Wait(1)
	same => n,Hangup()

[sub-dial]
exten => s,1,Dial(Local/noanswer@sub-dial/n,1)
	same => n,Hangup()
exten => noanswer,1,Progress() ; progress to trigger the digits to start sending
	same => n,Wait(0.5)
	same => n,Hangup()
```

However, if Dial execution aborts while sending digits due to progress, we currently return -1 from Dial (incorrectly) causing the call to hang up:

```
[prog-test]
exten => s,1,Dial(Local/ss@prog-test/n,,D(:::*1234567890#)) ; <--- only changed this line
	same => n,NoOp(got here) ; <------- we never get here
	same => n,Hangup()
exten => ss,1,Progress()
	same => n,Wait(1)
	same => n,Hangup()

[sub-dial]
exten => s,1,Dial(Local/noanswer@sub-dial/n,1)
	same => n,Hangup()
exten => noanswer,1,Progress() ; progress to trigger the digits to start sending
	same => n,Wait(0.5)
	same => n,Hangup()
```

This is in the case that the callee does not answer prior to hanging up. If the callee *does* answer, the behavior is also wrong, as app_dial never answers the calling channel in response but does correctly disconnect (although that was only by accident).

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

InterLinked1:
If we are sending digits (either DTMF, MF, or SF) to the called channel
after receiving progress or a wink, and the callee hangs up before we
have finished sending it digits, there are several problems that can ensue:

* If the callee hung up without answering, the calling channel would
  hang up and not continue in the dialplan.
* If the callee *did* answer before hanging up, the answer was never
  passed through to the caller, since this gets "eaten" by the various
  digit streaming functions and is never processed by app_dial.

This is generally an edge case that occurs due to some kind of signaling
failure, but to better handle this:

* Set to_answer to 0 to prevent hangup on the exit path, just like other
  parts of wait_for_answer.
* Better document this usage of to_answer.
* If the channel did answer while it was receiving digits, manually
  answer the calling channel before we abort. The call would not continue
  in the dialplan anyways (either before or after this fix), but technically
  the call was answered, so the CDRs should probably reflect that, and this
  mirrors the behavior of calls which normally do not continue.

Resolves: #1915

UserNote: If a called channel sends progress or wink and the caller begins
sending digits but the callee answers and then hangs up before digit
sending can finish, the call is now answered before being disconnected.
If the callee hangs up without answering, the call now continues in
the dialplan.

### DIFF
--- a/apps/app_dial.c
+++ b/apps/app_dial.c
@@ -1291,7 +1291,7 @@ static struct ast_channel *wait_for_answer(struct ast_channel *in,
 			} else {
 				ast_verb(3, "No one is available to answer at this time (%d:%d/%d/%d)\n", numlines, num.busy, num.congestion, num.nochan);
 			}
-			*to_answer = 0;
+			*to_answer = 0; /* Continue in the dialplan, since nobody answered */
 			if (is_cc_recall) {
 				ast_cc_failed(cc_recall_core_id, "Everyone is busy/congested for the recall. How sad");
 			}
@@ -1603,6 +1603,17 @@ static struct ast_channel *wait_for_answer(struct ast_channel *in,
 						}
 						if (res) {
 							ast_log(LOG_WARNING, "Called channel %s hung up post-progress before all digits could be sent\n", ast_channel_name(c));
+							if (ast_channel_state(c) == AST_STATE_UP) {
+								/* The called channel answered while we were sending it digits, so the answer never got processed by app_dial.
+								 * The channel is dying now, but better to answer late than never? */
+								ast_debug(1, "Channel %s answered while we were sending it digits, answering %s retroactively\n", ast_channel_name(c), ast_channel_name(in));
+								/* Indicate answer supervision to the caller before we exit.
+								 * We're not going to bridge, but this way at least the CDRs are correct, etc. */
+								ast_raw_answer(in);
+								strcpy(pa->status, "ANSWER");
+							} else {
+								*to_answer = 0; /* Continue in the dialplan, since nobody answered */
+							}
 							goto wait_over;
 						}
 					}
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

apps/app_dial.c
```diff
@@ -1603,6 +1603,17 @@ static struct ast_channel *wait_for_answer(struct ast_channel *in,
 						}
 						if (res) {
 							ast_log(LOG_WARNING, "Called channel %s hung up post-progress before all digits could be sent\n", ast_channel_name(c));
+							if (ast_channel_state(c) == AST_STATE_UP) {
+								/* The called channel answered while we were sending it digits, so the answer never got processed by app_dial.
```
jcolp, line 1607: The answer changes in this need a UserNote

jcolp (changes requested):

jcolp (approved):

PR #1918, opened 2026-05-06
res_rtp_asterisk: Add option to control stun host resolution when TTL = 0

### LINKED ISSUE #1858: [bug]: DNS records with a TTL of zero are permanently cached
### Severity

Minor

### Versions

20.4.0

### Components/Modules

res_rtp_asterisk.c

### Operating Environment

Linux 4.18

### Frequency of Occurrence

Constant

### Issue Description

I have asterisk using the STUN/TURN service from twilio, which uses a domain name that resolves to a different IP regularly. Asterisk resolves the name on startup and logs this line:

```
WARNING[114889] res_rtp_asterisk.c: Resolution for stunaddr 'global.stun.twilio.com' returned TTL = 0. Recurring resolution was cancelled.
```

The relevant DNS record (global.stun.twilio.com) has a TTL of zero. The [DNS spec](https://datatracker.ietf.org/doc/html/rfc1035#section-3.2.1) says "Zero values are interpreted to mean that the RR can only be used for the transaction in progress, and should not be cached" but it seems asterisk assumes a value of 0 means it should be cached indefinitely.

I was unable to find a way to work around recurring resolution being disabled for the name other than restarting asterisk every 10 minutes.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

gtjoseph:
If a hostname is specified for stunaddr in rtp.conf, periodic DNS resolution
is enabled based on the TTL returned in the DNS results.  If the TTL returned
is 0, it means that the next time the IP address is needed, it must be
looked up again.  I.E.  Don't cache.  Historically (and incorrectly) however,
res_rtp_asterisk stopped the periodic resolution and never re-resolved the
hostname again.

Besides what's mentioned in the user notes...
* Additional debugging was added in various STUN/DNS functions.
* The `rtp show settings` CLI command shows more detailed STUN info.
* Some debugging was added to dns_core.c and dns_recurring.c.

UserNote: A new `stunaddr_reresolve_ttl_0` parameter has been added to rtp.conf
that allows control over what happens when a STUN server hostname lookup
returns a TTL of 0.  The values can be set as follows:
- 'no': This is the historical (and current default) behavior of not doing
any further lookups and continuing to use the last successful result until
Asterisk is restarted or rtp.conf is reloaded.
- 'yes': Use the last cached result for the current call but trigger
re-resolution in the background for the benefit of future calls.
If the result of the background lookup is a ttl > 0, periodic resolution
will be restarted otherwise the next call will use the new cached value
and will trigger a background lookup again.

UserNote: A new CLI command `rtp resolve stun hostname` has been added
that will force a resolution of the STUN hostname and (re)start periodic
resolution if the result has a TTL > 0.

Resolves: #1858

### DIFF
--- a/configs/samples/rtp.conf.sample
+++ b/configs/samples/rtp.conf.sample
@@ -70,6 +70,25 @@ rtpend=20000
 ;
 ; stunaddr=
 ;
+; If a hostname is specified for stunaddr periodic DNS resolution is enabled
+; based on the TTL returned in the queries.  If the TTL returned is 0, it
+; means that the next time the IP address is needed, it should be looked up
+; again.  I.E.  Don't cache.  Historically (and incorrectly) however, we
+; stopped the periodic resolution and never re-resolved the hostname again.
+; The stunaddr_reresolve_ttl_0 parameter can address this:
+;
+; Possible values:
+; - 'no': This is the historical (and current default) behavior of not doing
+;    any further lookups and continuing to use the last successful result until
+;    Asterisk is restarted or rtp.conf is reloaded.
+; - 'yes': Use the last cached result for the current call but trigger
+;    re-resolution in the background for the benefit of future calls.
+;    If the result of the background lookup is a ttl > 0, periodic resolution
+;    will be restarted otherwise the next call will use the new cached value
+;    and will trigger a background lookup again.
+; 
+; stunaddr_reresolve_ttl_0 = no
+;
 ; Some multihomed servers have IP interfaces that cannot reach the STUN
 ; server specified by stunaddr.  Blacklist those interface subnets from
 ; trying to send a STUN packet to find the external IP address.
@@ -85,6 +104,8 @@ rtpend=20000
 ; to a specific IP address.  Blacklisting is done via ACL infrastructure
 ; so it's possible to whitelist as well.
 ;
+
+
 ; stun_acl = named_acl
 ; stun_deny = 0.0.0.0/0
 ; stun_permit = 1.2.3.4/32
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -228,7 +228,6 @@ static int dtls_mtu = DEFAULT_DTLS_MTU;
 #ifdef HAVE_PJPROJECT
 static int icesupport = DEFAULT_ICESUPPORT;
 static int stun_software_attribute = DEFAULT_STUN_SOFTWARE_ATTRIBUTE;
-static struct sockaddr_in stunaddr;
 static pj_str_t turnaddr;
 static int turnport = DEFAULT_TURN_PORT;
 static pj_str_t turnusername;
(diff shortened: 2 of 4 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

seanbright:
Can we make the default in master be the RFC compliant behavior?

gtjoseph:
> Can we make the default in master be the RFC compliant behavior?

Hmmm.  It's a small enough tweak that it shouldn't cause cherry-pick issues later. Let me chat with @jcolp in the morning.  

EDIT: I think it'd have to be "trigger_only" otherwise unexpected delays in call setup could occur.

jcolp:
I think changing the default in master to something more reasonable is fine.

gtjoseph:
I'm moving this back to draft.  I need to think a bit more about the "trigger_only" option.  If the TTL is always 0 and we keep triggering with a large call volume there's going to be a lot of churn going on and I'm not quite sure I like how locking is happening.  Maybe I need to check if a resolution is already in-flight before scheduling another one.

gtjoseph:
@seanbright @jcolp I'm wondering if we really need 3 options and can just do a simple boolean.

* false: current behavior
* true: if ttl == 0, call setup will use the cached result and trigger an async lookup.  If the async lookup gets a new TTL > 0, future calls will use the cached result.  This avoids having to do a synchronous lookup but it also means that a few calls might use a stale cached result.  I can't imagine that to be an issue however.

Thoughts?

Oh, To keep things simple right now, I'm going to cherry-pick to all branches and once merged, submit a new master-only PR to change the default.

jcolp:
I thinkkkk a boolean is fine, and same for using a stale entry.

gtjoseph:
The last force-push fixed a bug where a failed lookup was never retried and added a `rtp resolve stun hostname` CLI command.

It also removed the re-resolution when a module reload (even via `core reload`) was done but no changes were detected to rtp.conf.  I was kinda iffy on that in the first place because it could cause a change that wasn't expected.

configs/samples/rtp.conf.sample
```diff
@@ -70,6 +70,25 @@ rtpend=20000
...
+; - 'no': This is the historical (and current default) behavior of not doing
+;    any further lookups and continuing to use the last successful result until
+;    Asterisk is restarted or rtp.conf is reloaded.
+; - 'yes': Use the last cached result for the current call but trigger
+;    re-resolution in the background for the benefit of future calls.
+;    If the result of the background lookup is a ttl > 0, periodic resolution
+;    will be restarted otherwise the next call will use the new cached value
+;    and will trigger a background lookup again.
+; 
+; stunaddr_reresolve_ttl_0 = never
```
mbradeen: never?

gtjoseph: oops, that was a lefover.

mbradeen (approved):

res/res_rtp_asterisk.c
```diff
@@ -244,9 +243,16 @@ static ast_rwlock_t ice_acl_lock = AST_RWLOCK_INIT_VALUE;
...
+static struct sockaddr_in stunaddr;
 /*! stunaddr recurring resolution */
 static ast_rwlock_t stunaddr_lock = AST_RWLOCK_INIT_VALUE;
 static struct ast_dns_query_recurring *stunaddr_resolver = NULL;
+/*! TTL from last successful query */
+static int stunaddr_ttl = 0;
+/*! The current hostname if stunaddr isn't an IP address */
+static char *stun_hostname = NULL;
+/*! Re-resolve hostname if TTL = 0? */
+int stunaddr_reresolve_ttl_0 = 0;
```
jcolp: Why isn't this static?

jcolp (dismissed):

res/res_rtp_asterisk.c
gtjoseph: I was testing you to see if you'd notice.

jcolp (approved):

PR #1920, opened 2026-05-06
res_odbc: stop disrupting in-flight consumers on reload

### LINKED ISSUE #1919: [bug]: res_odbc reload disrupts in-flight ODBC consumers (func_odbc, res_config_odbc, PJSIP realtime) under heavy load
### Severity

Major

### Versions

22, 21, 20

### Components/Modules

res_odbc, func_odbc

### Operating Environment

Linux (reproduced on Debian bookworm + PostgreSQL/psqlodbc; mechanism is environment-independent)

### Frequency of Occurrence

Frequent

### Issue Description

This is a very long-standing issue with Asterisk Realtime, and highly scaled systems. I can recall running into this in production systems going back at least a decade .. but this typically manifests in a minority - but significant enough to be a problem when it happens - with remote ODBC connections AND a very large number of Asterisk Realtime operations.

`reload()` in `res/res_odbc.c` always destroys and recreates every `odbc_class` on every reload, regardless of whether the class's configuration actually changed. The new class starts with an empty connection pool. Under load this produces a thundering herd of fresh `SQLConnect`s after every reload, with the database returning `FATAL: too many clients already` for some, and `last_negative_connect` getting set on the new class as a result.

There is a second, additive bug in the request path: `aoro2_class_cb` (used by `_ast_odbc_request_obj2()` and `ast_odbc_get_max_connections()`) filters out classes with `delme=1`. The reload sequence sets `delme=1` on every class **before** `load_odbc_config()` has linked replacements, so during that window every concurrent `ast_odbc_request_obj()` returns NULL and queries fail with "Class not found".

**Blast radius.** Every consumer of res_odbc shares the same `class_container`:

- **func_odbc** — ODBC dialplan functions return empty/incorrect results for the duration of the storm.
- **res_config_odbc** — realtime config lookups (sorcery / `ast_load_realtime`) return no rows. With PJSIP realtime backed by ODBC (`ps_endpoints`/`ps_auths`/`ps_aors`/`ps_endpoint_id_ips`), this manifests as **SIP peers becoming unreachable / unavailable** during and after the reload window.
- Anything else that calls `ast_odbc_request_obj()` (cdr_odbc, etc.) sees the same disruption.

**Two related, smaller issues found while investigating** (will be fixed by the same PR):

- **Use After Free in `acf_odbc_read()`** at `funcs/func_odbc.c:873-877`. The queries rwlock is released at line 873 before the per-DSN execution loop, but `query->readhandle[dsn_num]` is dereferenced at line 876. A concurrent `module reload func_odbc.so` frees the query in that window. With glibc's tcache the freed `acf_odbc_query` is typically immediately reused for the *replacement* query, so this doesn't appear to crash in the real world. The unload path acknowledges this race with a `usleep` workaround at line 1933; the reload path does not.

- **Uninitialised `start` in `ast_odbc_direct_execute()` and `ast_odbc_prepare_and_execute()`** at `res/res_odbc.c:365-460`. Each function checks `obj->parent->logging` twice — once before the SQL exec to sample the start timestamp, and once after to compute `execution_time`. The local `struct timeval start` is only initialised inside the first branch. Latent race that pre-existing code dodged because `logging` was only ever set at module load; becomes triggerable as soon as anything mutates `logging` on a published class.

## Reproduction

A self-contained reproducer is at <https://github.com/joshelson/asterisk-repros/tree/main/res_odbc-reload-under-load>.

Two ways to demonstrate the bug:

1. **Standalone C model** (no DB or asterisk needed) at `repro.c` — `./build.sh && ./run.sh` produces a buggy-vs-fixed comparison table in 5 seconds. Also runs cleanly under ThreadSanitizer + AddressSanitizer (`./build-sanitized.sh`).
2. **Real-Asterisk Docker harness** — Linux container with Asterisk built from current `master`, talking to PostgreSQL with `max_connections=15`. AMI-driven load fires Originates into one of two dialplan contexts:
   - `odbc-loop` — uses `${ODBC_LOOKUP(...)}` (func_odbc → res_odbc path)
   - `realtime-loop` — uses `${REALTIME_FIELD(rt_test,id,...,value)}` (res_config_odbc → res_odbc path; same code path PJSIP realtime uses for `ps_endpoints` etc.)

Configurable matrix: 18 `module reload`s fired over 18 seconds; 7 seconds of pure observation tail. `cps=8` originates per second, 1500 ODBC queries per call. Postgres `max_connections=15`, Asterisk `max_connections=10` per class. Run `scripts/run-matrix.sh` for the full buggy-vs-patched matrix; see the repo README for tunables.

### Buggy results (current Asterisk)

| Scenario | failures / total queries | tail (post-reload-stop) |
|---|---|---|
| `ODBC_LOOKUP` + `module reload res_odbc.so` | 243 / 134,560 (0.18%) | 61 (0.26%) |
| `ODBC_LOOKUP` + `module reload func_odbc.so` | 0 / 170,853 | 0 |
| `ODBC_LOOKUP` + `module reload` (everything) | 167 / 123,812 (0.13%) | 18 (0.08%) |
| `REALTIME` + `module reload res_odbc.so` | 325 / 123,590 (0.26%) | 42 (0.18%) |
| `REALTIME` + `module reload func_odbc.so` | 0 / 114,138 | 0 |
| `REALTIME` + `module reload` (everything) | **380 / 90,720 (0.42%)** | **123 (0.65%)** |

The realtime path is hit *harder* than func_odbc by the same bug (more `ast_odbc_request_obj()` calls per unit of work). `func_odbc.so` reload alone is benign — confirms the trigger is res_odbc reload, and that bare `module reload` (which iterates all reload-capable modules) is the worst case.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

joshelson:
**Status:** Draft. Depends on #1924 (res_odbc aco_* conversion). Once
#1924 merges this branch will be rebased on top and the commit list
restructured to fit the new aco-based parser. The empirical results
below were captured against master before #1924 existed; they show the
bug and demonstrate the fix mechanism, and will be re-captured against
the post-rebase commits before this PR moves out of draft.

## Summary

`module reload res_odbc.so` (or bare `module reload`, which iterates
all reload-capable modules) under concurrent load disrupts every
consumer of `res_odbc` — `func_odbc` queries fail, `res_config_odbc`
realtime lookups return no rows, and PJSIP-realtime-driven SIP peer
lookups (with `ps_endpoints`/`ps_auths`/`ps_aors` backed by ODBC) make
peers go unreachable. The trigger is that `reload()` always destroys
and recreates every `odbc_class` regardless of whether the class's
config actually changed, dumping the cached connection pool every time
and producing a thundering herd of fresh `SQLConnect`s under load. An
additive request-path race where `aoro2_class_cb` filters out delme=1
classes during the brief window where every class is marked delme=1
before replacements have been linked compounds the failure rate.

See #1919 for the full diagnosis.

## Plan after #1924 lands

#1924 replaces the hand-rolled parser with `config_options` (aco_*) but
preserves the destroy-and-recreate reload semantics, so the bug
described above is still present after that lands. This PR's work
moves to the aco-based codebase as:

1. **Preserve-on-reload reconciliation** — modify
   `odbc_reconcile_config()` to look up each cfg item's existing
   runtime class in `class_container` and update safe-to-mutate
   fields in place (under `class->lock`) when no connection-affecting
   field changed. Replaces the original two-pass `aoro2_class_cb` and
   the legacy preserve-on-reload patches.
2. **`prewarm_pool` per-class aco option** — add the option to
   `struct odbc_class_cfg` + the `odbc_register_options` table; build
   the new class and pre-fill its connection pool before linking it
   into `class_container`. On pre-warm failure, keep the previous
   class alive.

Two related fixes identified during the original investigation could split into
single-purpose PRs since they are orthogonal to the
reload-under-load work and to the aco conversion:

- `func_odbc: snapshot readhandle before releasing queries rwlock`
  (UAF in `acf_odbc_read` between rwlock release and `query->readhandle`
  deref).
- `res_odbc: snapshot logging flag during execute paths` (latent
  uninitialised-`start` exposed once `logging` becomes mutable on a
  published class).

The exact commit shape will be locked in when #1924 merges and the
rebase happens.

## Reproduction

Self-contained reproducer (Docker harness + standalone C model + AMI
loadgen) at
https://github.com/joshelson/asterisk-repros/tree/main/res_odbc-reload-under-load .

| Scenario (18 reloads in 18s + 7s observation tail) | Buggy fail / total | Buggy tail | Patched fail / total | Patched tail |
|---|---|---|---|---|
| `ODBC_LOOKUP` + `module reload res_odbc.so` | 243 / 134,560 (0.18%) | 61 (0.26%) | 0 / 93,736 | 0 |
| `ODBC_LOOKUP` + `module reload func_odbc.so` | 0 / 170,853 | 0 | 0 / 57,134 | 0 |
| `ODBC_LOOKUP` + `module reload` (everything) | 167 / 123,812 (0.13%) | 18 (0.08%) | 0 / 64,588 | 0 |
| `REALTIME` + `module reload res_odbc.so` | 325 / 123,590 (0.26%) | 42 (0.18%) | 0 / 102,613 | 0 |
| `REALTIME` + `module reload func_odbc.so` | 0 / 114,138 | 0 | 0 / 132,364 | 0 |
| **`REALTIME` + `module reload` (everything)** | **380 / 90,720 (0.42%)** | **123 (0.65%)** | **0 / 112,328** | **0** |

The realtime path (the PJSIP-relevant one) is hit harder than
`func_odbc` by the same bug. Bare `module reload` against the realtime
path is the worst case and matches the production symptom most closely.
All patched runs show zero failures, including in the
post-reload-stop observation window.

Resolves: #1919

### DIFF
--- a/configs/samples/res_odbc.conf.sample
+++ b/configs/samples/res_odbc.conf.sample
@@ -33,6 +33,20 @@ dsn => asterisk
 ; Build a connection at startup?
 pre-connect => yes
 ;
+; On a reload that replaces this class (i.e. the dsn, username, password, or
+; sanitysql in res_odbc.conf was changed), pre-warm the entire connection pool
+; on the new class BEFORE publishing it. This eliminates the post-reload
+; thundering-herd of fresh SQLConnects under load by doing those connects up
+; front while the existing class continues to serve traffic. If pre-warm fails
+; (new DSN unreachable, bad credentials) the new configuration is rejected and
+; the previous class continues serving requests uninterrupted. This option has
+; NO effect at initial module load or on reloads that do not change a
+; connection-affecting field. Trade-off when triggered: reload latency
+; increases by max_connections * connect_latency for the affected class, and
+; the database briefly sees double the connection count for the affected class
+; during the swap. Defaults to no.
+;prewarm_pool => no
+;
 ; What should we execute to ensure that our connection is still alive?  The
 ; statement should return a non-zero value in the first field of its first
 ; record.  The default is "select 1".
--- a/funcs/func_odbc.c
+++ b/funcs/func_odbc.c
@@ -772,6 +772,14 @@ static int acf_odbc_read(struct ast_channel *chan, const char *cmd, char *s, cha
 	struct ast_str *sql = ast_str_thread_get(&sql_buf, 16);
 	const char *status = "FAILURE";
 	struct dsn *dsn = NULL;
+	/*
+	 * Snapshot of the per-DSN read handle list. The queries rwlock is
+	 * dropped before doing the slow SQL execution so that concurrent
+	 * reloads aren't blocked by long-running reads. But that means we
+	 * cannot dereference `query` after the unlock — a concurrent
+	 * func_odbc reload would free it. Copy what we still need.
+	 */
+	char readhandle[ARRAY_LEN(query->readhandle)][sizeof(query->readhandle[0])];
 
 	if (!sql || !colnames) {
 		if (chan) {
@@ -870,11 +878,12 @@ static int acf_odbc_read(struct ast_channel *chan, const char *cmd, char *s, cha
 			AST_LIST_HEAD_INIT(resultset);
 		}
 	}
+	memcpy(readhandle, query->readhandle, sizeof(readhandle));
 	AST_RWLIST_UNLOCK(&queries);
 
-	for (dsn_num = 0; dsn_num < 5; dsn_num++) {
-		if (!ast_strlen_zero(query->readhandle[dsn_num])) {
-			obj = get_odbc_obj(query->readhandle[dsn_num], &dsn);
+	for (dsn_num = 0; dsn_num < ARRAY_LEN(readhandle); dsn_num++) {
+		if (!ast_strlen_zero(readhandle[dsn_num])) {
+			obj = get_odbc_obj(readhandle[dsn_num], &dsn);
 			if (!obj) {
 				continue;
 			}
--- a/res/res_odbc.c
+++ b/res/res_odbc.c
@@ -886,6 +886,14 @@ int ast_odbc_backslash_is_escape(struct odbc_obj *obj)
 	return obj->parent->backslash_is_escape;
 }
 
+/*!
+ * \internal
+ * \brief ao2 matcher that returns only active classes (delme=0).
+ *
+ * The primary lookup used by request paths. Prefers a freshly-linked
+ * replacement class over an outgoing class when the two briefly
+ * coexist during reload.
+ */
 static int aoro2_class_cb(void *obj, void *arg, int flags)
 {
 	struct odbc_class *class = obj;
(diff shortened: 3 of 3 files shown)


joshelson:
multiple-commits: standalone

cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22

jcolp:
Did you explore moving this to the config_options API at all?

jcolp:
How much of that response was from an LLM?

joshelson:
SORRY WOW. The wayward child went crazy there. Didn't want to send that along... automatically monitoring for comments is crazy. Apologies @jcolp. 

> Did you explore moving this to the config_options API at all?

I didn't look at this in specific detail though we did use the aco_* stuff last time I made some changes.

Are you open to doing a full migration here as a part of this PR, or do you want to consider that separately? It isn't a huge effort to put it into the new API framework.

jcolp:
With res_odbc being a core module and the preferred way to do databases if the configuration handling is being touched (it is here) I would prefer at the very least the config_options API be explored as it provides atomic reloads, field handling, automatic documentation on the docs.asterisk.org site, and has proven to be extremely reliable. I don't know if it would fulfill all needs though, thus why I asked if it was explored and discarded for some reason. In general I do not trust the existing configuration handling and reconcillation.

jcolp:
I'd also add this PR can stay up in draft until you have an answer on that. That does mean all commits would be in draft, which is why if there are independent fixes then those should generally be done as individual PRs with their own independent review and timeline.

joshelson:
> I'd also add this PR can stay up in draft until you have an answer on that. That does mean all commits would be in draft, which is why if there are independent fixes then those should generally be done as individual PRs with their own independent review and timeline.

Moving to draft and looking into feasibility here. Thanks for the initial comments here! Very helpful.

joshelson:
@jcolp  - put up a PR here with a few caveats that might make for interesting discussion on the ACO changes.

#1924 should have the info and several caveats.

Would love some thoughts there and whether we can mvoe forward with this.

jcolp:
I am fully aware that you put up another PR, but there are numerous other PRs ahead of yours for me to look at. It will get looked at when it gets looked at. Maybe this week, maybe not.

joshelson:
> I am fully aware that you put up another PR, but there are numerous other PRs ahead of yours for me to look at. It will get looked at when it gets looked at. Maybe this week, maybe not.

Thank you. Didn't mean to be annoying over here. But let me know if I can help with anything. :)

PR #1925, opened 2026-05-07
pjsip_configuration: Show actual dtls_verify config.

jkroonza:
Rather than merely showing
    
dtls_verify : Yes/No
    
in pjsip show endpoint xxx it will now be shown what exactly is being
checked, ie, one of:
    
dtls_verify : No
dtls_verify : Fingerprint
dtls_verify : Certificate
dtls_verify : Yes
    
Where Yes implies both Fingerprint and Certificate.
    
Signed-off-by: Jaco Kroon <jaco@uls.co.za>

### DIFF
--- a/res/res_pjsip/pjsip_configuration.c
+++ b/res/res_pjsip/pjsip_configuration.c
@@ -1021,7 +1021,19 @@ static int dtls_handler(const struct aco_option *opt,
 static int dtlsverify_to_str(const void *obj, const intptr_t *args, char **buf)
 {
 	const struct ast_sip_endpoint *endpoint = obj;
-	*buf = ast_strdup(AST_YESNO(endpoint->media.rtp.dtls_cfg.verify));
+	if (endpoint->media.rtp.dtls_cfg.verify == AST_RTP_DTLS_VERIFY_NONE) {
+		*buf = ast_strdup("No");
+	} else if (endpoint->media.rtp.dtls_cfg.verify == AST_RTP_DTLS_VERIFY_FINGERPRINT) {
+		*buf = ast_strdup("Fingerprint");
+	} else if (endpoint->media.rtp.dtls_cfg.verify == AST_RTP_DTLS_VERIFY_CERTIFICATE) {
+		*buf = ast_strdup("Certificate");
+	} else if (endpoint->media.rtp.dtls_cfg.verify == (AST_RTP_DTLS_VERIFY_FINGERPRINT|AST_RTP_DTLS_VERIFY_CERTIFICATE)) {
+		*buf = ast_strdup("Yes");
+	} else {
+		*buf = NULL;
+		ast_assert(0);
+		return 1;
+	}
 	return 0;
 }
 


jkroonza:
Example output after this from pjsip show endpoint (rather than simply Yes/No):

```
 dtls_verify                        : fingerprint
```

jkroonza:
Okay, so since there are test failures here, the implication is that some external stuff may rely on the specific Yes/No response.  As such either this update should not be made, or tests should be updated along with UpgradeNotes?

seanbright:
I think your output should match what the parser consumes. So one of:

- yes
- no
- fingerprint
- certificate

jkroonza:
> I think your output should match what the parser consumes. So one of:
> 
>     * yes
> 
>     * no
> 
>     * fingerprint
> 
>     * certificate

What you say makes sense, but it doesn't align with reality, there are 4 possible values for the option where the values in brackets is what would currently be output:

0 => no validation (No)
1 => fingerprint (Yes)
2 => certificate (Yes)
3 => fingerprint,certificate - ie, check both (Yes)

The patch adjusts to for 1 to 3 output as above, for 0 it says None - but sure this can go back to No, which will solve part of the validation.  I can adjust the other options as preferred, but there really should be distinction between them.

The ari failures states:

"error": "Allocation failed"

Which allocation?  That doesn't sound like a string comparison issue, but possibly something bigger.

For the pjs2 test:

asterisk.ami:207 __check_result: Event occurred 1 times, which is out of the allowable range

That does sound like a validation issue.

seanbright:
I’m not following, but here is the parsing code I was talking about:

https://github.com/asterisk/asterisk/blob/176ecdad3cc97437d552d180936faf54c8f2b432/main/rtp_engine.c#L3280-L3287

Alternatively the config parser could be extended to also handle the `fingerprint, certificate` format.

res/res_pjsip/pjsip_configuration.c
```diff
@@ -1021,7 +1021,23 @@ static int dtls_handler(const struct aco_option *opt,
 static int dtlsverify_to_str(const void *obj, const intptr_t *args, char **buf)
 {
 	const struct ast_sip_endpoint *endpoint = obj;
-	*buf = ast_strdup(AST_YESNO(endpoint->media.rtp.dtls_cfg.verify));
+	switch (!endpoint->media.rtp.dtls_cfg.verify) {
+	case AST_RTP_DTLS_VERIFY_NONE:
+		*buf = ast_strdup("no");
```
seanbright: `"no"` -> `"No"`

res/res_pjsip/pjsip_configuration.c
```diff
@@ -1021,7 +1021,23 @@ static int dtls_handler(const struct aco_option *opt,
...
+		break;
+	case AST_RTP_DTLS_VERIFY_CERTIFICATE:
+		*buf = ast_strdup("Certificate");
+		break;
+	case AST_RTP_DTLS_VERIFY_FINGERPRINT|AST_RTP_DTLS_VERIFY_CERTIFICATE:
+		*buf = ast_strdup("Yes");
+		break;
+	default:
+		*buf = ast_strdup("Unknown Value");
+		break;
```
seanbright: If `verify` got a value we don't recognize, that is a logical error that should be addressed. I would do something like this instead:

```
*buf = NULL;
ast_assert(0);
return 1;
```

res/res_pjsip/pjsip_configuration.c
```diff
@@ -1021,7 +1021,23 @@ static int dtls_handler(const struct aco_option *opt,
...
+	case AST_RTP_DTLS_VERIFY_NONE:
+		*buf = ast_strdup("no");
+		break;
+	case AST_RTP_DTLS_VERIFY_FINGERPRINT:
+		*buf = ast_strdup("Fingerprint");
+		break;
+	case AST_RTP_DTLS_VERIFY_CERTIFICATE:
+		*buf = ast_strdup("Certificate");
+		break;
+	case AST_RTP_DTLS_VERIFY_FINGERPRINT|AST_RTP_DTLS_VERIFY_CERTIFICATE:
```
seanbright: This won't work with `switch`ing over an enum - at least with gcc 15.

seanbright (changes requested):

res/res_pjsip/pjsip_configuration.c
jkroonza: Not generating an error?

gcc version 15.2.1 20260214 (Gentoo 15.2.1_p20260214 p5) 

Does it just not work?

seanbright:
> Workflow Check failed [master-build](https://github.com/asterisk/asterisk/actions/runs/25756666014/job/75647196307): FAILED: Build

Yes. It literally fails to build.

res/res_pjsip/pjsip_configuration.c
seanbright: https://github.com/asterisk/asterisk/pull/1925#issuecomment-4433965713

jkroonza:
> > Workflow Check failed [master-build](https://github.com/asterisk/asterisk/actions/runs/25756666014/job/75647196307): FAILED: Build
> 
> Yes. It literally fails to build.

With -Wall -Werror sure.  So this correct warning once I fix the boolean situation (which for some reason I missed):

```
res_pjsip/pjsip_configuration.c: In function ‘dtlsverify_to_str’:
res_pjsip/pjsip_configuration.c:1034:9: warning: case value ‘3’ not in enumerated type ‘enum ast_rtp_dtls_verify’ [-Wswitch]
 1034 |         case AST_RTP_DTLS_VERIFY_FINGERPRINT|AST_RTP_DTLS_VERIFY_CERTIFICATE:
      |         ^~~~
```

Two fixes:

1.  Add a BOTH value to the enum.  Which frankly in my opinion is not the right way since this enum (by virtue of using bit-shifts in the enum definition) clearly represents a bitmask.
2.  Explicit cast to int so the enum value check isn't performed.  This is also nasty but the less nasty of the two options, and thus the route I went with.

Seems I screwed up copying back from the test host to where I run my git push from during previous push.  Re-verified. And will push now.

jcolp:
Or you could have just used ifs instead...

jkroonza:
> Or you could have just used ifs instead...

Right, that's to my surprise but sure.  Given that the warning really says you're comparing apples and oranges I'd have thought it would affect if statements too, not only switch .. case.  But alas, you learn something every day.

jkroonza:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

seanbright (approved):

jcolp (approved):

seanbright:
Sorry, I accidentally made a change to this PR that I did not intend to. I'm working to fix.

seanbright (approved):

PR #1934, opened 2026-05-18
WebSocket Enhancements: Proxies and Keepalives for ARI and Media Outbound Websockets.

### LINKED ISSUE #1881: [improvement]:  Add HTTP proxy support for outbound WebSocket connections in websocket_client.conf (chan_websocket / ARI ExternalMedia)
## Summary

`websocket_client.conf` (used by `chan_websocket` and ARI `externalMedia` for outbound WebSocket connections) does not currently expose any configuration for routing the outbound connection through an HTTP proxy. In restricted / corporate network environments where all outbound traffic must egress via a forward HTTP proxy (HTTP `CONNECT` tunneling), this makes it impossible to use these features without ugly workarounds.

This was discussed on the community forum here: https://community.asterisk.org/t/add-http-proxy-support-for-outbound-websocket-connections-in-websocket-client-conf-chan-websocket-ari-externalmedia/112661 — @gjoseph suggested opening an Improvement issue.

## Current Behavior

- `websocket_client.conf` has no proxy-related configuration keys.
- Outbound WebSocket connections initiated by `chan_websocket` or ARI `externalMedia` are attempted directly to the configured `uri`.
- In environments where direct egress is blocked and only an HTTP proxy is permitted (common in enterprise / on-prem / PCI-scoped deployments), the connection simply fails with no way to route it through a proxy.

## Proposed Behavior

Extend `websocket_client.conf` with HTTP proxy support for outbound `ws://` / `wss://` connections. Suggested configuration keys on each client section:

- `http_proxy` — proxy URL, e.g. `http://proxy.corp.example:3128`
- `http_proxy_user` — optional username for proxy authentication (Basic)
- `http_proxy_password` — optional password for proxy authentication
- `no_proxy` — optional comma-separated list of hosts/domains to bypass the proxy
- (Optionally) honor the standard `HTTP_PROXY` / `HTTPS_PROXY` / `NO_PROXY` environment variables as a fallback when no explicit config is set.

Implementation should use the standard HTTP `CONNECT` method to establish a tunnel, then perform the WebSocket handshake (and the TLS handshake for `wss://`) end-to-end over that tunnel.

This should apply uniformly to every consumer of `websocket_client.conf`:

- `chan_websocket` outbound client sessions
- ARI `externalMedia` outbound WebSocket channels
- Any other module that initiates an outbound WS/WSS connection via `res_websocket_client`

## Use Case

Our deployment runs Asterisk inside a network where direct outbound internet access is blocked by policy; all outbound HTTP(S) / WSS traffic must go through a corporate forward proxy. We want to use ARI `externalMedia` over WebSocket to stream audio to an external media server that lives outside this network. Because `websocket_client.conf` has no proxy option, the outbound `wss://` connection currently cannot be established at all.

This is a common pattern for:

- Enterprise / corporate on-prem deployments
- PCI / regulated environments with strict egress restrictions
- Cloud deployments behind an egress proxy (e.g. Squid, corporate NGFW)

## Alternatives Considered

- **Transparent proxy via `iptables REDIRECT` / `DNAT`** — fragile, breaks TLS verification for `wss://`, and requires privileged host-level configuration that isn't always available.
- **Local forward proxy / sidecar** that re-terminates the connection — adds an extra process to maintain, complicates TLS, and still doesn't cleanly solve proxy authentication.
- **Locally patching `chan_websocket`** — not portable and has to be re-applied on every upgrade.

None of these are acceptable for a production deployment; first-class support in `websocket_client.conf` is the right fix.

## Environment

- Asterisk version: `22.8.2`
- OS / distro: `Ubuntu 22.04`
- Proxy in use: `http://test.com:5005`

## Additional Context

Happy to help test once an implementation is available, and to contribute a patch if useful — would appreciate guidance on whether the preferred place to add this is at the `res_websocket_client` layer or lower in the HTTP client stack.

### LINKED ISSUE #1933: [improvement]: Outbound Websockets: Add keepalive mechanisms
### Improvement Description

Right now there's no way to detect issues (especially network issues) with persistent outbound websockets when they're idle.  They have to actually attempt to send data and fail but this can result in a call failure.  Instead WebSocket PING/PONGs and/or TCP keepalives should be used to detect an issue and possibly trigger a failover before a call needs the connection.

gtjoseph:
See the notes below for high-level descriptions of the new features.

* Proxies

Rather than attempting to modify the existing websocket client code in
res_http_websocket to add proxy support, we're using libcurl to provide both
the websocket client and proxy capabilities. However, because libcurl didn't
add support for websockets until 7.88, support is dependent on having that
version or greater installed. For that reason, the existing websocket client
implementation using the core tcptls facility remains intact and the curl
based client resides along side it. To make this easier, a new directory
res/res_http_websocket was created and contains curl_client.c, tcptls_client.c
and tcptls_server.c with common code remaining in res_http_websocket.c.

To manage the dependency on libcurl, autoconf/libcurl.m4, makeopts.in, and
res/Makefile were updated to detect websocket support and make compiling
curl_client.c dependent on it being available. Both the curl_client and
tcptls_client make a virtual function table available that the public
ast_websocket_* functions use to call the lower level functions.  If both
clients are available at compile time, the curl version is used. If not,
the tcptls version is used. This currently isn't selectable at runtime but
probably could be.

If you try to configure a proxy and the curl client isn't avaiable, you'll
get a configuration error on startup.

* Keepalives

Both TCP-level and WebSocket PING/PONG keepalives can be configured and are
available with either the curl or tcptls client implementations. The TCP
keepalives are handled entirely by the operating system and require no
resources from Asterisk but by their very nature, they can't traverse proxies.
WebSocket PING/PONGs are implemented in the Asterisk websocket code and require
a scheduler thread to keep track of them so they're a bit more complicated but
they do traverse proxies.  Which one is used is completely up to the admin.
You could use both.

Both methods are implemented in a way that allows them to be used on incoming
websocket connections but there's currently no way to configure it.
Something for the future.

* Other Changes

A few changes were needed to res/ari/ari_websockets to add explicit calls to
ast_websocket_close.  It had been assuming that the websocket session
destructor would close the websocket when it unreffed it but the keepalive
process now holds a reference so the destructor wouldn't actually run
without the call to ast_websocket_close to stop the keepalives.

There was a bug in sorcery.h where the ast_sorcery_register_uint macro
was referencing _stringify (which doesn't exist) instead of _sorcery_stringify.

Since autoconf/libcurl.m4 was modified, bootstrap.sh had to be run so
./configure and include/asterisk/autoconfig.h.in were modified.

Resolves: #1881
Resolves: #1933

UserNote: A forward proxy can now be specified for outbound websockets. Support
is dependent on libcurl version 7.88 or greater being available in both the
environment Asterisk is built in and run in. See the websocket_client.conf.sample
file for configuration information.

UserNote: TCP-level or WebSocket PING/PONG keepalives can now be enabled on
outbound websockets.  They can help detect network failures even when a
persistent connection is idle. See the websocket_client.conf.sample
file for configuration information.

DeveloperNote: The addition of the proxy and keepalive configuration parameters
pushed the websocket client parameter count over 32. This necessitated changing
the size of the ast_ws_client_fields enum from a 32 bit bitfield to a 64-bit
bitfield with a corresponding change to the ast_websocket_client structure.


gtjoseph:
I'm leaving this in DRAFT for now. There's still some cleanup and test work to be done but I wanted to get some initial feedback early on to make sure it fits everyone's needs.

It's OK to apply in a test environment BUT NOT IN PRODUCTION!

gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
testsuite-test-pr: 136

gtjoseph:
Well, res_http_websocket is now dependent on res_curl even if curl itself doesn't support websockets so I need to re-think that a bit.  I might have to move curl_client.c into its own module.

gtjoseph:
The curl_client.c file has been moved to a new res_websocket_client_curl.c module which registers itself to res_http_websocket as a client backend.  This makes res_http_websocket no longer dependent on curl or res_curl.  It also means that you can keep the original tcptls implementation if you don't need proxies but I wouldn't recommend it.  The curl implementation is more robust.

I'm not sure what the deal with the failing tests is as they all pass for me locally.

I'll update the PR description and commit message when things are more stable.

gtjoseph:
Unfortunately, proxies will never work with ubuntu 24 and earlier because for some bizarre reason, they don't enable websocket support in curl.  They do build curl with the wwebsocket APIs, which is what configure checks for, but you can't connect with a `ws://` URI.

/me rethings everything.

PR #1937, opened 2026-05-19
WebSocket Enhancements: Proxies and Keepalives for ARI and Media Outbound Websockets.

### LINKED ISSUE #1881: [improvement]:  Add HTTP proxy support for outbound WebSocket connections in websocket_client.conf (chan_websocket / ARI ExternalMedia)

### LINKED ISSUE #1933: [improvement]: Outbound Websockets: Add keepalive mechanisms

gtjoseph:
See the notes below for high-level descriptions of the new features.

* Proxies

Outbound/forward HTTP proxies are now supported and configurable in
websocket_client.conf. You can specify a host:port plus optional proxy_username
and proxy_password. Because WebSockets aren't consistently supported among
proxies (specifically passing through UPGRADEs), the CONNECT method is always
used to establish a TCP tunnel through the proxy. This is required if a TLS
session is to be established with the WebSocket server anyway.  It's important
to understand that that negotiation with the proxy is ALWAYS unsecured. Once
the proxy establishes the tunnel, the TLS session will be negotiated directly
with the remote WebSocket server via the tunnel.

* Keepalives

Both TCP-level and WebSocket PING/PONG keepalives can be configured and are
available with either the curl or tcptls client implementations. The TCP
keepalives are handled entirely by the operating system and require no
resources from Asterisk but by their very nature, they can't traverse proxies.
WebSocket PING/PONGs are implemented in the Asterisk websocket code and require
a scheduler thread to keep track of them so they're a bit more complicated but
they do traverse proxies.  Which one is used is completely up to the admin.
You could use both.

* Other Changes

A few changes were needed to res/ari/ari_websockets and
res/res_aeap/transport_websocket to add explicit calls to ast_websocket_close.
They had been assuming that the websocket session destructor would close the
websocket when it unreffed it but the keepalive process now holds a reference
so the destructor wouldn't actually run without the call to ast_websocket_close
to stop the keepalives.

A few new methods were added to tcptls.c to allow switching an existing
connection from unsecured to TLS.  These were required because the initial
connection and handshake with a proxy is always unsecured but then needs
to be switched to TLS if required for the remote WebSocket server.

There was a bug in sorcery.h where the ast_sorcery_register_uint macro
was referencing _stringify (which doesn't exist) instead of _sorcery_stringify.

Resolves: #1881
Resolves: #1933

UserNote: Forward/outbound proxies can now be specified for outbound websockets.
See the websocket_client.conf.sample file for configuration information.

UserNote: TCP-level or WebSocket PING/PONG keepalives can now be enabled on
outbound websockets.  They can help detect network failures even when a
persistent connection is idle. See the websocket_client.conf.sample
file for configuration information.

DeveloperNote: The addition of the proxy and keepalive configuration parameters
pushed the websocket client parameter count over 32. This necessitated changing
the size of the ast_ws_client_fields enum from a 32 bit bitfield to a 64-bit
bitfield with a corresponding change to the ast_websocket_client structure.


gtjoseph:
It's back!
No dependency on curl for proxies.
In draft mode until I see that the tests pass.

gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

configs/samples/websocket_client.conf.sample
```diff
@@ -49,3 +49,109 @@
...
+; All WebSocket server implementation are REQUIRED to respond with a PONG frame.
+; Once the configured number of PONG replies are missed, Asterisk will close the
+; connection and attempt to reconnect using the configured connection_timeout,
+; reconnect_interval and reconnect_attempts parameters.  Unlike TCP heepalives
+; however, Asterisk must create a scheduler thread and manage the process so it's
+; not without overhead, however small.
+;
+; In contrast to TCP keepalives, WebSocket PING/PONGs will traverse a proxy.
+;
+;enable_pingpongs     ; Enable Websocket PING/PONGs
```
mbradeen: Should this be an `= yes` ?

gtjoseph: Yes it should.

configs/samples/websocket_client.conf.sample
```diff
@@ -49,3 +49,109 @@
...
+; system will automatically close the TCP connection.  Asterisk will detect
+; that immediately and attempt to re-establish the connection using the configured
+; connection_timeout, reconnect_interval and reconnect_attempts parameters.
+;
+; However if the host server itself dies or packets are being dropped in the
+; network, Asterisk won't be able to detect this until the next time it tries
+; to send data.  To address this, there are two capabilities available:
+; TCP Keepalives and WebSocket PING/PONG.
+;
+; Which method you chose is entirely up to you and your environment.  Nothing
```
jcolp: choose

configs/samples/websocket_client.conf.sample
```diff
@@ -49,3 +49,109 @@
...
+; resources from Asterisk. The same can be said for the server side as replying to
+; the keepalives is also handled entirely by the operating system without any
+; involvement from application servers.  Once the configured number of replies are
+; missed, the kernel will close the socket which notifies Asterisk and causes an
+; attempt to reconnect using the configured connection_timeout, reconnect_interval
+; and reconnect_attempts parameters.
+;
+; The one downside to TCP keepalives is that they won't traverse a proxy so
+; you'll really only be testing if you can reach the proxy, not the websocket
+; server itself.
```
jcolp: I'd add a comment that if the proxy can be configured to do keepalives then this can still be a viable option.

configs/samples/websocket_client.conf.sample
```diff
@@ -49,3 +49,109 @@
...
+;
+;enable_tcp_keepalives = yes  ; Enable TCP keepalives on outgoing WebSockets.
+                              ; Default: no
+
+;tcp_keepalive_time = 20  ; Start sending keepalives when no data has been sent on
+                          ; the connection for this many seconds.
+                          ; Default: 20 seconds
+
+;tcp_keepalive_intvl = 20 ; Send keepalives at this interval in seconds.
```
jcolp: Let's not shorten things... _interval

configs/samples/websocket_client.conf.sample
```diff
@@ -49,3 +49,109 @@
...
+; reconnect_interval and reconnect_attempts parameters.  Unlike TCP heepalives
+; however, Asterisk must create a scheduler thread and manage the process so it's
+; not without overhead, however small.
+;
+; In contrast to TCP keepalives, WebSocket PING/PONGs will traverse a proxy.
+;
+;enable_pingpongs = yes ; Enable Websocket PING/PONGs
+                        ; Default: no
+
+;pingpong_intvl = 20    ; Send WebSocket PINGs at this interval in seconds.
```
jcolp: _interval

res/res_http_websocket.c
```diff
@@ -78,6 +79,42 @@
 #define MAX_WS_HDR_SZ 14
 #define MIN_WS_HDR_SZ 2
 
+#define WS_PING_PAYLOAD "WS_CLIENT_PING"
```
jcolp, line 87: Add a comment explaining why this uses a ping payload

res/res_http_websocket.c
```diff
@@ -319,23 +360,59 @@ static void websocket_mask_payload(struct ast_websocket *session, char *frame, c
...
+{
+	int enabled = ARE_PINGPONGS_ENABLED(session);
+	SCOPE_ENTER(2, "%s: Cancelling PING/PONG keepalives\n", WS_SESSION_REMOTE(session));
 
-/*! \brief Close function for websocket session */
-int AST_OPTIONAL_API_NAME(ast_websocket_close)(struct ast_websocket *session, uint16_t reason)
+	if (!enabled) {
+		SCOPE_EXIT_RTN("%s: Not enabled, cancel not needed\n", WS_SESSION_REMOTE(session));
+	}
+	AST_SCHED_DEL(ping_scheduler, session->client->ping_sched_timer);
```
jcolp, line 416: This should use AST_SCHED_DEL_UNREF

res/res_http_websocket.c
```diff
@@ -1526,21 +1849,85 @@ struct ast_websocket *AST_OPTIONAL_API_NAME(ast_websocket_client_create)
...
+		setsockopt(sockfd, IPPROTO_TCP, TCP_KEEPINTVL, &options->tcp_keepalive_intvl, sizeof(&options->tcp_keepalive_intvl));
+		setsockopt(sockfd, IPPROTO_TCP, TCP_KEEPCNT, &options->tcp_keepalive_probes, sizeof(&options->tcp_keepalive_probes));
+		ast_trace(-1, "%s: Enabled TCP keepalives\n", ws->client->options->uri);
+	}
+
+	if (ws->client->options->pingpongs) {
+		ws->client->ping_sched_timer = ast_sched_add(ping_scheduler, ws->client->options->pingpong_intvl * 1000,
+			ping_scheduler_callback, ao2_bump(ws));
+		if (ws->client->ping_sched_timer < 0) {
+			ast_log(LOG_WARNING, "%s: Unable to schedule PING/PONG keepalives\n", ws->client->options->uri);
```
jcolp, line 1994: Leak of ws reference

res/res_http_websocket.c
```diff
@@ -1526,21 +1849,85 @@ struct ast_websocket *AST_OPTIONAL_API_NAME(ast_websocket_client_create)
 	return ast_websocket_client_create_with_options(&options, result);
 }
 
+static int ping_scheduler_callback(const void *obj)
+{
+	struct ast_websocket *session = (struct ast_websocket *)obj;
```
jcolp, line 1884: How does the reference count handling for session work in this?

jcolp (dismissed):

configs/samples/websocket_client.conf.sample
gtjoseph: Done

configs/samples/websocket_client.conf.sample
gtjoseph: Done

configs/samples/websocket_client.conf.sample
gtjoseph: Done

configs/samples/websocket_client.conf.sample
gtjoseph: Done

res/res_http_websocket.c
gtjoseph: Done

res/res_http_websocket.c
gtjoseph: AST_SCHED_DEL_UNREF only runs the refcall if the delete succeeds.  This caused the session to leak in some circumstances.

res/res_http_websocket.c
gtjoseph: Done.

res/res_http_websocket.c
gtjoseph: The session was bumped when it was added to the scheduler so there should always be at least one reference to it when this callback runs.  If ping_scheduler_cancel() gets called while the callback is running, there's a check in AST_SCHED_DEL to delay a very short time to let the callback finish but I suppose it could still be running when AST_SCHED_DEL returns. However, I think the locking in websocket_close() and ping_scheduler_callback() prevents the last reference from being cleaned up.  This was one of the reasons for having ari_websockets.c and res_aeap/transport_websocket.c explicitly call ast_websocket_close() instead of just releasing the last reference and letting the session destructor close the websocket.

res/res_websocket_client.c
```diff
@@ -526,6 +696,26 @@ enum ast_ws_client_fields ast_websocket_client_get_field_diff(
...
+			changed |= AST_WS_CLIENT_FIELD_PROXY_HOST;
+		} else if (ast_strings_equal(v->name, "proxy_username")) {
+			changed |= AST_WS_CLIENT_FIELD_PROXY_USERNAME;
+		} else if (ast_strings_equal(v->name, "proxy_password")) {
+			changed |= AST_WS_CLIENT_FIELD_PROXY_PASSWORD;
+		} else if (ast_strings_equal(v->name, "enable_tcp_keepalives")) {
+			changed |= AST_WS_CLIENT_FIELD_TCP_KEEPALIVES;
+		} else if (ast_strings_equal(v->name, "tcp_keepalive_time")) {
+			changed |= AST_WS_CLIENT_FIELD_TCP_KEEPALIVE_TIME;
+		} else if (ast_strings_equal(v->name, "tcp_keepalive_intvl")) {
```
mbradeen: did this get changed to interval?

res/res_websocket_client.c
```diff
@@ -526,6 +696,26 @@ enum ast_ws_client_fields ast_websocket_client_get_field_diff(
...
+			changed |= AST_WS_CLIENT_FIELD_TCP_KEEPALIVE_INTERVAL;
+		} else if (ast_strings_equal(v->name, "tcp_keepalive_probes")) {
+			changed |= AST_WS_CLIENT_FIELD_TCP_KEEPALIVE_PROBES;
+		} else if (ast_strings_equal(v->name, "enable_pingpongs")) {
+			changed |= AST_WS_CLIENT_FIELD_PINGPONGS;
+		} else if (ast_strings_equal(v->name, "pingpong_intvl")) {
```
mbradeen: same as line 709

res/res_websocket_client.c
```diff
@@ -221,6 +221,128 @@ verify_server_hostname = no
...
+						<version>20.20.0</version>
+						<version>22.10.0</version>
+						<version>23.4.0</version>
+					</since>
+					<synopsis>Close the connection after this many missed PONG replies. (optional)</synopsis>
+					<description>
+						<para>
+							If a reply isn't received by the time the next PING is due
+							to be sent, it's considered missepingd.  The time to detect a failure
+							is therefore between (ping_cnt * ping_intvl) and
```
mbradeen: changed to interval?

configs/samples/websocket_client.conf.sample
```diff
@@ -49,3 +49,111 @@
...
+
+;pingpong_interval = 20 ; Send WebSocket PINGs at this interval in seconds.
+                        ; If a reply isn't received by the time the next PING is due
+                        ; to be sent, it's considered missed so this option also controls
+                        ; how long it takes to detect a failure.  See below.
+                        ; Default: 20 seconds
+
+;pingpong_probes = 3    ; Close the connection after this many missed PONG replies.
+                        ; The time to detect a failure is will be between
+                        ; (ping_cnt * ping_intvl) and ((ping_cnt + 1) * ping_intvl)
```
mbradeen: changed to _interval

res/res_http_websocket.c
```diff
@@ -319,23 +365,59 @@ static void websocket_mask_payload(struct ast_websocket *session, char *frame, c
...
+	}
+
+	if (force) {
+		ast_trace(-1, "%s: Forcing close. Handle: %p FD: %d\n", WS_SESSION_REMOTE(session),
+			session->stream, fd);
+		ast_iostream_close(session->stream);
+		session->stream = NULL;
+		ao2_unlock(session);
+		SCOPE_EXIT_RTN_VALUE(-1, "%s: Forced close\n", WS_SESSION_REMOTE(session));
+		return -1;
```
mbradeen: double return

mbradeen (dismissed):

res/res_websocket_client.c
gtjoseph: Yep, thought I got them all.

gtjoseph:
Target releases were updated to  20.21.0, 22.11.0, and 23.5.0.

res/res_http_websocket.c
jcolp: This is complicated enough that I think a large comment in the code explaining it is warranted

jcolp (dismissed):

res/res_http_websocket.c
gtjoseph: Yep.  In progress.  I also need to test a possible leak scenario.

gtjoseph: Much info added.

res/res_ari.c
```diff
@@ -1235,6 +1235,6 @@ AST_MODULE_INFO(ASTERISK_GPL_KEY, AST_MODFLAG_GLOBAL_SYMBOLS | AST_MODFLAG_LOAD_
 	.load = load_module,
 	.unload = unload_module,
 	.reload = reload_module,
-	.requires = "http,res_stasis,res_http_websocket,res_websocket_client",
+	.requires = "http,res_stasis,res_ari_model,res_http_websocket,res_websocket_client",
```
jcolp: You've got a separate change for res_ari_model to be optional and then in here it's requires. That seems conflicting.

gtjoseph: Yeah.  I had forgotten that I ran into this before vacation and added the dependency in this PR.  I ran into it yesterday on the base master branch which is what triggered the new PR.  I'll rebase this one off of new master when that one merges.

gtjoseph: Done

mbradeen (commented):
I think it's just the two option mis-names in the warning messages in wc_apply

gtjoseph:
> I think it's just the two option mis-names in the warning messages in wc_apply

I thought I fixed those. :(

gtjoseph:
I guess not.

res/res_websocket_client.c
```diff
@@ -221,6 +221,128 @@ verify_server_hostname = no
...
+						<version>20.21.0</version>
+						<version>22.11.0</version>
+						<version>23.5.0</version>
+					</since>
+					<synopsis>Close the connection after this many missed replies. (optional)</synopsis>
+					<description>
+						<para>
+							If a reply isn't received by the time the next keepalive is due
+							to be sent, it's considered missed.  The time to detect a failure
+							is therefore between (tcp_keepcnt * tcp_keepintvl) and
```
mbradeen: Sorry, found one more pair of intvl refs.  I think tcp_keepcnt was updated to tcp_keepalive_probes as well at some point unless I am reading it wrong.

gtjoseph: Done

mbradeen (approved):

jcolp (approved):

daninmadison:
I made a version using the 20.20.1 release and applied this patch (nothing more).

When using ARI, we are seeing asterisk crash and the gdb of the crash is showing it happening in res_http_websocket.c line 513

Either I'm applying the patch incorrectly (seeing an act match to the 1937 commit patch).

if (opcode == AST_WEBSOCKET_OPCODE_PONG) {
		/*
		 * If it's from our own PING, reset the missed count.
		 */
here---->		if (session->client->missed_pong_count
			&& payload_len == WS_PING_PAYLOAD_LEN

Not sure if some future patch fixed this problem.
Wanted to mention it because the crash means we can't test this fix.

jcolp:
It would be nice if you could provide the actual backtrace as well.

@gtjoseph I don't think there were any other fixes specifically, were there?

gtjoseph:
@daninmadison PR #2086 should fix the issue.  Please give it a try and report back in that PR.

daninmadison:
@jcolp Apologies, here is the backtrace showing it was that specific line.

#0  websocket_handled_pong_or_close (session=0x7feb2400c630)  res_http_websocket.c:513
#1  __ast_websocket_read (...)                                res_http_websocket.c:885
#2  session_read (session=0x7feb2400c210)                     ari/ari_websockets.c:224
#3  websocket_established_cb (ast_ws_session=0x7feb2400c630)  ari/ari_websockets.c:856
#4  __ast_websocket_uri_cb (...)                              res_http_websocket.c:1174
#5  ari_handle_websocket (ser=0x7feb28050bb0, uri="events")   ari/ari_websockets.c:276
#6  ast_ari_invoke (source=ARI_INVOKE_SOURCE_REST, uri="events", method=AST_HTTP_GET)  res_ari.c:784
#7  ast_ari_callback (ser=0x7feb28050bb0, uri="events")       res_ari.c:1154
#8  handle_uri (uri="events", method=AST_HTTP_GET)            http.c:1586
#9  httpd_process_request (ser=0x7feb28050bb0)                http.c:2072
#10 httpd_helper_thread (data=0x7feb28050bb0)                 http.c:2135
#11 dummy_start (...)                                         utils.c:1641
#12 start_thread (...)                                        nptl/pthread_create.c:442
#13 clone3 ()                                                 sysdeps/unix/sysv/linux/x86_64/clone3.S:81

@gtjoseph , I will make a version with your 2086 PR and get it to the testers.  Hopefully, this time they don't take 2+ months to get around to testing the original version I made for them when you wrote the original back in early June.

Have a good day!
Dan

daninmadison:
I posted on the PR 2086, but also posting here.

We ran a test with @gtjoseph new PR 2086 and it resolves the crash issue we were encountering.

Thank you for the expedited PR on this.

PR #1949, opened 2026-05-23
app_record: Fix hangup handling during beep playback

### LINKED ISSUE #1950: [bug]: app_record does not detect channel hangup during beep playback
### Severity

Minor

### Versions

master

### Components/Modules

apps/app_record

### Operating Environment

N/A - bug is in application source code, reproducible on any platform running Asterisk master branch.

### Frequency of Occurrence

Constant

### Issue Description

When a hangup occurs while `app_record` is playing the initial beep sound (via `ast_streamfile()` / `ast_waitstream()`), the application does not detect the hangup and continues running until the `maxduration` timeout expires.

In `apps/app_record.c`, after calling `ast_streamfile()` and `ast_waitstream()` for the beep, the return value of `ast_waitstream()` (which returns -1 on hangup) is not checked. As a result, the application proceeds to the recording loop and waits for the full timeout duration before exiting.

The fix is to check the return value of `ast_waitstream()` and, if it returns -1 (indicating hangup), set `RECORD_STATUS` to `HANGUP` and exit immediately via `goto out`.

**Steps to Reproduce:**
1. Configure a dialplan that uses `Record()` with the beep option
2. Call into the extension
3. Hang up during the beep playback (before recording starts)
4. Observe that Asterisk continues waiting for the full `maxduration` timeout instead of exiting immediately

**Expected:** app_record exits immediately with RECORD_STATUS=HANGUP
**Actual:** app_record waits for the full maxduration timeout

Related: PR #1790 (previous fix attempt), PR #1949 (current fix)

### Relevant log output

```shell
N/A
```

### Asterisk Issue Guidelines

UpBeta:
When a hangup occurs while app_record is playing the initial beep,
the application does not detect the hangup and continues running
until the maxduration timeout expires.

Replace the manual ast_streamfile() + ast_waitstream() sequence with
ast_stream_and_wait(), which properly detects hangup and returns
non-zero, allowing the application to exit immediately with
RECORD_STATUS set to HANGUP.

Resolves: #1950

### DIFF
--- a/apps/app_record.c
+++ b/apps/app_record.c
@@ -457,15 +457,11 @@ static int record_exec(struct ast_channel *chan, const char *data)
 	}
 
 	if (!ast_test_flag(&flags, OPTION_QUIET)) {
-		/* Some code to play a nice little beep to signify the start of the record operation */
-		res = ast_streamfile(chan, "beep", ast_channel_language(chan));
-		if (!res) {
-			res = ast_waitstream(chan, "");
-		} else {
-			ast_log(LOG_WARNING, "ast_streamfile(beep) failed on %s\n", ast_channel_name(chan));
-			res = 0;
+		/* Play a beep to signify the start of the record operation */
+		if (ast_stream_and_wait(chan, "beep", "")) {
+			status_response = "HANGUP";
+			goto out;
 		}
-		ast_stopstream(chan);
 	}
 
 	/* The end of beep code.  Now the recording starts */


InterLinked1:
#1790 is a PR, not an issue, so I don't know that you can resolve it. It doesn't look like there is an issue open for this either.

apps/app_record.c
```diff
@@ -511,6 +524,16 @@ static int record_exec(struct ast_channel *chan, const char *data)
 	if (maxduration <= 0)
 		maxduration = -1;
 
+	/* Safety net: check for hangup before entering the recording loop.
```
InterLinked1: Has testing shown that this block is actually necessary? I think that was the hangup on this before (no pun intended). With the `ast_check_hangup` above, it seems redundant to have additional `ast_check_hangup`'s that are probably not needed. I don't see any operations between the previous check and this one that would cause a hangup to get mixed, and the testing I did did not show this to be necessary, but would like to hear what you encountered.

UpBeta:
Thank you for the review, @InterLinked1.
Regarding the Resolves: #1790 — you're right, #1790 is a PR and not an issue. I'll create a proper issue to track this bug and update the reference accordingly.
Regarding the safety-net ast_check_hangup( ) before the recording loop: you raise a valid point. The primary motivation was to cover the case where the q (quiet) option is used and no beep is played at all — in that scenario, the first check (inside the OPTION_QUIET block) is never reached, so a hangup that occurs before the loop would go undetected.
However, if you feel this is an unlikely edge case and the code is cleaner without it, I'm happy to remove that block and keep only the ast_waitstream() return value handling. Let me know your preference and I'll update the PR.

InterLinked1:
> Thank you for the review, @InterLinked1. Regarding the Resolves: #1790 — you're right, #1790 is a PR and not an issue. I'll create a proper issue to track this bug and update the reference accordingly. Regarding the safety-net ast_check_hangup( ) before the recording loop: you raise a valid point. The primary motivation was to cover the case where the q (quiet) option is used and no beep is played at all — in that scenario, the first check (inside the OPTION_QUIET block) is never reached, so a hangup that occurs before the loop would go undetected.

Hang up from what, exactly? If the user hung up before `Record()` started running, the PBX core would disconnect the call before it got here.

> However, if you feel this is an unlikely edge case and the code is cleaner without it, I'm happy to remove that block and keep only the ast_waitstream() return value handling. Let me know your preference and I'll update the PR.

I'm not necessarily saying you have to remove it, but I'd like to understand what hangup scenario that guards against, before we start adding `ast_check_hangup` willy nilly all over the code "just in case". If it can't be explained, then I would prefer the last check not be added. Certainly, one of the checks is redundant with the other.

UpBeta:
> > Thank you for the review, @InterLinked1. Regarding the Resolves: #1790 — you're right, #1790 is a PR and not an issue. I'll create a proper issue to track this bug and update the reference accordingly. Regarding the safety-net ast_check_hangup( ) before the recording loop: you raise a valid point. The primary motivation was to cover the case where the q (quiet) option is used and no beep is played at all — in that scenario, the first check (inside the OPTION_QUIET block) is never reached, so a hangup that occurs before the loop would go undetected.
> 
> Hang up from what, exactly? If the user hung up before `Record()` started running, the PBX core would disconnect the call before it got here.
> 
> > However, if you feel this is an unlikely edge case and the code is cleaner without it, I'm happy to remove that block and keep only the ast_waitstream() return value handling. Let me know your preference and I'll update the PR.
> 
> I'm not necessarily saying you have to remove it, but I'd like to understand what hangup scenario that guards against, before we start adding `ast_check_hangup` willy nilly all over the code "just in case". If it can't be explained, then I would prefer the last check not be added. Certainly, one of the checks is redundant with the other.

You're right — the PBX core would handle the hangup before Record() is even invoked in that scenario. I'll remove the redundant safety-net check and keep only the ast_waitstream() return value handling, which is the actual fix for the issue. I'll push the update shortly.

UpBeta:
Thanks for the feedback @InterLinked1. I've made the following updates:

1. **Removed the redundant safety-net block** — you're right, the PBX core would handle the hangup before `Record()` is even invoked in that scenario. The fix now only checks the `ast_waitstream()` return value during beep playback.

2. **Created a proper Issue** — #1950 tracks this bug. I've updated the commit message to reference `Resolves: #1950` instead of the incorrect `#1790` (which was a PR, not an issue).

The updated commit keeps only the essential fix: checking `ast_waitstream()` return value (-1 = hangup during beep) and the `ast_check_hangup()` fallback when `ast_streamfile()` fails.

InterLinked1:
@UpBeta Your interactions smell heavily of AI as well, please disclose any AI usage in accordance with this project's policies as well.

UpBeta:
> @UpBeta Your interactions smell heavily of AI as well, please disclose any AI usage in accordance with this project's policies as well.

@InterLinked1 You're right, I used AI to help with the PR description and comments. The code fix itself is simple — checking the ast_waitstream() return value that was being ignored. I can own and debug it. Apologies for not disclosing upfront.

seanbright (changes requested):
Why not just change this whole block to the following?

```
if (ast_stream_and_wait(chan, "beep", "")) {
    status_response = "HANGUP";
    goto out;
}
```

UpBeta:
> Why not just change this whole block to the following?
> 
> ```
> if (ast_stream_and_wait(chan, "beep", "")) {
>     status_response = "HANGUP";
>     goto out;
> }
> ```

@seanbright Good call — much cleaner. Updated to use ast_stream_and_wait() as suggested. Force-pushed.

jcolp (approved):

seanbright (approved):

InterLinked1 (approved):

UpBeta:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seanbright:
@UpBeta the commit message still does not match the PR description. You should update the commit message and force push.

UpBeta:
Commit message updated to match PR description (force-pushed). The CI failure is on an unrelated PJSIP transfer test — could someone trigger a re-run? Thanks!

PR #1958, opened 2026-05-26
res_pjsip: Don't allow a leading period when wildcard matching

seanbright:
The reference identifier (what the client provides - in this case a hostname) must start with a domain label, not a `.`.

The current implementation will match `.seanbright.com` against `*.seanbright.com` which is incorrect.


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

mbradeen (approved):

PR #1967, opened 2026-06-02
Audiohooks: Allow whisper audiohooks to work without an underlying stream

### LINKED ISSUE #1966: [improvement]: Allow whisper audiohooks to work without an underlying audio stream
### Improvement Description

Audiohook whisper works by mixing it's audio with the current stream of frames being written to the hooked channel.  If there are no frames being sent to the channel outside of the audiohook, then the whisper audio is not heard on the hooked channel.  The improvment is to remove that requirement and allow whispered audio to be heard even when there is no underlying audio.

mbradeen:
Audiohook whispers work by mixing audio with the current stream of frames being
written to the hooked channel. Prior to this change if there were no audio
frames being written to the channel outside of the audiohook, then any whisper
audio has nothing to mix with and so is not heard on the hooked channel.

This change removes that requirement by creating a framehook driven timer on the
channel that creates silent frames the whispered frames can be mixed into.

The framehook is automatically added with the first attached whisper audiohook
and is removed when the last one is detached.  If there are no attached streams
being written to the channel, the timer is used to generate a default one. If
a stream is attached; the timer is ignored until the stream stops. This stop
and start is automatically triggered by the presence of a written stream and does
not require any external action.

Resolves: #1966

UserNote: There is no longer a requirement to play silence or otherwise have
a stream of audio being sent to a hooked channel for whispered frames to be
heard.

### DIFF
--- a/main/audiohook.c
+++ b/main/audiohook.c
@@ -39,17 +39,34 @@
 #include "asterisk/frame.h"
 #include "asterisk/translate.h"
 #include "asterisk/format_cache.h"
+#include "asterisk/framehook.h"
+#include "asterisk/timing.h"
 #include "asterisk/test.h"
 
 #define AST_AUDIOHOOK_SYNC_TOLERANCE 100 /*!< Tolerance in milliseconds for audiohooks synchronization */
 #define AST_AUDIOHOOK_SMALL_QUEUE_TOLERANCE 100 /*!< When small queue is enabled, this is the maximum amount of audio that can remain queued at a time. */
 #define AST_AUDIOHOOK_LONG_QUEUE_TOLERANCE 500 /*!< Otheriwise we still don't want the queue to grow indefinitely */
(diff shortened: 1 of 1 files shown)


mbradeen:
testsuite-test-pr: 139

mbradeen:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

InterLinked1:
This would be huge! This has been one of my long running peeves about Asterisk...

jcolp:
Broad testing would be good.

InterLinked1:
> Broad testing would be good.

I can do some testing later this month, but not for another 3 weeks. Not sure if that's worth waiting for here, but I'll give it a spin when I can.

jcolp:
@InterLinked1 Have you done any testing?

While I think this change is good, having various people and deployments test it would be ideal.

jcolp (approved):

Issue #1849, opened 2026-04-01
[improvement]: chan_local: Update references to chan_local for Local channels


InterLinked1:
### Improvement Description

The code references `chan_local` in a few places, which no longer exists since Local channels are built into the core. Update the outdated references so they're correct.

PR #1969, opened 2026-06-03
chan_local: Update chan_local references for Local channels.

jcolp:
chan_local no longer exists since Local channels are built into the core (core_local), but there are still comments which reference it, including in the configs. Update these to avoid confusion.

Resolves: https://github.com/asterisk/asterisk/issues/1849


jcolp:
cherry-pick-to: none

gtjoseph (approved):

PR #1973, opened 2026-06-04
app_queue.c: Index member device states to avoid scanning on every event

### LINKED ISSUE #1972: [bug]: app_queue: device_state_cb floods stasis/m:devicestate:all taskprocessor on reload of queues with many members
### Severity

Minor

### Versions

20.19.0

### Components/Modules

app_queue

### Operating Environment

Linux x86_64 (e.g. Ubuntu 24.04). Reproducible independent of OS.

### Frequency of Occurrence

Frequent

### Issue Description

When a queue with a large number of members (several thousand) is reloaded,
the single stasis subscription app_queue uses for device state
(stasis_subscribe(ast_device_state_topic_all(), device_state_cb) in
load_module) backs its dedicated taskprocessor "stasis/m:devicestate:all"
up past the 500-task high water mark.

Root cause is a combination of producer volume and per-message consumer cost:

* The reload publishes a cachable "Queue:<name>_pause_<iface>" device state
  per member (rt_handle_member_record / member_add_to_queue), i.e. thousands
  of device-state messages onto the devicestate:all topic in a tight burst.

* device_state_cb() handles each incoming device-state message by iterating
  ALL queues and, for each, ALL members (O(queues * members)). The reload's
  pause/avail hints match no member's state_interface, so every one of those
  messages forces a full scan that finds nothing.

* device_state_cb() also re-publishes "Queue:<name>_avail" device states,
  which forward back onto devicestate:all and re-enter the same callback,
  adding to the volume.

Because the single subscription serializer cannot drain faster than the
reload thread produces, the taskprocessor crosses its high water mark and
logs the warning below. Tasks are not dropped, but crossing the mark also
raises the process-global taskprocessor congestion alert
(ast_taskprocessor_alert_get). With PJSIP's default
taskprocessor_overload_trigger=global, a large reload can therefore briefly
cause PJSIP to shed inbound SIP (ignore UDP / 503), so on large deployments
the impact can be more than cosmetic.

The cost scales with member count; small queues do not show it.

Steps to reproduce:
1. Configure one or more queues with several thousand members (realtime
   queue_members reproduces it most reliably, since each reload re-publishes
   the per-member pause hint).
2. Trigger a reload: CLI "queue reload all" (or a realtime member reload).
3. Watch the devicestate taskprocessor depth during the reload:
   asterisk -rx "core show taskprocessor name <full devicestate:all name>"
  
Expected: the reload does not push the device-state taskprocessor near its
high water mark.

Actual: "stasis/m:devicestate:all" reaches 500 queued tasks and logs the
high water warning; "stasis" appears in
"core show taskprocessor alerted subsystems" for the duration.

### Relevant log output

```shell
message="2026-05-27 03:00:04.887239327  [May 27 03:00:04] WARNING[318927]: taskprocessor.c:1356 taskprocessor_push: Taskprocessor 'stasis/m:devicestate:all-00000009' queue reached 500 scheduled tasks (high water mark: 500)."

message="2026-05-27 03:00:04.887252486  [May 27 03:00:04] WARNING[318927]: taskprocessor.c:1356 taskprocessor_push: Taskprocessor 'stasis/m:devicestate:all-00000008' queue reached 500 scheduled tasks (high water mark: 500)."
```

### Asterisk Issue Guidelines

PujaGediya:
device_state_cb() iterated every queue and every member for each device
state message on the devicestate:all topic. Reloading a queue with
thousands of members floods that single subscription's taskprocessor
(stasis/m:devicestate:all) with the per-member pause/avail hints
app_queue publishes, tripping the 500 high water mark and raising the
global taskprocessor congestion alert.

Maintain a reference-counted index of the device-state identifiers that
queue members actually watch (via state_interface) and consult it at the
top of device_state_cb(). Device states no member watches are dropped in
O(1) instead of triggering an O(queues * members) scan, and the
Queue:..._avail hints the callback republishes no longer re-enter it.
Behavior for watched devices is unchanged.

Also fix a race in rt_handle_member_record(): when a realtime reload
changes a member's state_interface, start watching the new device before
storing it on the member (and before unwatching the old). Previously the
member was pointed at the new interface first and only then added to the
watcher set, leaving a brief window where m->state_interface referred to
a device not yet watched. Watching before publishing closes the window:
any device_state_cb() that passes the watch check then serializes on the
queue lock and observes the committed state_interface.

Resolves: #1972

UserNote: app_queue now handles device-state changes efficiently when
reloading queues with large member counts, avoiding a flood of the
stasis/m:devicestate:all taskprocessor past its high water mark.

Co-authored-by: Thomas <1258170+ThomasSevestre@users.noreply.github.com>

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -2800,6 +2800,135 @@ static int is_member_available(struct call_queue *q, struct member *mem)
 	return available;
 }
 
+/*!
+ * \internal
+ * \brief Set of device-state identifiers that at least one queue member is
+ *        watching through its state_interface (reference counted).
+ *
+ * device_state_cb() consults this set as a fast path: device-state messages for
+ * devices no member cares about are discarded in O(1) instead of forcing a full
+ * O(queues * members) scan.  This matters during a bulk member (re)load, which
+ * floods the devicestate:all topic with "Queue:..._pause_..." and
+ * "Queue:..._avail" hints app_queue itself publishes -- none of which match any
+ * member's state_interface.  As a side effect it also severs the feedback loop:
+ * the "Queue:..._avail" hints that device_state_cb republishes are not watched,
+ * so they are dropped here before they can re-enter the callback.
+ */
+static struct ao2_container *device_state_watchers;
+
+#define DEVICE_STATE_WATCHER_BUCKETS 547
+
+struct device_state_watcher {
+	int count;                          /*!< Number of members watching this device */
+	char interface[AST_CHANNEL_NAME];   /*!< Normalized device-state identifier */
+};
+
+/*!
+ * \internal
+ * \brief Compute the device-state identifier app_queue matches a member against.
+ *
+ * This MUST stay identical to the comparison performed in device_state_cb():
+ * for Local channels the trailing "/n" option is stripped so the member is
+ * matched by its base "Local/exten@context" device.
+ */
+static void queue_normalized_devstate_id(char *buf, size_t size, const char *state_interface)
+{
+	char *slash_pos;
+
+	ast_copy_string(buf, state_interface, size);
+	if ((slash_pos = strchr(buf, '/'))) {
+		if (!strncasecmp(buf, "Local/", 6) && (slash_pos = strchr(slash_pos + 1, '/'))) {
+			*slash_pos = '\0';
+		}
+	}
+}
+
+static int device_state_watcher_hash_fn(const void *obj, const int flags)
+{
+	const struct device_state_watcher *watcher = obj;
+	const char *interface = (flags & OBJ_KEY) ? obj : watcher->interface;
+
+	return ast_str_case_hash(interface);
+}
+
+static int device_state_watcher_cmp_fn(void *obj, void *arg, int flags)
+{
+	struct device_state_watcher *watcher = obj;
+	const char *interface = (flags & OBJ_KEY) ? arg : ((struct device_state_watcher *) arg)->interface;
+
+	return !strcasecmp(watcher->interface, interface) ? (CMP_MATCH | CMP_STOP) : 0;
+}
+
+/*! \internal \brief Start watching device state for a member's state_interface. */
+static void watch_state_interface(const char *state_interface)
+{
+	char interface[AST_CHANNEL_NAME];
+	struct device_state_watcher *watcher;
+
+	if (!device_state_watchers || ast_strlen_zero(state_interface)) {
+		return;
+	}
+
+	queue_normalized_devstate_id(interface, sizeof(interface), state_interface);
+
+	ao2_lock(device_state_watchers);
+	watcher = ao2_find(device_state_watchers, interface, OBJ_KEY | OBJ_NOLOCK);
+	if (watcher) {
+		++watcher->count;
+		ao2_ref(watcher, -1);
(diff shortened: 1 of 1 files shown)


jcolp:
Additionally, if AI was used in the creation of patches or issues its usage must be disclosed.

PujaGediya:
cherry-pick-to: 20

PujaGediya:
### AI usage disclosure

This contribution was prepared with AI assistance (Claude Code / Anthropic Claude). 
AI was used to:
- investigate the root cause in `device_state_cb()` and the device-state stasis path, and confirm the taskprocessor high-water behavior;
- draft the code change in `apps/app_queue.c` (the reference-counted `device_state_watchers` index and the `device_state_cb` fast-path early return);

I have reviewed the change in full and understand it, and I can debug and own it.
It was built against [master / 20.19] and tested on my test server a reload of a queue with
5000 members no longer pushes the `stasis/m:devicestate:all` taskprocessor past
its high water mark.

jcolp:
Instead of this approach, I wonder if instead app_queue should subscribe each queue member using a pool subscription to get the respective device or hint state. This would eliminate the processing of updates that app_queue is not interested in, and spread out updates across the pool instead of via a single thread.

PujaGediya:
Thanks for the suggestion. I want to make sure I understand the proposed approach correctly before pivoting.

Current state: the flood is caused by the reload publishing ~5000 `Queue:%s_pause_%s`
cachable hints onto `devicestate:all` synchronously. With the index, each of those events
hits an O(1) lookup and returns; no member watches those hint identifiers so they're
discarded immediately.

For the per-member pool subscription approach, a few questions:

1. **Overhead at scale**: 5000 members would mean 5000 `stasis_subscribe_pool()` calls
   creating 5000 ao2 objects and 5000 taskprocessors. Is that considered acceptable? The
   flip side is that `Queue:..._pause_...` hints would naturally not match any member's
   subscription topic, so the flood would be invisible at the stasis routing layer.

2. **Reload churn**: On a reload, would we unsubscribe all existing members and
   resubscribe? The reload path that originally caused the flood would instead do
   5000 unsubscribe + 5000 subscribe operations — each of those involves taskprocessor
   teardown/create. Is there a lighter-weight way to hand off the subscription across
   a member replacement (e.g. reuse the subscription if state_interface doesn't change)?

3. **Hint vs device topic**: Members can have a `state_interface` like `hint:Hint/1001` —
   should those subscribe to `ast_hint_topic()` and SIP members to `ast_device_state_topic()`?

Happy to rewrite to the suggestion if it's preferred. I just want to understand the expected
behaviour for the reload case before re-rolling.

jcolp:
1. I don't know the impact at that scale, it has to be tested out. The most heavy part would be if each taskprocessor had a thread, which using a pool it won't. I think the actual cost would be only a little more than the hash table approach.

2. I'm not that familiar with app_queue reload so I can't comment on that, but there's nothing from a stasis perspective that prohibits keeping the subscription if it's unchanged.

3. I would leave hints untouched for now until the impact of the change is understood.

PujaGediya:
@jcolp I've opened #2029 implementing the per-device pool subscription approach you suggested here, in place of the index-based approach in this PR.
Happy to close this PR in favor of #2029 whenever you'd prefer.

jcolp:
Closed this out per prior comment.

PR #1974, opened 2026-06-04
app_queue: Avoid full scans for irrelevant device states

seifzadeh:
Summary:
- Add a fast relevance check before app_queue scans all queues and members in device_state_cb().
- Avoid expensive no-op scans for device-state messages that are not used by any queue member state_interface.
- Reduce devicestate Stasis taskprocessor backlog during large queue reloads.

Reproduction:
- Generated a queue with several thousand members.
- Ran "queue reload all".
- Before the change, stasis/m:devicestate:all reached the 500 high-water mark.

Testing:
- Built app_queue successfully.
- Verified normal PJSIP endpoint registration.
- Verified a caller entering a queue.
- Verified queue member ringing and answering.
- Verified queue member state with "queue show".
- Verified pause/unpause behavior.
- Re-ran the large queue reload reproduction.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -2058,6 +2058,164 @@ static AST_LIST_HEAD_STATIC(rule_lists, rule_list);
 
 static struct ao2_container *queues;
 
+struct queue_member_state_interface {
+	char interface[AST_CHANNEL_NAME];
+	unsigned int count;
+};
+
+AO2_STRING_FIELD_CASE_HASH_FN(queue_member_state_interface, interface)
+AO2_STRING_FIELD_CASE_CMP_FN(queue_member_state_interface, interface)
+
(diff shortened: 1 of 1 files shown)


jcolp:
If AI was used for the creation of this then it must be disclosed.

seifzadeh:
Thanks. I reviewed #1973 and agree that both PRs are solving the same problem with the same general approach: keeping a reference-counted set of watched queue member state_interface values and returning early from device_state_cb() for unrelated device-state messages.

#1973 appears to be further along: CLA is complete, CI has passed, and the implementation also documents the realtime state_interface update ordering/race consideration.

To avoid duplicate review work, I am happy to close this PR and continue by contributing my reproduction and functional testing results to #1973.

PR #1977, opened 2026-06-06
res_pjsip_session: Bound delayed BYE behind UAC INVITE

### LINKED ISSUE #1965: [bug]: Memory leak when receiving malformed 200 OK in response to re-INVITE
### Severity

Minor

### Versions

23.3.0

### Components/Modules

PJSIP

### Operating Environment

OS:

```
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"
NAME="Debian GNU/Linux"
VERSION_ID="12"
VERSION="12 (bookworm)"
VERSION_CODENAME=bookworm
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

Kernel:

```
Linux 7e904979794b 6.8.0-100-generic #100-Ubuntu SMP PREEMPT_DYNAMIC Tue Jan 13 16:39:21 UTC 2026 aarch64 GNU/Linux
```

### Frequency of Occurrence

Constant

### Issue Description

When Asterisk sends a re-INVITE (e.g. for direct media) and the UAS replies with a malformed response (i.e.  missing a required header), PJSIP returns ``PJSIP_EMISSINGHDR`` and the associated transaction/dialog memory appears to never be freed. Memory keeps growing with each call that sees such a malformed response.

Steps to Reproduce:

1. Have two PJSIP endpoints for a "caller" and "callee", both with ``direct_media = yes``.
2. Caller places a call through Asterisk to callee, by entering a dialplan that immediately dials the callee using ``Dial``.
3. Asterisk bridges the call and sends a re-INVITE to the callee for direct media.
4. Callee responds to the re-INVITE with ``100 Trying`` and then a ``200 OK``. The latter missing the Via header and Contact header (the response is valid otherwise)
5. Call is torn down normally (``BYE`` + ``200 OK``)
6. Repeat this a few times and watch Asterisk's memory usage grow.

Excerpt from ``memory show summary`` (with ``MALLOC_DEBUG`` enabled) before doing 50 calls:

```
  26414880 bytes in        179 allocations in file ../src/pj/pool_policy_malloc.c
```

And after the calls have completed:

```
  41704800 bytes in       1843 allocations in file ../src/pj/pool_policy_malloc.c
```

### Relevant log output

```
[Jun  2 13:55:59] ERROR[40]: pjproject: <?>:                sip_endpoint.c Error processing packet from 172.20.0.30:5060: Missing required header(s) (PJSIP_EMISSINGHDR)  Via [code 171050]:
SIP/2.0 200 OK
From: "Caller" <sip:caller@172.20.0.10>;tag=25e1a2a3-730c-42a0-ad7c-a2094b54169b
To: <sip:172.20.0.30>;tag=1SIPpTag0129
Call-ID: 6cebdabd-b035-40b1-867f-02c30ae398fa
CSeq: 5791 INVITE
Content-Type: application/sdp
Content-Length:   133
asterisk  |
v=0
o=callee 1 1 IN IP4 172.20.0.30
s=call
c=IN IP4 172.20.0.30
t=0 0
m=audio 6112 RTP/AVP 0
a=rtpmap:0 PCMU/8000
a=sendrecv
asterisk  |
-- end of packet.
```

### Asterisk Issue Guidelines

seifzadeh:
When a confirmed session is being terminated while an outgoing in-dialog
INVITE transaction is still outstanding, the BYE is delayed until the
outstanding transaction terminates.

If that INVITE has already received a provisional response and the final
response is malformed and rejected before transaction processing, the
transaction can remain outstanding and the delayed BYE can keep the
session, media state, RTP instance, and PJPROJECT pools referenced after
the channels are gone.

When a BYE is delayed behind an outstanding UAC INVITE, set a PJPROJECT
transaction timeout on that INVITE so the delayed cleanup path has a
bounded wait. If PJPROJECT terminates the dialog as a result of the
timeout, discard the delayed BYE instead of sending a duplicate BYE.

Fixes: #1965

### DIFF
--- a/res/res_pjsip_session.c
+++ b/res/res_pjsip_session.c
@@ -1394,6 +1394,19 @@ static void delayed_request_free(struct ast_sip_session_delayed_request *delay)
 	ast_free(delay);
 }
 
+/*
+ * Delayed requests own pending/active media states. Keep cleanup centralized so
+ * timeout-driven teardown releases the same state as normal session teardown.
+ */
+static void flush_delayed_requests(struct ast_sip_session *session)
+{
+	struct ast_sip_session_delayed_request *delay;
+
+	while ((delay = AST_LIST_REMOVE_HEAD(&session->delayed_requests, next))) {
+		delayed_request_free(delay);
+	}
+}
+
 /*!
  * \internal
  * \brief Send a delayed request
@@ -1429,6 +1442,8 @@ static int send_delayed_request(struct ast_sip_session *session, struct ast_sip_
 		delay->active_media_state = NULL;
 		SCOPE_EXIT_RTN_VALUE(res, "%s\n", ast_sip_session_get_name(session));
 	case DELAYED_METHOD_BYE:
+		/* The delayed BYE is being sent now; timeout cleanup is no longer needed. */
+		session->terminate_on_invite_timeout = 0;
 		ast_sip_session_terminate(session, 0);
 		SCOPE_EXIT_RTN_VALUE(0, "%s: Terminating session on delayed BYE\n", ast_sip_session_get_name(session));
 	}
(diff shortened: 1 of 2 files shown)


seifzadeh:
Testing performed:

- Reproduced the issue with a one-call SIPp scenario.
- Confirmed the call was fully cleared after the wait period:
  - 0 active channels
  - 0 active calls
  - 1 call processed
- Confirmed the malformed re-INVITE final response triggered:
  - PJSIP_EMISSINGHDR
  - Missing required header(s) Via
- Confirmed the leak before the fix:
  - PJ pool grew from 26,511,904 bytes / 203 allocations
  - to 26,863,552 bytes / 248 allocations
  - delta: +351,648 bytes / +45 allocations
- Confirmed the retained allocations included SIP session, delayed request,
  media state, RTP instance, session supplements, and PJPROJECT pools.
- Re-tested after the fix and confirmed the stuck callee-side SIP session is
  destroyed and the repeatable per-call PJ pool growth no longer remains.
- Ran git diff --check.
- Built res/res_pjsip_session.so.

AI assistance disclosure:

I used AI assistance to organize the reproduction notes, summarize the memory
evidence, and review wording for the PR description/comment. I personally
reproduced the issue, reviewed the code change, built and tested the modified
module, verified the before and after behavior, and take responsibility for the
submitted change.

seifzadeh:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

res/res_pjsip_session.c
```diff
@@ -1629,6 +1644,57 @@ static int delay_request(struct ast_sip_session *session,
...
+		return 0;
+	}
+
+	tsx = session->inv_session->invite_tsx;
+	if (tsx->role != PJSIP_ROLE_UAC || tsx->method.id != PJSIP_INVITE_METHOD
+		|| tsx->state >= PJSIP_TSX_STATE_COMPLETED) {
+		return 0;
+	}
+
+	status = pjsip_tsx_set_timeout(tsx, 64 * pjsip_cfg()->tsx.t1);
```
gtjoseph: Rather than forcing the timeout to be `64 * pjsip_cfg()->tsx.t1`, I think just using `pjsip_cfg()->tsx.td` would be better.  This is set from the `timer_b` parameter in the pjsip.conf system section and defaults to `64 * pjsip_cfg()->tsx.t1`.

gtjoseph (dismissed):

gtjoseph:
@seifzadeh @jcolp pointed something out to me just now...  Is it possible that pjproject is already doing this but we're either missing the callback or the state change that would tell us to clean up in this case?

seifzadeh:
> @seifzadeh @jcolp pointed something out to me just now... Is it possible that pjproject is already doing this but we're either missing the callback or the state change that would tell us to clean up in this case?

Thanks, I’ll check that path.

Of course, I had a brief look at pjproject itself before reviewing it. Now, by putting more logs in the transaction state callbacks, it might be done through pjproject itself.

seifzadeh:
> @seifzadeh @jcolp pointed something out to me just now... Is it possible that pjproject is already doing this but we're either missing the callback or the state change that would tell us to clean up in this case?

I checked this against vanilla Asterisk before applying my patch.

I added log after session variable:
- session_on_rx_response()
- session_on_tsx_state()
- session_inv_on_tsx_state_changed()

For normal INVITE responses, the log confirms that the callbacks are reached and the transactions transition to Completed/Terminated as expected.

For the malformed final response case, the affected UAC INVITE transaction is visible before the malformed response in Proceeding/status 100. PJPROJECT then logs:

Missing required header(s) (PJSIP_EMISSINGHDR) Via

for a SIP/2.0 200 OK with CSeq INVITE.

After that validation error, I do not see session_on_rx_response() for that malformed final response, and I do not see the affected UAC INVITE transaction transition to Completed or Terminated through session_inv_on_tsx_state_changed() or session_on_tsx_state(). It remains visible only in Proceeding/status 100.

The delayed BYE path is reproduced in the same run with:

Delay sending BYE because of outstanding transaction...

Also, this cannot be a bug in pjsproject, because transaction management functionality is already included in it and this should be managed on the asterisk side.
[full.txt](https://github.com/user-attachments/files/28749432/full.txt)
[res_pjsip_session.c](https://github.com/user-attachments/files/28749467/res_pjsip_session.c)


So from what I can see, PJPROJECT rejects the malformed final response before Asterisk receives a usable response callback or transaction state transition for cleanup. I’ll still update the timeout value to use pjsip_cfg()->tsx.td as suggested.

gtjoseph:
Thanks for double-checking.

seifzadeh:
Updated. I changed the timeout value to use pjsip_cfg()->tsx.td instead of open-coding 64 * pjsip_cfg()->tsx.t1, so it now follows the configured PJSIP transaction timeout value.

gtjoseph (approved):

jcolp (approved):

PR #1982, opened 2026-06-10
app_mixmonitor: Fix duplex recording for non 8K codecs

### LINKED ISSUE #1779: [bug]: MixMonitor with flag D produces garbage in a 16KHz bridge
### Severity

Minor

### Versions

22.8.2

### Components/Modules

apps/app_mixmonitor.ch

### Operating Environment

Linux

### Frequency of Occurrence

None

### Issue Description

When using MixMonitor with a .raw file and the Flags Db, the output is garbage as soon as the bridge is operating with 16KHz (for example with a G722 codec). I have created a workaround (see attached .txt, what in fact is a diff, however adding a .diff is not supported), but I won't create a pull request with this code as my solution is quite naive (downsampling by simply discarding every second sample)... Have a look at the changes in the diff wich produces a "usable" output for me. But proper downsampling would be a better solution, so someone who knows better how the audiohook and ast_writefile works should probably create a better solution.

[mix_monitor.txt](https://github.com/user-attachments/files/25388795/mix_monitor.txt)

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

mbradeen:
The native sampling of duplex recording is set to match the raw 8K
output format. If one or more of the streams being recorded is above
8K, the frame size coming into the mixmonitor is too large and needs
to be translated to 8K before being mixed into the stereo frame to
avoid garbled and mistimed audio

Fixes: #1779

### DIFF
--- a/apps/app_mixmonitor.c
+++ b/apps/app_mixmonitor.c
@@ -60,6 +60,7 @@
 #include "asterisk/mixmonitor.h"
 #include "asterisk/format_cache.h"
 #include "asterisk/beep.h"
+#include "asterisk/translate.h"
 
 /*** DOCUMENTATION
 	<application name="MixMonitor" language="en_US">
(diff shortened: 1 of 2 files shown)


mbradeen:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

mbradeen:
testsuite-test-pr: 142

apps/app_mixmonitor.c
```diff
@@ -875,6 +881,8 @@ static void *mixmonitor_thread(void *obj)
 			if (ast_test_flag(mixmonitor, MUXFLAG_INTERLEAVED)) {
 				/* The 'D' option is set, so mix the frame as an interleaved dual channel frame */
 				int i;
+				/* We are fed by a call to ast_audiohook_read_frame_all that uses the fixed SAMPLES_PER_FRAME,
```
jcolp: Update the API documentation for ast_audiohook_read_frame_all so it is explicitly stated that fr_read and fr_write may be in a different format than the one specified to the function call

apps/app_mixmonitor.c
```diff
@@ -883,20 +891,107 @@ static void *mixmonitor_thread(void *obj)
...
 
-				if (fr_read) {
-					memcpy(read_buf, fr_read->data.ptr, sizeof(read_buf));
+				/*
+				 * Depending on the input codec's rate (which may change during the call) we may get frames in
+				 * a slin format that does not match the native format_slin of the mixmonitor.  In that case
+				 * we need to translate.
+				 */
+				 if (fr_read) {
+					/* We only need to worry about translating if the frame format doesn't match the native. */
```
jcolp: Clarify further here - doesn't match the format expected by the file format we are writing

apps/app_mixmonitor.c
```diff
@@ -883,20 +891,107 @@ static void *mixmonitor_thread(void *obj)
...
+						if (ast_format_cmp(last_format_read, fr_read->subclass.format) == AST_FORMAT_CMP_NOT_EQUAL ||
+							!trans_pvt_read) {
+							ast_debug(3, "Read frame format changed from %s to %s, building translator path to %s\n",
+								last_format_read ? ast_format_get_name(last_format_read) : "none",
+								ast_format_get_name(fr_read->subclass.format),
+								ast_format_get_name(format_slin));
+							if (trans_pvt_read) {
+								ast_translator_free_path(trans_pvt_read);
+							}
+							trans_pvt_read = ast_translator_build_path(format_slin, fr_read->subclass.format);
```
jcolp: This could fail, if it does then the most that can be done is silence until the next round where it can attempt to build again.

apps/app_mixmonitor.c
```diff
@@ -883,20 +891,107 @@ static void *mixmonitor_thread(void *obj)
...
+						}
+					}
+					last_format_read = fr_read->subclass.format;
 				} else {
 					memset(read_buf, 0, sizeof(read_buf));
 				}
 
 				if (fr_write) {
-					memcpy(write_buf, fr_write->data.ptr, sizeof(write_buf));
+					/* We only need to worry about translating if the frame format doesn't match the native. */
```
jcolp: Same here

apps/app_mixmonitor.c
```diff
@@ -883,20 +891,107 @@ static void *mixmonitor_thread(void *obj)
...
+						if (ast_format_cmp(last_format_write, fr_write->subclass.format) == AST_FORMAT_CMP_NOT_EQUAL ||
+							!trans_pvt_write) {
+							ast_debug(3, "Write frame format changed from %s to %s, building translator path to %s\n",
+								last_format_write ? ast_format_get_name(last_format_write) : "none",
+								ast_format_get_name(fr_write->subclass.format),
+								ast_format_get_name(format_slin));
+							if (trans_pvt_write) {
+								ast_translator_free_path(trans_pvt_write);
+							}
+							trans_pvt_write = ast_translator_build_path(format_slin, fr_write->subclass.format);
```
jcolp: Same here

jcolp (changes requested):

jcolp (approved):

apps/app_mixmonitor.c
```diff
@@ -883,20 +894,127 @@ static void *mixmonitor_thread(void *obj)
...
 				}
 
-				if (fr_read) {
-					memcpy(read_buf, fr_read->data.ptr, sizeof(read_buf));
+				/*
+				 * Depending on the input codec's rate (which may change during the call) we may get frames in
+				 * a slin format that does not match the native format_slin of the mixmonitor.  In that case
+				 * we need to translate.
+				 */
+				 if (fr_read) {
```
seanbright: This block and the `if (fr_write) {` block at 966 are structurally identical. Could this be de-duplicated?

mbradeen: moved to a helper

jcolp (approved):

seanbright (approved):

PR #1983, opened 2026-06-10
res_pjsip_registrar: Resolve AOR for identified endpoints

### LINKED ISSUE #1398: [bug]:  AOR matching for inbound registration not working with domain involved and identifying by contact header
### Severity

Minor

### Versions

22.3.0

### Components/Modules

res_pjsip_endpoint_identifier_ip.so

### Operating Environment

Ubuntu Noble

### Frequency of Occurrence

Constant

### Issue Description

I've been trying to match an endpoint using the contact header field and a regular expression.
Registration kept failing so I assumed my regexp was wrong  but then discovered that INVITE messages are correctly matched to the endpoint so I can make calls.

I then discovered an old email thread about this but cant find it as a current known issue in this database.
https://lists.digium.com/pipermail/asterisk-bugs/2022-June/219151.html

On closer inspection I can see that the endpoint must be matched for registration because it challenges with a realm specific to the matched endpoint.
But registration fails with:
 WARNING[650]: res_pjsip_registrar.c:1239 find_registrar_aor: AOR '' not found for endpoint 'Student2@mysipdom' (192.0.2.20:58931)

I am not using WEBRTC - I'm using regular SIP over UDP

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

seifzadeh:
When an inbound REGISTER is matched to an endpoint using an
identification method such as IP, header, or request URI, the endpoint can
be identified and authenticated successfully while registrar AOR
resolution still fails.

find_registrar_aor() only resolved AOR names for username and
auth_username identify methods. A header-only endpoint therefore left
aor_name unset and failed registration with "AOR '' not found" even when
the REGISTER To URI matched one of the endpoint's configured AORs.

Resolve the REGISTER AOR from the To URI for endpoint identification
methods that do not directly provide an AOR name, while still
constraining the match to the already-identified endpoint's configured AOR
list.

Fixes: #1398

### DIFF
--- a/res/res_pjsip_registrar.c
+++ b/res/res_pjsip_registrar.c
@@ -1179,55 +1179,54 @@ static struct ast_sip_aor *find_registrar_aor(struct pjsip_rx_data *rdata, struc
 {
 	struct ast_sip_aor *aor = NULL;
 	char *aor_name = NULL;
+	pjsip_sip_uri *uri;
+	pj_str_t username;
 	int i;
 
-	for (i = 0; i < AST_VECTOR_SIZE(&endpoint->ident_method_order); ++i) {
-		pj_str_t username;
-		pjsip_sip_uri *uri;
-		pjsip_authorization_hdr *header = NULL;
+	/*
+	 * RFC 3261: The To header contains the Address-of-Record
+	 * whose registration is being created, queried, or modified.
+	 */
+	uri = pjsip_uri_get_uri(rdata->msg_info.to->uri);
+	pj_strassign(&username, &uri->user);
 
-		switch (AST_VECTOR_GET(&endpoint->ident_method_order, i)) {
-		case AST_SIP_ENDPOINT_IDENTIFY_BY_USERNAME:
-			uri = pjsip_uri_get_uri(rdata->msg_info.to->uri);
+	if (ast_sip_get_ignore_uri_user_options()) {
+		pj_ssize_t semi = pj_strcspn2(&username, ";");
 
-			pj_strassign(&username, &uri->user);
-
-			/*
-			 * We may want to match without any user options getting
-			 * in the way.
-			 *
-			 * Logic adapted from AST_SIP_USER_OPTIONS_TRUNCATE_CHECK for pj_str_t.
-			 */
-			if (ast_sip_get_ignore_uri_user_options()) {
-				pj_ssize_t semi = pj_strcspn2(&username, ";");
-				if (semi < pj_strlen(&username)) {
-					username.slen = semi;
-				}
-			}
-
-			aor_name = find_aor_name(&username, &uri->host, endpoint->aors);
-			if (aor_name) {
-				ast_debug(3, "Matched aor '%s' by To username\n", aor_name);
-			}
-			break;
-		case AST_SIP_ENDPOINT_IDENTIFY_BY_AUTH_USERNAME:
-			while ((header = pjsip_msg_find_hdr(rdata->msg_info.msg, PJSIP_H_AUTHORIZATION,
-				header ? header->next : NULL))) {
-				if (header && !pj_stricmp2(&header->scheme, "digest")) {
-					aor_name = find_aor_name(&header->credential.digest.username,
-						&header->credential.digest.realm, endpoint->aors);
-					if (aor_name) {
-						ast_debug(3, "Matched aor '%s' by Authentication username\n", aor_name);
-						break;
-					}
-				}
-			}
-			break;
-		default:
-			continue;
+		if (semi < pj_strlen(&username)) {
+			username.slen = semi;
 		}
+	}
 
-		if (aor_name) {
+	aor_name = find_aor_name(&username, &uri->host, endpoint->aors);
+	if (aor_name) {
+		ast_debug(3, "Matched aor '%s' by REGISTER To URI\n", aor_name);
+	}
+
+	if (!aor_name) {
+		for (i = 0; i < AST_VECTOR_SIZE(&endpoint->ident_method_order); ++i) {
+			pjsip_authorization_hdr *header = NULL;
+
+			if (AST_VECTOR_GET(&endpoint->ident_method_order, i)
+				!= AST_SIP_ENDPOINT_IDENTIFY_BY_AUTH_USERNAME) {
+				continue;
+			}
+
+			while ((header = pjsip_msg_find_hdr(rdata->msg_info.msg,
+				PJSIP_H_AUTHORIZATION, header ? header->next : NULL))) {
+				if (pj_stricmp2(&header->scheme, "digest")) {
+					continue;
+				}
+
+				aor_name = find_aor_name(&header->credential.digest.username,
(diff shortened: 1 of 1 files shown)


seifzadeh:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seifzadeh:
Testing performed:

Reproduced the original failure on current master using a Cisco
CP-8841-3PCC registering over UDP.

Environment:
- Asterisk: GIT-master-16bbc5e0d0
- Phone: Cisco CP-8841-3PCC/12.0.1
- Phone IP: 192.168.17.113
- Asterisk IP: 192.168.17.141
- Test domain: mysipdom

Original reproduction:

[Student2@mysipdom]
type=identify
endpoint=Student2@mysipdom
match_header=contact: /.*sip:Student2@192\.168\.17\.113:.*/

[Student2@mysipdom]
type=endpoint
identify_by=header
auth=Student2@mysipdom
aors=Student2@mysipdom
context=issue1398-internal
disallow=all
allow=ulaw,alaw
direct_media=no

[Student2@mysipdom]
type=auth
auth_type=userpass
username=Student2
password=sessame
realm=mysipdom

[Student2@mysipdom]
type=aor
support_path=yes
max_contacts=1
remove_existing=yes

The phone sent a REGISTER with:

REGISTER sip:mysipdom SIP/2.0
From: "Shiela" <sip:Student2@mysipdom>
To: "Shiela" <sip:Student2@mysipdom>
Contact: "Shiela" <sip:Student2@192.168.17.113:5061>;expires=300

Before the change:
- Asterisk identified the endpoint and challenged with
  realm="mysipdom".
- The second REGISTER included valid Authorization.
- Digest authentication succeeded.
- Registration still failed with 404 and:
  AOR '' not found for endpoint 'Student2@mysipdom'.

Temporary debug logging confirmed that find_registrar_aor() saw:

endpoint='Student2@mysipdom'
endpoint_aors='Student2@mysipdom'
methods=1
method[0]=8
final resolved aor_name='<null>'

The enum value 8 is AST_SIP_ENDPOINT_IDENTIFY_BY_HEADER, confirming that
header endpoint identification succeeded but registrar AOR resolution had
no path for that identify method.

After the change:
- identify_by=header registration succeeds.
- The REGISTER To URI is resolved to the configured endpoint AOR.
- The contact is stored under the correct AOR.
- A REGISTER using a To URI outside endpoint->aors still fails.

Regression tests:
- identify_by=username: PASS.
- identify_by=auth_username: PASS.
- identify_by=auth_username,header: PASS.
- Normal endpoint registration without a domain: PASS.
- Domain endpoint registration with mysipdom: PASS.

Build checks:
- git diff --check: PASS.
- make -j$(nproc) res/res_pjsip_registrar.so: PASS.

seifzadeh:
AI assistance disclosure: I used AI assistance to help organize debugging notes, testing summaries, and review the proposed patch structure. I reproduced the issue locally, added temporary debug logging, reviewed the code path in res_pjsip_registrar.c, built the change, and verified the registration behavior myself. use chatgpt 5.5

res/res_pjsip_registrar.c
```diff
@@ -1175,37 +1175,58 @@ static char *find_aor_name(const pj_str_t *pj_username, const pj_str_t *pj_domai
...
+ * Endpoint identification methods such as IP, header, and request URI select
+ * an endpoint but do not provide an AOR name. Use the REGISTER To URI to select
+ * the requested AOR, constrained to the AORs configured on that endpoint.
+ */
+static char *find_aor_name_from_to_uri(struct pjsip_rx_data *rdata, const char *endpoint_aors)
+{
+	pj_str_t username;
+	pjsip_sip_uri *uri;
+
+	if (!rdata || !rdata->msg_info.to || !rdata->msg_info.to->uri) {
```
jcolp: All of this is guaranteed to exist, so this check is not needed

res/res_pjsip_registrar.c
```diff
@@ -1175,37 +1175,58 @@ static char *find_aor_name(const pj_str_t *pj_username, const pj_str_t *pj_domai
...
 	int i;
 
 	for (i = 0; i < AST_VECTOR_SIZE(&endpoint->ident_method_order); ++i) {
-		pj_str_t username;
-		pjsip_sip_uri *uri;
 		pjsip_authorization_hdr *header = NULL;
 
 		switch (AST_VECTOR_GET(&endpoint->ident_method_order, i)) {
 		case AST_SIP_ENDPOINT_IDENTIFY_BY_USERNAME:
-			uri = pjsip_uri_get_uri(rdata->msg_info.to->uri);
```
jcolp, line 1191: Why did this need to be broken out? All of these newly added methods match based on the same thing, and the 2 log messages don't need to be different.

res/res_pjsip_registrar.c
```diff
@@ -1223,6 +1244,14 @@ static struct ast_sip_aor *find_registrar_aor(struct pjsip_rx_data *rdata, struc
...
 			break;
+		case AST_SIP_ENDPOINT_IDENTIFY_BY_IP:
+		case AST_SIP_ENDPOINT_IDENTIFY_BY_HEADER:
+		case AST_SIP_ENDPOINT_IDENTIFY_BY_REQUEST_URI:
+			aor_name = find_aor_name_from_to_uri(rdata, endpoint->aors);
+			if (aor_name) {
+				ast_debug(3, "Matched aor '%s' by To URI\n", aor_name);
+			}
+			break;
 		default:
```
jcolp: Did you do any research or investigation as to why this logic was the way it was in the first place - not accepting a REGISTER essentially.

jcolp (changes requested):

res/res_pjsip_registrar.c
seifzadeh: Thanks for your advice
I am reviewing again
Also, by studying the codes before res_pjsip_registrar.c, I am suspecting that checking the aor based on IP may not be necessary at all. And simply adding the header section is sufficient. Also, to check the header, simply adding AST_SIP_ENDPOINT_IDENTIFY_BY_HEADER
to the first case itself will solve the problem. But I need to investigate the cases that will cause segfault more. And maybe this is the reason for not checking the header in this section.

seifzadeh:
rfc-3261 sections 10.2 and 10.3 define the To URI as the AOR whose registration is being created, queried, or modified, and state that the registrar extracts the AOR from the To header.

Based on that, the updated implementation now:

- Resolves the AOR from the REGISTER To URI first.
- Restricts the result to the AORs configured on the already identified endpoint.
- Retains Asterisk's auth_username lookup as a fallback only when auth_username is configured(identify_by) and the To URI did not match.
- Removes the unnecessary null and URI checks, since those values are guaranteed by the caller.
- Removes the separate helper and duplicate matching paths.

I also reproduced the original header-identified REGISTER failure and verified the updated behavior, including username, auth_username, combined auth_username/header, domain, and non-domain registration cases.

jcolp (approved):

PR #1984, opened 2026-06-12
manager: Move away from shared linked list for events.

jcolp:
This change moves manager away from a shared linked list for events to a per-session vector of pending events. This allows early filtering of events which eliminates the need to wake up threads for events that they have no interest in.

TCP based connections have also been moved to an alert pipe based wakeup mechanism to reduce locking and simplify usage.

UserNote: The "manager show eventq" CLI command has been removed as there is no longer a single event queue to display.

### DIFF
--- a/main/manager.c
+++ b/main/manager.c
@@ -127,36 +127,24 @@ enum add_filter_result {
 	FILTER_FORMAT_ERROR,
 };
 
-/*!
- * Linked list of events.
- * Global events are appended to the list by append_event().
- * The usecount is the number of stored pointers to the element,
- * excluding the list pointers. So an element that is only in
- * the list has a usecount of 0, not 1.
+/*! \brief An AMI event message
  *
- * Clients have a pointer to the last event processed, and for each
- * of these clients we track the usecount of the elements.
- * If we have a pointer to an entry in the list, it is safe to navigate
- * it forward because elements will not be deleted, but only appended.
- * The worst that can happen is seeing the pointer still NULL.
- *
- * When the usecount of an element drops to 0, and the element is the
- * first in the list, we can remove it. Removal is done within the
- * main thread, which is woken up for the purpose.
- *
- * For simplicity of implementation, we make sure the list is never empty.
+ * Events are created as reference counted immutable objects. Each
+ * active session that should receive the event is given a pointer to
+ * the event with a reference. Once the reference reaches 0 the event
+ * is destroyed. These events are stored in a vector on each session
+ * so that each session does not need to allocate a linked list
+ * wrapper for each event. Instead the vector grows as needed with
+ * excess each time to reduce the number of overall reallocations.
  */
 struct eventqent {
-	int usecount;		/*!< # of clients who still need the event */
+	/*! \brief Category this event belongs to */
 	int category;
-	unsigned int seq;	/*!< sequence number */
-	struct timeval tv;  /*!< When event was allocated */
+	/*! \brief Hash of the name of the event */
 	int event_name_hash;
-	AST_RWLIST_ENTRY(eventqent) eq_next;
-	char eventdata[1];	/*!< really variable size, allocated by append_event() */
+	/*! \brief The contents of the AMI event message */
+	struct ast_str *message;
 };
-
-static AST_RWLIST_HEAD_STATIC(all_events, eventqent);
 
 static int displayconnects = 1;
 static int allowmultiplelogin = 1;
@@ -226,6 +214,8 @@ static const struct {
 	{{ "restart", "gracefully", NULL }},
 };
 
+static struct eventqent *eventqent_alloc(const char *event_name, int category);
+
 static void acl_change_stasis_cb(void *data, struct stasis_subscription *sub, struct stasis_message *message);
 
 static void acl_change_stasis_subscribe(void)
@@ -281,7 +271,8 @@ struct mansession_session {
 	struct ast_iostream *stream;	/*!< AMI stream */
 	int inuse;		/*!< number of HTTP sessions using this entry */
 	int needdestroy;	/*!< Whether an HTTP session should be destroyed */
-	pthread_t waiting_thread;	/*!< Sleeping thread using this descriptor */
+	ast_mutex_t http_thread_lock; /*!< Lock for protecting the HTTP thread waiting on this session */
+	pthread_t http_thread;	/*!< HTTP thread waiting on this session */
 	uint32_t managerid;	/*!< Unique manager identifier, 0 for AMI sessions */
 	time_t sessionstart;    /*!< Session start time */
 	struct timeval sessionstart_tv; /*!< Session start time */
@@ -296,16 +287,16 @@ struct mansession_session {
 	struct ao2_container *includefilters;	/*!< Manager event filters - include list */
 	struct ao2_container *excludefilters;	/*!< Manager event filters - exclude list */
 	struct ast_variable *chanvars;  /*!< Channel variables to set for originate */
-	int send_events;	/*!<  XXX what ? */
-	struct eventqent *last_ev;	/*!< last event processed. */
+	int send_events;	/*!< Event categories to send to this session */
 	int writetimeout;	/*!< Timeout for ast_carefulwrite() */
 	time_t authstart;
-	int pending_event;         /*!< Pending events indicator in case when waiting_thread is NULL */
 	time_t noncetime;	/*!< Timer for nonce value expiration */
 	unsigned long oldnonce;	/*!< Stale nonce value */
 	unsigned long nc;	/*!< incremental  nonce counter */
 	unsigned int kicked:1;	/*!< Flag set if session is forcibly kicked */
-	ast_mutex_t notify_lock; /*!< Lock for notifying this session of events */
+	int alert_pipe[2]; /*!< Pipe for alerting this session */
+	AST_VECTOR(, struct eventqent *) pending_events; /*!< Queue of pending events to send */
+	ast_mutex_t pending_events_lock; /*!< Lock for pending events queue */
 	AST_LIST_HEAD_NOLOCK(mansession_datastores, ast_datastore) datastores; /*!< Data stores on the session */
 	AST_LIST_ENTRY(mansession_session) list;
 };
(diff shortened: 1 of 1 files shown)


jcolp:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
Getting up for initial eyes, but working on the tests as they directly manipulate things that have changed.

main/manager.c
```diff
@@ -7507,33 +7536,52 @@ static int purge_sessions(int n_max)
...
+	event->event_name_hash = ast_str_hash(event_name);
 
-	AST_RWLIST_WRLOCK(&all_events);
-	AST_RWLIST_INSERT_TAIL(&all_events, tmp, eq_next);
-	AST_RWLIST_UNLOCK(&all_events);
+	/*
+	 * This event will most likely be queued to at least one session so we do message
+	 * creation directly inside the event to avoid having to duplicate the string.
+	 */
+	event->message = ast_str_create(MANAGER_EVENT_BUF_INITSIZE);
```
seanbright, line 7595: Would there be any benefit to using stringfields instead of an ast_str here? I presume not but thought I’d ask.

jcolp: I haven't tested it, but I suspect not - it might even be slightly worse - since stringfields have overhead themselves since they're made for different addressable strings sharing a pool, instead of for building a single string like ast_str.

seanbright: Thanks. Just glanced at the header file and noticed there was no “append” function for stringfields anyway.

qnic11:
Thanks for the refactor. Moving AMI event delivery away from the global event queue and toward refcounted event objects with per-session pending queues looks like a reasonable direction to me.

That said, I think this still needs another pass before it is safe to merge.

The first thing I would fix is set_eventmask. Previously, send_events was only updated if the parsed event mask was valid. With this patch, send_events is assigned readperm & maskint unconditionally. If strings_to_mask returns -1 for an invalid or empty mask, that turns into readperm & -1, which effectively enables all read-permitted events even though the action may still return an error.

The TCP WaitEvent path also looks off. Timeout is still parsed as seconds, but the value is passed directly to poll, which expects milliseconds. That changes the behavior pretty significantly. For example, Timeout: 30 should behave like roughly 30 seconds, not a handful of short poll intervals

There is another issue in the TCP WaitEvent path: it polls the manager stream fd, but does not do anything with revents for that fd. It only handles the alert pipe. If the client sends input, disconnects, or the socket reports ERR while WaitEvent is active, this can wake repeatedly without reading anything or closing the session.

Finally, I also noticed what looks like an ActionID formatting regression in WaitEventComplete. action_waitevent still builds idText with the ActionID header, but the helpers are passed the raw id instead. The helpers then append that raw value directly into the AMI event body. That would produce a malformed line instead of an ActionID header.

Reviewed with the assistance of ChatGPT AI tool.

jcolp:
Those should now be taken care of.

jcolp:
I also ask that if you have used AI for reviewing things that you disclose it including the model.

jcolp:
I know what's going on. :D

main/manager.c
```diff
@@ -985,19 +941,29 @@ static struct mansession_session *build_mansession(const struct ast_sockaddr *ad
...
 
+	AST_VECTOR_INIT(&newsession->pending_events, 0);
+	ast_alertpipe_clear(newsession->alert_pipe);
+	ast_mutex_init(&newsession->pending_events_lock);
+	AST_LIST_HEAD_INIT_NOLOCK(&newsession->datastores);
+
 	newsession->includefilters = ao2_container_alloc_list(AO2_ALLOC_OPT_LOCK_MUTEX, 0, NULL, NULL);
 	newsession->excludefilters = ao2_container_alloc_list(AO2_ALLOC_OPT_LOCK_MUTEX, 0, NULL, NULL);
 	if (!newsession->includefilters || !newsession->excludefilters) {
 		ao2_ref(newsession, -1);
```
phoneben, line 953: When session is destroyed, `ast_mutex_destroy(&session->http_thread_lock) `is always called, but at this stage `session->http_thread_lock` was not yet initialized — that only happens at line 961


AI (Claude and GPT) was used to help review this, but the analysis and writeup are mine — these reviews take real time

jcolp: Fixed. Looks like it was a pre-existing bug.

phoneben:
question today i configure 

``
disabledevents = AGIExecEnd,AGIExecStart,AOC-D,AOC-E,AOC-S,AgentCalled,AgentDump,AgentLogin,AgentLogoff,AgentRingNoAnswer,Agents,AgentsComplete,Alarm,AlarmClear,AorDetail,AorList,AsyncAGIEnd,AsyncAGIExec,AsyncAGIStart,AttendedTransfer,AuthDetail,AuthList,AuthMethodNotAllowed,BlindTransfer,BridgeCreate,BridgeDestroy,BridgeEnter,BridgeLeave,BridgeMerge,BridgeVideoSourceUpdate,CEL,Cdr,ChallengeResponseFailed,ChallengeSent,ChanSpyStart,ChanSpyStop,ChannelTalkingStart,ChannelTalkingStop,ConfbridgeEnd,ConfbridgeJoin,ConfbridgeLeave,ConfbridgeList,ConfbridgeListRooms,ConfbridgeMute,ConfbridgeRecord,ConfbridgeStart,ConfbridgeStopRecord,ConfbridgeTalking,ConfbridgeUnmute,ContactList,ContactStatusDetail,CoreShowChannel,CoreShowChannelMapComplete,CoreShowChannelsComplete,DAHDIChannel,DNDState,DTMFBegin,DTMFEnd,DeadlockStart,DeviceStateChange,DialBegin,DialEnd,DialState,EndpointDetail,EndpointList,ExtensionStatus,FAXSession,FAXSessionsComplete,FAXSessionsEntry,FAXStats,FAXStatus,FailedACL,Flash,FullyBooted,HangupHandlerPop,HangupHandlerPush,HangupHandlerRun,HangupRequest,Hold,IdentifyDetail,InboundRegistrationDetail,InboundSubscriptionDetail,InvalidAccountID,InvalidPassword,InvalidTransport,Load,LoadAverageLimit,LocalBridge,LocalOptimizationBegin,LocalOptimizationEnd,LogChannel,MCID,MWIGet,MWIGetComplete,MeetmeEnd,MeetmeJoin,MeetmeLeave,MeetmeList,MeetmeListRooms,MeetmeMute,MeetmeTalkRequest,MeetmeTalking,MemoryLimit,MessageWaiting,MiniVoiceMail,MixMonitorMute,MixMonitorStart,MixMonitorStop,MonitorStart,MonitorStop,MusicOnHoldStart,MusicOnHoldStop,NewAccountCode,NewCallerid,NewConnectedLine,NewExten,Newchannel,OriginateResponse,OutboundRegistration,OutboundSubscription,ParkedCallSwap,Pickup,PresenceStateChange,PresenceStatus,QueueEntry,QueueMemberAdded,QueueMemberPenalty,QueueMemberRemoved,QueueMemberRinginuse,QueueMemberStatus,RTCPReceived,RTCPSent,ReceiveFAX,Registry,Reload,Rename,RequestBadFormat,RequestNotAllowed,RequestNotSupported,ResourceListDetail,SIPQualifyPeerDone,SendFAX,SessionLimit,SessionTimeout,Shutdown,SoftHangupRequest,SpanAlarm,SpanAlarmClear,Status,StatusComplete,SuccessfulAuth,TransportDetail,UnexpectedAddress,Unhold,Unload,VarSet,VoicemailPasswordChange,Wink
``

really think it would benefit is we could use a hash lookup to compare  instead of  string scan as now .


attached is a ai idea of how it could be implemented (did not review it or test it).
[manager_disabledevents (2).patch](https://github.com/user-attachments/files/28929587/manager_disabledevents.2.patch)

jcolp:
I'm not going to do such a thing as part of this PR. I purposely kept this change targeted to queueing and some ancillary improvements as a result of the queueing change to make it easier to review and to better understand the impact of it.

main/manager.c
```diff
@@ -296,16 +287,16 @@ struct mansession_session {
...
-	int send_events;	/*!<  XXX what ? */
-	struct eventqent *last_ev;	/*!< last event processed. */
+	int send_events;	/*!< Event categories to send to this session */
 	int writetimeout;	/*!< Timeout for ast_carefulwrite() */
 	time_t authstart;
-	int pending_event;         /*!< Pending events indicator in case when waiting_thread is NULL */
 	time_t noncetime;	/*!< Timer for nonce value expiration */
 	unsigned long oldnonce;	/*!< Stale nonce value */
 	unsigned long nc;	/*!< incremental  nonce counter */
 	unsigned int kicked:1;	/*!< Flag set if session is forcibly kicked */
```
phoneben, line 296: `session->kicked` is read without consistent synchronization in a few paths, e.g. around lines 3152 and 7290.

This is a existing issue rather than something introduced by this patch. 
But since this refactor is already changing the session wakeup/locking model, it may be worth cleaning up.

suggesting a fix
If we want to make this safe, `kicked` probably needs to stop being a bitfield and become a plain `int`, then be accessed through Asterisk's atomic helpers or protected consistently by the same lock.

InterLinked1: Does that actually matter, considering `kicked` can only change from 0 to 1 anyways?

jcolp: I lean towards no, which is probably why it was the way it was before.

main/manager.c
```diff
@@ -3252,54 +3251,88 @@ static int action_waitevent(struct mansession *s, const struct message *m)
...
 		astman_send_response(s, m, "Success", "Waiting for Event completed.");
-		while ((eqe = advance_event(eqe))) {
-			if (((s->session->readperm & eqe->category) == eqe->category)
-				&& ((s->session->send_events & eqe->category) == eqe->category)
-				&& should_send_event(s->session->includefilters, s->session->excludefilters, eqe)) {
-				astman_append(s, "%s", eqe->eventdata);
+		for (x = 0; x < events_count; x++) {
+			struct eventqent *eqe = events[x];
+
+			if (((s->session->send_events & eqe->category) == eqe->category)) {
```
mbradeen: could make this an && instead of a nested if

main/manager.c
```diff
@@ -6424,35 +6454,53 @@ AST_TEST_DEFINE(eventfilter_test_matching)
...
+	for (i = 0; i < events_count; i++) {
+		struct eventqent *eqe = events[i];
+
+		if (eqe->category == EVENT_FLAG_SHUTDOWN) {
+			ast_debug(3, "Received CloseSession event\n");
+			ret = -1;
+			/* We purposely don't break here so that we drop the reference on remaining events */
+		}
+
+		if (!ret && ((s->session->send_events & eqe->category) == eqe->category)) {
```
mbradeen: Why the move to nested if's instead of &&'s?

mbradeen (approved):

phoneben:
are you sure you want to merge this ?
https://community.asterisk.org/t/pickupchan-not-working/113477/12

jcolp:
We won't be doing normal releases for a month probably, so merging now will get it into active testing amongst all the other tests and PRs and for those who use from the branch. If it causes issues (or you can get a backtrace) we can resolve or revert before then.

I've reviewed the change again and I don't see how it could directly cause an issue - the only thing it could do is make things more efficient to then make an already existing issue more likely to occur like I mentioned on the forum. Without a backtrace to go with that, it would be like looking for a needle in a haystack.

PR #1991, opened 2026-06-22
test_performance: Add performance experimentation test module.

jcolp:
This is a module which is only built when TEST_FRAMEWORK is enabled
and provides CLI commands for testing performance of certain things
on the system they are invoked on. The first CLI commands added
cover the most common container usage: storage of objects with a
lookup based on a string key. These commands take various arguments
and allow you to see how they perform. There is also an "all" command
named "performance test container_key_lookup_all" that will execute
all of these container tests and pass through the given arguments,
which makes it easy to run all of the tests for given usage.

To facilitate a vector bsearch test a new AST_VECTOR_BSEARCH macro
has been added that allows more efficient searching of sorted vectors.

### DIFF
--- a/tests/test_performance.c
+++ b/tests/test_performance.c
@@ -1,0 +1,1013 @@
+/*
+ * Asterisk -- An open source telephony toolkit.
+ *
+ * Copyright (C) 2026, Sangoma Technologies Inc
+ *
+ * Joshua Colp <jcolp@sangoma.com>
+ *
+ * See http://www.asterisk.org for more information about
+ * the Asterisk project. Please do not directly contact
+ * any of the maintainers of this project for assistance;
+ * the project provides a web site, mailing lists and IRC
+ * channels for your use.
+ *
+ * This program is free software, distributed under the terms of
+ * the GNU General Public License Version 2. See the LICENSE file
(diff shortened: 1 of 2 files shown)


jcolp:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

gtjoseph (changes requested):
An option to run them all would be handy.

gtjoseph (approved):

tests/test_performance.c
```diff
@@ -0,0 +1,1013 @@
...
+}
+
+/*!
+ * \brief Compares two strings for finding in the RBTree
+ * \param lhs The first string to compare
+ * \param rhs The second string to compare
+ * \param flags The search flags
+ * \return The result of the comparison
+ */
+static int rbtree_str_cmp(void *lhs, void *rhs, int flags)
```
mbradeen, line 317: should lhs and rhs be const in this function as well?

jcolp: The definition for ao2_callback_fn, which this implements, is non-cost so alas - it can not be.

mbradeen (approved):

PR #1993, opened 2026-06-23
format_cap: guard against NULL src in *_from_cap helpers

### LINKED ISSUE #1992: [bug]: Asterisk crashes (SIGSEGV) when CHANNEL(videonativeformat)/CHANNEL(audionativeformat) is evaluated on a dummy channel
### Severity

Major

### Versions

22.10.0

### Components/Modules

format_cap

### Operating Environment

Debian 12 (bookworm)

### Frequency of Occurrence

Constant

### Issue Description

`ast_format_cap_append_from_cap()` (and its sibling `ast_format_cap_replace_from_cap()`)
dereference the `src` capability pointer without a NULL check:

```c
/* main/format_cap.c:275 */
for (idx = 0; (idx < AST_VECTOR_SIZE(&src->preference_order)) && !res; ++idx) {
```

A **dummy channel** created via `ast_dummy_channel_alloc()` never sets a native-format
capability (`__ast_dummy_channel_alloc()` in `main/channel.c` does not call
`ast_channel_nativeformats_set()`), so `ast_channel_nativeformats()` returns `NULL`
for such a channel.

`func_channel_read()` passes that value straight through for the
`audionativeformat` / `videonativeformat` keys:

```c
/* funcs/func_channel.c:417 */
ast_format_cap_append_from_cap(tmpcap, ast_channel_nativeformats(chan), AST_MEDIA_TYPE_VIDEO);
```

When `chan` is a dummy channel, `src == NULL`, and the loop dereferences NULL
(`src->preference_order`, fault address `0x28`), causing a SIGSEGV.

## How it is triggered in practice
The crash is reached through the voicemail email-notification path combined with an
ARI configuration that publishes `CHANNEL(videonativeformat)` (and/or
`audionativeformat`) as a channel variable (`ari.conf` `channelvars=`):

1. A caller leaves a voicemail => `app_voicemail` `leave_voicemail()` =>
   `notify_new_message()` => `sendmail()` => `make_email_file()`.
2. `make_email_file()` allocates a **dummy channel** with `ast_dummy_channel_alloc()`
   to perform variable substitution for the email template, then calls
   `prep_email_sub_vars()`.
3. `prep_email_sub_vars()` sets variables on the dummy channel via
   `pbx_builtin_setvar_helper(ast, "VM_NAME", ...)`.
4. Setting a channel variable publishes a Stasis **VarSet** event
   (`ast_channel_publish_varset()`), which gathers the configured ARI channel
   variables (`ast_channel_get_ari_vars()` => `channel_get_external_vars()`).
5. Evaluating `CHANNEL(videonativeformat)` invokes `func_channel_read()`, which calls
   `ast_format_cap_append_from_cap()` with the dummy channel's NULL native formats.
6. NULL dereference that causes a SIGSEGV.

This report was generated with the help of AI, using Claude Opus 4.8.
The issue was reproduced every time with the voicemail email notification path.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

bloom1:
ast_format_cap_append_from_cap() and ast_format_cap_replace_from_cap()
dereference 'src' (src->preference_order) without checking it for NULL.

A dummy channel allocated with ast_dummy_channel_alloc() never sets a
native-format capability, so ast_channel_nativeformats() returns NULL on
such channels. When CHANNEL(audionativeformat) / CHANNEL(videonativeformat)
is evaluated against a dummy channel (e.g. via ARI channelvars during a
Stasis VarSet event raised while app_voicemail builds the notification
email on a dummy channel), func_channel_read() passes that NULL straight
into ast_format_cap_append_from_cap(), causing a NULL dereference at
offset 0x28 and a SIGSEGV.

Guard both helpers against a NULL source. A NULL source simply means
"no formats to copy", so appending/replacing nothing is the correct
no-op behaviour. This also protects all other callers.

Fixes https://github.com/asterisk/asterisk/issues/1992

AI disclosure: this was generated using Claude Opus 4.8, tested to fix the issue. Not sure if it is the *right* way to do it.

### DIFF
--- a/main/format_cap.c
+++ b/main/format_cap.c
@@ -271,6 +271,10 @@ int ast_format_cap_append_from_cap(struct ast_format_cap *dst, const struct ast_
 {
 	int idx, res = 0;
 
+	if (!src) {
+		return 0;
+	}
+
 	/* NOTE:  The streams API is dependent on the formats being in "preference" order */
 	for (idx = 0; (idx < AST_VECTOR_SIZE(&src->preference_order)) && !res; ++idx) {
 		struct format_cap_framed *framed = AST_VECTOR_GET(&src->preference_order, idx);
@@ -307,6 +311,10 @@ void ast_format_cap_replace_from_cap(struct ast_format_cap *dst, const struct as
 	enum ast_media_type type)
 {
 	int idx;
+
+	if (!src) {
+		return;
+	}
 
 	for (idx = 0; (idx < AST_VECTOR_SIZE(&src->preference_order)); ++idx) {
 		struct format_cap_framed *framed = AST_VECTOR_GET(&src->preference_order, idx);


main/format_cap.c
```diff
@@ -271,6 +271,10 @@ int ast_format_cap_append_from_cap(struct ast_format_cap *dst, const struct ast_
 {
 	int idx, res = 0;
 
+	if (!src) {
+		return 0;
```
gtjoseph, line 275: `-1` is the normal return code for an error.  That's what ast_format_cap_append returns for an error.

Please also address the PR Checklist items.

gtjoseph (dismissed):

bloom1:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

gtjoseph (approved):

main/format_cap.c
seanbright: `ast_format_cap_append` returns a `-1` only when an allocation error occurs.

- If `src` is `NULL` semantically means that the `src` is empty, then returning `0` here is more appropriate - we appended nothing successfully.
- If `src` is `NULL` is an error condition then it should be an `ast_assert` instead and the call sites should be updated. As it is, many callers of this function ignore the return value completely.

seanbright (changes requested):

main/format_cap.c
bloom1: @seanbright indeed, semantically it's not an error. I will put back `return 0`.

seanbright (approved):

gtjoseph:
@bloom1 The merge failure didn't have anything to do with you.  GitHub was once again being stupid and claiming something it just created didn't exist. :)

bloom1:
> @bloom1 The merge failure didn't have anything to do with you. GitHub was once again being stupid and claiming something it just created didn't exist. :)

Alright, thank you! I saw that the branch was not up-to-date with master so I rebased it just to be sure.

gtjoseph:
You did the right thing even if it was GitHub's fault. :)

PR #1998, opened 2026-06-24
extensions.ael.sample: Restore removed macros

seanbright:
Commit e8f548c1 removed AEL `macro` definition and calls from the sample configuration file, but those do not use the deprecated/removed `Macro` app - they use `Gosub` under the hood.

### DIFF
--- a/configs/samples/extensions.ael.sample
+++ b/configs/samples/extensions.ael.sample
@@ -174,6 +174,18 @@ context ael-dundi-e164-lookup {
 };
 
 //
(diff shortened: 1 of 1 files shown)


seanbright:
cherry-pick-to: 22
cherry-pick-to: 23

gtjoseph (approved):

gtjoseph (dismissed):
Doesn't cherry-pick to 20 probably because it's not needed there.

PR #2009, opened 2026-06-30
chan_websocket: Use leftover data if no frames are available when the timer fires.

### LINKED ISSUE #2006: [bug]: WebSocket Channel, odd frame is dropped after `STOP_MEDIA_BUFFERING`
### Severity

Major

### Versions

23.4.1

### Components/Modules

chan_websocket

### Operating Environment

VM with Debian13 
Asterisk 23.4.1 ===ws==> Python websocket server

```
import asyncio
import websockets

async def receive_messages(websocket):
    async for message in websocket:
        if isinstance(message, str):
            print(f"EVENT: {message}")

async def send_audio(websocket):
    while True:
        await websocket.send("START_MEDIA_BUFFERING")
        print("sent START_MEDIA_BUFFERING")
        with open("1-192-max.raw", "rb") as f:
            sent_bytes = 0
            while True:
                chunk = f.read(50000)
                if not chunk:
                    break
                await websocket.send(chunk)
                print(".", end="", flush=True)
                await asyncio.sleep(0.1)
                sent_bytes += len(chunk)
        await websocket.send("STOP_MEDIA_BUFFERING")
        print("\nFinished sending audio.")
        await asyncio.sleep(3)

async def handler(websocket):
    print(f"Client connected: {websocket.remote_address}")
    await asyncio.gather(
        receive_messages(websocket),
        send_audio(websocket),
    )

async def main():
    async with websockets.serve(handler, "0.0.0.0", 8765):
        await asyncio.Future() 

if __name__ == "__main__":
    asyncio.run(main())
```

Dialplan:
```
exten = 999,1,MixMonitor(/tmp/voice_test.wav,b)
same=n,Dial(WebSocket/ws_local/c(slin192))
```

[1-192-max.raw.zip](https://github.com/user-attachments/files/29399292/1-192-max.raw.zip)

### Frequency of Occurrence

Constant

### Issue Description

There are two things: 
1. When we send `START_MEDIA_BUFFERING`, then audio chunks, and don't send   `STOP_MEDIA_BUFFERING` the last odd frame is kept in the buffer and played when the next audio chunk arrives (in this case, after 3 secs). Formally, it's not a bug, as documentation doesn't explicitly define the behaviour for this case, but from a practical point of view, it doesn't make sense to keep this leftover (less than 20 ms) in the buffer after we have already started writing silence to the core. Typically, the next audio chunk is the next inference from TTS, which starts from silence, so having this leftover will introduce click-like artifacts. 
Image

2. When we send `START_MEDIA_BUFFERING`, then audio chunks, and send  `STOP_MEDIA_BUFFERING` the last odd frame is dropped, but based on the documentation, it should be concealed with silence. If the last odd frame is concealed and played, then tone in the bottom track should be a bit loner compared to the top track.  
> it'll take whatever data is left in the buffer that couldn't make a full frame with, append silence to it to make up a full frame, and send that to the core.

The second screenshot is a zoomed version of the first one,  top track w/o `STOP_MEDIA_BUFFERING`, bottom with `STOP_MEDIA_BUFFERING`

Image

The issue can be reproduced with the Python script attached.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

gtjoseph:
When the 20ms channel timer fires but there are no frames available in
the queue, we now check for leftover data in the buffer and if there is
any, we create a frame with it and send it to the core. This resolves an
issue with the leftover data being delayed if a STOP_MEDIA_BUFFERING
command is delayed. Some existing comments were also clarified to
account for the new behavior.

Resolves: #2006

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -469,9 +469,7 @@ static struct ast_frame *dequeue_frame(struct websocket_pvt *instance)
 		AST_LIST_UNLOCK);
 
 	/*
-	 * If the queue is paused, don't read a frame.  Processing
-	 * will continue down the function and a silence frame will
-	 * be sent in its place.
+	 * If the queue is paused, don't read a frame.
 	 */
 	if (instance->queue_paused) {
 		return NULL;
@@ -493,10 +491,9 @@ static struct ast_frame *dequeue_frame(struct websocket_pvt *instance)
 	queued_frame = AST_LIST_REMOVE_HEAD(&instance->frame_queue, frame_list);
 
 	/*
-	 * If there are no frames in the queue, we need to
-	 * return NULL so we can send a silence frame.  We also need
-	 * to send the QUEUE_DRAINED notification if we were requested
-	 * to do so.
+	 * If there are no frames in the queue, we need to return NULL.
+	 * We also need to send the QUEUE_DRAINED notification if we
+	 * were requested to do so.
 	 */
 	if (!queued_frame) {
 		if (instance->report_queue_drained) {
@@ -550,8 +547,7 @@ static struct ast_frame *dequeue_frame(struct websocket_pvt *instance)
 
 	/*
 	 * If, after reading all control frames,  there are no frames
-	 * left in the queue, we need to return NULL so we can send
-	 * a silence frame.
+	 * left in the queue, we need to return NULL.
 	 */
 	if (!queued_frame) {
 		return NULL;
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

mbradeen (approved):

channels/chan_websocket.c
```diff
@@ -561,6 +557,28 @@ static struct ast_frame *dequeue_frame(struct websocket_pvt *instance)
...
+static struct ast_frame *create_frame_from_buffer(struct websocket_pvt *instance,
+	char *buffer, size_t len)
+{
+	struct ast_frame fr = { 0, };
+	struct ast_frame *duped_frame = NULL;
+
+	AST_FRAME_SET_BUFFER(&fr, buffer, 0, len);
+	fr.frametype = AST_FRAME_VOICE;
+	fr.subclass.format = instance->native_format;
+	fr.samples = instance->native_codec->samples_count(&fr);
```
jcolp: I know this was already present before, but looking at the ast_codec_samples_count shows that samples_count is not guaranteed to be non-NULL.

jcolp (dismissed):

channels/chan_websocket.c
gtjoseph: resolved

jcolp (approved):

sohankhan19974-hub (approved):

PR #2015, opened 2026-07-06
res_pjsip: Use unique payload types for bundled streams

### LINKED ISSUE #2010: [bug]: Ambiguous RTP stream association in bundled WebRTC sessions with multiple streams with the same payload types
### Severity

Minor

### Versions

20.17.x

### Components/Modules

WebRTC

### Operating Environment

Ubuntu 22.04.5 LTS

### Frequency of Occurrence

Constant

### Summary

When Asterisk generates SDP containing multiple bundled media sections, it may reuse the same payload type values across those sections.

When SDES is not supported (which asterisk does not support as far as I know), [RFC 8843](https://www.rfc-editor.org/rfc/rfc8843.html#name-associating-rtp-rtcp-stream) requires payload type values to uniquely identify the associated media section. Reusing the same payload types across bundled media sections makes RTP stream association ambiguous and causes interoperability issues with sins WebRTC implementations that rely on RFC 8843 payload-type-based demultiplexing.

### Reproduction Steps

1. Configure a WebRTC endpoint in Asterisk with bundle enabled.
2. Establish a session containing multiple bundled streams.
3. Observe that the generated SDP contains multiple `m=` sections using the same payload types.

Example:

```sdp
a=group:BUNDLE 0 1 video-2

m=video ... 96 103
a=mid:1

m=video ... 96 103
a=mid:video-2
```

4. Receive RTP packets from asterisk.
5. Attempt to associate the RTP packets received from asterisk with a media section using the payload type, as specified by RFC 8843.

### Details

[RFC 8843](https://www.rfc-editor.org/rfc/rfc8843.html#name-associating-rtp-rtcp-stream) defines how RTP streams are associated with media sections in a bundled session. When SDES is not supported, the payload type is used to identify the associated media section.

The RFC states:
> In this situation, each "m=" section needs to use unique payload type values, in order for the payload type to be a reliable indicator of the relevant "m=" section for the RTP stream. If an implementation fails to ensure unique payload type values, it will be impossible to associate the RTP stream using that payload type value to a particular "m=" section. 

Asterisk may generate multiple bundled media sections that advertise identical payload types (e.g. `96` and `103`). Since the same payload type appears in more than one `m=` section, a receiver cannot uniquely determine which media section an RTP packet belongs to based solely on the payload type.

Some WebRTC clients use the payload type as the primary identifier in this situation and therefore cannot reliably associate packets with the correct media section. This can lead to packets being processed using the wrong SRTP context, resulting in decryption failures and loss of media for the affected stream. 
Clients that perform additional SSRC-based matching on top of the payload matching are not affected by this.

**Expected behavior**
When bundle is used without SDES support, payload types should be unique across bundled media sections.

**Actual behavior**
Asterisk generates bundled media sections that reuse the same payload types, resulting in ambiguous RTP stream association for implementations that strictly follow RFC 8843 payload-type-based demultiplexing.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

seifzadeh:
This update fixes BUNDLE handling for multiple media streams in Asterisk.

Asterisk previously assigned the same dynamic payload type to multiple bundled streams using the same codec. The change assigns unique dynamic payload types to local streams within the BUNDLE group.

The issue also exposed pjproject rejecting valid zero-port `a=bundle-only` media sections before codec negotiation. The pjproject fix has been submitted upstream as pjsip/pjproject#5067.

The bundled pjproject patch is retained temporarily so the fix works with Asterisk's current bundled pjproject version. It can be removed after a future pjproject upgrade includes the upstream change.

The Asterisk changes preserve valid BUNDLE-only streams in the session topology and apply unique payload types in SDP and RTP.

Tested with the original WebRTC multi-stream scenario and pjproject's PJMEDIA tests.

Fixes: #2010

### DIFF
--- a/res/res_pjsip_sdp_rtp.c
+++ b/res/res_pjsip_sdp_rtp.c
@@ -441,6 +441,150 @@ static void get_codecs(struct ast_sip_session *session, const struct pjmedia_sdp
 	SCOPE_EXIT_RTN();
 }
 
+
+/*! \brief Determine whether a payload type is already advertised in a BUNDLE group. */
+static int bundle_payload_type_used(struct ast_sip_session *session,
+	const pjmedia_sdp_session *sdp, const pjmedia_sdp_media *media,
+	int bundle_group, int payload)
+{
+	int media_index;
+
+	for (media_index = 0; media_index < sdp->media_count; ++media_index) {
+		struct ast_sip_session_media *other_session_media;
+		int format_index;
+
+		if (media_index >= AST_VECTOR_SIZE(&session->pending_media_state->sessions)) {
+			break;
+		}
+
+		other_session_media = AST_VECTOR_GET(&session->pending_media_state->sessions, media_index);
+		if (!other_session_media || other_session_media->bundle_group != bundle_group) {
+			continue;
+		}
+
+		for (format_index = 0; format_index < sdp->media[media_index]->desc.fmt_count; ++format_index) {
+			if (pj_strtoul(&sdp->media[media_index]->desc.fmt[format_index]) == payload) {
+				return 1;
+			}
+		}
+	}
+
+	/* The current media section is not in the SDP yet, so inspect it separately. */
+	for (media_index = 0; media_index < media->desc.fmt_count; ++media_index) {
+		if (pj_strtoul(&media->desc.fmt[media_index]) == payload) {
+			return 1;
+		}
+	}
+
+	return 0;
+}
+
+/*! \brief Check whether a payload slot is either free or already mapped to the requested format. */
+static int payload_type_compatible(struct ast_rtp_codecs *codecs, int payload,
+	struct ast_format *format)
+{
+	struct ast_rtp_payload_type *type;
+	int compatible;
+
+	type = ast_rtp_codecs_get_payload(codecs, payload);
+	if (!type) {
+		return 1;
+	}
+
+	compatible = type->asterisk_format
+		&& ast_format_cmp(type->format, format) == AST_FORMAT_CMP_EQUAL;
+	ao2_ref(type, -1);
+
+	return compatible;
+}
+
+/*! \brief Ensure an RTP codecs structure can receive a format using a payload type. */
+static int ensure_payload_type_mapping(struct ast_rtp_codecs *codecs, int payload,
+	struct ast_format *format)
+{
+	struct ast_rtp_payload_type *type;
+	int matches;
+
+	type = ast_rtp_codecs_get_payload(codecs, payload);
+	if (type) {
+		matches = type->asterisk_format
+			&& ast_format_cmp(type->format, format) == AST_FORMAT_CMP_EQUAL;
+		ao2_ref(type, -1);
+		return matches ? 0 : -1;
+	}
+
+	return ast_rtp_codecs_payload_set_rx(codecs, payload, format) < 0 ? -1 : 0;
+}
+
+/*! \brief Assign a payload type that is unique across the generated BUNDLE group. */
+static int assign_unique_bundle_payload(struct ast_sip_session *session,
+	struct ast_sip_session_media *session_media,
+	struct ast_sip_session_media *session_media_transport,
+	const pjmedia_sdp_session *sdp, const pjmedia_sdp_media *media,
+	struct ast_format *format)
+{
+	struct ast_rtp_codecs *transport_codecs =
+		ast_rtp_instance_get_codecs(session_media_transport->rtp);
+	struct ast_rtp_codecs *stream_codecs =
+		ast_rtp_instance_get_codecs(session_media->rtp);
+	int preferred;
+	int candidate;
+	int offset;
+
+	preferred = ast_rtp_codecs_payload_code(stream_codecs, 1, format, 0);
+	if (preferred < 0) {
+		return -1;
+	}
+
+	/* Preserve the stream's normal payload when it is not used by another m= section. */
+	if (!bundle_payload_type_used(session, sdp, media, session_media->bundle_group, preferred)
+		&& payload_type_compatible(transport_codecs, preferred, format)
+		&& payload_type_compatible(stream_codecs, preferred, format)) {
+		candidate = preferred;
+		goto configure;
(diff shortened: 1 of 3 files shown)


seifzadeh:
multiple-commits: standalone
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seifzadeh:
AI usage disclosure:
I used AI-assisted tools during this work for:
- Reviewing the relevant sections of RFC 9143.
- Generating the initial WebRTC and Asterisk test configuration.
- Developing the reproduction procedure.
- Helping inspect and summarize the SIP, SDP, RTP, and Asterisk debug logs.
- Reviewing possible implementation approaches.
- Assisting with the initial code changes.
- Converting the tested pjproject source change into an Asterisk bundled-pjproject patch.
ai model: chatgpt 5.5

seifzadeh:
Reproduction and test details:

Environment:

- Asterisk built from the current master branch.
- Bundled pjproject.
- Firefox WebRTC client.
- JsSIP-based Cyber Mega Phone 2K client.
- VP8 video using generated Canvas media tracks.
- SIP over WebSocket with ICE, DTLS, SRTP, and BUNDLE enabled.

jcolp:
Has the PJSIP change been provided upstream for consideration of inclusion and review?

seifzadeh:
> Has the PJSIP change been provided upstream for consideration of inclusion and review?

Not yet. I checked the asterisk source and saw that it is possible to patch the pjsip source at configure and compile time. Should I send these parts to pjsip itself?

jcolp:
We don't fork PJSIP if we can help it, so if there's changes which are not specific to us (which this isn't) then it should go to upstream if possible.

seifzadeh:
> We don't fork PJSIP if we can help it, so if there's changes which are not specific to us (which this isn't) then it should go to upstream if possible.

Thanks for the advice. I will make the changes in pjproject. The sources of this pr will also change subsequently.

seifzadeh:
The related pjproject change has been submitted upstream as pjsip/pjproject#5067. I removed the bundled pjproject patch from this PR, so it now contains only the Asterisk-specific BUNDLE topology and payload type changes.

jcolp:
The patch still has to be here to work. What will happen is when we next upgrade PJSIP it will then be removed, since it is no longer required.

silentindark:
@seifzadeh Your PR https://github.com/pjsip/pjproject/pull/5067 has been merged into the pjproject upstream. Please update the Asterisk commit by replacing the existing patch with the merged patch: https://github.com/pjsip/pjproject/commit/58666750d6bfe19dc213d010be63b1ab6b33ef65.patch

seifzadeh:
> The patch still has to be here to work. What will happen is when we next upgrade PJSIP it will then be removed, since it is no longer required.

the changed on pjproject, merged to master
then removed this path on the pr?
https://github.com/pjsip/pjproject/pull/5067

silentindark:
@seifzadeh The current Asterisk releases use PJSIP 2.17. Your change will be included in the next PJSIP release 2.18, so you'll need to backport it from that release to the current one. Once PJSIP 2.18 is released, the Asterisk releases will be updated to use PJSIP 2.18, and your patch 0030-pjmedia-handle-bundle-only-offers.patch will be removed from third-party/pjproject/patches because PJSIP 2.18 will already include this change.

jcolp (approved):

seifzadeh:
@jcolp The PR checklist is still using the old cherry-pick-to: none value. The metadata comment has been updated to 23, 22, and 20. Could this be rechecked?

gtjoseph (dismissed):
This PR doesn't build because the pjproject patch is malformed.

You need to use `git format-patch` in your pjproject fork to get the  actual merged commit's patch including the full commit message.

Please then test locally before updating the PR.

silentindark:
@seifzadeh FYI https://github.com/pjsip/pjproject/pull/5192

seifzadeh:
> @seifzadeh FYI [pjsip/pjproject#5192](https://github.com/pjsip/pjproject/pull/5192)

Thank you. This was just a recheck of the media with the pjmedia_sdp_validate2 function. It will not disrupt the current pr.

gtjoseph:
@seifzadeh You still need to fix the malformed patch as it won't apply.

```
patching file pjmedia/src/pjmedia/sdp_neg.c
patch: **** malformed patch at line 93:                                 const pjmedia_sdp_media *offer,
```

Make sure to use `git format-patch` on the version that was actually merged in pjsip/pjproject#5067

seifzadeh:
> @seifzadeh You still need to fix the malformed patch as it won't apply.
> 
> ```
> patching file pjmedia/src/pjmedia/sdp_neg.c
> patch: **** malformed patch at line 93:                                 const pjmedia_sdp_media *offer,
> ```
> 
> Make sure to use `git format-patch` on the version that was actually merged in [pjsip/pjproject#5067](https://github.com/pjsip/pjproject/pull/5067)
> 
> https://docs.asterisk.org/Development/Policies-and-Procedures/Code-Contribution/

thanks, Fixed the patch.

gtjoseph:
I don't _think_ the test failures are related to the PR but I'm re-running the checks to make sure.

gtjoseph (changes requested):
Unfortunately, those 3 test failures are legitimate.  I can reproduce them locally with this PR but they pass without it.  You're going to have to run those 3 tests locally to troubleshoot.  Make sure you run `./configure --enable-dev-mode` and set DO_CRASH in menuselect compiler flags and also run `make install-headers` after the normal build and install. When you run the testsuite, use the `--keep-core` and `--keep-full-logs` options to hang on to the everything.  The two pjsip tests may actually indicate that they passed but you'll see `FRACK!, Failed assertion...` in the output.

seifzadeh:
@gtjoseph I found that the failures were coming from the branch being behind the current master. After rebasing onto the latest master, the assertions no longer reproduce in the failing tests locally.

Also, thanks for pointing me to the Asterisk testsuite. I spent some time working with it and learned a lot from reproducing and debugging the failures there.

PR #2017, opened 2026-07-06
res_pjsip: Use common payload mappings for direct media

### LINKED ISSUE #950: [bug]: No audio in call between 2 pjsip endpoints using opus codec
### Severity

Minor

### Versions

18.23, 20.9

### Components/Modules

chan_pjsip

### Operating Environment

OS is Ubuntu 22.04 LTS

### Frequency of Occurrence

Constant

### Issue Description

call between chan_pjsip endpoints using direct media and codec as opus, has no audio. But when I change codec to ulaw it works fine and also when I change chan_pjsip to chan_sip, direct media using opus works fine. When I use chan_pjsip and opus codec,  directmedia is not working.
[ast_verbose_full_PJSIPendpoints_Noaudio.txt](https://github.com/user-attachments/files/17410869/ast_verbose_full_PJSIPendpoints_Noaudio.txt)
[ast_verbose_full_SIPendpoints_workingFine.txt](https://github.com/user-attachments/files/17410870/ast_verbose_full_SIPendpoints_workingFine.txt)

### Relevant log output

_No response_

### Asterisk Issue Guidelines

remiquezada:
Direct media reINVITEs could independently adopt the opposite leg's
dynamic payload mappings. When endpoints used different payload numbers
for the same codec, each leg could consequently negotiate a different
mapping and cause one-way audio.

Build a common payload mapping for both legs using a stable payload
layout. Retain only mutually compatible formats, negotiate codec
attributes, and intersect telephone-event clock rates and event ranges.

Add coverage for different Opus and telephone-event payload mappings.

Resolves: #950.

### DIFF
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -328,26 +328,37 @@ static int check_for_rtp_changes(struct ast_channel *chan, struct ast_rtp_instan
 		struct ast_sip_session_media *media, struct ast_sip_session *session)
 {
 	int changed = 0, position = -1;
+	int payloads_changed;
 
 	if (media->rtp) {
 		position = rtp_find_rtcp_fd_position(session, media->rtp);
 	}
 
 	if (rtp) {
+		payloads_changed = ast_sip_session_media_set_direct_media_payloads(media, rtp);
+		if (payloads_changed < 0) {
+			ast_log(LOG_ERROR, "Unable to snapshot direct media payloads for channel '%s'\n",
+				ast_channel_name(chan));
+			return 0;
+		}
 		changed = ast_rtp_instance_get_and_cmp_remote_address(rtp, &media->direct_media_addr);
+		changed |= payloads_changed;
 		if (media->rtp) {
 			if (position != -1) {
 				ast_channel_set_fd(chan, position + AST_EXTENDED_FDS, -1);
 			}
 			ast_rtp_instance_set_prop(media->rtp, AST_RTP_PROPERTY_RTCP, 0);
 		}
-	} else if (!ast_sockaddr_isnull(&media->direct_media_addr)){
-		ast_sockaddr_setnull(&media->direct_media_addr);
-		changed = 1;
-		if (media->rtp) {
-			ast_rtp_instance_set_prop(media->rtp, AST_RTP_PROPERTY_RTCP, 1);
-			if (position != -1) {
-				ast_channel_set_fd(chan, position + AST_EXTENDED_FDS, ast_rtp_instance_fd(media->rtp, 1));
+	} else {
+		changed |= ast_sip_session_media_set_direct_media_payloads(media, NULL);
+		if (!ast_sockaddr_isnull(&media->direct_media_addr)) {
+			ast_sockaddr_setnull(&media->direct_media_addr);
+			changed = 1;
+			if (media->rtp) {
+				ast_rtp_instance_set_prop(media->rtp, AST_RTP_PROPERTY_RTCP, 1);
+				if (position != -1) {
+					ast_channel_set_fd(chan, position + AST_EXTENDED_FDS, ast_rtp_instance_fd(media->rtp, 1));
+				}
 			}
 		}
 	}
--- a/channels/chan_pjsip.c
+++ b/channels/chan_pjsip.c
@@ -424,26 +424,37 @@ static int check_for_rtp_changes(struct ast_channel *chan, struct ast_rtp_instan
 		struct ast_sip_session_media *media, struct ast_sip_session *session)
 {
 	int changed = 0, position = -1;
+	int payloads_changed;
 
 	if (media->rtp) {
 		position = rtp_find_rtcp_fd_position(session, media->rtp);
 	}
 
 	if (rtp) {
+		payloads_changed = ast_sip_session_media_set_direct_media_payloads(media, rtp);
+		if (payloads_changed < 0) {
+			ast_log(LOG_ERROR, "Unable to snapshot direct media payloads for channel '%s'\n",
+				ast_channel_name(chan));
+			return 0;
+		}
 		changed = ast_rtp_instance_get_and_cmp_remote_address(rtp, &media->direct_media_addr);
+		changed |= payloads_changed;
 		if (media->rtp) {
 			if (position != -1) {
 				ast_channel_set_fd(chan, position + AST_EXTENDED_FDS, -1);
 			}
 			ast_rtp_instance_set_prop(media->rtp, AST_RTP_PROPERTY_RTCP, 0);
 		}
-	} else if (!ast_sockaddr_isnull(&media->direct_media_addr)){
-		ast_sockaddr_setnull(&media->direct_media_addr);
-		changed = 1;
-		if (media->rtp) {
-			ast_rtp_instance_set_prop(media->rtp, AST_RTP_PROPERTY_RTCP, 1);
-			if (position != -1) {
-				ast_channel_set_fd(chan, position + AST_EXTENDED_FDS, ast_rtp_instance_fd(media->rtp, 1));
+	} else {
+		changed |= ast_sip_session_media_set_direct_media_payloads(media, NULL);
+		if (!ast_sockaddr_isnull(&media->direct_media_addr)) {
+			ast_sockaddr_setnull(&media->direct_media_addr);
+			changed = 1;
+			if (media->rtp) {
+				ast_rtp_instance_set_prop(media->rtp, AST_RTP_PROPERTY_RTCP, 1);
+				if (position != -1) {
+					ast_channel_set_fd(chan, position + AST_EXTENDED_FDS, ast_rtp_instance_fd(media->rtp, 1));
+				}
 			}
 		}
 	}
--- a/include/asterisk/res_pjsip_session.h
+++ b/include/asterisk/res_pjsip_session.h
@@ -51,6 +51,8 @@ struct pjmedia_sdp_media;
 struct pjmedia_sdp_session;
 struct ast_dsp;
 struct ast_udptl;
+struct ast_rtp_payload_type;
+struct ast_sip_session_media_rtp_payloads;
 
 /*! \brief T.38 states for a session */
 enum ast_sip_session_t38state {
@@ -77,6 +79,8 @@ typedef int (*ast_sip_session_media_write_cb)(struct ast_sip_session *session, s
 struct ast_sip_session_media {
 	/*! \brief RTP instance itself */
 	struct ast_rtp_instance *rtp;
+	/*! \brief Snapshot of the common direct media transmit payload mappings */
+	struct ast_sip_session_media_rtp_payloads *direct_media_payloads;
 	/*! \brief UDPTL instance itself */
 	struct ast_udptl *udptl;
 	/*! \brief Direct media address */
@@ -128,6 +132,27 @@ struct ast_sip_session_media {
 	/*! \brief Stream name */
 	char *stream_name;
 };
+
+/*!
+ * \brief Replace the direct media payload snapshot with mappings common to two RTP instances
+ *
+ * \param session_media Session media to update
+ * \param rtp Direct media peer, or NULL to clear the snapshot
+ *
+ * \retval 1 Snapshot changed
+ * \retval 0 Snapshot did not change
+ * \retval -1 Allocation failure
+ */
+int ast_sip_session_media_set_direct_media_payloads(
+	struct ast_sip_session_media *session_media, struct ast_rtp_instance *rtp);
+
+/*!
+ * \brief Retrieve a payload from the direct media peer snapshot
+ *
+ * \note The returned payload has its reference count increased.
+ */
+struct ast_rtp_payload_type *ast_sip_session_media_get_direct_media_payload(
+	const struct ast_sip_session_media *session_media, int payload);
 
 /*!
  * \brief Structure which contains read callback information
--- a/include/asterisk/rtp_engine.h
+++ b/include/asterisk/rtp_engine.h
@@ -304,6 +304,8 @@ struct ast_rtp_payload_type {
 	/*! If asterisk_format is set, this is the internal
 	 * asterisk format represented by the payload */
 	struct ast_format *format;
+	/*! Raw SDP format parameters associated with the payload */
+	char *fmtp;
 	/*! Is this an Asterisk value */
 	int asterisk_format;
 	/*! Actual internal RTP specific value of the payload */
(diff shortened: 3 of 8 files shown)


jcolp:
Was an LLM used for creating this? Additionally this should have test coverage added if possible.

jcolp:
I would also ask: Disregarding DTMF, are codecs that use dynamic payloads working for direct media without this change?

remiquezada:
Yes I did use an LLM as an assistant while investigating and making the patch.  Before submitting the PR I tested the patch with direct_media enabled and can confirm DTMF is working now.  I can look into adding test coverage.

jcolp:
We require disclosure of LLM usage:

https://docs.asterisk.org/Development/Policies-and-Procedures/AI-Policy/

You should also ensure you are aware of our code contribution process:

I would also appreciate an answer to my question regarding non-DTMF, because if that also does not work then this change is not really fixing things over all - and if it is working, then this is likely being done in the wrong way.

remiquezada:
I just tried with opus codec and when direct media is enabled it does not work with or without this change.

jcolp (changes requested):
I'm marking this as request changes as based on your own testing, this solves it for your specific DTMF use case but illustrates a larger issue regarding dynamic payloads and direct media.

We need to exchange dynamic payloads across for all codecs (including DTMF) alongside any attributes, provided the negotiated ones are compatible (I'm thinking of opus).

remiquezada:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

remiquezada:
OpenAI Codex (GPT-5) was used to analyze the RTP payload-mapping behavior, assist with the implementation, and develop the unit test.  

With this PR both cases DTMF and opus codec calls are now working with reINVITE.

jcolp (commented):
This change also needs direct media testsuite coverage to verify the resulting SDP for each.

remiquezada:
testsuite-test-pr: 151

include/asterisk/res_pjsip_session.h
```diff
@@ -77,6 +79,8 @@ typedef int (*ast_sip_session_media_write_cb)(struct ast_sip_session *session, s
 struct ast_sip_session_media {
 	/*! \brief RTP instance itself */
 	struct ast_rtp_instance *rtp;
+	/*! \brief Snapshot of the common direct media transmit payload mappings */
```
jcolp: This is a public structure definition, this needs to go at the end and the comment should explain when this is set

include/asterisk/rtp_engine.h
```diff
@@ -304,6 +304,8 @@ struct ast_rtp_payload_type {
 	/*! If asterisk_format is set, this is the internal
 	 * asterisk format represented by the payload */
 	struct ast_format *format;
+	/*! Raw SDP format parameters associated with the payload */
```
jcolp: Why does this use the raw instead of the version that can be created from the underlying format like normal SDP creation does?

Additionally this is a public structure so it has to go to the end

res/res_pjsip_session.c
```diff
@@ -475,6 +481,333 @@ static int stream_destroy(void *obj, void *arg, int flags)
...
+			}
+			continue;
+		}
+
+		if (left_type->asterisk_format != right_type->asterisk_format
+			|| left_type->rtp_code != right_type->rtp_code
+			|| left_type->sample_rate != right_type->sample_rate
+			|| (left_type->asterisk_format
+				&& ast_format_cmp(left_type->format, right_type->format) == AST_FORMAT_CMP_NOT_EQUAL)
+			|| strcmp(S_OR(left->fmtp[payload], ""), S_OR(right->fmtp[payload], ""))) {
```
jcolp: The ast_format_cmp should already handle the underlying fmtp comparison with the attributes being contained on the format. If you want it to be the same, you'd change the ast_format_cmp to != AST_FORMAT_CMP_EQUAL

channels/chan_pjsip.c
```diff
@@ -424,26 +424,37 @@ static int check_for_rtp_changes(struct ast_channel *chan, struct ast_rtp_instan
...
 {
 	int changed = 0, position = -1;
+	int payloads_changed;
 
 	if (media->rtp) {
 		position = rtp_find_rtcp_fd_position(session, media->rtp);
 	}
 
 	if (rtp) {
+		payloads_changed = ast_sip_session_media_set_direct_media_payloads(media, rtp);
```
jcolp: Why do you need to store this independently in payloads_changed?

include/asterisk/res_pjsip_session.h
```diff
@@ -129,6 +133,27 @@ struct ast_sip_session_media {
...
+/*!
+ * \brief Replace the direct media payload snapshot with mappings common to two RTP instances
+ *
+ * \param session_media Session media to update
+ * \param rtp Direct media peer, or NULL to clear the snapshot
+ *
+ * \retval 1 Snapshot changed
+ * \retval 0 Snapshot did not change
+ * \retval -1 Allocation failure
+ */
```
jcolp, line 149: These will need to have doxygen for the versions they were added in

include/asterisk/rtp_engine.h
```diff
@@ -1699,6 +1701,18 @@ int ast_rtp_codecs_payloads_set_rtpmap_type_rate(struct ast_rtp_codecs *codecs,
...
+ * \brief Record tx payload format parameters from an a=fmtp SDP line
+ *
+ * \param codecs Codecs structure to update
+ * \param payload Numerical payload from the a=fmtp SDP line
+ * \param fmtp Format parameters without the payload number
+ *
+ * \retval 0 on success
+ * \retval -1 on failure
+ */
+int ast_rtp_codecs_payload_set_fmtp(struct ast_rtp_codecs *codecs, int payload, const char *fmtp);
```
jcolp: See prior comment about why this has to exist

include/asterisk/rtp_engine.h
```diff
@@ -1751,6 +1765,20 @@ enum ast_media_type ast_rtp_codecs_get_stream_type(struct ast_rtp_codecs *codecs
...
 
+/*!
+ * \brief Retrieve a transmit payload mapping by payload type
+ *
+ * \param codecs Codecs structure to look in
+ * \param payload Numerical payload to look up
+ *
+ * \return Payload information.
+ * \retval NULL if payload does not exist.
+ *
```
jcolp: This will need doxygen for version

include/asterisk/rtp_engine.h
```diff
@@ -2032,6 +2060,19 @@ int ast_rtp_codecs_payload_set_rx(struct ast_rtp_codecs *codecs, int code, struc
...
+ * \param codecs Codecs structure to manipulate
+ * \param type Payload mapping to install
+ *
+ * \retval 0 Payload mapping was installed
+ * \retval -1 Invalid payload mapping
+ *
+ * \note The codecs structure takes its own reference to type.
+ */
+int ast_rtp_codecs_payload_set_rx_type(struct ast_rtp_codecs *codecs, struct ast_rtp_payload_type *type);
```
jcolp: Why is a set_rx_type needed for this work?

main/rtp_engine.c
```diff
@@ -310,6 +310,7 @@ static void rtp_payload_type_dtor(void *obj)
 	struct ast_rtp_payload_type *payload = obj;
 
 	ao2_cleanup(payload->format);
+	ast_free(payload->fmtp);
```
jcolp: See prior comment about fmtp in general

res/res_pjsip_sdp_rtp.c
```diff
@@ -760,6 +760,41 @@ static pjmedia_sdp_attr* generate_fmtp_attr(pj_pool_t *pool, struct ast_format *
 	return attr;
 }
 
+static struct ast_rtp_payload_type *find_direct_media_payload(
```
jcolp: Is there any reason that things were built this way instead of adding a common function to rtp_engine.c that would essentially create a new ast_rtp_codecs which represented the direct media payloads, that would then just be placed on the session_media. If present then use that, keeping the existing logic and minimizing the changes here?

res/res_pjsip_sdp_rtp.c
```diff
@@ -2000,6 +2035,8 @@ static int create_outgoing_sdp_stream(struct ast_sip_session *session, struct as
 
 	for (index = 0; index < ast_format_cap_count(caps); ++index) {
 		struct ast_format *format = ast_format_cap_get_format(caps, index);
+		struct ast_format *sdp_format = format;
```
jcolp: Why two variables for this?

res/res_pjsip_sdp_rtp.c
```diff
@@ -2079,33 +2130,77 @@ static int create_outgoing_sdp_stream(struct ast_sip_session *session, struct as
...
+						ast_rtp_instance_get_codecs(session_media->rtp), 0, NULL, index);
+				}
 				if (rtp_code == -1) {
+					ao2_cleanup(direct_payload);
 					continue;
 				} else if ((attr = generate_rtpmap_attr(session, media, pool, rtp_code, 0, NULL, index))) {
 					media->attr[media->attr_count++] = attr;
 				}
+				ao2_cleanup(direct_payload);
+			} else if (direct_media_enabled) {
```
jcolp: This feels like a lot of duplication, which goes back to my question prior to this about these new structures and function for getting it

res/res_pjsip_sdp_rtp.c
```diff
@@ -2079,33 +2130,77 @@ static int create_outgoing_sdp_stream(struct ast_sip_session *session, struct as
...
+					ao2_cleanup(direct_payload);
+				}
 			} else if (build_dtmf_sample_rates) {
 				/*
 				 * Walk through the possible bitrates for the RFC 2833/4733 digits and generate the rtpmap
 				 * attributes.
 				 */
 				int i, found_default_offer = 0;
 				for (i = 0; i < AST_VECTOR_SIZE(&sample_rates); i++) {
+					unsigned int sample_rate = AST_VECTOR_GET(&sample_rates, i);
```
jcolp: This seems like an unrelated change?

res/res_pjsip_session.c
```diff
@@ -62,6 +63,11 @@
 /* Most common case is one audio and one video stream */
 #define DEFAULT_NUM_SESSION_MEDIA 2
 
+struct ast_sip_session_media_rtp_payloads {
```
jcolp: This is just a continuation of my prior question about why there's a new structure

res/res_pjsip_session.c
```diff
@@ -475,6 +481,333 @@ static int stream_destroy(void *obj, void *arg, int flags)
...
+			return 0;
+		}
+	}
+
+	return 1;
+}
+
+#define DTMF_EVENT_COUNT 256
+
+static int direct_media_dtmf_fmtp_parse(const char *fmtp, unsigned char events[DTMF_EVENT_COUNT])
```
jcolp: We're not embedding a parser for DTMF fmtp like this...

Can this be accomplished using existing functionality? We already have to negotiate between Asterisk and the remote side, so we already do some negotiation.

tests/test_res_pjsip_session_payload.c
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * \brief PJSIP direct media payload tests
+ *
+ * \author Remi Quezada
+ */
+
+/*** MODULEINFO
+	<depend>TEST_FRAMEWORK</depend>
+	<depend>pjproject</depend>
+	<depend>res_format_attr_opus</depend>
```
jcolp, line 27: This is a build time dependency and not a runtime dependency, it needs to be added to the modules .requires

tests/test_res_rtp.c
```diff
@@ -26,6 +26,7 @@
 
 /*** MODULEINFO
 	<depend>TEST_FRAMEWORK</depend>
+	<depend>res_format_attr_opus</depend>
```
jcolp, line 29: This is a build time dependency, it needs to be added to .requires

jcolp (changes requested):
Deeper look.

tests/test_res_pjsip_session_payload.c
remiquezada: I have res_format_attr_opus already listed in the module's .require alongside res_pjsip_session and res_rtp_asterisk.  Do you just want me to remove the <depend>res_format_attr_opus</depend> build-time declaration and keep only the runtime dependency?

jcolp: Ah so it is there, I'll mark this as resolved. They should both exist.

res/res_pjsip_session.c
remiquezada: I checked res_format_attr_opus.c, and opus_interface currently has no format_cmp callback. In that case, ast_format_cmp() returns AST_FORMAT_CMP_EQUAL for the same codec even when its attributes differ. 

This comparison is used to detect changes to the stored direct-media mappings, including attribute-only changes that require updated SDP. Removing the separate attribute comparison could therefore miss an Opus attribute change.

Would you prefer adding an Opus format_cmp implementation, or another approach for detecting those changes?

jcolp: The format_cmp approach is the defined way within the project to do such things, so it should be implemented/used.

res/res_pjsip_session.c
remiquezada: I haven't found an existing helper that intersects the supported event ranges. The normal SDP generation path appears to advertise 0-16.  The parser was intended to preserve a common supported range for direct media, for example when a peer advertises only 0-15.  

Is there existing event-range negotiation functionality I should reuse? Or do want me to use existing DTMF handling and address the event-range negotiation separately?

jcolp: There isn't as we just use fixed values. Actually supporting event-range negotiation should be separate.

res/res_pjsip_sdp_rtp.c
remiquezada: That makes sense. I’ll move the common payload-mapping construction into rtp_engine.c and represent the result with ast_rtp_codecs, stored on the session media. SDP generation can then select that structure for direct media and reuse the existing generation path, removing the custom container and duplicated logic.

I’ll preserve the same payload mappings on both legs and the negotiated codec attributes.

res/res_pjsip_sdp_rtp.c
remiquezada: This will get resolved as part of the ast_rtp_codecs restructuring discussed.

res/res_pjsip_sdp_rtp.c
remiquezada: This will get resolved as part of the ast_rtp_codecs restructuring discussed.

res/res_pjsip_sdp_rtp.c
remiquezada: Yes is unrelated to fix.

tests/test_res_rtp.c
remiquezada: .requires now added.

include/asterisk/res_pjsip_session.h
remiquezada: versions added.

remiquezada:
I ran those 3 tests locally and all passed.  There seems to be an issue with the Github CI environment.

PR #2019, opened 2026-07-09
res_http_websocket: Unref session when it fails to establish.

### LINKED ISSUE #2020: [bug]: Possible leak of inbound websocket session
### Severity

Minor

### Versions

22.10.1

### Components/Modules

res_http_websocket

### Operating Environment

Debian 13

### Frequency of Occurrence

None

### Issue Description

There are two error cases in res_http_websocket, happening after an ast_websocket session is allocated, where an unref of the session is missing, resulting on a leak.

The failure to generate a session id case is probably only theorical as you need to be totally OOM to reach it.

On the other hand, the second case where the ast_websocket_pre_callback session_attempted callback fails can be reached when using ari_websockets, as it defines that callback and there are two possibilities for it to fail.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

escatg:
Two error paths in res_http_websocket could leak an ast_websocket session by failing to drop a reference after the session had been allocated.

The first occurs when session ID generation fails. While this is largely theoretical, as it would require an out-of-memory condition, the session reference should still be released correctly.

The second occurs when the ast_websocket_pre_callback session_attempted callback returns an error. This path is reachable through ari_websockets, which implements this callback and can legitimately fail under certain conditions.

Fixes: #2020

### DIFF
--- a/res/res_http_websocket.c
+++ b/res/res_http_websocket.c
@@ -1089,6 +1089,7 @@ int AST_OPTIONAL_API_NAME(ast_websocket_uri_cb)(struct ast_tcptls_session_instan
 			ast_log(LOG_WARNING, "WebSocket connection from '%s' could not be accepted - failed to generate a session id\n",
 				ast_sockaddr_stringify(&ser->remote_address));
 			ast_http_error(ser, 500, "Internal Server Error", "Allocation failed");
+			ao2_ref(session, -1);
 			ao2_ref(protocol_handler, -1);
 			return 0;
 		}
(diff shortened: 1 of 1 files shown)


escatg:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

gtjoseph:
You may want to open an issue for this and reference it in the PR description and the commit message.  This way it'll get a mention at the top of the ChangeLog for the next release.

See:

PR #2029, opened 2026-07-16
app_queue.c: Subscribe to member device state per-device via a pool

### LINKED ISSUE #1972: [bug]: app_queue: device_state_cb floods stasis/m:devicestate:all taskprocessor on reload of queues with many members

PujaGediya:
device_state_cb() subscribed once to the aggregate devicestate:all topic
and iterated every queue and every member for each device-state message.
Reloading a queue with thousands of members floods that single
subscription's taskprocessor (stasis/m:devicestate:all) with the
per-member pause/avail hints app_queue publishes, tripping the 500 high
water mark and raising the taskprocessor congestion alert.

Instead, subscribe to each watched device's own topic with a
reference-counted pool subscription (device_state_watchers), created when
the first member starts watching a state_interface and released when the
last member stops. app_queue now receives only the events for devices a
member actually watches, and those events are spread across the stasis
pool threads instead of being serialized through a single taskprocessor.
Device states no member watches - including the Queue:..._pause_... hints
app_queue itself publishes during a reload - are never delivered, so the
reload flood disappears at the stasis routing layer.

publish_queue_avail() now deduplicates the Queue:..._avail hint by last
published value, so a burst of member updates collapses to one publish per
real transition. This also keeps queue chaining working safely: a member
whose state_interface is another queue's Queue:..._avail device is watched
like any other device, and the dedup bounds the resulting callback chain
so even a cyclic chain converges instead of looping. Hint members
(state_interface "hint:...") are left to extension_state_cb() as before.

Also fix a race in rt_handle_member_record(): when a realtime reload
changes a member's state_interface, start watching the new device before
storing it on the member (and before unwatching the old), closing the
window where m->state_interface referred to a device not yet watched.

Resolves: #1972

UserNote: app_queue now subscribes to member device state per-device using
a stasis pool, so reloading queues with large member counts no longer
floods the stasis/m:devicestate:all taskprocessor past its high water
mark, and device-state updates are processed in parallel across the pool.

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -1750,8 +1750,6 @@ static int shared_lastcall;
 /*! \brief queuerules.conf [general] option */
 static int realtime_rules;
 
-/*! \brief Subscription to device state change messages */
-static struct stasis_subscription *device_state_sub;
 
 /*! \brief queues.conf [general] option */
 static int negative_penalty_invalid;
@@ -2007,6 +2005,7 @@ struct call_queue {
 	unsigned int autopauseunavail:1;
 	enum empty_conditions joinempty;
 	enum empty_conditions leavewhenempty;
+	int avail_devstate;                 /*!< Last published Queue:<name>_avail device state, or -1 if never published (see publish_queue_avail()) */
 	int announcepositionlimit;          /*!< How many positions we announce? */
 	int announcefrequency;              /*!< How often to announce their position */
 	int minannouncefrequency;           /*!< The minimum number of seconds between position announcements (def. 15) */
@@ -2062,6 +2061,7 @@ static void update_realtime_members(struct call_queue *q);
 static struct member *interface_exists(struct call_queue *q, const char *interface);
 static int set_member_paused(const char *queuename, const char *interface, const char *reason, int paused);
 static int update_queue(struct call_queue *q, struct member *member, int callcompletedinsl, time_t starttime);
+static int num_available_members(struct call_queue *q);
 
 static struct member *find_member_by_queuename_and_interface(const char *queuename, const char *interface);
 /*! \brief sets the QUEUESTATUS channel variable */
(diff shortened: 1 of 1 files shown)


PujaGediya:
cherry-pick-to: 20

PujaGediya:
### AI usage disclosure

This contribution was prepared with AI assistance (Claude Code / Anthropic Claude), used to investigate the root cause in `device_state_cb()` and the device-state stasis path, draft the per-device pool subscription change in `apps/app_queue.c`, and verify behavior. I have reviewed the change in full, understand it, and can debug and own it.

Tested on a preprod server: reloading a queue with thousands of members no longer pushes the `stasis/m:devicestate:all` taskprocessor past its high-water mark; device-state updates are processed in parallel across the pool. This supersedes the index-based approach in #1973, reworked to the per-device pool subscription approach suggested by @jcolp.

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+ * Using pool subscriptions also lets concurrent device-state changes for
+ * different members be processed in parallel instead of serializing everything
+ * through one stasis/m:devicestate:all thread.
+ *
+ * The subscription is created when the first member starts watching a device
+ * and released (via the ao2 destructor) when the last member stops watching.
+ * Hint-based members (state_interface starts with "hint:") are excluded here;
+ * they are handled by extension_state_cb() instead.
+ */
+static struct ao2_container *device_state_watchers;
```
jcolp, line 2826: Why did you opt for the approach of having a container of subscriptions, instead of a subscription on each agent?

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+
+#define DEVICE_STATE_WATCHER_BUCKETS 547
```
jcolp, line 2828: Where did 547 come from?

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+
+struct device_state_watcher {
+	struct stasis_subscription *sub;    /*!< Pool subscription to this device's stasis topic */
+	int count;                          /*!< Number of members watching this device */
```
jcolp, line 2832: unsigned int

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+	char interface[AST_CHANNEL_NAME];   /*!< Normalized device-state identifier */
```
jcolp, line 2833: Each watcher could also have a container of objects containing references to the queue+member, eliminating the need to iterate all queues and members on each device state update

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+	 * device_state_watchers_shutdown() uses stasis_unsubscribe_and_join() for
+	 * every subscription it owns before queue memory is freed, so any race window
+	 * there is handled at the module-unload layer, not here.
+	 */
+	if (watcher->sub) {
+		stasis_unsubscribe(watcher->sub);
+	}
+}
+
+static int device_state_watcher_hash_fn(const void *obj, const int flags)
```
jcolp, line 2906: This can use the AO2_STRING_FIELD_CASE_HASH_FN macro

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+	case OBJ_SEARCH_OBJECT:
+	default:
+		watcher = obj;
+		interface = watcher->interface;
+		break;
+	}
+	return ast_str_case_hash(interface);
+}
+
+static int device_state_watcher_cmp_fn(void *obj, void *arg, int flags)
```
jcolp, line 2924: This can use the AO2_STRING_FIELD_CASE_CMP_FN macro

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+	 * stasis_subscribe_pool() only allocates a serializer and enqueues a
+	 * subscription-change message asynchronously; it does not call back into
+	 * device_state_watchers, so holding the lock here is safe.
+	 */
+	ao2_lock(device_state_watchers);
+	watcher = ao2_find(device_state_watchers, interface, OBJ_KEY | OBJ_NOLOCK);
+	if (watcher) {
+		++watcher->count;
+	} else if ((watcher = ao2_alloc_options(sizeof(*watcher), device_state_watcher_destructor,
+			AO2_ALLOC_OPT_LOCK_NOLOCK))) {
```
jcolp, line 2975: Add a comment stating why the watcher doesn't need a lock

apps/app_queue.c
```diff
@@ -2800,14 +2800,300 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+ * Per the queue_avail hint convention: AST_DEVICE_INUSE means no member is
+ * available, AST_DEVICE_NOT_INUSE means at least one member is available.
+ *
+ * \note Callers must hold the queue lock (q->avail_devstate is protected by it).
+ */
+static void publish_queue_avail(struct call_queue *q, int available)
+{
+	int state = available ? AST_DEVICE_NOT_INUSE : AST_DEVICE_INUSE;
+
+	if (q->avail_devstate == state) {
```
jcolp, line 3082: Is storing of the avail devstate and this comparison still required after this change went in:

https://github.com/asterisk/asterisk/commit/bd5c4093714e53738c9c0bfec8d2b0179ae7fbb4

And if so - do you know why?

jcolp (changes requested):

jcolp:
I may also have additional comments, I'm digesting this and thinking things through.

apps/app_queue.c
PujaGediya: Because a device is rarely watched by exactly one member. The same agent in N queues is N distinct struct member objects, several members can share a state_interface, and the Local channel normalization collapses Local/x@y/1 and Local/x@y/2 onto one device. A subscription per member would give:

one stasis pool serializer per member — 5k members means 5k serializers, which is the same taskprocessor pressure this PR is trying to remove, just moved;
N dispatches of device_state_cb() for a single device change, each doing the full scan;
subscribe/unsubscribe churn on every reload, because the member set is torn down and rebuilt via kill_dead_members() even when the underlying devices are unchanged. Reference counting by device means an unchanged device keeps its subscription straight across the reload.
One refcounted subscription per normalized device gives one subscription and one dispatch per device. I've added this rationale to the block comment.

It also lines up with your suggestion below: the watcher is already the device→interested-parties index, it just doesn't carry the parties yet.

apps/app_queue.c
PujaGediya: A prime picked for the scale that motivated the change — a few thousand distinct member devices lands around 10 entries per bucket — following the existing MAX_CALL_ATTEMPT_BUCKETS (353) idiom in this file. It isn't derived from anything measured, so if you'd rather have a different value I'll change it.

apps/app_queue.c
PujaGediya: Done

apps/app_queue.c
PujaGediya: Agreed — that's the right next step, and the watcher is the natural place for it. Two things I want to get right first:

Where watch/unwatch happen. Today watch_state_interface() runs in create_queue_member(), before the member is linked into any queue, and unwatch_state_interface() runs in the member's ao2 destructor. Neither has a queue in hand. member_add_to_queue() / member_remove_from_queue() are the obvious replacement hooks since they take both, but not every path goes through them: reload_single_queue() does a raw ao2_link()/ao2_unlink() pair when replacing an existing member, and kill_dead_members() unlinks via ao2_callback(..., OBJ_UNLINK, ...). Those need routing through the pair.

apps/app_queue.c
PujaGediya: no, the comparison isn't needed, but the -1 sentinel is, so it becomes unsigned int avail_hint_seeded:1

apps/app_queue.c
jcolp: one stasis pool serializer per member — 5k members means 5k serializers, which is the same taskprocessor pressure this PR is trying to remove, just moved;

Can you elaborate further? In operation this would actually spread device state updates across more stasis taskpool threads.

jcolp: Oh and for:

subscribe/unsubscribe churn on every reload, because the member set is torn down and rebuilt via kill_dead_members() even when the underlying devices are unchanged.

There is nothing to say that a subscription couldn't be reused as part of the reload process

apps/app_queue.c
jcolp: I think add and remove make sense, since ultimately it's only when part of a queue that this matters and for even the existing logic it is examined.

apps/app_queue.c
PujaGediya: You're right and my phrasing was wrong. ast_taskpool_serializer() creates a taskprocessor backed by the shared pool, not a thread, so N serializers spread work across the pool's threads — per-member subscriptions would give more parallelism, not less.

What I should have said is object count, not thread pressure: each subscription is a stasis_subscription, a serializer registered as an ast_taskprocessor in the global container, and an entry in the per-device topic's subscriber vector. Keyed by member that's one per member; keyed by device it's one per distinct device, several times fewer on a large install, and it keeps core show taskprocessors usable.

The other difference is on the publisher side: publish_msg() iterates the subscriber vector under ao2_lock(topic), so N members sharing a device means the thread that changed the device state does N dispatches under that lock before it continues. One subscription per device makes that one, with the fan-out happening off that thread.

Neither is decisive. If you'd rather have per-member subscriptions I'll rework it — see the other thread, because per-member actually composes better with your index suggestion than the container does.

jcolp: I think a small test implementation should be done to understand the actual ramifications instead of guessing or assuming. This is what I've done with all my performance work, and in some cases the result has been the opposite of what you'd expect.

apps/app_queue.c
PujaGediya: Agreed, member_add_to_queue() / member_remove_from_queue() it is. Two sites bypass them today and would need routing through: reload_single_member() does a raw ao2_link()/ao2_unlink() pair when replacing a member, and kill_dead_members() unlinks via ao2_callback(..., OBJ_UNLINK, ...).

Moving it there also resolves the reference-cycle problem whichever subscription model we pick: while watch/unwatch are driven by the member destructor, anything holding a member or queue reference prevents that destructor from running. Driven from add/remove, the watcher can hold real references — and the same applies to a per-member subscription carrying the member as callback data, which is what makes per-member viable at all, since you can't stasis_unsubscribe_and_join() from a destructor the callback itself can reach.

Which would you prefer for this PR — per-member subscriptions, or the shared container plus the index?

apps/app_queue.c
PujaGediya: Built all four and measured. You were right — the result was the opposite of what I argued.

5000 member entries held constant, burst of 5000 device state changes, only `app_queue.so` swapped, `AST_DEVMODE` off (its per-dispatch statistics would have biased this toward my own argument). Wall-clock for the system to finish processing the burst:

| topology | devices | master | this PR (per-device) | per-device + index | per-member |
|---|---|---|---|---|---|
| 1 q × 5000 m | 5000 (1:1) | 3.2–3.6 s | 3.5–3.8 s | 0.03–0.04 s | 0.04 s |
| 5 q × 1000 m | 1000 (5:1) | 3.8 s | 1.87 s | 0.25 s | 0.045 s |
| 20 q × 250 m | 1000 (5:1) | 6.7–7.1 s | 2.6–2.9 s | 0.24 s | 0.03–0.05 s |
| 100 q × 50 m | 1000 (5:1) | 8.0–8.2 s | 3.3–3.7 s | 0.02–0.25 s | 0.042 s |
| 500 q × 10 m | 1000 (5:1) | 10.1 s | 4.5 s | 0.014 s | 0.039 s |
| 1000 q × 5 m | 1000 (5:1) | 13.6 s | 5.8 s | 0.016 s | 0.034 s |

Object count at 20 × 250: 14 / 1013 / 1013 / 5013 taskprocessors; RSS 85 / 88 / 88 / 92 MB.

**The scan was the dominant cost, not the subscription shape** — your index comment is what mattered, worth ~97% of the available gain, with no change in object count. The scanning variants degrade ~4× as queue count rises because they walk every queue on every message; the indexed ones are flat across a 50× range. And with one device per member this PR as written buys nothing over master: it pays for the subscriptions and still scans.

My serializer claim was wrong as you said, and the object-count argument I fell back on only holds when members share a device — at 1:1 all variants create the same 5013 taskprocessors and per-member uses the least RSS.

I've switched the branch to per-member. It and per-device+index are within ~0.2 s, so this isn't mainly a performance choice: dropping the container also removes the mutex held across `ast_device_state_topic()` and `stasis_subscribe_pool()`, an unsynchronised global pointer at unload, and a refcount imbalance on allocation failure. It retires the `547` question too. Watch/unwatch are driven from `member_add_to_queue()`/`member_remove_from_queue()` per your other comment, with `reload_single_member()` and `kill_dead_members()` routed through the same pair.

Building it turned up four subscription-lifetime defects — the main one being that a SELECTIVE subscription must also accept `stasis_subscription_change_type()` or the unsubscribe final message never reaches the callback and the callback data leaks (`module unload` went 10.03 s → 0.02 s). All fixed and verified; re-running the sweep afterwards left per-member unchanged.

apps/app_queue.c
```diff
@@ -2800,91 +2800,315 @@ static int is_member_available(struct call_queue *q, struct member *mem)
...
+	ao2_lock(watch->q);
+	/*
+	 * The unsubscribe is non-joining, so a message queued before it can still
+	 * arrive here.  By then the member may have been removed from the queue, or
+	 * re-pointed at a different device by a realtime update -- in either case
+	 * mem->devstate_sub no longer refers to this subscription, and applying the
+	 * old device's state would clobber the current one.
+	 */
+	if (watch->mem->devstate_sub == sub) {
+		update_status(watch->q, watch->mem, dev_state->state);
```
jcolp, line 3108: I'm starting to look at this again and as I was just getting my feeling for this I believe I've found a fundamental issue that is stems from having multiple subscriptions and requires work.

This function calls update_status with the lock for "watch->q" held.
In update_status this results in update_queue being called if the status is device not in use.
In update_queue if shared_lastcall is enabled the configured queues are iterated.
The lock of each queue is taken as it iterates.

If this happens across two threads with differing queues both configured for shared_call, it is possible for a deadlock to occur.

Previously this wouldn't have been possible because there would only be a single thread.

I'll continue to review, but that needs resolving.

jcolp (changes requested):

PR #2030, opened 2026-07-16
pjsip: Eliminate some unique taskprocessors.

jcolp:
When placing outbound calls each call would get its own taskprocessor for things related to the call. In practice this is overkill as few things actually occur. Inbound calls on the other hand already use one of the fixed number of distributor taskprocessors. This also occurred for OPTIONS requests with each AOR having its own taskprocessor.
    
This change moves both to using a distributor taskprocessor instead.

### DIFF
--- a/include/asterisk/res_pjsip.h
+++ b/include/asterisk/res_pjsip.h
@@ -2051,6 +2051,32 @@ struct ast_taskprocessor *ast_sip_create_serializer_group(const char *name, stru
 struct ast_taskprocessor *ast_sip_get_distributor_serializer(pjsip_rx_data *rdata);
 
 /*!
+ * \brief Determine the distributor serializer for the SIP dialog.
+ * \since 20.21.0
+ * \since 22.11.0
+ * \since 23.5.0
+ *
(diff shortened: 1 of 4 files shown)


jcolp:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
I'm marking this as draft, as George had the thought about moving PJSIP OPTIONS as well since each AOR results in its own taskprocessor which like this shouldn't really be needed as they are generally infrequent and quick tasks.

mbradeen:
seems pretty straightforward

gtjoseph (approved):

bkford (approved):

jcolp:
Hrm, I'm uncertain if these sporadic test failures are the result of this change or not...

gtjoseph:
> Hrm, I'm uncertain if these sporadic test failures are the result of this change or not...

I don't _think_ so.  I did a stress test yesterday with this PR and didn't see any issues.  

I'll trigger the cherry-pick tests again and see what happens.

mbradeen (approved):

PR #998, opened 2024-11-18
Enabling Real Time Text (RTT) for PJSIP library

bnandaku:
Implement changes based on archived issue: https://issues-archive.asterisk.org/ASTERISK-28654

I'm the original author and I've signed CLA.

Changes are:

    chan_pjsip - added write_text function to the pjsip_tech driver
    Added case for AST_FRAME_TEXT in chan_pjsip_write_stream
    Added case to handle text in chan_pjsip_write
    pjsip/dialplan_functions.c: added logic to handle media for TEXT
    res_pjsip.h: added variables to deal with endpoint media
    channel.c: added logic to add a stream_num for a frame
    frame.c: added code to handle and fix issues with red.
    res_pjsip_sdp_rtp.c: added logic to setup SDP.
    Added logic to handle tos and cos text
    Added logic to RED for red enabled
    res_pjsip_session.c: aded logic for max_text_streams
    res_rtp_asterisk.c: Added logic to hanle Red and dealing with
    a repeating RTP issue
    pjsip_configuration.c: Added logic to handle text for PJSIP library

Original code-review can be found here:
http://lists.digium.com/pipermail/asterisk-code-review/2019-December/094886.html

PR #2031, opened 2026-07-17
Introduces support for real-time text conversation over RTP (RFC 4103).

### LINKED ISSUE #997: [improvement]: PJSIP support for Real Time Text
### Improvement Description

Enable PJSIP to support RTT with t140 and t140red codes

### LINKED ISSUE #1968: [bug]: Disabled audio stream stops RTP processing also for text stream in native bridging
### Severity

Trivial

### Versions

master

### Components/Modules

chan_pjsip

### Operating Environment

Ubuntu Linux 24.04

### Frequency of Occurrence

Constant

### Issue Description

Disabled audio stream stops RTP processing also for text stream in native bridging.

This is related to the RTT PR at https://github.com/asterisk/asterisk/pull/1128

A disabled audio stream (m-line audio with port 0) in SDP stops RTP processing also for text stream in native bridging.

Expected outcome would be that the disabled audio stream did not cause problems for the text stream.

### Relevant log output

```shell
Log output will be attached, as its too long.
```

### Asterisk Issue Guidelines

henningw:
* Co-authored-by: Xenofon Karamanos <xk@gilawa.com>
* Co-authored-by: Alexios Hadjisotiriou <ah@gilawa.com>
* Co-authored-by: Mike Bradeen <mbradeen@sangoma.com>

Based on a PR from Bharat Ramaswamy Nandakumar <bharat@bramsoft.com>
* Co-authored-by: Kevin Basov <kevin.basov@cgi.com>
* Co-authored-by: Sean Bright <sean@seanbright.com>

Previous changes from PR #998
* chan_pjsip - added write_text function to the pjsip_tech driver
* Added case for AST_FRAME_TEXT in chan_pjsip_write_stream
* Added case to handle text in chan_pjsip_write
* pjsip/dialplan_functions.c: added logic to handle media for TEXT
* res_pjsip.h: added variables to deal with endpoint media
* channel.c: added logic to add a stream_num for a frame
* frame.c: added code to handle and fix issues with red.
* res_pjsip_sdp_rtp.c: added logic to setup SDP.
* Added logic to handle tos and cos text
* Added logic to RED for red enabled
* res_pjsip_session.c: aded logic for max_text_streams
* res_rtp_asterisk.c: Added logic to handle Red and dealing with a repeating RTP issue
* pjsip_configuration.c: Added logic to handle text for PJSIP library

New additions:
* Maintain ABI compatibility
* Add alembic and docs
* Adjust length based on stream
* refactoring existing SDP attribute logic to use the existing framework
* fix redundant generation SDP payload mismatch
* remove not needed t140 debug fix, the t140 lengths are reset in the red case
* extend error logging and add some debug messages
* add more comment documentation regarding edge cases and workarounds
* add missing reference counting decrement for scheduler deletion
* add native bridging support for real-time RTP support
* add streaming support for real-time text
* additional logs and checks for RED generations overflow
* fix transmission of empty T140 frames
* add example settings for real-time text endpoint
* chan_sip compatiblity for real-time text, add back write, unused for chan_pjsip
* properly remove red scheduler reference
* fix non-codec formats addition for RTT
* count samples for text frames as well
* add text media type to RTP dialplan functions
* Handle text frames in frame lifecycle helpers
* Rework T.140/RED redundancy de-packetization
* Route real-time text frames through stream logic
* fix regression related to SIP MESSAGE text bridging
* fix RTP processing for text stream in native bridging
* fix forwarding for empty packets needed for RED idle transmissions

UserNote: This introduces support for real-time text conversation over RTP (RFC 4103). To enable it you need to add t140 and preferable also red as supported codecs. You can configure this functionality by three new parameters in the endpoint section of pjsip.conf: tos_text, cos_text and max_text_streams. If you are using a database, please have a look to the upgrade notes.

UpgradeNote: This change introduce a schema modification for the ps_endpoint table by alembic. If you don't use the scripts to update your database, please adapt the table manually to the new structure. New fields were tos_text, cos_text and max_text_streams.

Resolves: #997
Resolves: #1128
Resolves: #1968
Resolves: #2031

### DIFF
--- a/bridges/bridge_native_rtp.c
+++ b/bridges/bridge_native_rtp.c
@@ -69,6 +69,7 @@ struct rtp_glue_data {
 	struct ast_rtp_glue *cb;
 	struct rtp_glue_stream audio;
 	struct rtp_glue_stream video;
+	struct rtp_glue_stream text;
 	/*! Combined glue result of both bridge channels. */
 	enum ast_rtp_glue_result result;
 };
@@ -115,6 +116,8 @@ static void rtp_glue_data_init(struct rtp_glue_data *glue)
 	glue->audio.result = AST_RTP_GLUE_RESULT_FORBID;
 	glue->video.instance = NULL;
 	glue->video.result = AST_RTP_GLUE_RESULT_FORBID;
+	glue->text.instance = NULL;
+	glue->text.result = AST_RTP_GLUE_RESULT_FORBID;
 	glue->result = AST_RTP_GLUE_RESULT_FORBID;
 }
 
@@ -125,6 +128,7 @@ static void rtp_glue_data_destroy(struct rtp_glue_data *glue)
 	}
 	ao2_cleanup(glue->audio.instance);
 	ao2_cleanup(glue->video.instance);
+	ao2_cleanup(glue->text.instance);
 }
 
 static void rtp_glue_data_reset(struct rtp_glue_data *glue)
@@ -160,6 +164,28 @@ static struct native_rtp_bridge_channel_data *native_rtp_bridge_channel_data_all
 
 /*!
  * \internal
+ * \brief Helper function for rtp_glue_data_get for common bridging mode decision logic
+ *
+ * \param stream0 The first RTP glue stream
+ * \param stream1 The second RTP glue stream
+ * \return The combined result
+ */
+static enum ast_rtp_glue_result rtp_glue_data_get_helper(struct rtp_glue_stream stream0, struct rtp_glue_stream stream1)
+{
+	if (stream0.result == AST_RTP_GLUE_RESULT_FORBID
+		|| stream1.result == AST_RTP_GLUE_RESULT_FORBID) {
+		/* If any sort of bridge is forbidden just completely bail out and go back to generic bridging */
+		return AST_RTP_GLUE_RESULT_FORBID;
+	} else if (stream0.result == AST_RTP_GLUE_RESULT_LOCAL
+		|| stream1.result == AST_RTP_GLUE_RESULT_LOCAL) {
+		return AST_RTP_GLUE_RESULT_LOCAL;
+	} else {
+		return AST_RTP_GLUE_RESULT_REMOTE;
+	}
+}
+
+/*!
+ * \internal
  * \brief Helper function which gets all RTP information (glue and instances) relating to the given channels
  *
  * \retval 0 on success.
@@ -170,7 +196,7 @@ static int rtp_glue_data_get(struct ast_channel *c0, struct rtp_glue_data *glue0
 {
 	struct ast_rtp_glue *cb0;
 	struct ast_rtp_glue *cb1;
-	enum ast_rtp_glue_result combined_result;
+	enum ast_rtp_glue_result combined_result = AST_RTP_GLUE_RESULT_FORBID;
 
 	cb0 = ast_rtp_instance_get_glue(ast_channel_tech(c0)->type);
 	cb1 = ast_rtp_instance_get_glue(ast_channel_tech(c1)->type);
@@ -185,11 +211,15 @@ static int rtp_glue_data_get(struct ast_channel *c0, struct rtp_glue_data *glue0
 	glue0->audio.result = cb0->get_rtp_info(c0, &glue0->audio.instance);
 	glue0->video.result = cb0->get_vrtp_info
 		? cb0->get_vrtp_info(c0, &glue0->video.instance) : AST_RTP_GLUE_RESULT_FORBID;
+	glue0->text.result = cb0->get_trtp_info
+		? cb0->get_trtp_info(c0, &glue0->text.instance) : AST_RTP_GLUE_RESULT_FORBID;
 
 	glue1->cb = cb1;
 	glue1->audio.result = cb1->get_rtp_info(c1, &glue1->audio.instance);
 	glue1->video.result = cb1->get_vrtp_info
 		? cb1->get_vrtp_info(c1, &glue1->video.instance) : AST_RTP_GLUE_RESULT_FORBID;
+	glue1->text.result = cb1->get_trtp_info
+		? cb1->get_trtp_info(c1, &glue1->text.instance) : AST_RTP_GLUE_RESULT_FORBID;
 
 	/*
 	 * Now determine the combined glue result.
@@ -204,6 +234,8 @@ static int rtp_glue_data_get(struct ast_channel *c0, struct rtp_glue_data *glue0
 			glue0->audio.result = glue1->audio.result = AST_RTP_GLUE_RESULT_LOCAL;
 		}
 	}
+	ast_debug(2, "%p, From media bridging, glue audio result 1: %d, instance %p, 2: %d, instance %p\n",
+			glue0->cb->allow_rtp_remote, glue0->audio.result, &glue0->audio.instance, glue1->audio.result, &glue1->audio.instance);
 	if (glue0->video.result == glue1->video.result && glue1->video.result == AST_RTP_GLUE_RESULT_REMOTE) {
 		if (glue0->cb->allow_vrtp_remote && !glue0->cb->allow_vrtp_remote(c0, glue1->video.instance)) {
 			/* If the allow_vrtp_remote indicates that remote isn't allowed, revert to local bridge */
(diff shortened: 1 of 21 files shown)


henningw:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

henningw:
This is the squashed version of PR #1128, no changes done otherwise.

Right now we are aware of the following issues:
- completely empty "idle" RED packets are not forwarded in (it was tested in simple bridge mode). Almost empty RED "idle" packets are forwarded, e.g. when a one BOM marker is included. We still need to reproduce this in our test suite.
- switch to voice-only in SRTP mode causes problems when switched back to RTT. The RTT is not properly decrypted anymore after the switch back. In our tests we were able to successfully transmit text again after a waiting time of 3-4 seconds. This could not be reproduced in another setup that also shows the problem, so it might be also a issue in the code

mbradeen:
testsuite-test-pr: 140

bridges/bridge_native_rtp.c
```diff
@@ -212,32 +244,67 @@ static int rtp_glue_data_get(struct ast_channel *c0, struct rtp_glue_data *glue0
...
+		}
+	}
+	ast_debug(2, "%p, From media bridging, glue text result 1: %d, instance %p, 2: %d, instance %p\n",
+			glue0->cb->allow_trtp_remote, glue0->text.result, &glue0->text.instance, glue1->text.result, &glue1->text.instance);
 	/* If we are carrying video, and both sides are not going to remotely bridge... fail the native bridge */
-	if (glue0->video.result != AST_RTP_GLUE_RESULT_FORBID
-		&& (glue0->audio.result != AST_RTP_GLUE_RESULT_REMOTE
-			|| glue0->video.result != AST_RTP_GLUE_RESULT_REMOTE)) {
-		glue0->audio.result = AST_RTP_GLUE_RESULT_FORBID;
+	if (glue0->video.instance && glue1->video.instance) {
```
mbradeen, line 260: Why do we need to modify this code?

bridges/bridge_native_rtp.c
```diff
@@ -338,14 +425,18 @@ static void native_rtp_bridge_start(struct ast_bridge *bridge, struct ast_channe
...
 	case AST_RTP_GLUE_RESULT_LOCAL:
-		if (ast_rtp_instance_get_engine(glue0->audio.instance)->local_bridge) {
-			ast_rtp_instance_get_engine(glue0->audio.instance)->local_bridge(glue0->audio.instance, glue1->audio.instance);
-		}
-		if (ast_rtp_instance_get_engine(glue1->audio.instance)->local_bridge) {
-			ast_rtp_instance_get_engine(glue1->audio.instance)->local_bridge(glue1->audio.instance, glue0->audio.instance);
+		/* check if we actually have audio, video or text - in this order */
+		if (glue0->audio.instance && glue1->audio.instance) {
+			remote_rtp_bridge_start_helper(glue0->audio.instance, glue1->audio.instance);
+		} else {
```
mbradeen: if / else if / else if instead of nesting context.

bridges/bridge_native_rtp.c
```diff
@@ -491,14 +601,18 @@ static void native_rtp_bridge_stop(struct ast_bridge *bridge, struct ast_channel
...
 	case AST_RTP_GLUE_RESULT_LOCAL:
-		if (ast_rtp_instance_get_engine(glue0->audio.instance)->local_bridge) {
-			ast_rtp_instance_get_engine(glue0->audio.instance)->local_bridge(glue0->audio.instance, NULL);
-		}
-		if (ast_rtp_instance_get_engine(glue1->audio.instance)->local_bridge) {
-			ast_rtp_instance_get_engine(glue1->audio.instance)->local_bridge(glue1->audio.instance, NULL);
+		/* check if we actually have audio, video or text - in this order */
+		if (glue0->audio.instance && glue1->audio.instance) {
+			remote_rtp_bridge_stop_helper(glue0->audio.instance,  glue1->audio.instance);
+		} else {
```
mbradeen: same as above, if / else if / else if instead of nesting context.

bridges/bridge_native_rtp.c
```diff
@@ -675,21 +789,25 @@ static int native_rtp_bridge_compatible_check(struct ast_bridge *bridge, struct
 		return 0;
 	}
 
-	if (ao2_container_count(bc0->features->dtmf_hooks)
+	/* do we actually have audio, for DTMF */
```
mbradeen, line 785: what prompted these new checks?

main/bridge_channel.c
```diff
@@ -2435,6 +2436,19 @@ static void bridge_channel_handle_write(struct ast_bridge_channel *bridge_channe
...
+		 * The below logic was added to support old SIP MESSAGE based text sending.
+		 * This is still used, so we need to keep it.
+		 */
+		if (ast_format_cmp(fr->subclass.format, ast_format_t140) == AST_FORMAT_CMP_EQUAL ||
+				ast_format_cmp(fr->subclass.format, ast_format_t140_red) == AST_FORMAT_CMP_EQUAL) {
+					ast_debug(1, "Sending TEXT frame with default method to '%s': %*.s\n",
+			ast_channel_name(bridge_channel->chan), fr->datalen, (char *)fr->data.ptr);
+			goto T140_RED;
+		}
+		ast_debug(1, "Sending TEXT frame with old (SIP MESSAGE)  method to '%s': %*.s\n",
```
mbradeen: I think this can say 'via SIP MESSAGE' without saying 'old'

main/channel.c
```diff
@@ -3716,7 +3717,11 @@ static struct ast_frame *__ast_read(struct ast_channel *chan, int dropaudio, int
 			default:
 				break;
 			}
-		} else if (f->frametype == AST_FRAME_VOICE || f->frametype == AST_FRAME_VIDEO) {
+		/* Text messages over SIP MESSAGE have no stream codec topology, leaving f->subclass.format as NULL.
+		 * We must bypass this block for standalone text frames to prevent them from being dropped.
+		 */
+		} else if (f->frametype == AST_FRAME_VOICE || f->frametype == AST_FRAME_VIDEO ||
+					( f->frametype == AST_FRAME_TEXT && f->subclass.format)) {
```
mbradeen: space after (

main/channel.c
```diff
@@ -3765,7 +3774,11 @@ static struct ast_frame *__ast_read(struct ast_channel *chan, int dropaudio, int
 			/* Since this channel driver does not support multistream determine the default stream this frame
 			 * originated from and update the frame to include it.
 			 */
-			if (f && (f->frametype == AST_FRAME_VOICE || f->frametype == AST_FRAME_VIDEO)) {
+			/* Text messages based on SIP MESSAGE) have no stream codec topology, leaving f->subclass.format as NULL.
+			 * We must bypass this block for standalone text frames to prevent them from being dropped.
+			 */
+			if (f && (f->frametype == AST_FRAME_VOICE || f->frametype == AST_FRAME_VIDEO ||
+				( f->frametype == AST_FRAME_TEXT && f->subclass.format))) {
```
mbradeen: space after (

main/channel.c
```diff
@@ -5048,6 +5065,18 @@ int ast_write_video(struct ast_channel *chan, struct ast_frame *fr)
 	return res;
 }
 
+/* currently unused */
```
mbradeen, line 5068: if unused, add a comment explaining why it was left in.

main/channel.c
```diff
@@ -5284,12 +5313,23 @@ int ast_write_stream(struct ast_channel *chan, int stream_num, struct ast_frame
 		break;
 	case AST_FRAME_TEXT:
 		CHECK_BLOCKING(chan);
-		if (ast_format_cmp(fr->subclass.format, ast_format_t140) == AST_FORMAT_CMP_EQUAL) {
-			res = (ast_channel_tech(chan)->write_text == NULL) ? 0 :
-				ast_channel_tech(chan)->write_text(chan, fr);
+		if (ast_channel_tech(chan)->write_stream) {
+			if (stream) {
```
mbradeen, line 5319: might be a good place to use ye ol ternary operator

res/res_pjsip/pjsip_config.xml
```diff
@@ -448,6 +448,45 @@
 						</enumlist>
 					</description>
 				</configOption>
+				<configOption name="max_text_streams" default="1">
+					<since>
+						<version>22.7.0</version>
```
mbradeen: Versions need updating before final acceptance.

res/res_format_attr_red.c
```diff
@@ -0,0 +1,282 @@
...
+	/* Get the resulting attributes for the joint format */
+	attr_res = ast_format_get_attribute_data(jointformat);
+
+	/* Determine joint attributes */
+	if (attr1->red_num_gen == 0 || attr2->red_num_gen == 0) {
+		/* If either format has no redundancy, set joint to no redundancy */
+		attr_res->red_num_gen = 0;
+	} else {
+		/*
+		* If both formats have redundancy, take the minimum number of generations.
```
mbradeen: This comment is a little oddly worded, I would say something like "If both formats have redundancy, take the minimum number of generations to save bandwidth and avoid compatibility issues."

res/res_pjsip_sdp_rtp.c
```diff
@@ -1555,6 +1594,51 @@ static void set_session_media_remotely_held(struct ast_sip_session_media *sessio
 	}
 }
 
+
+/*! \brief Function which checks for RED in a session and initialize it if needed */
+static void check_red_support(struct ast_sip_session *session, struct ast_sip_session_media *session_media,
+	struct ast_format *format_parsed)
+{
+	/*
+	 * To keep API compatibility, we are using an array here, even if the red codec
```
mbradeen, line 1603: To keep API compatibility, we are using an array here, even if the red codec
is using only an integer internally - similar to the other codec attributes.
Rtp red support from Asterisk expects the primary payload to be added to
the array, therefore we need one more element.

res/res_pjsip_sdp_rtp.c
```diff
@@ -1555,6 +1594,51 @@ static void set_session_media_remotely_held(struct ast_sip_session_media *sessio
...
+			ast_sip_session_get_name(session));
+		return;
+	}
+
+	if (ast_format_cap_has_type(session->endpoint->media.codecs, AST_MEDIA_TYPE_TEXT)) {
+		if (ast_format_attribute_get(format_parsed, "red_payload") != NULL &&
+				strcasecmp(ast_format_get_name(format_parsed), "RED") == 0) {
+			int red_payload = *(int*)ast_format_attribute_get(format_parsed, "red_payload");
+			int red_num_gen = *(int*)ast_format_attribute_get(format_parsed, "red_num_gen");
+			/* Already checked in the red sdp parser, just to be save */
```
mbradeen: "to be safe"

res/res_pjsip_sdp_rtp.c
```diff
@@ -2056,7 +2174,7 @@ static int create_outgoing_sdp_stream(struct ast_sip_session *session, struct as
 	}
 
 	/* Add non-codec formats */
-	if (ast_sip_session_is_pending_stream_default(session, stream) && media_type != AST_MEDIA_TYPE_VIDEO
+	if (ast_sip_session_is_pending_stream_default(session, stream) && media_type == AST_MEDIA_TYPE_AUDIO
```
mbradeen, line 2185: if we are trying to exclude text, I would add it as an extra && - unless you are sure this will not cause unintended consequences.

res/res_rtp_asterisk.c
```diff
@@ -4413,10 +4414,14 @@ static int ast_rtp_destroy(struct ast_rtp_instance *instance)
 	}
 
 	/* Destroy RED if it was being used */
-	if (rtp->red) {
+	/* only needed for chan_sip compatibility, should be removed after completely getting rid of it */
```
mbradeen, line 4417: If we don't need this in Asterisk 20, we can drop the chan_sip stuff and only cherry-pick to 22&23.  We can discuss this in the PR conversation.

mbradeen (changes requested):

henningw:
Thank you for the review @mbradeen. I will have a look to the comments and address them. 

Regarding the chan_sip compatiblity topic, we can drop it, there is no special requirement from our side to include this into asterisk 20. 

As a related note, an updated RTT patch was merged into pjsua git master (https://github.com/pjsip/pjproject/pull/4984). Its currently tested from @AlexisHadj.

mbradeen:
@henningw thank you for the update on pjproject, I will have to update my phone app.

I took a deeper Codex assisted review and found a few things, which I have addressed in the attached patch.  Please give it a shot in your testing and we can decide what we want to include it in the PR.

[rtt_reviewed.patch](https://github.com/user-attachments/files/30244733/rtt_reviewed.patch)

henningw:
> Regarding the chan_sip compatiblity topic, we can drop it, there is no special requirement from our side to include this into asterisk 20.

Hello @mbradeen, regarding the chan_sip topic please have a look at [this](https://github.com/asterisk/asterisk/pull/1128#issuecomment-4337499350) discussion from April. Please consult with @gtjoseph if his position on this has changed. If not its probably not necessary to discuss it here again.

Thanks for the additional patch in the latest comment, had a quick look to it. Many things are straight-forwarded and I will integrate them directly. Some of them needs a closer look and more tests.

bridges/bridge_native_rtp.c
henningw: If I remember correctly, there were some crashes without this checks. But they might be now resolved by additional additions, I will test to remove them.

main/channel.c
henningw: For the same reasons why there is a ast_write_video(..) above, to have a consistent internal API. I have added the comment also to this function.

main/channel.c
henningw: I wanted to stay in the same structure/logic as for the AST_FRAME_VIDEO below.

res/res_pjsip_sdp_rtp.c
henningw: Basically this adds non-codec types (e.g. telephony event) to the SDP. This is of course not valid for VIDEO, but also not for TEXT (previously there was no TEXT support). Therefore I added the check for AUDIO, as its only valid for this.

res/res_rtp_asterisk.c
henningw: Replied in https://github.com/asterisk/asterisk/pull/2031#issuecomment-5042616782

res/res_pjsip/pjsip_config.xml
henningw: Of course, I updated them for now already.

bridges/bridge_native_rtp.c
henningw: The reason for generally touching this code was to add native bridging support for text. I was observing crashes without this additional guards in certain situations. For example by having only a text stream without audio, if I remember correctly. I have added the guards to the text case, and thought it would be also a good idea (and better to keep it synchronised) to add them to the video case. If you think there are not necessary, let me know. I need to re-test it then.

bridges/bridge_native_rtp.c
henningw: Actually, I think it was also related to text only calls without audio.

henningw:
@mbradeen The review comments were addressed in the latest version and I also integrated your changes from the second review. Especially the chan_pjsip.c direct media related changes were good, we missed them before. The parser improvement for the RED are also a good addition, makes the large interpret function more readable. 

I have run the RTT test suite several times and also did a tests in different scenarios with three different user-agents, looked fine.

One question though, can you provide a bit more background for this changes:
```
@@ -2441,8 +2449,12 @@ static int apply_negotiated_sdp_stream(struct ast_sip_session *session,
                if (format_parsed) {
                        check_red_support(session, session_media, format_parsed);
                        ao2_ref(format_parsed, -1);
+               } else {
+                       ast_rtp_red_init(session_media->rtp, 0, NULL, -1);
                }
                ao2_ref(joint, -1);
+       } else {
+               ast_rtp_red_init(session_media->rtp, 0, NULL, -1);
```

Why we need to call the ast_rtp_init in all cases? To make sure the eventually earlier started RED support get properly tear-down e.g. on a re-INVITE? Otherwise this will not do much with this parameters, I think.

mbradeen:
@henningw - "To make sure the eventually earlier started RED support get properly tear-down e.g. on a re-INVITE?" - Yes, this exactly.  We want to make sure that on an REI any previous red state is cleaned up properly.

mbradeen (approved):
The versions may need to be updated again before this makes it in, but otherwise I think this is ready for wider review!

henningw:
Thank you @mbradeen for the review and the feedback, appreciated. The versions will be updated, I will also add the new major release to the cherry-picks later this week.

To give an update on the previously mentioned issues:

>     * completely empty "idle" RED packets are not forwarded (it was tested in simple bridge mode). Almost empty RED "idle" packets are forwarded, e.g. when a one BOM marker is included. We still need to reproduce this in our test suite.

This was fixed and some additional tests were added to the test suite PR.

>     * switch to voice-only in SRTP mode causes problems when switched back to RTT. The RTT is not properly decrypted anymore after the switch back. In our tests we were able to successfully transmit text again after a waiting time of 3-4 seconds. This could not be reproduced in another setup that also shows the problem, so it might be also a issue in the code

This is still open, we opened issue #2066 to track it further.

We are not aware of other issues right now.

henningw:
As the last review was one month ago, is there any update from your side? I have commented on the #2066, it seems to only affect a specific setup, as we were not able to reproduce it so far.

jcolp:
If there's an update, then an update will appear on it. As it is we're extremely busy and other stuff takes priority.

jcolp:
And to be more specific, there is an issue in our queue for dedicated time for a second set of eyes to do a review due to the size and complexity, since once this is merged it is then something we have to maintain as a project.

PR #2037, opened 2026-07-22
res_rtp_asterisk: Avoid two lock inversion deadlocks with pj project

### LINKED ISSUE #1946: [bug]: Deadlock between RTP instance and ICE session
### Severity

Critical

### Versions

20.17.0, 22.9.0

### Components/Modules

chan_pjsip, res_pjsip_session, res_pjsip, res_ari, res_ari_bridges, stasis, bridge_softmix, threadpool, taskprocessor, channel.c, channelstorage_ao2_legacy, astobj2

### Operating Environment

Ubuntu 22.04.5 LTS - 5.15.0-177-generic - x86_64 x86_64 x86_64 GNU/Linux

### Frequency of Occurrence

Frequent

### Issue Description

Asterisk appears to enter a deadlock or severe lock-contention state during concurrent processing of:
- inbound PJSIP INVITEs
- ARI bridges/addChannel operations
- channel allocation
- AO2 channel container traversal

One thread becomes blocked in:
__ast_channel_alloc_ap()

during inbound SIP INVITE handling via:
chan_pjsip_new()
chan_pjsip_incoming_request()

At the same time, other threads are blocked around:
- AO2 channel traversal (internal_ao2_traverse)
- bridge command synchronization (command_join)
- ARI bridge addChannel operations

Multiple threads accumulate waiting on futexes and mutexes, indicating a possible lock ordering inversion or deadlock involving:
- channel allocation locks
- AO2 container locks
- bridge/stasis locks

Symptoms:
- inbound SIP calls stop progressing
- ARI bridge operations hang
- worker threads block indefinitely
- Asterisk responsiveness degrades significantly

Could this be a known deadlock or lock-order inversion between channel allocation, AO2 channel traversal, and ARI bridge operations during concurrent PJSIP INVITE handling? We would appreciate guidance on where the root cause is most likely located and whether there are known fixes, patches, or recommended debugging steps for this scenario. Is this something that should be investigated primarily in `channel.c`, `channelstorage_ao2_legacy`, `stasis/control`, or the ARI bridge handling path?

[asterisk_bt_full 1.txt](https://github.com/user-attachments/files/28102376/asterisk_bt_full.1.txt)

### Relevant log output

See above attachment: asterisk_bt_full 1.txt

### Asterisk Issue Guidelines

mbradeen:
Fixes two related instance, group lock inversion deadlocks.

When writing the RTP stream via rtp_sendto, the RTP instance was only unlocked
before being passed to pj project when the instance was un-bundled (ie video
was not bundled onto the audio stream's ICE transport.)

The first change unlocks the instance whether bundled or unbundled to avoid the
deadlock, then re-checks the bundled status upon return to know wether or not the
transport should be used.

For the second change, when an incoming RTCP NACK record was recieved, the parent
(transport) and child (instance) locks were both held before the call to rtp_sendto,
which is only able to release the child lock. Now in the un-bundled case the parent
and child will be unlocked and re-locked in the correct order via a new helper
function before and after the call to send_to.

Fixes: #1946

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -3503,18 +3503,26 @@ static int __rtp_sendto(struct ast_rtp_instance *instance, void *buf, size_t siz
 		/* Release the instance lock to avoid deadlock with PJPROJECT group lock */
 		ice = transport_rtp->ice;
 		ao2_ref(ice, +1);
-		if (instance == transport) {
-			ao2_unlock(instance);
-		}
+		ao2_ref(transport, +1);
+		ao2_unlock(instance);
 		status = pj_ice_sess_send_data(ice->real_ice, component, temp, len);
 		ao2_ref(ice, -1);
-		if (instance == transport) {
-			ao2_lock(instance);
-		}
+		ao2_lock(instance);
 		if (status == PJ_SUCCESS) {
 			*via_ice = 1;
+			ao2_ref(transport, -1);
 			return len;
 		}
+		if (transport != (rtp->bundled ? rtp->bundled : instance)) {
+			/*
+			 * In case the transport was bundled or un-bundled while we were unlocked don't
+			 * fall through to sending using the transport instance as we may no longer be
+			 * associated with it.
+			 */
+			ao2_ref(transport, -1);
+			return 0;
+		}
+		ao2_ref(transport, -1);
 	}
 #endif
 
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

bkford (approved):

PR #2040, opened 2026-07-23
file.c: Fix write-format race when opening a stream for playback

### LINKED ISSUE #1623: [bug]: `Bridge` dialplan randomly fails to play courtesy tone
### Severity

Trivial

### Versions

20.11.1, 20.17.0

### Components/Modules

file.c

### Operating Environment

Debian 13 - frequent
Debian 12 - occasional

### Frequency of Occurrence

Frequent

### Issue Description

We randomly experience errors playing a courtesy tone while bridging two channels via the `Bridge` dialplan command. The bridge itself works fine, audio is flowing both ways.

The problem was initially occasional but has become very frequent upon updating a server from Debian 12 to Debian 13. The EC2 instances in both case have the exact same characteristics, serving the same traffic and were both running Asterisk 20.11.0. An upgrade to Asterisk 20.17.0 did not change the situation.

The error reads:

`file.c: Unable to open beep (format (alaw)): No such file or directory`

## Analysis to date ##

We have two problems here:
- The error message looks misleading (we have double and triple checked - the "beep" sound file is correctly installed).
- It is unclear what is triggering the problem, logs with debug=5, verbose=5 do not provide any insights as to the random behaviour. It does however appear to be sensitive to the system load.

The error message originates here:

https://github.com/asterisk/asterisk/blob/a6539cb9af3094790f3f08b92667d2cfae7c4b6f/main/file.c#L1342

The "No such file or directory" is most likely misleading, if the file really did not exist, the error would not be random. Furthermore, we should have tripped the following error:

https://github.com/asterisk/asterisk/blob/a6539cb9af3094790f3f08b92667d2cfae7c4b6f/main/file.c#L832

A possible explanation for why we get the "No such file or directory" errors is that `errno` is not initialised in `filehelper` before the `stat` and other file operations:

https://github.com/asterisk/asterisk/blob/a6539cb9af3094790f3f08b92667d2cfae7c4b6f/main/file.c#L588

=> it seems possible for `filehelper` to return successfully, while still having `errno` set.

Insights on how to progress the analysis are most welcome.

### Relevant log output

```shell
[2025-12-03 12:28:23] WARNING[2836930][C-00054564] file.c: Unable to open beep (format (alaw)): No such file or directory```

### Asterisk Issue Guidelines

jlaine:
openstream_internal() set the channel write format under the channel lock, released the lock, then re-checked that format in filehelper() (ACTION_OPEN) while holding only the formats RWLIST.  The bridging and format-negotiation code changes the same channel's write format concurrently (always under the channel lock), so the re-check could observe a different format and wrongly reject a valid sound file with:

  Unable to open beep (format (alaw)): No such file or directory

Because the reader never took the channel lock, the set and the re-check were two separate critical sections with the lock dropped in between.

Hold the channel lock across both the write-format set and the filehelper(ACTION_OPEN) re-check so the two accesses share a common lock.

Confirmed with ThreadSanitizer: the data race on the channel write format no longer appears and the intermittent warning no longer reproduces.

Resolves: #1623

### DIFF
--- a/main/file.c
+++ b/main/file.c
@@ -840,19 +840,26 @@ static struct ast_filestream *openstream_internal(struct ast_channel *chan,
 		return NULL;
 	}
 
-	/* Set the channel to a format we can work with and save off the previous format. */
+	/*
+	 * Keep the channel locked across setting the write format and filehelper(ACTION_OPEN)
+	 * to avoid race conditions with the bridging / format-negotiation code. Otherwise the
+	 * write format may be changed concurrently, leading to a valid file being rejected as
+	 * not supported.
+	 */
 	ast_channel_lock(chan);
+	/* Set the channel to a format we can work with and save off the previous format. */
 	ast_channel_set_oldwriteformat(chan, ast_channel_writeformat(chan));
 	/* Set the channel to the best format that exists for the file. */
 	res = ast_set_write_format_from_cap(chan, file_fmt_cap);
-	ast_channel_unlock(chan);
 	/* don't need this anymore now that the channel's write format is set. */
 	ao2_ref(file_fmt_cap, -1);
 
 	if (res == -1) {	/* No format available that works with this channel */
+		ast_channel_unlock(chan);
 		return NULL;
 	}
 	res = filehelper(buf, chan, NULL, ACTION_OPEN);
+	ast_channel_unlock(chan);
 	if (res >= 0)
 		return ast_channel_stream(chan);
 	return NULL;


jlaine:
@mbradeen I have not yet had a chance to use this patch in production so handle with care. I'm putting it up here to get some feedback on whether the approach is valid.

jlaine:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jlaine:
I would appreciate any feedback on the potential downsides of holding the channel lock across the "filehelper" call.

jcolp:
It should be safe.

PR #2041, opened 2026-07-27
res_pjsip: Qualify all resolved contact targets

### LINKED ISSUE #1927: [bug]: PJSIP OPTIONS not adhering to SRV failover
### Severity

Major

### Versions

PJSIP 2.15.1 on Asterisk 22.4.1

### Components/Modules

SRV Record Processing

### Operating Environment

FreePBX 15/16/17 on CentOS and Debian - this bug seems to be present on all installations.

### Frequency of Occurrence

Constant

### Issue Description

This morning my carrier had a Datacenter go down, and we should have failed over cleanly to another center with the SRV records that we were registered to - but we didn’t. Looking at packet-traces and such, here is the list of centers in the SRV record:

Endpoint:  customer.name.cloud                             Not in use    0 of inf
OutAuth:  customer.name.cloud/334a911ce092
Aor:  customer.name.cloud                           0
Contact:  customer.name.cloud/sip:334a911ce092@ 03af184822 Avail        60.557
Transport:  0.0.0.0-udp               udp      3     96  0.0.0.0:54321
Identify:  customer.name.cloud/customer.name.cloud
**Match: 162.252.251.41/32**
**Match: 162.252.251.43/32**
Match: 162.252.250.230/32
Match: 162.252.250.43/32

First two IP’s are at the dead data center, next two are alive and kicking.

But here is the problem - PJSIP properly registers with the first available gateway 162.252.250.230/32 but it insists on sending the Qualify packets to the first (dead) server 162.252.251.41/32 - so Incoming works, but outgoing fails.

The Solution is to configure an outbound proxy in Advanced Settings - sip:162.252.250.230 and then it registers and qualifies and Outbound proceeds.

This behavior totally kills the redundancy of the SIP trunk and causes an outage.

SRV Redundancy is the standard for creating a redundant SIP connection spanning multiple Data Centers and SBC's - PJSIP has to look at the IP that actually took the registration attempt and direct all Qualify packets to the server it is actually registered to.  Packet traces showed that no matter what server of the four listed in this SRV record PJSIP ended up registering to, all Qualify Packets were sent to the first server in the List and so they failed - Inbound works because the trunk is registered, but outbound fails because the trunk looks down to Asterisk.

### Relevant log output

```shell
I have lots of logs if you need them, but this bug is 100% consistent - Any PJSIP trunk with a SRV record with multiple hosts will exhibit the same behavior.
```

### Asterisk Issue Guidelines

mbradeen:
Previously, PJSIP endpoint qualification was offloaded to pjproject which
used a single, internally selected contact based on DNS results.
This could result in unused SRV, A, or AAAA targets, leaving a contact
marked as unreachable without probing the remaining targets.

Now, we resolve the complete set of records before sending and creating
OPTIONS transactions for every resolved target.  The contact will be marked
as reachable as soon as any target succeeds and unreachable only after every
target fails.

Fixes: #1927

UserNote: DNS-backed PJSIP contacts are now considered reachable when any resolved
SRV, A, or AAAA target responds to qualification.


mbradeen:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

mbradeen:
testsuite-test-pr: 149

res/res_pjsip/pjsip_options.c
```diff
@@ -548,6 +548,21 @@ struct sip_options_contact_callback_data {
 	struct timeval rtt_start;
 	/*! \brief The new status of the contact */
 	enum ast_sip_contact_status_type status;
+	/*! \brief Number of resolved targets waiting for a response */
+	unsigned int pending;
+	/*! \brief Mark to prevent more than one contact state publication per qualify */
+	unsigned int completed;
```
jcolp: Being directly below pending makes this seem as though it is a counter and not just a binary yes/no.

res/res_pjsip/pjsip_options.c
```diff
@@ -848,13 +933,114 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+
+	if (status != PJ_SUCCESS || !addresses || !addresses->count) {
+		batch->pending = 1;
+		sip_options_queue_target_result(batch, UNAVAILABLE);
+		ao2_ref(batch, -1);
+		return;
+	}
+
+	batch->pending = addresses->count;
+	pjsip_tx_data_dec_ref(batch->resolve_tdata);
```
jcolp, line 1053: Why dec ref here instead of letting the eventual destructor handle it?

res/res_pjsip/pjsip_options.c
```diff
@@ -848,13 +933,114 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+
+		/*
+		 * Build a new request per resolved target so each send gets its own
+		 * transaction and Via branch. Setting exactly one pre-resolved
+		 * destination here avoids PJSIP re-resolving and choosing only the
+		 * first DNS result.
+		 */
+		tdata->dest_info.addr.count = 1;
+		tdata->dest_info.cur_addr = 0;
+		pj_memcpy(&tdata->dest_info.addr.entry[0], &addresses->entry[idx],
```
jcolp: pj_sockaddr_cp

res/res_pjsip/pjsip_options.c
```diff
@@ -782,7 +797,83 @@ static int sip_options_contact_status_notify_task(void *obj)
...
+		AO2_ALLOC_OPT_LOCK_NOLOCK);
+	if (!result) {
+		ast_log(LOG_WARNING, "Unable to allocate qualify target result for '%s'\n",
+			ast_sorcery_object_get_id(batch->contact));
+		return;
+	}
+
+	result->batch = ao2_bump(batch);
+	result->status = status;
+	/*
```
jcolp, line 883: Explain the overall threading for this more

res/res_pjsip/pjsip_options.c
```diff
@@ -910,14 +1102,28 @@ static int sip_options_qualify_contact(void *obj, void *arg, int flags)
...
+		|| sip_options_apply_transport_to_destination(tdata, &destination)
+			!= PJ_SUCCESS) {
+		ast_log(LOG_ERROR, "Unable to resolve destination to qualify contact %s on AOR %s\n",
 			contact->uri, aor_options->name);
+		pjsip_tx_data_dec_ref(tdata);
 		ao2_ref(contact_callback_data, -1);
+		return 0;
 	}
 
+	contact_callback_data->endpoint = ao2_bump(endpoint);
```
jcolp: The contact_callback_data already has an allocator which accepts arguments, why aren't endpoint and tdata passed in there to continue the existing usage pattern?

res/res_pjsip/pjsip_options.c
```diff
@@ -848,13 +933,114 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+
+/*! \brief DNS callback used to fan out a qualify to every resolved target */
+static void sip_options_qualify_resolve_cb(pj_status_t status, void *token,
+	const pjsip_server_addresses *addresses)
+{
+	struct sip_options_contact_callback_data *batch = token;
+	unsigned int idx;
+
+	if (status != PJ_SUCCESS || !addresses || !addresses->count) {
+		batch->pending = 1;
```
jcolp, line 1041: Add a comment explaining why batch->pending is set here

res/res_pjsip/pjsip_options.c
```diff
@@ -848,13 +933,114 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+		sip_options_queue_target_result(batch, UNAVAILABLE);
+		ao2_ref(batch, -1);
+		return;
+	}
+
+	batch->pending = addresses->count;
```
jcolp, line 1048: Add a comment explaining that every returned address must cause a batch result to occur

jcolp (changes requested):

jcolp (approved):

gtjoseph (commented):
I was testing this locally and I wonder if we should cancel any other in-flight transactions when any one of them complete.  I'm testing where a contact has both A and AAAA records and the remote stops listening on the A record.  In this case, the contact still shows as Avail because the AAAA address is responding but the A address attempts just keep retrying.  If the qualify interval is 30 seconds and timer_b is 32 seconds we're basically just continually retrying the A address every few seconds.

mbradeen:
Converting to draft while we attempt to figure out how to handle George's comment.

res/res_pjsip/pjsip_options.c
```diff
@@ -782,7 +797,88 @@ static int sip_options_contact_status_notify_task(void *obj)
...
+	 * target onto the AOR serializer, the single owner of the batch's processing.
+	 *
+	 * Sends and callbacks can execute synchronously while we are already
+	 * running on that serializer thread. Re-queueing to the same serializer in
+	 * that case would defer updates and can break processing.
+	 *
+	 * Process inline when already on the serializer to preserve immediate,
+	 * in-order state transitions.
+	 */
+	if (ast_taskprocessor_is_task(batch->aor_options->serializer)) {
```
jcolp: I think this should use ast_taskpool_serializer_get_current instead, which will return the current serializer within taskpool - using the taskprocessor API can return the wrong thing

jcolp (changes requested):

mbradeen:
AI disclosure:  Codex 5.6 Sol Light was used in adapting to the provided pjsip patch as well as in overall review.

res/res_pjsip/pjsip_options.c
```diff
@@ -843,18 +978,142 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+		}
+
+		/* ast_sip_send_out_of_dialog_request_with_tsx consumes the tdata reference. */
+		if (ast_sip_send_out_of_dialog_request_with_tsx(tdata, batch->endpoint,
+			(int)(batch->aor_options->qualify_timeout * 1000),
+			ao2_bump(batch), qualify_contact_cb, &tsx)) {
+			ao2_ref(batch, -1);
+			sip_options_queue_target_result(batch, UNAVAILABLE);
+		} else if (tsx && AST_VECTOR_APPEND(&batch->target_transactions, tsx)) {
+			pj_grp_lock_dec_ref(tsx->grp_lock);
```
jcolp, line 1104: This would yield a transaction that can't be cancelled.

I think a safer option is to pre-initialize the target_transactions vector before sending any of these OPTIONS in this function to the expected count, so that AST_VECTOR_APPEND is guaranteed to not fail. You can also change sip_options_contact_callback_data_alloc to just initialize it to 0 instead of the magical 4.

This way there is explicit defined behavior in the case where there's not enough memory - OPTIONS will go unavailable, otherwise I fear it could go awry.

res/res_pjsip/pjsip_options.c
```diff
@@ -843,18 +978,142 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+			via->branch_param.slen = 0;
+		}
+
+		/* ast_sip_send_out_of_dialog_request_with_tsx consumes the tdata reference. */
+		if (ast_sip_send_out_of_dialog_request_with_tsx(tdata, batch->endpoint,
+			(int)(batch->aor_options->qualify_timeout * 1000),
+			ao2_bump(batch), qualify_contact_cb, &tsx)) {
+			ao2_ref(batch, -1);
+			sip_options_queue_target_result(batch, UNAVAILABLE);
+		} else if (tsx && AST_VECTOR_APPEND(&batch->target_transactions, tsx)) {
```
jcolp: Does with_tsx guarantee a tsx?

res/res_pjsip.c
```diff
@@ -1887,10 +1894,12 @@ static void send_request_cb(void *token, pjsip_event *e)
 		switch(e->body.tsx_state.type) {
 		case PJSIP_EVENT_TRANSPORT_ERROR:
 		case PJSIP_EVENT_TIMER:
+		case PJSIP_EVENT_USER:
```
jcolp: Does PJSIP_EVENT_USER really belong here? This goes into failover checking via check_request_status, which I don't think we want.

jcolp (changes requested):

res/res_pjsip/pjsip_options.c
jcolp: Oh, or an alternative is to initialize AST_VECTOR_INIT originally with the fixed #define for maximum number of resolved targets instead of 4, which would also guarantee AST_VECTOR_APPEND can't fail.

res/res_pjsip/pjsip_options.c
mbradeen: no and ast_sip_send_out_of_dialog_request calls ast_sip_send_out_of_dialog_request_with_tsx with a NULL tsx.  Based on this comment I am updating endpt_send_request to only call the new underlying pjsip pjsip_endpt_send_request2 if we have one to avoid pjproject potentially upping the ref count on something we are ignoring.

res/res_pjsip/pjsip_options.c
```diff
@@ -843,18 +978,144 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+		sip_options_queue_target_result(batch, UNAVAILABLE);
+		ao2_ref(batch, -1);
+		return;
+	}
+
+	/* Each resolved address must contribute exactly one batch result. */
+	batch->pending = addresses->count;
+	/*
+	 * Clean up the resolve transaction data as we don't need it for the rest of the batch
+	 * processing. Do the same in the destructor for the contact callback data as a fallback.
```
jcolp, line 1051: This is only true because our own resolver implementation doesn't use the passed in pool. If we ever do, it won't be. I think it's safer to ensure that the pool remains alive throughout the usage here.

res/res_pjsip.c
```diff
@@ -1753,7 +1758,14 @@ static pj_status_t endpt_send_request(struct ast_sip_endpoint *endpoint,
 	 * transaction callback is executed.
 	 */
 	ao2_ref(req_wrapper, +1);
-	ret_val = pjsip_endpt_send_request(endpt, tdata, -1, req_wrapper, endpt_send_request_cb);
+	if (tsx) {
+		ret_val = pjsip_endpt_send_request2(endpt, tdata, -1, req_wrapper,
```
jcolp, line 1762: Argh I just remembered we allow building against third party PJSIP, and some people actually do it. That means we need a configure.ac check for whether this exists or not and a fallback behavior if it does not - doubly so since this functionality isn't even released by Teluu yet.

res/res_pjsip/pjsip_options.c
```diff
@@ -843,18 +978,144 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+		/* ast_sip_send_out_of_dialog_request_with_tsx consumes the tdata reference. */
+		if (ast_sip_send_out_of_dialog_request_with_tsx(tdata, batch->endpoint,
+			(int)(batch->aor_options->qualify_timeout * 1000),
+			ao2_bump(batch), qualify_contact_cb, &tsx)) {
+			ao2_ref(batch, -1);
+			sip_options_queue_target_result(batch, UNAVAILABLE);
+		} else if (tsx) {
+			/* The Append should never fail as we init the vector to PJSIP_MAX_RESOLVED_ADDRESSES */
+			AST_VECTOR_APPEND(&batch->target_transactions, tsx);
+			pj_grp_lock_dec_ref(tsx->grp_lock);
```
jcolp, line 1104: Er, I don't think releasing the reference is correct here since you are storing the tsx which will later release the reference.

res/res_pjsip/pjsip_options.c
```diff
@@ -843,18 +978,144 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
...
+			!= (transport_type | PJSIP_TRANSPORT_IPV6))) {
+		return PJSIP_ETPNOTSUITABLE;
+	}
+
+	destination->type = transport_type;
+	return PJ_SUCCESS;
+}
+
+/*! \brief DNS callback used to fan out a qualify to every resolved target */
+static void sip_options_qualify_resolve_cb(pj_status_t status, void *token,
```
jcolp, line 1029: Is this guaranteed to run on the same serializer as the AOR options serializer? If not, then this would be mutating state unprotected as the same time that a result could be.

jcolp (changes requested):

res/res_pjsip/pjsip_options.c
```diff
@@ -843,18 +1089,156 @@ static struct sip_options_contact_callback_data *sip_options_contact_callback_da
 
 	contact_callback_data->contact = ao2_bump(contact);
 	contact_callback_data->aor_options = ao2_bump(aor_options);
+	contact_callback_data->endpoint = ao2_bump(endpoint);
+	contact_callback_data->resolve_tdata = resolve_tdata;
```
jcolp, line 1098: In off-nominal this can cause a double unref. The caller to the _alloc function drops the ref if this can't be allocated, and the destructor for contact_callback_data also drops the ref. Move the setting to after the AST_VECTOR_INIT, or move the AST_VECTOR_INIT to earlier such as the first thing done after the ao2_alloc_options.

res/res_pjsip.c
```diff
@@ -1933,9 +1966,9 @@ static void send_request_cb(void *token, pjsip_event *e)
 	ao2_ref(req_data, -1);
 }
 
-int ast_sip_send_out_of_dialog_request(pjsip_tx_data *tdata,
+int ast_sip_send_out_of_dialog_request_with_tsx(pjsip_tx_data *tdata,
 	struct ast_sip_endpoint *endpoint, int timeout, void *token,
-	void (*callback)(void *token, pjsip_event *e))
+	void (*callback)(void *token, pjsip_event *e), pjsip_transaction **tsx)
```
jcolp, line 1971: The header file documents that *tsx will be NULL if an error occurs, but nothing actually guarantees this so there are cases where it won't be set by endpt_send_request. The best thing to do is to just set it to NULL at the start of the function so only in success will it be non-NULL.

jcolp (changes requested):

res/res_pjsip.c
mbradeen: endpt_send_request does it at the beginning:
https://github.com/asterisk/asterisk/pull/2041/changes#diff-295139a66ec5fbd8034bd80497e78247306b94118fcaf2bf64f75ac6c00de461R1702
- Do we need it here too?

jcolp: Yes, if send_request_data_alloc fails then *tsx is not set to NULL.

jcolp (approved):

gtjoseph (approved):

PR #2042, opened 2026-07-28
build: Fix GCC discarded-qualifiers const errors.

seanbright:
GCC 15.2.1 pays attention to the discarding of the const qualifier when strchr, strrchr, memchr, or memrchr are now used. This change fixes numerous errors with this throughout chan_sip. The fixes can be broken down into the following:

1. The value passed to strchr or strrchr can be cast as it is expected and allowed to be modified.
2. The pointer passed to strchr or strrchr is not meant to be modified and so the contents must be duplicated.
3. It was declared const and never should have been.

### DIFF
--- a/channels/chan_sip.c
+++ b/channels/chan_sip.c
@@ -11594,7 +11594,7 @@ static int process_sdp_a_audio(const char *a, struct sip_pvt *p, struct ast_rtp_
 	int debug = sip_debug_test_pvt(p);
 
 	if (!strncasecmp(a, "ptime", 5)) {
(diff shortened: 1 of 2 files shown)


seanbright:

gtjoseph (approved):

gtjoseph:
@seanbright Do you have time to fix a few more issues reported by GCC 16 in this PR?
In chan_sip.c:reload_config() registry_count and peer_count are set but never used.

gtjoseph:
Thanks!

PR #2045, opened 2026-07-30
app_queue: Update lastpause on realtime pause transition

### LINKED ISSUE #1760: [bug]: queue incorrectly updates the pause time for realtime members
### Severity

Minor

### Versions

20+

### Components/Modules

app_queue

### Operating Environment

Any

### Frequency of Occurrence

Constant

### Issue Description

Function **rt_handle_member_record** (example, start from line 3773 in version 22.6.0-lts file app_queue.c)

```c
if (paused && m->lastpause == 0) {
    time(&m->lastpause); /* XXX: Should this come from realtime? */
}
```

The condition will only work if the member has not been paused before (m->lastpause == 0)

If an external application unpaused a member and then paused him again, the time spent in the pause will be counted from the very first pause, which is incorrect.

It seems to me that it is necessary to divide the conditions in this way:

```c
if (paused) {
    if(m->lastpause == 0)
	time(&m->lastpause); /* set lastpause = now, if not set */
} else {
    m->lastpause = 0; /* reset lastpause if RT return unpause state */
}
```

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

seifzadeh:
Realtime queue members used lastpause to determine whether a member
had transitioned from unpaused to paused. However, lastpause also
stores the timestamp of the member's most recent pause and should not
be cleared when the member is unpaused.

Compare the existing paused state with the new realtime value instead,
and update lastpause only when the member transitions from unpaused to
paused.

Resolves: #1760

### DIFF
--- a/apps/app_queue.c
+++ b/apps/app_queue.c
@@ -3826,8 +3826,12 @@ static void rt_handle_member_record(struct call_queue *q, char *category, struct
 			ast_copy_string(m->rt_uniqueid, rt_uniqueid, sizeof(m->rt_uniqueid));
 			if (paused_str) {
 				m->paused = paused;
-				if (paused && m->lastpause == 0) {
-					time(&m->lastpause); /* XXX: Should this come from realtime? */
+				if (paused) {
+					if (m->lastpause == 0) {
+						time(&m->lastpause); /* XXX: Should this come from realtime? */
+					}
+				} else {
+					m->lastpause = 0; /* Reset lastpause when unpaused from Realtime */
 				}
 				ast_devstate_changed(m->paused ? QUEUE_PAUSED_DEVSTATE : QUEUE_UNPAUSED_DEVSTATE,
 					AST_DEVSTATE_CACHABLE, "Queue:%s_pause_%s", q->name, m->interface);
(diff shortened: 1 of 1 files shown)


seifzadeh:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seifzadeh:
AI usage: ChatGPT was used to review the coding style, commit message, and pull request workflow. I reviewed and understand the code change and verified the behavior manually.

apps/app_queue.c
```diff
@@ -3826,8 +3826,12 @@ static void rt_handle_member_record(struct call_queue *q, char *category, struct
 			ast_copy_string(m->rt_uniqueid, rt_uniqueid, sizeof(m->rt_uniqueid));
 			if (paused_str) {
 				m->paused = paused;
-				if (paused && m->lastpause == 0) {
-					time(&m->lastpause); /* XXX: Should this come from realtime? */
+				if (paused) {
```
jcolp: This change worries me - because I think lastpause is being used in two different ways.

The first way is that realtime is trying to use it to determine if state has transitioned from unpaused to paused and thus lastpause should be changed. I think perhaps instead the m->paused and paused should be compared and if it is transitioning to paused then m->lastpause should be set. I don't think it should use lastpause for the check.

The second case is that it is literally used to store the last time the agent was paused for displaying elsewhere.

Using this change as-is would break the second case.

jcolp (changes requested):

jcolp (approved):

PR #2048, opened 2026-07-31
Fix GCC compile issues in branch certified/20.7

gtjoseph:
Cherry-picks of existing fixes:

- **build: Fix unused-but-set-variable warnings with gcc 16.**
- **chan_dahdi: Fix discarded-qualifiers errors.**
- **chan_sip.c: Fix __sip_reliable_xmit build error**

Standalone fix that was fixed in other branches as part of a larger change which we don't want to include in this branch:

- **res_stir_shaken.c: Fix GCC set-but-not-used error in unload_module.**

### DIFF
--- a/apps/app_externalivr.c
+++ b/apps/app_externalivr.c
@@ -309,12 +309,11 @@ static void ast_eivr_getvariable(struct ast_channel *chan, char *data, char *out
 
 	char *inbuf, *variable;
 	const char *value;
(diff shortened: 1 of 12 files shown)


gtjoseph:
cherry-pick-to: none
multiple-commits: standalone

jcolp (approved):

PR #2052, opened 2026-08-02
res_musiconhold: Fix mohclass reference leak on answeredonly early return.

### LINKED ISSUE #2051: [bug]: res_musiconhold: answeredonly leaks a mohclass reference on unanswered channels
### Severity

Minor

### Versions

master, 23, 22, 21, 20

### Components/Modules

res/res_musiconhold.c

### Operating Environment

Linux x86_64 (Manjaro, kernel 6.12)

### Frequency of Occurrence

Constant

### Issue Description

While diagnosing https://github.com/asterisk/asterisk/issues/2035 and reading through local_ast_moh_start() I noticed the answeredonly check returns without releasing the mohclass reference. This is unrelated to that race, so filing it separately (res/res_musiconhold.c:1917 on current master):

```c
	if (mohclass->answeredonly && (ast_channel_state(chan) != AST_STATE_UP)) {
		ast_verb(3, "The channel '%s' is not answered yet. Ignore the moh request.\n", ast_channel_name(chan));
		return -1;
	}
```

The function has 12 return statements, and this is the only one that returns while still holding a class reference. The reference held here came from get_mohbyname(), from moh_class_malloc() for realtime classes, or in one corner case from a bump of the class already attached to the channel, and nothing else ever releases it. The channel has not taken a reference for this request yet (that happens later, inside ast_activate_generator()), so the refcount can never drop back and moh_class_destructor() never runs for that object.

What that means in practice:

- static classes: the refcount grows with every suppressed request. Once the class is replaced by a reload (or removed on shutdown), the old object can never be destroyed; for app-mode classes (mp3/custom) its monitor thread and external application process stay alive for the rest of the Asterisk process.
- realtime without cachertclasses (the default): every suppressed request builds a fresh class, and for the app modes the monitor thread is already running before the check is reached. So each request leaks the object, one thread, one external process and two fds, and this keeps growing for the lifetime of the process.

Note this is the option working as intended, not an error path: any MOH request against a not-yet-answered channel with answeredonly=yes leaks once. Using Dial's m option before the call is answered is enough to hit it.

I reproduced it on a clean master checkout (2026-08-02). Setup: realtime musiconhold via sqlite3 (any driver should do), so extconfig.conf has "musiconhold => sqlite3,asterisk", cachertclasses not set. The shipped alembic schema has no answeredonly column, so add it first:

```
ALTER TABLE musiconhold ADD COLUMN answeredonly VARCHAR(5);
INSERT INTO musiconhold (name,mode,directory,application,format,answeredonly)
VALUES ('rtmohao','custom','nodir','/etc/asterisk/mohgen.sh','slin','yes');
```

mohgen.sh is just "exec cat /dev/zero". Dialplan; 700 never answers before starting MOH, 710 is the same thing with Answer() first as a control:

```
[moh-test]
exten => 700,1,Set(CHANNEL(musicclass)=rtmohao)
 same => n,StartMusicOnHold()
 same => n,Wait(0.2)
 same => n,StopMusicOnHold()
 same => n,Hangup()

exten => 710,1,Answer()
 same => n,Set(CHANNEL(musicclass)=rtmohao)
 same => n,StartMusicOnHold()
 same => n,Wait(0.2)
 same => n,StopMusicOnHold()
 same => n,Hangup()
```

driven from the CLI with "channel originate Local/700@moh-test application Wait 1", 20 times. With "core set debug 1" so the destructor message is visible. Results: threads went 75 to 95, child processes 1 to 21, open fds 21 to 61, and "Destroying MOH class" was never logged for any of the 20. The same 20 calls to 710 stay essentially flat (+1 thread, +1 fd, no new children) and log the destructor 20 times.

I then added the missing mohclass_unref() before that return, rebuilt only res_musiconhold and ran the identical test again: threads 75 to 75, children 1 to 1, and the destructor now logs for the unanswered requests as well (20 times instead of 0). So the missing unref is what leaks the class, the thread and the child.

```
before fix, 20 unanswered MOH requests: threads +20, children +20, fds +40, destructor runs 0
after fix,  same 20 requests:           threads +0,  children +0,  fds +4..6, destructor runs 20
```

A few stray fds do remain after the fix, but they are not from this path: this bug leaks exactly 2 fds per call (40 over 20 calls), while the leftover is 4 to 6 per whole run rather than 2 per call. An fd listing shows they are pipes with both ends still open. I haven't fully chased them down; they look like the pipe/fork cancellation window in spawn_mp3 that now gets exercised when the destructor runs (same family as #2035).

The unref has been missing since the option was introduced in fc63688e3b (ASTERISK-30135, 2022), and the same check is present in the 20 through 23 branches. I'll send a PR with the one-line fix.

### Relevant log output

```shell
[Aug  2 19:22:08] VERBOSE[4082544][C-00000012] res_musiconhold.c: The channel 'Local/700@moh-test-00000011;2' is not answered yet. Ignore the moh request.
[Aug  2 19:22:21] DEBUG[4083059][C-00000026] res_musiconhold.c: Destroying MOH class 'rtmohao'

summary of the same run (measured via /proc/<pid>):
baseline:                  threads=75 children=1  fds=21
after 20 unanswered (700): threads=95 children=21 fds=61   "is not answered yet" x20, destructor x0
after 20 answered (710):   threads=96 children=21 fds=62   destructor x20 (all from the answered calls)
```

### Asterisk Issue Guidelines

aabolfazl:
local_ast_moh_start() returns -1 from the answeredonly check without
releasing the mohclass reference it holds, unlike every other exit path
in the function. Nothing else ever releases that reference, so
moh_class_destructor() never runs for the object. With realtime music
on hold and cachertclasses disabled, each suppressed request leaks the
class object, its monitor thread, the external application process and
two file descriptors for the lifetime of Asterisk. For static classes
the stale references prevent the class from ever being destroyed after
it is replaced by a reload.

Release the reference before returning, matching the other exit paths.

Fixes: #2051

### DIFF
--- a/res/res_musiconhold.c
+++ b/res/res_musiconhold.c
@@ -1916,6 +1916,7 @@ static int local_ast_moh_start(struct ast_channel *chan, const char *mclass, con
 
 	if (mohclass->answeredonly && (ast_channel_state(chan) != AST_STATE_UP)) {
 		ast_verb(3, "The channel '%s' is not answered yet. Ignore the moh request.\n", ast_channel_name(chan));
+		mohclass = mohclass_unref(mohclass, "unreffing local reference to mohclass (channel not answered)");
 		return -1;
 	}
 


aabolfazl:
Testing: with a realtime custom-mode class with answeredonly=yes and 20 MOH requests against an unanswered channel per run: unpatched leaks 20 threads, 20 child processes and 40 fds with 0 destructor runs; patched (4 runs) leaks none and the destructor runs for all 20. Full reproduction steps are in #2051.

I used AI assistance to audit the reference counting and verify the fix.

seanbright:
All of the `cherry-pick-to` directives should be in a single comment.

aabolfazl:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

aabolfazl:
@seanbright thanks, consolidated.

jcolp (approved):

PR #2060, opened 2026-08-04
res_pjsip_sdp_rtp: Fix intermittent call drop on reINVITE sdp change

### LINKED ISSUE #2059: [bug]: Occasional call drops during re-INVITE when codec changes
### Severity

Minor

### Versions

20, 22, 23, master

### Components/Modules

rtp_engine, res_pjsip

### Operating Environment

all supported

### Frequency of Occurrence

Occasional

### Issue Description

During SDP re-negotiation due to a RE-INVITE, a call can be dropped under specific timing if Asterisk attempts to send an rtp packet using a codec that the re-negotiation had removed before that re-negotiation finalizes.
For example, if Alice is using alaw and Bob is using ulaw, then Bob switches to alaw, a ulaw rtp packet in-flight can result in Asterisk tearing down the call.

### Relevant log output

```shell
res_rtp_asterisk.c: Don't know how to send format ulaw packets with RTP
```

### Asterisk Issue Guidelines

mbradeen:
A reINVITE that changed the offered codec set (e.g. ulaw+alaw -> alaw
only) could lead to call teardown when a bridge write still using the
old format occurred during the re-negotiation.

This change avoids the termination by making the following changes.

First, Asterisk now holds the lock across both the tx payload update
and the channel format update so they commit atomically.

Second, when processing the sdp on the new incoming offer, merge the
new offer with the old until the negotiation is complete. Only replace
the sdp if and when the negotion completes. Roll back if it fails.

Also Adds unit tests for the merge/rollback/commit semantics and pending
media-state transaction ownership.

Fixes: #2059

### DIFF
--- a/res/res_pjsip_sdp_rtp.c
+++ b/res/res_pjsip_sdp_rtp.c
@@ -509,8 +509,19 @@ static struct ast_format_cap *set_incoming_call_offer_cap(
 	 */
 	ast_rtp_codecs_payloads_xover(&codecs, &codecs, NULL);
 
-	ast_rtp_codecs_payloads_copy(&codecs,
-		ast_rtp_instance_get_codecs(session_media->rtp), session_media->rtp);
+	/*
+	 * This only merges the newly offered payloads into the live RTP instance
+	 * rather than doing a destructive replace, so any format the channel is
+	 * still actively sending in (e.g. while a reINVITE is in flight) remains
+	 * usable right up until set_caps() commits the channel to the new
+	 * negotiated format.
+	 */
+	if (ast_sip_session_media_state_payloads_merge(session->pending_media_state,
+			&codecs, ast_rtp_instance_get_codecs(session_media->rtp),
+			session_media->rtp)) {
+		ao2_cleanup(incoming_call_offer_cap);
+		incoming_call_offer_cap = NULL;
+	}
 
 	ast_rtp_codecs_payloads_destroy(&codecs);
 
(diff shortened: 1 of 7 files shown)


mbradeen:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

mbradeen:
testsuite-test-pr: 150

seanbright:
AI disclosure?

res/res_pjsip_sdp_rtp.c
```diff
@@ -509,8 +509,19 @@ static struct ast_format_cap *set_incoming_call_offer_cap(
...
-	ast_rtp_codecs_payloads_copy(&codecs,
-		ast_rtp_instance_get_codecs(session_media->rtp), session_media->rtp);
+	/*
+	 * This only merges the newly offered payloads into the live RTP instance
+	 * rather than doing a destructive replace, so any format the channel is
+	 * still actively sending in (e.g. while a reINVITE is in flight) remains
+	 * usable right up until set_caps() commits the channel to the new
+	 * negotiated format.
+	 */
+	if (ast_sip_session_media_state_payloads_merge(session->pending_media_state,
```
jcolp, line 519: I've got a question before considering: Why do we mutate state on RTP instance here in the first place? Is this necessary, or could that be done differently instead? For example instead of using codecs on the RTP instance we instead use a codecs on the ast_sip_session_media instead, and only commit it once negotiated - like is done for other things? That would keep the RTP codecs stuff simple without this intermediary merging, and also make it so no container of payload merge buckets would need to exist.

jcolp (changes requested):

jcolp:
@mbradeen Indeed, per Sean this needs AI usage disclosure.

PR #2067, opened 2026-08-05
pjsip: Eliminate some unique taskprocessors.

gtjoseph:
When placing outbound calls each call would get its own
taskprocessor for things related to the call. In practice this
is overkill as few things actually occur. Inbound calls on
the other hand already use one of the fixed number of
distributor taskprocessors. This also occurred for OPTIONS
requests with each AOR having its own taskprocessor.

This change moves both to using a distributor taskprocessor
instead.


gtjoseph:
cherry-pick-to: none

gtjoseph:
This is a cherry-pick of @jcolp's original PR to the cert-22.8 branch.

jcolp (approved):

PR #2068, opened 2026-08-05
chan_websocket: Fix NULL requestor dereference in webchan_request.

SvenKube:
Two log statements in webchan_request() passed `requestor` straight to ast_channel_name(), which dereferences the channel with no NULL check. `requestor` is NULL whenever no originator channel is supplied: ARI POST /channels/externalMedia always passes NULL, and POST /channels/create passes NULL when `originator` is omitted.


SvenKube:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

jcolp (approved):

PR #2070, opened 2026-08-05
res_pjsip_sdp_rtp: Fix intermittent call drop on reINVITE sdp change

### LINKED ISSUE #2059: [bug]: Occasional call drops during re-INVITE when codec changes

mbradeen:
A reINVITE that changed the offered codec set (e.g. ulaw+alaw -> alaw
only) could lead to call teardown when a bridge write still using the
old format occurred during the re-negotiation.

This change avoids the termination by making the following changes.

First, Asterisk now holds the lock across both the tx payload update
and the channel format update so they commit atomically.

Second, when processing the sdp on the new incoming offer, merge the
new offer with the old until the negotiation is complete.

Also adds unit tests for the sdp merge

Fixes: #2059

### DIFF
--- a/res/res_pjsip_sdp_rtp.c
+++ b/res/res_pjsip_sdp_rtp.c
@@ -509,7 +509,16 @@ static struct ast_format_cap *set_incoming_call_offer_cap(
 	 */
 	ast_rtp_codecs_payloads_xover(&codecs, &codecs, NULL);
 
-	ast_rtp_codecs_payloads_copy(&codecs,
+	/*
+	 * This only merges the newly offered payloads into the live RTP instance
+	 * rather than doing a destructive replace, so any format the channel is
+	 * still actively sending in (e.g. while a reINVITE is in flight) remains
+	 * usable right up until set_caps() commits the channel to the new
+	 * negotiated format. A destructive replace here, before the channel's
+	 * own read/write format has caught up, is what let a concurrent bridge
+	 * write briefly land on a payload mapping that no longer had its format.
+	 */
+	ast_rtp_codecs_payloads_merge(&codecs,
 		ast_rtp_instance_get_codecs(session_media->rtp), session_media->rtp);
 
 	ast_rtp_codecs_payloads_destroy(&codecs);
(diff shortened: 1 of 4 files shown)


mbradeen:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

mbradeen:
AI Disclosure: Codex 5.6 Sol was used to review and to assist in generating the associated unit tests.

res/res_pjsip_sdp_rtp.c
```diff
@@ -531,6 +540,7 @@ static int set_caps(struct ast_sip_session *session,
 	int direct_media_enabled = !ast_sockaddr_isnull(&session_media->direct_media_addr) &&
 		ast_format_cap_count(session->direct_media_cap);
 	int dsp_features = 0;
+	int session_channel_locked = 0;
```
jcolp: Why is this needed? If session->channel is present won't it be locked?

res/res_pjsip_sdp_rtp.c
```diff
@@ -581,13 +591,18 @@ static int set_caps(struct ast_sip_session *session,
 		 */
 		ast_rtp_codecs_payloads_xover(&codecs, &codecs, NULL);
 	}
+
+	if (session->channel){
```
jcolp: if (session->channel) {

jcolp (changes requested):

res/res_pjsip_sdp_rtp.c
mbradeen: I was paranoid about a situation where the channel would exist on the unlock but not the lock.  I have switched it to the simpler comparison.

jcolp (approved):

jcolp:
Looks like 20.7 will need a bespoke PR

gtjoseph:
> Looks like 20.7 will need a bespoke PR

After discussion, we're going to skip cert-20.7

gtjoseph (approved):

PR #2071, opened 2026-08-06
res_pjsip_header_funcs: Add stored headers to Transfer() REFER

youjelly:
Headers stored on a session with PJSIP_HEADER(add, ...) were previously
only applied to outbound INVITE requests by the header_funcs session
supplement. The REFER generated by the Transfer() dialplan application
is built directly on the dialog in chan_pjsip with pjsip_xfer_initiate(),
which bypasses the session supplement path entirely, so the stored
headers were never included on it. Inheritable headers stored on the
channel with PJSIP_INHERITABLE_HEADER(add, ...) were likewise never
included: they are applied only by the same INVITE-only supplement,
which additionally never runs on an established (CONFIRMED) dialog.

The header application logic has been refactored into a new function,
ast_sip_session_add_stored_pjsip_headers(), exported as an optional API
(AST_OPTIONAL_API) so no new hard module dependency is introduced.
chan_pjsip now calls it from transfer_refer() just before the REFER is
sent. As with the INVITE path, the datastore is consumed once applied.
Only headers explicitly added via PJSIP_HEADER(add, ...) are applied;
headers captured from incoming requests for PJSIP_HEADER(read, ...) are
never copied to outbound messages.

A second optional API, ast_sip_session_add_inheritable_pjsip_headers(),
applies the headers stored on the session's channel by
PJSIP_INHERITABLE_HEADER(add, ...) and is also called from
transfer_refer(). Unlike the PJSIP_HEADER store, the channel datastore
is not consumed: inheritable headers must survive on the channel so
they can still be inherited by future channels.

UserNote: Headers added with PJSIP_HEADER(add, ...) or
PJSIP_INHERITABLE_HEADER(add, ...) on a channel are now also included
on the SIP REFER generated when the Transfer() application is executed
on that channel.

DeveloperNote: res_pjsip_header_funcs now provides
ast_sip_session_add_stored_pjsip_headers() and
ast_sip_session_add_inheritable_pjsip_headers() (declared in
res_pjsip_session.h) as optional APIs to apply PJSIP_HEADER-stored and
PJSIP_INHERITABLE_HEADER-stored headers to a pjsip_tx_data built outside
the session supplement path.


youjelly:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

youjelly:
AI assistance disclosure (per the AI Policy): this change was analyzed, designed and implemented with AI coding assistance. The plan was verified against the source before implementation, and the result was verified with a full local build (gcc 12 / Debian 12, bundled pjproject): clean compile with no new warnings, and correct AST_OPTIONAL_API wiring confirmed on both the provider (res_pjsip_header_funcs) and consumer (chan_pjsip) sides via symbol inspection. I have reviewed the diff and own the change.

gtjoseph:
@youjelly Is there a reason this can't be cherry-picked to the 20 branch?

gtjoseph (dismissed):
You're going to have to investigate the two refer test failures and figure out whether the issue is in this PR or the tests need to be adjusted.

youjelly:
No particular reason — 20 works too. Added `cherry-pick-to: 20`. Thanks!

youjelly:
The `channels/pjsip/allow_sending_180_after_183/disabled` failure in master-pjs2-check-849 appears unrelated to this PR:

- This PR only touches the Transfer()/REFER path (`transfer_refer()` in chan_pjsip + the new optional APIs in res_pjsip_header_funcs); it has no effect on provisional (180/183) response handling.
- The same master-pjs2 gate (128 tests) passed on identical branch content an hour earlier (run 32321137130, sha 373488d4a). The failure occurred on the follow-up run against the auto-merge with master (d8b379b0); none of the 13 intervening master commits touch the provisional-response path.
- This test has a history of periodic failures (asterisk/testsuite@8daa369 "testsuite: fix periodic 180 after 183 failures"), and its `enabled` sibling passed in the same run.

Could a maintainer re-trigger the check (`recheckpr`)?

jcolp:
Fixing the checklist would also cause it to rerun:

https://github.com/asterisk/asterisk/pull/2071#pullrequestreview-4978397780

jcolp (changes requested):
I'm not open to accepting the change for a broader reason.

As a project what we need is overall header control in the dialplan for multiple use cases outside of INVITE such as BYE as well, or SIP responses, not just REFER. If we accept each individual case it's going to get messy and the development itself is going to end up organic instead of something that is more thought out and has user feedback as well as needs taken into account.

I'll leave this PR open so others can grab it if they wish, or perhaps even comment, but the broader thing will need to be looked at first and as this has come up a few times I will raise it up, but I don't have a timeframe.

asteriskteam:

PR #2073, opened 2026-08-07
alembic: Preserve sub-second precision in the queue_log time column.

### LINKED ISSUE #2072: [bug]: queue_log 'time' column discards the microseconds ast_queue_log() writes on MySQL/MariaDB
### Severity

Minor

### Versions

20, 21, 22, 23, master

### Components/Modules

contrib/ast-db-manage (queue_log)

### Operating Environment

Any

### Frequency of Occurrence

Constant

### Summary

ast_queue_log() stamps realtime queue_log rows with microsecond precision (%F %T.%6q), but the shipped Alembic schema declares the time column as SQLAlchemy's generic sa.DateTime().

On MySQL and MariaDB that renders as a bare DATETIME, whose default fractional-seconds precision (FSP) is 0, so the fraction is dropped on insert. PostgreSQL is unaffected: sa.DateTime() renders as TIMESTAMP, which keeps microseconds.

### Impact

Events misorder and land in the wrong time bucket.

MariaDB truncates, so every event for a queue entry collapses into the same second and a reader has nothing but the id column to order them by. MySQL rounds, which can push an event into a second it did not occur in. Neither server warns, even in STRICT mode.

### Steps to reproduce

1. Apply the schema on MySQL or MariaDB: 
   ```sh 
   alembic -c queue_log.ini upgrade head
   ```

2. Check the precision (returns 0):
   ```sql
   SELECT datetime_precision FROM information_schema.columns 
   WHERE table_name = 'queue_log' AND column_name = 'time';
   ```

3. Insert a row carrying a microsecond timestamp:
   ```sql
    INSERT INTO queue_log (time, callid, queuename, agent, event)
    VALUES ('2026-08-03 09:15:22.834869', 'c1', 'q1', 'NONE', 'ENTERQUEUE');
    SHOW WARNINGS;              -- returns zero rows
    SELECT time FROM queue_log; -- MariaDB: 09:15:22   MySQL: 09:15:23
   ```

### Workaround

Widen the column by hand after alembic upgrade head:
   ```sql
    ALTER TABLE queue_log MODIFY time DATETIME(6);
   ```
Rows already written at whole-second precision stay that way. Nothing recovers them; the fraction was discarded at insert, not at read.

### Proposed fix

Add an Alembic revision to the queue_log tree that widens time to DATETIME(6) on MySQL and MariaDB only, guarded on `op.get_context().bind.dialect.name == 'mysql'`, which is the dialect name both servers report. Every other engine skips the step. Asterisk writes queue_log and never queries it, so no existing read path depends on the column's type.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

civilcoder55:
ast_queue_log() formats realtime queue_log timestamps with microseconds
("%F %T.%6q" in main/logger.c), but the queue_log table declares 'time' as
a generic SQLAlchemy DateTime.  That renders on MySQL and MariaDB as bare
DATETIME, which has a fractional-seconds precision of 0, so MariaDB
truncates the fraction and MySQL rounds it.  Every event of a single queue
entry can collapse onto one second, or move to the next one, and neither
server warns, not even in strict mode.

Once that happens, queue_log rows can no longer be ordered against each
other, or against the CEL records that cel_odbc and cel_pgsql do stamp
with microseconds.

This adds a revision that widens the queue_log column to DATETIME(6) on
MySQL and MariaDB.  PostgreSQL is unaffected and left untouched: DateTime
already renders as TIMESTAMP, which keeps microseconds by default.

Fixes: #2072

UpgradeNote: (schema) On MySQL and MariaDB the queue_log 'time' column is widened
from DATETIME to DATETIME(6) so that the microsecond timestamps Asterisk
already writes are stored rather than discarded. This ALTER rewrites the
table, a datetime precision change is neither INSTANT nor INPLACE, so
queue_log writes will block for its duration on a large table. PostgreSQL
installations are unaffected. Rows already written at whole-second precision will remain as-is and won't be updated.

### DIFF
--- a/contrib/ast-db-manage/queue_log/versions/8e6a9c2f4b17_increase_queue_log_time_precision.py
+++ b/contrib/ast-db-manage/queue_log/versions/8e6a9c2f4b17_increase_queue_log_time_precision.py
@@ -1,0 +1,38 @@
+"""increase queue_log time precision
+
+Revision ID: 8e6a9c2f4b17
+Revises: 4105ee839f58
+Create Date: 2026-08-03 00:00:00.000000
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = "8e6a9c2f4b17"
+down_revision = "4105ee839f58"
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects.mysql import DATETIME as MYSQL_DATETIME
+
+
+def upgrade():
+    # MariaDB is reported as the 'mysql' dialect, so this covers both.
+    if op.get_context().bind.dialect.name == "mysql":
+        op.alter_column(
+            "queue_log",
+            "time",
+            existing_type=sa.DateTime(),
+            type_=MYSQL_DATETIME(fsp=6),
+            existing_nullable=True,
+        )
+
+
+def downgrade():
+    if op.get_context().bind.dialect.name == "mysql":
+        op.alter_column(
+            "queue_log",
+            "time",
+            existing_type=MYSQL_DATETIME(fsp=6),
+            type_=sa.DateTime(),
+            existing_nullable=True,
+        )


civilcoder55:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (changes requested):
This part of the commit message needs to be clarified:

"Rows already written at whole-second precision are not recovered."

It can read as though this will cause those rows to be removed, not that the whole-second precision of existing rows can not be recovered.

civilcoder55:
> This part of the commit message needs to be clarified:
> 
> "Rows already written at whole-second precision are not recovered."
> 
> It can read as though this will cause those rows to be removed, not that the whole-second precision of existing rows can not be recovered.

Old rows will still be as is, only new rows will get the microseconds precision. Would you like me to rephrase it? 

image

jcolp:
Yes, otherwise some users who read the message may go "wait - my rows might disappear?".

jcolp (approved):

gtjoseph:
pr-checklist-exception: alembic-upgrade

PR #2075, opened 2026-08-08
taskpool: count the in-flight task when selecting an executor

### LINKED ISSUE #2074: [bug]: Taskpool assigns work to an executor that is inside a long task while other executors in the pool are idle
### Severity

Minor

### Versions

22.10.1, 23.4.1

### Components/Modules

main/taskpool.c, main/taskprocessor.c, res_pjsip

### Operating Environment

Debian 13 (trixie), kernel 6.12.90+deb13.1-cloud-amd64, x86_64
Asterisk built from source, bundled pjproject 2.17, bundled jansson
Configured with --enable-dev-mode; build options DEBUG_THREADS, BUILD_NATIVE,
OPTIONAL_API, TEST_FRAMEWORK
Both 22.10.1 and 23.4.1 are the current releases of their branches, built and
tested with identical flags

### Frequency of Occurrence

Constant

### Issue Description

A task pushed to a taskpool serializer does not run while another executor in
the same pool is inside a long-running task, even when the rest of the pool is
idle. Attached are two unit tests that reproduce it and a patch. Both tests fail
on every run on 22.10.1 and 23.4.1; in the first one, three of the four
executors are free the whole time. I am not attributing any production
incident to this: everything below is from the lab.

`taskpool_least_full_selector()` picks an executor by queue depth alone, and a
task that is currently executing has already been removed from its queue. An
executor inside a long task therefore reports a size of 0, is treated as "the
best choice", and receives new work. This matters for res_pjsip in particular,
since it is the main consumer of the taskpool and its serializers carry
per-session signalling, though I have not reproduced it at the SIP level.

## The selector

    /* main/taskpool.c:302-307 */
    for (i = 0; i < AST_VECTOR_SIZE(&taskprocessors->taskprocessors); i++) {
            struct taskpool_taskprocessor *tp = AST_VECTOR_GET(&taskprocessors->taskprocessors, i);

            /* If this taskprocessor has no outstanding tasks, it is the best choice */
            if (!ast_taskprocessor_size(tp->taskprocessor)) {
                    *taskprocessor = tp;
                    return;
            }

`ast_taskprocessor_size()` returns the queue length only:

    /* main/taskprocessor.c:1087 */
    return (tps) ? tps->tps_queue_size : -1;

and `ast_taskprocessor_execute()` pops the task before running it:

    /* main/taskprocessor.c:1075-1076 */
    if ((task = AST_LIST_REMOVE_HEAD(&tps->tps_queue, list))) {
            --tps->tps_queue_size;

So "no outstanding tasks" is true for an executor that is busy. The first such
entry in the vector wins and the loop returns immediately.

## Why the pool does not compensate

Growth is evaluated from the same number:

    /* main/taskpool.c:311-313, the site that applies to PJSIP,
       which uses the default least_full selector */
    if (ast_taskprocessor_size(tp->taskprocessor) >= pool->options.growth_threshold) {
            *growth_threshold_reached = 1;
    }

PJSIP neither exposes `growth_threshold` in pjsip.conf nor sets it, so
`ast_taskpool_create()` fills it in at main/taskpool.c:371-372 with
`TASKPOOL_GROW_THRESHOLD`, that is 250 given the high water level of 500
(include/asterisk/taskprocessor.h:64). Long-running tasks do not build a queue
of that size, so growth is never requested.

Selection is also two-stage:

    /* main/taskpool.c:542-550 */
    pool->selector(pool, &pool->static_taskprocessors, &taskprocessor, &growth_threshold_reached);
    if (pool->options.auto_increment && growth_threshold_reached) {
            pool->selector(pool, &pool->dynamic_taskprocessors, &taskprocessor, &growth_threshold_reached);

Dynamic executors are only consulted once the static pass reports a queue at the
threshold. Below it they never receive work, `last_pushed` never advances, and
the shrink scheduler reclaims them, so a pool with a larger initial_size settles
back to minimum_size.

From reading the code rather than from a reproduction:
res_pjsip/pjsip_distributor.c:581 pushes `distribute()` onto a serializer, and
`distribute()` runs `pjsip_endpt_process_rx_data()`, the whole PJSIP module
chain, inside it. Some paths there can hold an executor for a while, for example
`ast_sockaddr_resolve()` on a host taken from an incoming message
(res_pjsip_sdp_rtp.c:1592 and :2287, res_pjsip_t38.c:886 and :1060,
res_pjsip_acl.c:179).

## Reproduction

Add the attached tests (test_taskpool_selector.c.txt) to tests/test_taskpool.c,
register them in load_module() and unload_module(), and build with
--enable-dev-mode. They are written to fail on an unpatched build; happy to
adapt them if they were to go in a PR.

The first isolates the selector, with growth and idle timeout disabled. One
executor of four is held inside a task; a trivial task is then pushed on a
different serializer:

    CLI> test execute category /main/taskpool/ name selector_skips_busy_executor

    START  /main/taskpool/ - selector_skips_busy_executor
    one executor is inside a task; 3 of 4 executors are free
    the task did not run within 3s while 3 executors were free, so it was assigned to the busy one
    END    /main/taskpool/ - selector_skips_busy_executor Time: 3201ms Result: FAIL

The second uses a pool with initial_size 32 and max_size 500, and 16
long-running tasks:

    CLI> test execute category /main/taskpool/ name decays_to_minimum

    pool created: minimum_size=4 initial_size=32 idle_timeout=3s -> 4 static + 28 dynamic executors
    after the quiet period, long running tasks that got an executor: 4 of 16
    probes pushed=32 executed=0 queued=32 busy_inside=4
    POOL DECAYED: after an idle period the pool is down to minimum_size (4) and cannot grow back.
    32 trivial task(s) stalled with max_size=500 unused.
    END    /main/taskpool/ - decays_to_minimum Time: 33003ms Result: FAIL

Both fail on 22.10.1 and 23.4.1, built with identical flags. The ten existing
/main/taskpool/ tests pass on both.

## Why this does not look like a tuning problem

I swept every taskpool_* option pjsip.conf exposes (defaults: minimum_size 4,
initial_size 4, auto_increment 1, idle_timeout 60, max_size 50). `max_size` and
`auto_increment` change nothing under long-running tasks. `initial_size` helps
only until the dynamic executors are reclaimed. Only `taskpool_minimum_size`
changes the outcome, and it has to be at least the number of concurrent
long-running tasks.

The behaviour changed within each branch at the same point: 22.7.0 has
`ast_threadpool_create("pjsip", ...)` in res_pjsip.c and 22.8.0 has
`ast_taskpool_create("pjsip", ...)`, and the same applies to 23.1.0 and 23.2.0.
The change comes from commit 2d288891 ("pjsip: Move from threadpool to
taskpool"), reverted in b2e17cb7 and reapplied in d23fcd30. The threadpool grew
on executor exhaustion rather than on queue depth:

    /* 22.7.0 main/threadpool.c:588 */
    /* If no idle threads could be transitioned to active grow the pool as permitted. */
    if (ao2_container_count(pool->active_threads) == existing_active) {
            grow(pool, pool->options.auto_increment);

With its defaults (initial 0, auto_increment 5, max 50) that tolerated up to 50
concurrent blocking tasks; the taskpool tolerates minimum_size, which is 4. Since
threadpool_* and taskpool_* write the same fields
(res_pjsip/config_system.c:205-222), a pjsip.conf carried across that upgrade
keeps its numbers and changes behaviour.

Is the use of queue depth alone in `taskpool_least_full_selector()` intentional,
given that `tps->executing` is available in the same structure? If serializers
are meant to stay bound to one executor for ordering reasons, that would explain
it, but the pool would then need another way to notice an executor is
unavailable, since queue depth cannot show it.

Searched the tracker, open and closed, for taskpool, taskpool_minimum_size,
least_full, growth_threshold, serializer_task_pushed, execute_tasks, was_empty
and threadpool_initial_size; reviewed #436, #1558, #1961, #1972 and #2036, none
of which describe this.

## Possible direction for a fix

Counting the in-flight task when the selector measures load, and treating "no
executor free" as a growth signal, changes the behaviour. The attached patch
(taskpool_selector.patch.txt) does that in three places: a new accessor
(`ast_taskprocessor_is_executing`), a helper in taskpool.c that adds the
in-flight task to the reported load, and the growth flag set when the least_full
loop finds no free executor.
`ast_taskprocessor_size()` is left alone so its public semantics do not change.

It applies cleanly to 22.10.1 and 23.4.1, and on both the whole /main/taskpool/
category passes with it (12 of 12). In the second test the pool grows to 16
executors rather than staying at 4.

It is a starting point rather than a proposal: the growth policy is yours to
decide, and there may be a better place for the signal than the selector. Two
trade-offs in it are worth discussing, which I will add as a comment.

## Workaround

Set taskpool_minimum_size to at least the expected number of concurrent PJSIP
sessions that can hold an executor for a long time. It requires a restart, since
the pool is created once in load_module().

---

AI disclosure: this report was prepared with AI assistance (Claude Opus 5, model
id claude-opus-5, via Claude Code). It ran the lab work and drafted the text
under my direction; I reviewed the findings, verified the code references
against the 22.10.1 and 23.4.1 sources, and posted this. A second independent AI
session was used to review and correct the analysis before submission.

[taskpool_selector.patch.txt](https://github.com/user-attachments/files/30844666/taskpool_selector.patch.txt)
[test_taskpool_selector.c.txt](https://github.com/user-attachments/files/30844665/test_taskpool_selector.c.txt)

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

giandiego:
The taskpool selectors measure load with ast_taskprocessor_size(), which
only reports queued tasks. A taskprocessor that is executing a long
running task has an empty queue, so it looks identical to an idle one and
the least full selector hands it more work while other executors sit free.

Add ast_taskprocessor_is_executing() and include the in-flight task in the
load the selectors compare.

This is not a new semantic. taskprocessor_push() already counts the
in-flight task as outstanding work when it notifies the listener:

	/* The currently executing task counts as still in queue */
	was_empty = tps->executing ? 0 : previous_size == 0;

The taskpool selectors are the one place that still measures only the
queue. This change applies the same rule there.

The accessor reads tps->executing without the object lock, the way the CLI
already does in main/taskprocessor.c, because taking the lock on every
push cost roughly six times the throughput: taskpool push efficiency went
from 324438 to 52966 tasks per second and the serializer variant from
217467 to 38313. Without the lock both are back to baseline. A stale read
only means one push is routed as if the taskprocessor had just changed
state, which the selector already tolerates.

Also add a unit test that holds one executor of a four executor pool
inside a long task and checks that a task pushed to a different
serializer runs on one of the three free executors. It fails on every run
without this change.

Fixes: #2074

UserNote: Taskpool executors that are running a long task are no longer
considered idle when work is distributed, so tasks are routed to free
executors instead of queueing behind a busy one.

### DIFF
--- a/include/asterisk/taskprocessor.h
+++ b/include/asterisk/taskprocessor.h
@@ -345,6 +345,15 @@ const char *ast_taskprocessor_name(struct ast_taskprocessor *tps);
 long ast_taskprocessor_size(struct ast_taskprocessor *tps);
 
 /*!
+ * \brief Return whether the taskprocessor is currently executing a task
+ *
+ * \param tps taskprocessor
+ * \retval 0 if the taskprocessor is not executing a task.
+ * \retval non-zero if the taskprocessor is executing a task.
+ */
+unsigned int ast_taskprocessor_is_executing(const struct ast_taskprocessor *tps);
+
+/*!
  * \brief Return the listener associated with the taskprocessor
  */
 struct ast_taskprocessor_listener *ast_taskprocessor_listener(struct ast_taskprocessor *tps);
--- a/main/taskprocessor.c
+++ b/main/taskprocessor.c
@@ -1087,6 +1087,22 @@ long ast_taskprocessor_size(struct ast_taskprocessor *tps)
 	return (tps) ? tps->tps_queue_size : -1;
 }
 
+unsigned int ast_taskprocessor_is_executing(const struct ast_taskprocessor *tps)
+{
+	if (!tps) {
+		return 0;
+	}
+
+	/*
+	 * Read without the object lock. This is called from the taskpool
+	 * selector on every push, and taking the lock there costs more than the
+	 * value is worth: a stale answer only means one push is routed as if the
+	 * taskprocessor had just changed state, which the selector already
+	 * tolerates.
+	 */
+	return tps->executing;
+}
+
 struct ast_taskprocessor_listener *ast_taskprocessor_listener(struct ast_taskprocessor *tps)
 {
 	return tps ? tps->listener : NULL;
--- a/tests/test_taskpool.c
+++ b/tests/test_taskpool.c
@@ -1067,6 +1067,214 @@ end:
 	return CLI_SUCCESS;
 }
 
+#define SEL_POOL 4
+
+struct sel_data {
+	ast_mutex_t lock;
+	ast_cond_t cond;
+	int blocker_running;
+	int blocker_release;
+	int blocker_done;
+	int probe_ran;
+};
+
+static int sel_blocking_task(void *data)
+{
+	struct sel_data *sd = data;
+
+	ast_mutex_lock(&sd->lock);
+	sd->blocker_running = 1;
+	ast_cond_broadcast(&sd->cond);
+	while (!sd->blocker_release) {
+		ast_cond_wait(&sd->cond, &sd->lock);
+	}
+	sd->blocker_done = 1;
+	ast_cond_broadcast(&sd->cond);
+	ast_mutex_unlock(&sd->lock);
+	return 0;
+}
+
+static int sel_probe_task(void *data)
+{
+	struct sel_data *sd = data;
+
+	ast_mutex_lock(&sd->lock);
+	sd->probe_ran = 1;
+	ast_cond_broadcast(&sd->cond);
+	ast_mutex_unlock(&sd->lock);
+	return 0;
+}
+
+/*!
+ * \internal
+ * \brief Run the scenario once.
+ *
+ * \param test The test being run.
+ * \param probe_first Create the probe serializer before the blocking task runs.
+ * \param label Description of the scenario, used in the test output.
+ * \retval 1 the probe task ran
+ * \retval 0 the probe task did not run within the timeout
+ */
+static int sel_run_case(struct ast_test *test, int probe_first, const char *label)
+{
+	struct ast_taskpool *pool = NULL;
+	struct sel_data *sd = NULL;
+	struct ast_taskpool_options options = {
+		.version = AST_TASKPOOL_OPTIONS_VERSION,
+		.idle_timeout = 0,
+		.auto_increment = 0,
+		.minimum_size = SEL_POOL,
+		.initial_size = SEL_POOL,
+		.max_size = SEL_POOL,
+	};
+	struct ast_taskprocessor *blocker_ser = NULL;
+	struct ast_taskprocessor *probe_ser = NULL;
+	int ran = 0;
+	int running;
+	struct timeval start;
+	struct timespec end;
+
+	sd = ast_calloc(1, sizeof(*sd));
+	if (!sd) {
+		return 0;
+	}
+	ast_mutex_init(&sd->lock);
+	ast_cond_init(&sd->cond, NULL);
+
+	pool = ast_taskpool_create(label, &options);
+	if (!pool) {
+		ast_test_status_update(test, "%s: could not create pool\n", label);
+		goto cleanup;
+	}
+
+	blocker_ser = ast_taskpool_serializer("blocker", pool);
+	if (!blocker_ser) {
+		goto cleanup;
+	}
+
+	if (probe_first) {
+		probe_ser = ast_taskpool_serializer("probe", pool);
+		if (!probe_ser) {
+			goto cleanup;
+		}
+	}
+
+	if (ast_taskprocessor_push(blocker_ser, sel_blocking_task, sd)) {
+		goto cleanup;
+	}
+
+	/* Wait until the blocker is definitely inside its task. */
+	start = ast_tvnow();
+	end.tv_sec = start.tv_sec + 5;
+	end.tv_nsec = start.tv_usec * 1000;
+
+	ast_mutex_lock(&sd->lock);
+	while (!sd->blocker_running
+		&& ast_cond_timedwait(&sd->cond, &sd->lock, &end) != ETIMEDOUT) {
+	}
+	running = sd->blocker_running;
+	ast_mutex_unlock(&sd->lock);
+
+	if (!running) {
+		ast_test_status_update(test, "%s: blocking task never started\n", label);
+		goto release;
+	}
+
+	/*
+	 * With probe_first == 0 the probe serializer is created here, while one
+	 * executor is already busy and three are free.
+	 */
+	if (!probe_first) {
+		probe_ser = ast_taskpool_serializer("probe", pool);
+		if (!probe_ser) {
+			goto release;
+		}
+	}
+
+	if (ast_taskprocessor_push(probe_ser, sel_probe_task, sd)) {
+		goto release;
+	}
+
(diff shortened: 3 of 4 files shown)


giandiego:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

giandiego:
This work was done with AI assistance (Claude). The AI ran the lab reproduction, drafted the patch and the unit test, and did the code search across the PJSIP receive path. I evaluated each hypothesis as it came up and discarded the ones that did not hold, verified the code references against the sources, and ran commands on the systems myself to confirm what was claimed. What went into this PR is what survived that. The growth signal came out after Joshua explained the design intent, and the decay test came out with it. I can debug and own this change.

include/asterisk/taskprocessor.h
```diff
@@ -344,6 +344,15 @@ const char *ast_taskprocessor_name(struct ast_taskprocessor *tps);
...
 long ast_taskprocessor_size(struct ast_taskprocessor *tps);
 
+/*!
+ * \brief Return whether the taskprocessor is currently executing a task
+ *
+ * \param tps taskprocessor
+ * \retval 1 a task is currently being executed
+ * \retval 0 otherwise
+ */
+int ast_taskprocessor_is_executing(struct ast_taskprocessor *tps);
```
jcolp: unsigned int, and tps can be const

main/taskprocessor.c
```diff
@@ -1087,6 +1087,22 @@ long ast_taskprocessor_size(struct ast_taskprocessor *tps)
...
+	}
+
+	/*
+	 * Read without the object lock. This is called from the taskpool
+	 * selector on every push, and taking the lock there costs more than the
+	 * value is worth: a stale answer only means one push is routed as if the
+	 * taskprocessor had just changed state, which the selector already
+	 * tolerates.
+	 */
+	return tps->executing ? 1 : 0;
```
jcolp: This can just return tps->executing, as it is always 1 or 0.

tests/test_taskpool.c
```diff
@@ -1067,6 +1069,201 @@ static char *handle_cli_taskpool_push_serializer_efficiency(struct ast_cli_entry
...
+		"%s: one executor busy, %d of %d free -> probe %s\n",
+		label, SEL_POOL - 1, SEL_POOL,
+		ran ? "ran on a free executor" : "did NOT run within 3s");
+
+release:
+	ast_mutex_lock(&sd->lock);
+	sd->blocker_release = 1;
+	ast_cond_broadcast(&sd->cond);
+	ast_mutex_unlock(&sd->lock);
+	usleep(200000);
```
jcolp: Is there any way to not use an explicit usleep here and instead coordinate so it can terminate faster when things complete?

jcolp (changes requested):

tests/test_taskpool.c
giandiego: Yes, that worked: the blocking task now signals when it leaves and the wait is a condvar with a timeout only as a guard, so the case ends as soon as the task completes. The test went from 403ms to 2ms.

jcolp (approved):

PR #2077, opened 2026-08-09
app_voicemail: Fix crash when VM_INFO reads an unset email address.

### LINKED ISSUE #2063: [bug]: VM_INFO segfaults when requesting an email address from a mailbox that doesn't have an email address
### Severity

Trivial

### Versions

22.6.0

### Components/Modules

 app_voicemail_odbc.c

### Operating Environment

Debian Trixie

### Frequency of Occurrence

Constant

### Issue Description

When using one touch recording and attempting to send recording to user by using the email generally configured in voicemail.conf
we implemented the following dialplan, when caller hangs up, the callback attempts to execute `same => n,Set(RECORDING_EMAIL=${VM_INFO(${ARG1}@default,email)})` and segfaults

This has been previously noted as an issue 
[Issue 629](https://github.com/asterisk/asterisk/issues/629)
> The cause is consistent, an attempt to call ast_copy_string on NULL in the test_voicemail_vm_info unit test:

voicemail.conf excert
8487 => 1234,Service

```
exten => 8123,1,NoOp(Hangup handler Testing)
 same => n,Set(CHANNEL(hangup_handler_push)=sub-email-recording,s,1(8487,05551234))
 same => n,Playback(redacted)
```

```

[sub-email-recording]
exten => s,1,NoOp(Emailing one-touch recording)
 same => n,StopMixMonitor()
 same => n,Set(RECORDING_FILE=${TOUCH_MIXMONITOR_OUTPUT})
 same => n,Set(CALLED_NUMBER=${ARG2})

 ; Check the mailbox before requesting its email address
 same => n,Set(VM_EXISTS=${VM_INFO(${ARG1}@default,exists)})
 same => n,NoOp(Mailbox=${ARG1}@default Exists=${VM_EXISTS})

 same => n,GotoIf($["${VM_EXISTS}" != "1"]?no_mailbox)

 same => n,Set(RECORDING_EMAIL=${VM_INFO(${ARG1}@default,email)})
 same => n,NoOp(File=${RECORDING_FILE} Email=${RECORDING_EMAIL} Called=${CALLED_NUMBER})

 same => n,ExecIf($["${RECORDING_FILE}" != "" & "${RECORDING_EMAIL}" != ""]?System(/redacted/email-recording.sh "${RECORDING_FILE}" "${RECORDING_EMAIL}" "${CALLED_NUMBER}" >/dev/null 2>&1 &))
 same => n,Return()

 same => n(no_mailbox),NoOp(No voicemail mailbox for ${ARG1}@default - recording not emailed)
 same => n,Return()

```

from app_voicemail_odbc.c you can see there are no checks to see if email is valid/present before ast_copy_string(buf, vmu->email, len);

```

         if (!strncasecmp(arg.attribute, "exists", 5)) {
                ast_copy_string(buf, vmu ? "1" : "0", len);
                free_user(vmu);
                return 0;
        }

        if (vmu) {
                if (!strncasecmp(arg.attribute, "password", 8)) {
                        ast_copy_string(buf, vmu->password, len);
                } else if (!strncasecmp(arg.attribute, "fullname", 8)) {
                        ast_copy_string(buf, vmu->fullname, len);
                } else if (!strncasecmp(arg.attribute, "email", 5)) {
                        ast_copy_string(buf, vmu->email, len);


```

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

aabolfazl:
The email member of struct ast_vm_user is a pointer rather than a fixed
array, and populate_defaults() leaves it NULL when a mailbox has no
email address configured. Every other attribute VM_INFO reads is a fixed
array, so only the email attribute is affected.

VM_INFO passed vmu->email straight to ast_copy_string(), which
dereferences its source unconditionally. Reading the email attribute of
a mailbox that has no email address therefore crashed Asterisk from the
dialplan.

Guard the copy with S_OR() so an unset email address yields an empty
string, matching how the language attribute already handles its
fallback. The make_email_file() call sites were already guarded and are
left alone.

Add a regression test for the unset case, and restore the voicemail
configuration when the VM_INFO test finishes. That test was the only one
in app_voicemail that did not do so, which left its test mailbox in the
users list and made the test fail if it ran a second time.

Fixes: #2063

### DIFF
--- a/apps/app_voicemail.c
+++ b/apps/app_voicemail.c
@@ -13658,7 +13658,7 @@ static int acf_vm_info(struct ast_channel *chan, const char *cmd, char *args, ch
 		} else if (!strncasecmp(arg.attribute, "fullname", 8)) {
 			ast_copy_string(buf, vmu->fullname, len);
 		} else if (!strncasecmp(arg.attribute, "email", 5)) {
-			ast_copy_string(buf, vmu->email, len);
+			ast_copy_string(buf, S_OR(vmu->email, ""), len);
 		} else if (!strncasecmp(arg.attribute, "pager", 5)) {
 			ast_copy_string(buf, vmu->pager, len);
 		} else if (!strncasecmp(arg.attribute, "language", 8)) {
@@ -16132,8 +16132,23 @@ AST_TEST_DEFINE(test_voicemail_vm_info)
 		}
 	}
 
+	ast_free(vmu->email);
+	vmu->email = NULL;
+
+	ast_copy_string(vminfo_args, "00000000@test,email", sizeof(vminfo_args));
+	test_ret = acf_vm_info(chan, vminfo_cmd, vminfo_args, vminfo_buf, sizeof(vminfo_buf));
+	if (!ast_strlen_zero(vminfo_buf)) {
+		ast_test_status_update(test, "VM_INFO response for a mailbox without an email address was: '%s', but expected: ''\n", vminfo_buf);
+		res = AST_TEST_FAIL;
+	}
+	if (test_ret != 0) {
+		ast_test_status_update(test, "VM_INFO return code for a mailbox without an email address was: '%i', but expected '0'\n", test_ret);
+		res = AST_TEST_FAIL;
+	}
+
 	chan = ast_channel_unref(chan);
 	free_user(vmu);
+	force_reload_config(); /* Restore original config */
 	return res;
 }
 #endif /* defined(TEST_FRAMEWORK) */


aabolfazl:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23

aabolfazl:
Testing summary.

Built with dev mode following `tests/CI/buildAsterisk.sh` (`TEST_FRAMEWORK`,
`MALLOC_DEBUG`, `DO_CRASH`, `DONT_OPTIMIZE`, `BETTER_BACKTRACES`) and installed
to a `DESTDIR` sandbox.

`test execute all`: 673 executed, 671 passed. `test_voicemail_vm_info` passes,
as do the other five `/apps/app_voicemail/` tests and the ten
`/main/voicemail_api/` tests. The two failures are
`test_message_queue_both_nominal` and `test_message_queue_dialplan_nominal`,
which time out waiting for dialplan user events; both reproduce identically on
an unmodified tree in the same environment, so they are pre-existing and
unrelated to this change.

To confirm the new test actually covers the bug, I reverted only the
`acf_vm_info` hunk while keeping the test, and Asterisk crashed during
`test_voicemail_vm_info`:

```
#0  ast_copy_string (dst=0x7f703c476ef0 "", src=0x0, size=256) at strings.h:412
#1  acf_vm_info (args="00000000@test,email", len=256) at app_voicemail.c:13661
#2  test_voicemail_vm_info at app_voicemail.c:16143
```

Restoring the guard makes it pass again.

`make validate-docs` passes. No XML documentation changed; the existing
`<option name="email">` text already describes returning the mailbox's e-mail
address, and an unset address now yields an empty string rather than crashing.

I also confirmed the unguarded `ast_copy_string(buf, vmu->email, len)` is
present on 20, 22, 23 and master, hence the cherry-pick requests.

aabolfazl:
I used AI assistance to audit the NULL handling in this code path and to verify the fix with unit tests.

jcolp (approved):

PR #2079, opened 2026-08-10
res_pjsip_config_wizard: Fix documentation errors.

maksym-t-didww:
The sample configuration and the module's XML documentation contained
several errors, two of which actively misled users:

* The sample said that sends_auth and accepts_auth require
  "outbound/username" and "inbound/username".  The wizard looks for
  "outbound_auth/username" and "inbound_auth/username", so a
  configuration written by following the sample fails to load with
  "Wizard 'xxx' must have 'outbound_auth/username' if it sends
  authentication."

* The sample claimed that hint_application defaults to "Dial(${HINT})".
  There is no such default.  When hint_application is not specified the
  wizard removes the priority 1 extension it manages rather than
  creating one.

The remaining corrections are a reversed object/field reference
(match/identify), a leftover line in the hint_exten description, an
option name (send_registrations), a stray space (remote _hosts), an
unbalanced quote in a synopsis, a misspelling (nneds), an unbalanced
bracket in the CLI usage text, and a copy/paste error in a log message
that referred to the "sangoma wizard" instead of pjsip_wizard.  Two
"(default = ...)" annotations became "(default: ...)" to match the
other fourteen in the same file.

No functional change.

UserNote: The pjsip_wizard.conf sample documented the wrong option
names for sends_auth and accepts_auth, and claimed a default for
hint_application that does not exist.  Configurations written by
following the sample would have failed to load.

### DIFF
--- a/configs/samples/pjsip_wizard.conf.sample
+++ b/configs/samples/pjsip_wizard.conf.sample
@@ -72,12 +72,12 @@
 ;  SYNOPSIS: Provides configuration wizard for common scenarios.
 ;sends_auth=    ; Will create an outbound auth object for the endpoint and
                 ; registration.
-                ; If yes, outbound/username must be specified.
-                ; (default = "no")
+                ; If yes, outbound_auth/username must be specified.
+                ; (default: "no")
 
 ;accepts_auth=  ; Will create an inbound auth object for the endpoint.
-                ; If yes, inbound/username must be specified.
-                ; (default = "no")
+                ; If yes, inbound_auth/username must be specified.
+                ; (default: "no")
 
 ;sends_registrations=    ; Will create an outbound registration object and an
                          ; identify match for each host in remote_hosts (which
@@ -102,7 +102,7 @@
                  ; If specified, a static contact for each host will be created
                  ; in the aor.  If accepts_registrations is no, an identify
                  ; object is also created with a match line for each remote host.
-                 ; If an aor/contact or match/identify is explicitly supplied,
+                 ; If an aor/contact or identify/match is explicitly supplied,
                  ; remote_hosts will not be used to automatically create contacts
                  ; or matches respectively.
                  ; (default: "")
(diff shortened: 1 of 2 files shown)


maksym-t-didww:
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

maksym-t-didww:
Two notes on the automated checklist, and an AI disclosure.

The sample-configs check will ask for an Alembic change because this touches
configs/samples/pjsip_wizard.conf.sample and the check treats every pjsip*.sample
the same way. There is no database schema behind these options: the wizard reads
them with ast_variable_find_last_in_list() rather than registering them with
sorcery, so pjsip_wizard.conf has no realtime representation and nothing in
contrib/ast-db-manage to update.

AI assistance disclosure, per the project's AI Policy: the defect list and the
wording of this change were produced with AI assistance. Every item was checked
against the source before submission:

* The option names come from res_pjsip_config_wizard.c:920 and the error text it
  logs at :921-922.
* The hint_application behaviour comes from :536-543, where an empty value calls
  ast_context_remove_extension2() rather than add_extension().
* "identify/match" is the ordering the module's own XML documentation already
  uses at :260.
* The two "(default = ...)" annotations were the only two of sixteen in the
  sample not written as "(default: ...)".

I was not able to run a full dev-mode build on this host (macOS). The module's
XML documentation block validates against doc/appdocsxml.dtd with xmllint, which
is the check make validate-docs performs, and the change touches only comment
text and two string literals with no format specifiers altered. I tested the
cherry-pick locally against 23, 22 and 20: it applies cleanly to all three and
the same defects are present in each.

jcolp (approved):

maksym-t-didww:
Re the Alembic checklist item: see my comment above — pjsip_wizard.conf
options are read with ast_variable_find_last_in_list() and are not
registered with sorcery, so they have no realtime schema and there is
nothing in contrib/ast-db-manage to update.

jcolp (approved):

PR #2080, opened 2026-08-12
pjsip_wizard.conf.sample: Add IP authentication and TLS examples.

maksym-t-didww:
The only ITSP example in the file shows a single trunk that registers and
authenticates.  Many providers, including DIDWW which motivated this
change, instead authenticate by IP address, deliver inbound calls from
several regional POPs, and terminate outbound calls through a separate
gateway that does require digest authentication.  Several things about
configuring that with the wizard are not obvious from the option
reference:

* remote_hosts alone generates the per-host identify matches, and is
  mandatory when registrations are not accepted.
* Inbound and outbound need one wizard each, because a wizard has a
  single remote_hosts list.
* A template carrying "type = wizard" is not turned into objects itself,
  so the two can share one.

TLS has a pitfall of its own: the wizard cannot create transport objects,
so the transport has to be defined in pjsip.conf and referenced here by
name, while media_encryption belongs to the endpoint and needs the
endpoint/ prefix.

Add commented-out examples for an IP authenticated trunk pair sharing a
template, and for a TLS trunk with SRTP.

No functional change.

UserNote: pjsip_wizard.conf.sample now includes examples for an ITSP that
authenticates by IP address and delivers calls from several hosts, and
for a TLS trunk with SRTP.

### DIFF
--- a/configs/samples/pjsip_wizard.conf.sample
+++ b/configs/samples/pjsip_wizard.conf.sample
@@ -67,6 +67,64 @@
 ;endpoint/context = default
 
 
+;============EXAMPLE WIZARD CONFIGURATION FOR AN IP AUTHENTICATED ITSP======
+
+; This ITSP doesn't use registration.  It authenticates by IP address and
+; delivers inbound calls from any of several regional POPs, while outbound
+; calls go to a separate termination gateway that does require digest
+; authentication.  Inbound and outbound need one wizard each because a
+; single wizard has only one remote_hosts list, so they share a template.
+
+; Settings common to both trunks.  A template is never turned into objects
+; itself, so "type = wizard" here is safe.
+
+;[myitsp-common](!)
+;type = wizard
+;transport = ipv4
+;endpoint/allow = !all,ulaw,alaw
+;endpoint/context = from-myitsp
+
+; Inbound only.  This config would create an endpoint, an aor with a static
+; contact for each remote host, and an identify with a match for each remote
+; host, so a call from any of the POPs matches this endpoint.  No auth and no
+; registration objects are created.  remote_hosts is required here because
+; accepts_registrations is "no".
+
+;[myitsp-in](myitsp-common)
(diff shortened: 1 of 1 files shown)


maksym-t-didww:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

maksym-t-didww:
One note on the automated checklist, and an AI disclosure.

As with #2079, the sample-configs check will ask for an Alembic change because
this touches configs/samples/pjsip_wizard.conf.sample and the check treats every
pjsip*.sample the same way. The wizard reads its options with
ast_variable_find_last_in_list() rather than registering them with sorcery, so
pjsip_wizard.conf has no realtime representation and nothing in
contrib/ast-db-manage to update. This change adds only comments in any case.

AI assistance disclosure, per the project's AI Policy: the wording of these
examples was produced with AI assistance. Every behavioural claim in them was
checked against the source first:

* remote_hosts is mandatory when registrations are not accepted, and produces
  one identify match per host: res_pjsip_config_wizard.c:775 and handle_identify
  below it.
* The aor gets one static contact per remote host through contact_pattern
  (:640), and registrations pick up the wizard's transport (:870).
* outbound_auth is attached to the endpoint only when sends_auth is set (:717).
* A template carrying "type = wizard" is not instantiated: the wizard browses
  with the filter "type=^wizard$" (:1105) and does_category_match rejects
  cat->ignored unless the filter names TEMPLATES (main/config.c:1148), while
  ast_category_inherit (:1569) still copies "type" into each child, so the
  children are picked up. configs/basic-pbx/pjsip.conf:58 uses the same pattern
  for plain endpoints.
* [transport-tls] is the name used in pjsip.conf.sample:152 and :187, and
  media_encryption=sdes appears there at :358.

Placeholders follow the file's existing conventions (*.myitsp.com, testname /
test password) plus an RFC 5737 documentation address. I could not run a
dev-mode build on this host (macOS); the change adds commented lines only and
does not touch the module's XML documentation block.

jcolp (approved):

PR #2090, opened 2026-08-17
res_audiosocket: Bound idle poll retries while reading header

### LINKED ISSUE #2078: [bug]: 20.1x] ARI /channels requests stall after immediate hangup with AudioSocket ExternalMedia
### Severity

Major

### Versions

20.18.0, 20.19.0

### Components/Modules

PJSIP, ARI, Local channels, ExternalMedia over TCP/AudioSocket

### Operating Environment

Ubuntu 22.04, Asterisk built from sources, Ari application 

### Frequency of Occurrence

Frequent

### Issue Description

Hi!
Here’s a quick overview: we use Asterisk 20.1.0 and the ARI application to manage channels. We decided to upgrade to version 20.18 because it includes a fix for a bug we needed resolved—but we ended up finding a different bug instead.

### Problem

Under high concurrent call load, Asterisk 20.18.0 and 20.19.0 may stop processing ARI requests involving `/ari/channels` for several minutes.

The same workload does not reproduce the problem on Asterisk 20.1.0.

Asterisk does not freeze completely: ARI WebSocket remains connected and requests such as `GET /ari/bridges/{id}` may still succeed. However, the following requests remain pending:

- `POST /ari/channels/externalMedia`;
- `POST /ari/channels/{id}/answer`;
- `DELETE /ari/channels/{id}`.

### Trigger sequence

The problem occurs intermittently under load with the following sequence:

1. A Local channel answers and enters Stasis.
2. An AudioSocket ExternalMedia channel is created and added to a mixing bridge.
3. The AudioSocket server (ARI application) accepts the TCP connection and receives the UUID handshake.
4. The ARI application immediately hangs up the Local channel.
5. Asterisk returns `204` for the DELETE and sends `ChannelHangupRequest`, but does not send `StasisEnd` or `ChannelDestroyed`.
6. Asterisk starts logging:

   ```
   Exceptionally long voice queue length (...) queuing to Local/<extension>@<context>-...;1
   ```

7. Subsequent `/ari/channels` requests remain pending. New AudioSocket TCP connections are accepted, but their UUID handshake is not sent.
8. When the original AudioSocket TCP connection is closed by the peer, the channels are destroyed and Asterisk sends the accumulated ARI responses in a burst.

### Suspected cause

The suspected code path is:

``
res/res_audiosocket.c
ast_audiosocket_receive_frame_with_hangup()
```

In recent Asterisk 20.x versions, the AudioSocket header and payload read loops retry `read()` after `EAGAIN` using `poll()`, but do not have a total deadline.

Because the channel technology read callback may execute while the channel is locked, an idle or partially read AudioSocket connection can keep the channel locked for an unbounded period. This prevents channel teardown and creates a lock convoy for other `/ari/channels` operations.

The relevant read-loop behavior appears in Asterisk 20.17 and later. This is a suspected regression range; a runtime git bisect has not yet been completed.

### Expected behavior

An idle AudioSocket connection must not indefinitely block channel teardown or unrelated ARI channel operations.

After `DELETE /ari/channels/{id}`, the channel should leave Stasis within a bounded time even if the AudioSocket peer remains connected but sends no audio.

### Possible fix

The AudioSocket read callback should be fully nonblocking:

1. Preserve partial header and payload data in the AudioSocket instance.
2. Read only currently available bytes.
3. On `EAGAIN`, save the parser state and return control to the channel core.
4. Resume parsing during the next read callback.
5. Check for channel hangup between read attempts.
6. Avoid polling indefinitely while the channel lock is held.

A simpler but less complete fix would be to introduce a strict total timeout for assembling an AudioSocket frame.

A regression test should create an ExternalMedia AudioSocket channel, keep the TCP connection open without sending media, immediately hang up the Local channel, and verify that channel teardown and other ARI `/channels` requests still complete.


Note: We used a GPT-5.6 Sol model to help identify the problem.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

### LINKED ISSUE #2089: [bug]: res_audiosocket header read loop has no total deadline, starving AudioSocket outbound audio
**Setup:** an ARI application using `ExternalMedia` with AudioSocket transport (slin16), bridged two-party with a PJSIP channel. The AudioSocket peer waits for its first inbound audio frame before producing any outbound audio. Stock Asterisk 23.3.0, no patches.

**Symptom:** intermittently — roughly 1 call in 3 to 5 — the call sets up correctly and then carries no audio in either direction for its whole duration. The caller's RTP arrives at Asterisk normally throughout.

**Reproduction.** A batch of 9 consecutive calls under identical conditions: 8 completed normally (3.3–3.9s each), 1 was silent for its full 27.9s. Counting the loop's own debug line (`res_audiosocket.c:309`, "Poll timed out while waiting for header data") per call:

- 8 healthy calls: **0** occurrences each
- the failing call: **2,328** occurrences at ~83.5/s — the loop's 5ms poll cadence — for the entire call

RTP-side debug counters confirm the caller's inbound audio was accepted and processed entirely normally on the failing call (~50 packets/s, statistically identical to the healthy calls). The AudioSocket peer received **zero** frames. Exactly one write failure was logged, at teardown, once the peer's own timeout closed the TCP connection — consistent with `audiosocket_write()` essentially never being called during the call.

**Suspected cause.** `ast_audiosocket_receive_frame_with_hangup()`'s header-read loop retries `read()` after `EAGAIN` via a 5ms `ast_wait_for_input()`, and on timeout executes a bare `continue` with no counter and no total deadline. Per `bridge_channel_wait()`'s own doc comment — *"Each channel does writing/reading in their own thread"* — the same per-channel thread that must drain this channel's outbound queue is the thread spinning in that loop, so it never returns to service the write side while the peer stays silent. `__ast_read()` also holds the channel lock across `tech->read()`, which is consistent with the ARI-stall symptom reported separately in #2078.

Note the asymmetry: the **payload** read loop in the same function treats a single poll timeout as fatal. Only the header loop retries unconditionally.

Same root cause as #2078, different trigger and symptom (mid-call audio starvation rather than ARI stalls after hangup), and on a currently-supported release rather than 20.x. Patch to follow.

daniel-sullivan:
ast_audiosocket_receive_frame_with_hangup() reads a 3-byte AudioSocket
frame header before it reads the payload. When read() returns EAGAIN
and no header bytes have been consumed yet (i == 0), a poll timeout
on ast_wait_for_input() falls through to a bare continue, with no
limit on how many times that may happen.

The AudioSocket channel and application both service a connection
from a single dedicated thread that does both the read and the write
side. While that thread is spinning in this loop it never returns to
drain its outbound write queue, so an idle peer that keeps the TCP
connection open without sending further data can starve the channel's
outbound audio indefinitely, even though nothing is actually wrong
with the connection.

Bound the retry, but only while i == 0. In that state no header bytes
have been consumed from the socket, so giving up and returning is
lossless: the caller can simply try again on its own schedule. Once
i > 0 a partial header is already in hand, so that path is left to
retry unconditionally as before, since discarding it would
desynchronize the stream. The idle counter resets on any real read
progress.

After IDLE_POLL_MAX (4, roughly 20ms of polling) consecutive idle
polls at i == 0, the function now returns &ast_null_frame instead of
continuing to loop. Both existing callers of this function already
tolerate that return value: apps/app_audiosocket.c passes it straight
to ast_write(), and channels/chan_audiosocket.c returns it from its
.read channel tech callback. AST_FRAME_NULL is an existing, explicitly
handled no-op frame type in both ast_write() and __ast_read() in
main/channel.c, and &ast_null_frame is already the established idiom
elsewhere in the tree for exactly this "nothing to deliver this cycle"
case.

The payload read loop later in the same function is intentionally
left unchanged: it already treats a single poll timeout as fatal, so
it has no equivalent unbounded-retry condition to fix.

UserNote: An idle AudioSocket peer (one that stops sending data
mid-call without closing the connection) can no longer prevent the
channel from delivering outbound audio to the far end.

UpgradeNote: ast_audiosocket_receive_frame_with_hangup() may now
return &ast_null_frame after a short bounded idle period instead of
continuing to block. Out-of-tree callers of this public API should
handle an AST_FRAME_NULL return the same way apps/app_audiosocket.c
and channels/chan_audiosocket.c do, rather than assuming every
non-NULL return carries audio data.

Fixes: #2089

### DIFF
--- a/res/res_audiosocket.c
+++ b/res/res_audiosocket.c
@@ -44,6 +44,9 @@
 
 #define MAX_CONNECT_TIMEOUT_MSEC 2000
 
+/*! \brief Maximum consecutive idle polls tolerated before a header read yields. */
+#define IDLE_POLL_MAX 4
+
 /*!
  * \internal
  * \brief Attempt to complete the audiosocket connection.
@@ -288,6 +291,7 @@ struct ast_frame *ast_audiosocket_receive_frame_with_hangup(const int svc,
 	int *const hangup)
 {
 	int i = 0, n = 0, ret = 0;
+	unsigned int idle_polls = 0;
 	struct ast_frame f = {
 		.frametype = AST_FRAME_VOICE,
 		.src = "AudioSocket",
@@ -312,6 +316,9 @@ struct ast_frame *ast_audiosocket_receive_frame_with_hangup(const int svc,
 					continue;
 				} else if (poll_result == 0) {
 					ast_debug(1, "Poll timed out while waiting for header data\n");
+					if (i == 0 && ++idle_polls > IDLE_POLL_MAX) {
+						return &ast_null_frame;
+					}
 					continue;
 				} else {
 					ast_log(LOG_WARNING, "Poll error: %s\n", strerror(errno));
@@ -324,6 +331,7 @@ struct ast_frame *ast_audiosocket_receive_frame_with_hangup(const int svc,
 		if (n == 0) {
 			break;
 		}
+		idle_polls = 0;
 		i += n;
 	}
 


daniel-sullivan:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

daniel-sullivan:
**AI assistance disclosure.** An AI coding agent was used to diagnose this and draft the patch — tracing the unbounded-retry condition, confirming it against Asterisk's single-thread-per-channel service model, and identifying that the zero-bytes-consumed position is the only safe place to yield without desynchronising the header stream. The reproduction evidence in the linked issue is from a real production deployment, not synthetic. I have reviewed and own every line of this change and am happy to walk through any part of it.

jcolp:
Did it also write the commit message? It is very... very... very... verbose.

daniel-sullivan:
> Did it also write the commit message? It is very... very... very... verbose.

Good afternoon! Apologies for the horribly thrown up fix yesterday. We were just about to start live testing a deployment when this bug bit us. I asked the bloody agent to follow the rules and submit it before I got pulled onto the next crisis and forgot about sharing it but as you can see, it took some liberties...

I've got some time now so more than happy to tear this PR/Issue down and redo them by hand if you prefer?

jcolp:
They can be updated. The commit message should be shortened a ton and doesn't need to go into implementation detail. Detail about why can be a short comment with the code itself.

I'd also appreciate in the future not unleashing an LLM on our issue and PR process, since in the case of issues it can/will bypass our templates and process causing more work - and in the case of the PR can be verbose (something we explicitly call out in our policy).

gtjoseph:
Resolves: #2078

jcolp (changes requested):
Please see my prior comment about commit message, and additional UpgradeNote is not used for developer notifications, DeveloperNote is[1].

[1] https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/#special-trailers-for-commit-messages

inferno-cc:
I can confirm this PR fixes the issue on Asterisk 22 mainstream as well.

PR #2094, opened 2026-08-18
main/file.c: Log the reason a file could not be opened.

### LINKED ISSUE #1629: [bug]: `ast_streamfile` logs a misleading error based on errno
### Severity

Trivial

### Versions

20.17.0

### Components/Modules

file.c

### Operating Environment

Debian 13

### Frequency of Occurrence

None

### Issue Description

While investigating #1623 I came to the conclusion that the following line issues misleading error messages based on `errno`:

https://github.com/asterisk/asterisk/blob/a6539cb9af3094790f3f08b92667d2cfae7c4b6f/main/file.c#L1342

For instance:

```shell
[2025-12-03 12:28:23] WARNING[2836930][C-00054564] file.c: Unable to open beep (format (alaw)): No such file or directory```
```

There are several problems here:

- The value of `errno` needs to be initialised to 0 before making any system calls, otherwise the value is not defined : it's whatever is left in that variable.
- The operations in `filehelper` involve iterating over **numerous** files, so using `errno` outside of the iteration is ambiguous : you have no idea which format / extension the message refers to.
- The problem is compounded by #1625, since even after successfully opening a file, we continue iterating over file formats. As a result, you are almost guaranteed that `stat` is going to hit a `ENOENT` error.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

jlaine:
When ast_streamfile() failed to open a file it reported strerror(errno), but errno was not set by anything on that path.  The value came from whatever syscall ran last, so the message routinely blamed an unrelated error.

Drop errno from that message and log at the point of failure instead: filehelper() now reports the fopen() error, which does carry a meaningful errno, and an allocation failure from get_filestream().  Candidate files whose format does not match the channel are logged at debug level only, since filehelper() walks every registered format and skipping the ones that do not match is normal: a sound installed in several formats would otherwise warn on every successful playback.

Failures inside the format module's open callback stay unlogged here, as those callbacks already report their own reason.

Resolves: #1629

### DIFF
--- a/main/file.c
+++ b/main/file.c
@@ -600,15 +600,18 @@ static int filehelper(const char *filename, const void *arg2, const char *fmt, c
 				if ((ast_format_cmp(ast_channel_writeformat(chan), f->format) == AST_FORMAT_CMP_NOT_EQUAL) &&
 				     !(((ast_format_get_type(f->format) == AST_MEDIA_TYPE_AUDIO) && fmt) ||
 					  ((ast_format_get_type(f->format) == AST_MEDIA_TYPE_VIDEO) && fmt))) {
+					ast_debug(3, "File %s format is not compatible with the channel\n", fn);
 					ast_free(fn);
 					continue;	/* not a supported format */
 				}
 				if ( (bfile = fopen(fn, "r")) == NULL) {
+					ast_log(LOG_WARNING, "Failed to open file %s due to: %s\n", fn, strerror(errno));
 					ast_free(fn);
 					continue;	/* cannot open file */
 				}
 				s = get_filestream(f, bfile);
 				if (!s) {
+					ast_log(LOG_WARNING, "Failed to open file %s due to: file stream creation failure\n", fn);
 					fclose(bfile);
 					ast_free(fn);	/* cannot allocate descriptor */
 					continue;
@@ -1344,8 +1347,8 @@ int ast_streamfile(struct ast_channel *chan, const char *filename,
 		if (!fs) {
 			struct ast_str *codec_buf = ast_str_alloca(AST_FORMAT_CAP_NAMES_LEN);
 			ast_channel_lock(chan);
-			ast_log(LOG_WARNING, "Unable to open %s (format %s): %s\n",
-					filename, ast_format_cap_get_names(ast_channel_nativeformats(chan), &codec_buf), strerror(errno));
+			ast_log(LOG_WARNING, "Unable to open %s (format %s)\n",
+					filename, ast_format_cap_get_names(ast_channel_nativeformats(chan), &codec_buf));
 			ast_channel_unlock(chan);
 			return -1;
 		}


jlaine:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

main/file.c
```diff
@@ -600,15 +600,18 @@ static int filehelper(const char *filename, const void *arg2, const char *fmt, c
...
 					continue;	/* not a supported format */
 				}
 				if ( (bfile = fopen(fn, "r")) == NULL) {
+					ast_log(LOG_WARNING, "File %s fopen failed: %s\n", fn, strerror(errno));
 					ast_free(fn);
 					continue;	/* cannot open file */
 				}
 				s = get_filestream(f, bfile);
 				if (!s) {
+					ast_log(LOG_WARNING, "File %s get_filestream failed\n", fn);
```
jcolp: This should be more user friendly and not reference a function name, as it can cause confusion.

main/file.c
```diff
@@ -600,15 +600,18 @@ static int filehelper(const char *filename, const void *arg2, const char *fmt, c
 				if ((ast_format_cmp(ast_channel_writeformat(chan), f->format) == AST_FORMAT_CMP_NOT_EQUAL) &&
 				     !(((ast_format_get_type(f->format) == AST_MEDIA_TYPE_AUDIO) && fmt) ||
 					  ((ast_format_get_type(f->format) == AST_MEDIA_TYPE_VIDEO) && fmt))) {
+					ast_debug(3, "File %s format is not compatible with the channel\n", fn);
 					ast_free(fn);
 					continue;	/* not a supported format */
 				}
 				if ( (bfile = fopen(fn, "r")) == NULL) {
+					ast_log(LOG_WARNING, "File %s fopen failed: %s\n", fn, strerror(errno));
```
jcolp: This should be more user friendly and not reference a function name, as it can cause confusion.

jcolp (changes requested):

jlaine:
@jcolp I'd appreciate guidance on the phrasing you'd be comfortable with but which still allows users to pinpoint exactly what is going on. I had a look at existing code and I see two patterns:

- The same style I used, referencing a function name for detailed debugging:
  https://github.com/asterisk/asterisk/blob/19f98688a302958c71614afe74d84023c7f64cc2/main/file.c#L657
  
- Chaining 3 function calls back-to-backed with no details about what failed:
  https://github.com/asterisk/asterisk/blob/19f98688a302958c71614afe74d84023c7f64cc2/main/file.c#L1435

To be clear: the warning messages I added should not occur in normal usage. If they do crop up, something is badly wrong: bad file permissions or a memory allocation failure.

EDIT: as an alternative, if you prefer, these messages could all be made `ast_debug` and retain their "technical" phrasing?

main/file.c
jlaine: Would `"File %s could not be opened: %s\n"` be better?

main/file.c
jlaine: Would `"File %s could not be streamed: %s\n"` be better?

jcolp:
The examples you reference are generally from the wild wild west era where anything went.

"ast_log(LOG_WARNING, "Failed to open file %s due to: file stream creation failure\n", fn);"

"ast_log(LOG_WARNING, "Failed to open file %s due to: %s\n", fn, strerror(errno));"

jlaine:
> The examples you reference are generally from the wild wild west era where anything went.
> 
> "ast_log(LOG_WARNING, "Failed to open file %s due to: file stream creation failure\n", fn);"
> 
> "ast_log(LOG_WARNING, "Failed to open file %s due to: %s\n", fn, strerror(errno));"

Understood. Would you like me to align the other messages in `filehelper` on this template?

jcolp:
No. If something comes up in the future that somehow impacts it then it can be changed.

jcolp (approved):

PR #2096, opened 2026-08-18
res_rtp_asterisk: Change MES calculation to use long-term jitter mean

### LINKED ISSUE #2095: [bug]: MES calculation skips the long-term jitter mean, using only the short-term RFC 3550 EWMA sample
### Severity

Trivial

### Versions

master, 20,22,24,24

### Components/Modules

res_rtp_asterisk

### Operating Environment

all

### Frequency of Occurrence

None

### Issue Description

calc_media_experience_score() expects a smoothed jitter value (its own doxygen comment for the normdev_rxjitter parameter says “The smoothed jitter”). Both call sites, however, pass a value that has only gone through the RFC 3550 per-packet exponential smoothing (Appendix A.8), not through Asterisk’s own longer-window running mean — so a single anomalous RTP/RTCP sample can still dominate the score.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

mbradeen:
The MES score used the reported jitter value instead of the smoothed statistic
value, which could cause a single high anomalous RTCP sample to unduly skew
the overall score.

Fixes: #2095

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -6505,7 +6505,7 @@ static void update_reported_mes_stats(struct ast_rtp *rtp)
 {
 	double mes = calc_media_experience_score(rtp->owner,
 		rtp->rtcp->normdevrtt,
-		rtp->rtcp->reported_jitter,
+		rtp->rtcp->reported_normdev_jitter,
 		rtp->rtcp->reported_stdev_jitter,
 		rtp->rtcp->reported_normdev_lost);
 
@@ -6526,7 +6526,7 @@ static void update_reported_mes_stats(struct ast_rtp *rtp)
 	ast_debug_rtcp(2, "%s: rtt: %.9f j: %.9f sjh: %.9f lost: %.9f mes: %4.1f\n",
 		ast_rtp_instance_get_channel_id(rtp->owner),
 		rtp->rtcp->normdevrtt,
-				rtp->rtcp->reported_jitter,
+				rtp->rtcp->reported_normdev_jitter,
 				rtp->rtcp->reported_stdev_jitter,
 				rtp->rtcp->reported_normdev_lost, mes);
 }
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (changes requested):
The commit message has misspelt "calculation"

jcolp (approved):

res/res_rtp_asterisk.c
```diff
@@ -7175,14 +7208,16 @@ static struct ast_frame *ast_rtcp_interpret(struct ast_rtp_instance *instance, s
...
+				valid_jitter_value = update_jitter_stats(rtp, report_block->ia_jitter);
 				update_lost_stats(rtp, report_block->lost_count.packets);
 				/*
 				 * update_reported_mes_stats must be called AFTER
 				 * update_rtt_stats, update_jitter_stats and
 				 * update_lost_stats.
 				 */
-				update_reported_mes_stats(rtp);
+				if (valid_jitter_value) {
+					update_reported_mes_stats(rtp);
```
jcolp: Should this still run regardless so that the MES stats are updated using any other things that updated, even if jitter didn't?

jcolp (changes requested):

res/res_rtp_asterisk.c
mbradeen: We would need to set the `reported_jitter` to _something_ in that case, possibly the `reported_normdev_jitter`?

jcolp: So I think there's two possible things:

1. If no prior valid jitter has been received, I think we can only assume there is no jitter at that point
2. If prior valid jitter was received then that would be used, even if this one isn't considered valid

jcolp (approved):

PR #2098, opened 2026-08-18
res_rtp_asterisk: Correct effective latency calculation

### LINKED ISSUE #2097: [bug]: effective_latency uses full RTT instead of one-way delay and mixes seconds with milliseconds
### Severity

Trivial

### Versions

master, 20,22,24,24

### Components/Modules

res_rtp_asterisk

### Operating Environment

all

### Frequency of Occurrence

Constant

### Issue Description

The effective latency calculation is using the full round trip time instead of the one-way, which causes the MES score to be worse than it would otherwise.  Additionally, the rx jitter calculation was not scaling the normdev and stdev values to milliseconds, essentially disregarding their values in the calculation

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

mbradeen:
The effective latency calculation was using the full round trip time instead of
the one-way, which causes the MES score to be worse than it would otherwise.

Additionally, the rx jitter calculation was not scaling the normdev and stdev
values to milliseconds, essentially disregarding their values in the calculation

Fixes: #2097

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -6452,9 +6452,16 @@ static double calc_media_experience_score(struct ast_rtp_instance *instance,
 	 * jitter scaled according to its standard deviation. The scaling is done in order
 	 * to increase jitter's weight since a higher deviation can result in poorer overall
 	 * quality.
+	 *
+	 * normdevrtt is the mean round trip time in seconds. The G.107's delay-impairment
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

PR #2102, opened 2026-08-18
res_rtp_asterisk: Fix reported packet loss to use current report's difference

### LINKED ISSUE #2101: [bug]: Reported packet loss feeds a cumulative counter into the MES mean, drifting the score toward its floor over a call’s duration
### Severity

Trivial

### Versions

master,20,22,23,24

### Components/Modules

res_rtp_asterisk

### Operating Environment

all

### Frequency of Occurrence

None

### Issue Description

update_lost_stats() calculates the lost packet statistics based on the cumulative lost packets statistics instead of the lost packets for the current report.  This leads to the MES score skewing over time.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

mbradeen:
The lost packet calculation was using the cumulative number of lost packets instead
of the current interval's lost packets.  This lead to the MES caculation skewing
over time.  We now keep track of the last reported value so we can determine the
current value vs the cumulative value.

Fixes: #2101

### DIFF
--- a/res/res_rtp_asterisk.c
+++ b/res/res_rtp_asterisk.c
@@ -559,6 +559,7 @@ struct ast_rtcp {
 	double rtt;			/*!< Last reported rtt */
 	double reported_jitter;	/*!< The contents of their last jitter entry in the RR in seconds */
 	unsigned int reported_lost;	/*!< Reported lost packets in their RR */
+	unsigned int last_reported_lost; /*!< Reported cumulative lost packets in the previous RR */
 
 	double reported_maxjitter; /*!< Maximum reported interarrival jitter */
 	double reported_minjitter; /*!< Minimum reported interarrival jitter */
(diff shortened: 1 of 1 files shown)


mbradeen:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

res/res_rtp_asterisk.c
```diff
@@ -6397,7 +6398,20 @@ static void update_lost_stats(struct ast_rtp *rtp, unsigned int lost_packets)
...
 	rtp->rtcp->reported_lost = lost_packets;
-	reported_lost = (double)rtp->rtcp->reported_lost;
+
+	/*
+	 * lost_packets contains the cumulative number of lost packets as reported in
+	 * the peer's RTCP RR/SR report block (RFC 3550). Calculate the number of lost
+	 * packets in the current interval based on the difference from the previous
+	 * count.
+	 */
+	reported_lost = (double)(lost_packets - rtp->rtcp->last_reported_lost);
```
jcolp: lost_packets and rtp->rtcp->last_reported_lost are both unsigned, so reported_lost can never be negative - instead it would just end up being a large value

You'd need to cast them individually to double and do the subtraction, instead of casting the result to double

jcolp (changes requested):

jcolp (approved):

PR #2107, opened 2026-08-20
res_cdrel_custom: Use extendable ast_str for JSON records.

### LINKED ISSUE #2106: [bug]:  res_cdrel_custom JSON output is truncated at 1024 bytes
### Severity

Major

### Versions

22.10.1

### Components/Modules

res_cdrel_custom

### Operating Environment

Ubuntu 24.04.4 LTS"

### Frequency of Occurrence

Constant

### Issue Description

Using Advanced Mapping with JSON:

```ini
[Master.json]
format = json
fields = ...
```

JSON records longer than 1024 bytes are silently truncated at exactly 1024 bytes.

Example:

```text
1024 bytes
JSON: invalid
```

The record is cut in the middle of a field, so the resulting JSON is invalid.

Records shorter than 1024 bytes are written correctly.

The same CDR with the same fields using:

```ini
format = dsv
```

is written completely and is not truncated.

### Expected behavior

JSON output should be written completely regardless of record size, or the module should return an explicit error instead of silently truncating the record.

### Actual behavior

Records larger than 1024 bytes are silently truncated at exactly 1024 bytes, producing invalid JSON.

**Steps to reproduce:**

1. Configure an Advanced Mapping with `format = json`.
2. Include enough CDR/uservar fields to produce a record larger than 1024 bytes.
3. Make a call.
4. Check the generated JSON file.

The generated record is exactly 1024 bytes and is invalid JSON.

### Asterisk Issue Guidelines

gtjoseph:
Since JSON records include the field names, the actual record size can be quite
lengthy and since the thread-local backed ast_str used by the DSV writer can't be
extended, we need to use a regular, extendable ast_str.

Resolves: #2106

### DIFF
--- a/res/cdrel_custom/writers.c
+++ b/res/cdrel_custom/writers.c
@@ -136,7 +136,16 @@ static int json_writer(struct cdrel_config *config, struct cdrel_values *values)
 {
 	int ix = 0;
 	int res = 0;
-	struct ast_str *str = ast_str_thread_get(&custom_buf, 1024);
+	/*
+	 * Since JSON records include the field names, the actual record size can be quite
+	 * lengthy and since the thread-local backed ast_str used by the DSV writer can't be
+	 * extended, we need to use a regular, extendable ast_str.
+	 */
+	RAII_VAR(struct ast_str *, str, ast_str_create(1024), ast_free);
+
+	if (!str) {
+		return -1;
+	}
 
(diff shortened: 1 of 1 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

jcolp (approved):

PR #2109, opened 2026-08-20
frame: Correct ast_frame_adjust_volume* documentation of 'adjustment'

### LINKED ISSUE #1167: frame.h: ast_frame_adjust_volume* Incorrect documention related to 'adjustment' parameter
https://github.com/asterisk/asterisk/blob/606d5d054afc7c79268a54d27256a443bf6cba8d/include/asterisk/frame.h#L613-L627

The documentation claims `adjustment` is in dB.  This value is used to multiply the audio frames which is NOT dB.
Generally this value is calculated from dB with `adjustment = 10^(dB/20)` or the inverse `20*log10(adjustment) = dB`
The adjustment to dB are as follows:

| Adjustment | dB   |
|------------|------|
| 10         | 20   |
| 9          | 19.1 |
| 8          | 18.1 |
| 7          | 16.9 |
| 6          | 15.6 |
| 5          | 14.0 |
| 4          | 12.0 |
| 3          | 9.5  |
| 2          | 6.0  |
| 1          | 0    |

jacky-w-j-li:
The documentation for ast_frame_adjust_volume() and
ast_frame_adjust_volume_float() described the 'adjustment' parameter as a
dB value. In reality the parameter is a linear gain factor applied
directly to the audio samples: a positive value multiplies each sample,
a negative value divides each sample, and 0 leaves the audio unchanged.

The documentation block has also been reformatted to use the standard
' *' Doxygen style (asterisk in the second column) for consistency with
the rest of the file.

To convert a dB change to this factor use adjustment = 10^(dB/20)
(for example, 2 ~= +6 dB and 10 ~= +20 dB).

Resolves: #1167

### DIFF
--- a/include/asterisk/frame.h
+++ b/include/asterisk/frame.h
@@ -612,18 +612,28 @@ void ast_frame_dump(const char *name, struct ast_frame *f, char *prefix);
 struct ast_frame *ast_frame_enqueue(struct ast_frame *head, struct ast_frame *f, int maxlen, int dupe);
 
 /*!
-  \brief Adjusts the volume of the audio samples contained in a frame.
-  \param f The frame containing the samples (must be AST_FRAME_VOICE and AST_FORMAT_SLINEAR)
-  \param adjustment The number of dB to adjust up or down.
-  \return 0 for success, non-zero for an error
+ * \brief Adjusts the volume of the audio samples contained in a frame.
+ * \param f The frame containing the samples (must be AST_FRAME_VOICE and AST_FORMAT_SLINEAR)
+ * \param adjustment The linear gain factor applied to each sample.
+ *
+ * \note This is NOT a dB value.  A positive value multiplies each sample by the factor
+ * (e.g. 2 doubles the volume), a negative value divides each sample by the magnitude
+ * (e.g. -2 halves it), and 0 leaves the audio unchanged. To convert a dB change
+ *  to this factor use adjustment = 10^(dB/20) (so 10 ~= +20 dB and 2 ~= +6 dB).
+ * \return 0 for success, non-zero for an error
  */
 int ast_frame_adjust_volume(struct ast_frame *f, int adjustment);
 
 /*!
-  \brief Adjusts the volume of the audio samples contained in a frame.
-  \param f The frame containing the samples (must be AST_FRAME_VOICE and AST_FORMAT_SLINEAR)
-  \param adjustment The number of dB to adjust up or down.
-  \return 0 for success, non-zero for an error
+ * \brief Adjusts the volume of the audio samples contained in a frame.
+ * \param f The frame containing the samples (must be AST_FRAME_VOICE and AST_FORMAT_SLINEAR)
+ * \param adjustment The linear gain factor applied to each sample.
+ *
+ * \note This is NOT a dB value.  A positive value multiplies each sample by the factor
+ * (e.g. 2 doubles the volume), a negative value divides each sample by the magnitude
+ * (e.g. -2 halves it), and 0 leaves the audio unchanged. To convert a dB change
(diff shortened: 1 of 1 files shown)


jacky-w-j-li:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jacky-w-j-li:
Note: I used an LLM to help draft parts of this change and its description; the correction itself was verified against the actual implementation in main/frame.c (ast_frame_adjust_volume multiplies/divides samples by the adjustment factor, so it is a linear gain, not dB). Happy to adjust the wording if needed.

jcolp (approved):

include/asterisk/frame.h
```diff
@@ -614,15 +614,15 @@ struct ast_frame *ast_frame_enqueue(struct ast_frame *head, struct ast_frame *f,
 /*!
   \brief Adjusts the volume of the audio samples contained in a frame.
   \param f The frame containing the samples (must be AST_FRAME_VOICE and AST_FORMAT_SLINEAR)
-  \param adjustment The number of dB to adjust up or down.
+  \param adjustment The linear gain factor applied to each sample. This is NOT a dB value. A positive value multiplies each sample by the factor (e.g. 2 doubles the volume), a negative value divides each sample by the magnitude (e.g. -2 halves it), and 0 leaves the audio unchanged. To convert a dB change to this factor use adjustment = 10^(dB/20) (so 10 ~= +20 dB and 2 ~= +6 dB).
```
gtjoseph: These comments end at column 382 which is just a bit long :)

How about...

```
 * \param adjustment The linear gain factor applied to each sample.
 *
 * \note This is NOT a dB value.  A positive value multiplies each sample by the factor
 * (e.g. 2 doubles the volume), a negative value divides each sample by the magnitude
 * (e.g. -2 halves it), and 0 leaves the audio unchanged. To convert a dB change
 *  to this factor use adjustment = 10^(dB/20) (so 10 ~= +20 dB and 2 ~= +6 dB).
```

Also, Would you mind fixing the rest of that documentation block so it has '*' in the second column?  It's actually been that way for 20 years but it would help if you could fix it while you're there.  Thanks!

include/asterisk/frame.h
```diff
@@ -614,15 +614,15 @@ struct ast_frame *ast_frame_enqueue(struct ast_frame *head, struct ast_frame *f,
...
   \return 0 for success, non-zero for an error
  */
 int ast_frame_adjust_volume(struct ast_frame *f, int adjustment);
 
 /*!
   \brief Adjusts the volume of the audio samples contained in a frame.
   \param f The frame containing the samples (must be AST_FRAME_VOICE and AST_FORMAT_SLINEAR)
-  \param adjustment The number of dB to adjust up or down.
+  \param adjustment The linear gain factor applied to each sample. This is NOT a dB value. A positive value multiplies each sample by the factor (e.g. 2 doubles the volume), a negative value divides each sample by the magnitude (e.g. -2 halves it), and 0 leaves the audio unchanged. To convert a dB change to this factor use adjustment = 10^(dB/20) (so 10 ~= +20 dB and 2 ~= +6 dB).
```
gtjoseph: Same as above.

gtjoseph (changes requested):

jacky-w-j-li:
Thanks @gtjoseph — both addressed: shortened the `\param adjustment` line (moved the explanation into a `\note`) and reformatted the whole block to the ` *` style. PTAL.

gtjoseph (dismissed):
Thanks for fixing the docs but now you have two commits.  You need to squash them down to one commit and do a `git push --force`.

jacky-w-j-li:
Squashed down to a single commit and force-pushed. PTAL.

PR #2111, opened 2026-08-21
res_prometheus: add toggle config for channels metrics

whitetreebug:
In high‑volume call scenarios, Prometheus exposes a large number of channels‑related metrics.

Not all users require channels_state and channels_duration.

Add configuration options to enable or disable these two metrics.

UserNote: "channels_detail_metrics_enabled" defaults to "yes" for backward‑compatibility.
Set it to "no" if you do not need channels detail metrics (channels_state and channels_duration).

### DIFF
--- a/res/prometheus/channels.c
+++ b/res/prometheus/channels.c
@@ -26,9 +26,11 @@
  */
 
 #include "asterisk.h"
+#include "asterisk/astobj2.h"
 #include "asterisk/res_prometheus.h"
 #include "asterisk/stasis_channels.h"
 #include "asterisk/pbx.h"
+#include "asterisk/utils.h"
 #include "prometheus_internal.h"
 
 #define CHANNELS_STATE_HELP "Individual channel states. 0=down; 1=reserved; 2=offhook; 3=dialing; 4=ring; 5=ringing; 6=up; 7=busy; 8=dialing_offhook; 9=prering."
@@ -129,6 +131,8 @@ static struct prometheus_metric global_channel_metrics[] = {
  */
 static void channels_scrape_cb(struct ast_str **response)
 {
+	RAII_VAR(struct prometheus_general_config *, config, prometheus_general_config_get(), ao2_cleanup);
+
 	struct ao2_container *channel_cache;
 	struct ao2_container *channels;
 	struct ao2_iterator it_chans;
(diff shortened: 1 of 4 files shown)


whitetreebug:
Below is the comparison before and after disabling the metrics.

image

whitetreebug:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

gtjoseph:
@whitetreebug Is there a reason this can't be cherry-picked to the 20 branch?

whitetreebug:
> @whitetreebug Is there a reason this can't be cherry-picked to the 20 branch?
I have added the cherry‑pick
Sorry, I’m not familiar with our branching strategy. Is there any guideline that tells me which branch I should cherry‑pick to?

res/prometheus/channels.c
```diff
@@ -176,6 +180,12 @@ static void channels_scrape_cb(struct ast_str **response)
 		return;
 	}
 
+	if (!config || config->channels_detail_metrics_enabled == 0) {
```
jcolp, line 183: We use !config->channels_detail_metrics_enabled

res/res_prometheus.c
```diff
@@ -991,6 +991,7 @@ static int load_module(void)
 	}
 	aco_option_register(&cfg_info, "enabled", ACO_EXACT, global_options, "no", OPT_BOOL_T, 1, FLDSET(struct prometheus_general_config, enabled));
 	aco_option_register(&cfg_info, "core_metrics_enabled", ACO_EXACT, global_options, "yes", OPT_BOOL_T, 1, FLDSET(struct prometheus_general_config, core_metrics_enabled));
+	aco_option_register(&cfg_info, "channels_detail_metrics_enabled", ACO_EXACT, global_options, "yes", OPT_BOOL_T, 1, FLDSET(struct prometheus_general_config, channels_detail_metrics_enabled));
```
jcolp, line 994: This needs to be documented at the top of res_prometheus.c in the XML including version numbers

jcolp (changes requested):
The UserNote should be more explicit. For example:

"Option X has been added which does Y. It defaults to yes to preserve existing behavior." Without further clarification a user has to also look through commit messages.

PR #2118, opened 2026-08-24
ARI: Add support for sending and receiving in-dialog SIP MESSAGE.

### LINKED ISSUE #1810: [new-feature]:  Channel-Scoped In-Dialog SIP MESSAGE via ARI
# Updated Feature Request: Channel-Scoped In-Dialog SIP MESSAGE via ARI (PJSIP)

  ## Summary

  This feature request is for **channel-scoped, in-dialog SIP `MESSAGE` support during an ongoing call** on **SIP/PJSIP channels**.

  The request started as a SIP INFO proposal, but but lead to the more practical API direction:

  - `POST /channels/{channelId}/sendMessage`
  - async delivery result event: `ChannelMessageDeliveryStatus`

  This keeps the operation tied to a specific live call leg while reusing existing Asterisk/PJSIP MESSAGE plumbing.

## Motivation

We need a way to exchange application payloads that are strictly tied to an active call context.
For this use case:

  - the payload must be associated with an existing live dialog
  - the selected `channelId` must determine the exact call leg used
  - if there is no active in-dialog PJSIP session, the request must be rejected
  - the Stasis application must receive final delivery status for the outbound request using events
  - this is **not** intended for random/out-of-dialog endpoint messaging
  - It is **not** intended to replace existing endpoints `MESSAGE` support.

## Use Cases

1. Mid-call assistant/control updates or switching features on/off that must stay attached to the active call.
2. In-call metadata/control signals valid only while the dialog exists.
4. Instead of (currently) overloading Q.850 codes, send explicit in-call messages for call classification/routing decisions:
  - Empty call
  - Call request
  - Emergency call request

## Why SIP INFO (and not MESSAGE) in this case

We recognize `MESSAGE` already exists in Asterisk core and PJSIP (/endpoints/.../sendMessage).
The key requirement is not just "send a MESSAGE to an endpoint", but:

1. Payload must be bound to an existing call channel/dialog, not just an endpoint identity.
2. Signaling must occur as in-dialog SIP requests.
3. Result handling must map to that active dialog/channel context.
4. If there is no active PJSIP in-dialog session, the request must be rejected.
5. The application needs request correlation and a **final delivery result** (result event).

Reusing MESSAGE - SIP transaction handling details already covered by PJSIP
MESSAGE internals is lower-risk than introducing a separate SIP INFO API family.

### Endpoint

  `POST /channels/{channelId}/sendMessage`

  ### Parameters

  - messages: array (optional; JSON payload for structured message content)
  - body: string (optional; alternative to `messages` when sending one payload only)
  - `content_type`: string (optional; defaults to `application/json` when `messages` is used)
  - `request_id`: string (required)

  At least one of `messages` or `body` must be provided.

  ### Error codes

  - `400` invalid or missing request parameters
  - `404` channel not found
  - `409` channel not in Stasis
  - `412` channel invalid state
  - `501` supported only for PJSIP channels

  ### Behavior

  This operation is **asynchronous**.

  - the REST call returns `202 Accepted` if Asterisk accepted the request for processing
  - Asterisk later emits a final delivery status event for that request
  - `request_id` is used to correlate the REST request with the final event

  ## Proposed Stasis Event

  `ChannelMessageDeliveryStatus`

  ### Purpose

  Emit the final delivery result for an outbound in-dialog channel-scoped SIP `MESSAGE`.
  Emmitted only if HTTP response was 202 before.

  ### Fields

  - `channel`: Channel
  - `request_id`: string
  - `status`: string
    - `sent`
    - `failed`
    - `transport_error`
  - `to`: string (optional)
  - `from`: string (optional)
  - `body`: string (optional, if caller sends messages[], the body contains the serialized payload)
  - `content_type`: string (optional)
  - `sip_status_code`: int (optional)
  - `reason`: string (optional)

  ## Historical context

  This issue originally started from a SIP INFO-based idea for mid-call signaling.

The direction shifted to a channel-scoped in-dialog SIP MESSAGE API instead. 
The current proposal is the therefore channel-scoped MESSAGE approach

## Open design points 

 The following details are still open and are part of this feature discussion:
(Perhaps there are preferred constraints and behavior)

- Max outbound message/body size  /  Max total payload
- Behavior when limits exceeded (which error)
- Encoding (e.g. UTF-8)
- Max timeout (default, allowed range) -> probably the default PJSIP transaction timeout?

owoschhodan:
Applications need a way to exchange payloads that are strictly tied to an
active call context: the message must be associated with an existing live
dialog, and the channel the ARI application selects must determine the exact
call leg used. The existing endpoint-scoped MESSAGE (PUT
/endpoints/sendMessage) targets an endpoint identity instead, which is
ambiguous when that endpoint has multiple registered devices, is involved in
multiple simultaneous calls, or the active dialog changes due to a transfer.
This feature does not replace endpoint-scoped MESSAGE support.

Add a new POST /channels/{channelId}/sendMessage endpoint that sends a SIP
MESSAGE on the channel's existing dialog, guaranteeing it reaches the exact
call leg the ARI application is tracking. Delivery result is reported
asynchronously via a new ChannelMessageDeliveryStatus event, correlated by
the caller-supplied request_id. Incoming in-dialog MESSAGE requests are
likewise published to the owning channel's Stasis application as a new
ChannelMessageReceived event.

Resolves: #1810

UserNote: ARI applications can now send a SIP MESSAGE on an existing
channel's dialog via POST /channels/{channelId}/sendMessage, with delivery
result reported via the new ChannelMessageDeliveryStatus event. Incoming
in-dialog SIP MESSAGE requests on a channel under Stasis control are now
published as a new ChannelMessageReceived event.

DeveloperNote: New ARI capability: POST /channels/{channelId}/sendMessage,
and two new Stasis/ARI events, ChannelMessageReceived (incoming in-dialog
MESSAGE) and ChannelMessageDeliveryStatus (async delivery result for an
outgoing in-dialog MESSAGE, correlated by request_id).


owoschhodan:
Implementation note on delivery correlation: in-dialog requests on a dialog
share one incrementing CSeq counter across all request types, and an
auth-challenged retry (401/407) gets a new CSeq and branch for what is
logically the same request. Keying delivery tracking by CSeq would mean
re-keying mid-flight on every auth retry, and risks collision with
concurrent/duplicate requests on the same dialog (see the
overlap_pending_request and duplicate_request_id testsuite cases).

Instead, each outgoing MESSAGE's delivery-tracking object is bound directly
via a pointer in PJSIP's per message mod_data slot on the exact
pjsip_tx_data for that attempt. 
On response, pjsip_rdata_get_tsx(rdata)->last_tx yields the literal tdata 
that produced it, giving the exact delivery via one dereference without key or lookup. 
For the auth-retry path, the delivery is briefly parked in the dialog's
mod_data slot so the retry's freshly-created tdata can be rebound to the
same delivery object when PJSIP builds it.

owoschhodan:
Testing performed:

- All 20 new asterisk-testsuite tests added for this feature pass
  (send_message/, recv_in_dialog_message, message_sendtext_no_messaging_module).
- Regression tests executed against the full message/ and rest_api/channels/ test
  directories plus in_dialog_invite_replaces, in_dialog_options, and
  session_timers.
- Additionally deployed on our company server(s) and passed our internal
  automatic feature and regression test suite.

owoschhodan:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

owoschhodan:
testsuite-test-pr: 152

include/asterisk/res_pjsip_session.h
```diff
@@ -737,6 +738,77 @@ int ast_sip_session_add_datastore(struct ast_sip_session *session, struct ast_da
  */
 struct ast_datastore *ast_sip_session_get_datastore(struct ast_sip_session *session, const char *name);
 
+/*!
+ * \brief Optional delivery tracker for tracking in-dialog MESSAGEs.
+ */
+struct ast_sip_session_message_tracker {
```
gtjoseph, line 744: I haven't looked at this in detail but why not just use a session datastore instead of creating the message tracking stuff?

gtjoseph (dismissed):

include/asterisk/res_pjsip_session.h
owoschhodan: This is actually a good question and we tried that approach before and failed because:
It lead to a dependency issue in res_pjsip_session.c when messaging was NOT used.
By breaking the build itself (menuselect) or worse: during runtime when modules.conf did not load res_pjsip_messaging. 

The message tracker is more an API hook (registry) with call-back less a plain data storage.
Using the current implementation, without res_pjsip_messaging, the message_tracker is just NULL leading to no-ops calls, nothing breaks.


btw: We got "inspiration" from ast_sip_session_supplement where similar activity is done also without the datastore.

I hope the brief answer suffices

gtjoseph: It does, for now. :)

res/res_pjsip_messaging.c
```diff
@@ -139,6 +139,16 @@ const pjsip_method pjsip_message_method = {PJSIP_OTHER_METHOD, {"MESSAGE", 7} };
 #define MAX_USER_SIZE 128
 
 static struct ast_taskprocessor *message_serializer;
+static struct ao2_container *in_dialog_message_deliveries;
+
+static pj_bool_t module_on_rx_request(pjsip_rx_data *rdata);
```
jcolp, line 144: Before I start to dive in and review this I need an explanation on why exactly this approach was taken, whether other approaches were attempted, issues with them/pitfalls. On initial reading it feels like code which certainly may work for your usage, but without going deeper in it feels like the wrong approach and I need to be convinced that this is the "right" approach and something the project should support. The handling of things is also complex enough that I'd be worried about off-nominals.

jcolp (changes requested):

PR #2120, opened 2026-08-25
chan_websocket:  Add "unbuffered" mode and fix a few bugs.

### LINKED ISSUE #2115: [bug]: chan_websocket: frame queue never drains, turning any backlog into permanent latency
### Severity

Minor

### Versions

22.10.1

### Components/Modules

chan_websocket

### Operating Environment

Debian GNU/Linux 13 (trixie), kernel 6.12.73+deb13-amd64, x86_64

### Frequency of Occurrence

Constant

### Description

`chan_websocket` delivers exactly one queued frame per timer tick and has no way to recover a slot it missed. As a result, any event that puts frames into `frame_queue` faster than they are drained produces latency that stays for the rest of the call: the queue never returns to its previous depth.

In `webchan_read()`:

```c
if (ast_timer_get_event(instance->timer) == AST_TIMING_EVENT_EXPIRED) {
        ast_timer_ack(instance->timer, 1);
}

native_frame = dequeue_frame(instance);
```

`ast_timer_ack()` reads the timerfd, which clears *every* expiration that accumulated, but `dequeue_frame()` removes a single frame. A tick the thread woke up late for is therefore a media slot that can never be recovered.

The only backpressure is `QUEUE_LENGTH_XOFF_LEVEL` (900 frames), which at the usual 20ms ptime is 18 seconds of buffered audio. XOFF freezes the backlog but does not reduce it.

### Environment

16 cores, 15 GB RAM, KVM guest. ExternalMedia over WebSocket (`transport=websocket`, ulaw, 20ms ptime) bridged to PJSIP.

### Steps to reproduce

Using an ExternalMedia WebSocket channel bridged to a normal call:

1. Send media to Asterisk at twice the nominal rate (100 fps for 20ms G.711) for a few seconds. Note: a sender that paces its own output to 50 fps will not reproduce this — the burst has to actually reach the channel.
2. Return to exactly the nominal rate (50 fps) and keep sending.
3. Poll `GET_STATUS` on the WebSocket and watch `queue_length`.

### Actual result

The queue grows during the burst and then stays there. Measured on an otherwise idle machine:

```
burst phase (6s at 100 fps):  queue 0 -> 263 frames  (5.26s of audio) steady phase (60s at 50 fps): queue 290 -> 281 frames
```

Nine frames drained in sixty seconds. At that rate recovering the backlog would take about 31 minutes, and the residual drain only happens because the test sender ran marginally below 50 fps; with an exactly nominal sender the queue stays where it is indefinitely.

### Expected result

The backlog should drain back down so the added latency is transient.

### Notes

This was originally observed in production as audio delay that grew during a call in the direction of the WebSocket channel and never recovered, reaching several seconds. The reverse direction was unaffected, since writes to the WebSocket are not gated by a timer-driven queue.

This may be the same underlying problem as this community report, which describes matching symptoms at scale (XOFF/XON cycling with 800+ frames buffered) and was never root-caused:
https://community.asterisk.org/t/audio-distortion-lag-with-chan-websocket-externalmedia-at-90-concurrent-calls-asterisk-22-8-2/112483

Possibly related: #1712 reports MEDIA_START delays that show up more often above 60% CPU.

#1304 fixed a different accounting problem with `frame_queue_length` (not reset after `FLUSH_MEDIA`); this report is about the queue depth itself never recovering, which is independent of that fix.

### Suggested fix
Deliver extra frames while the backlog is above a watermark, so the queue works its way back down instead of staying where it is. Delivering at most one extra frame per tick keeps the catch-up bounded so the far end's jitter buffer can absorb it. In our testing this took the reproduction above from a peak of 263 frames to 5-6, draining back to 0, and production queues now sit flat.

(Patch removed - per project guidelines patches belong in a pull request, not on an issue. The change is two constants plus a short block in `webchan_read()`; happy to take it through the PR process if wanted.)

### Relevant log output

```shell
chan_websocket queue_length sampled via GET_STATUS on the media WebSocket.
Single call, ExternalMedia/WebSocket bridged to PJSIP, ulaw, 20ms ptime.
Idle machine (load ~1, "timing test" reporting a clean 50/50 ticks).

=== UNPATCHED (22.10.1 stock) ===

Burst phase - sender at 100 fps for 6s:
  t=1s   queue_length=21    (0.42s of audio)
  t=2s   queue_length=70    (1.40s)
  t=3s   queue_length=118   (2.36s)
  t=4s   queue_length=166   (3.32s)
  t=5s   queue_length=215   (4.30s)
  t=6s   queue_length=263   (5.26s)
  -> grows ~48 frames/s, i.e. exactly the excess over the 50 fps consumed

Steady phase - sender back to exactly 50 fps (nominal rate):
  t= 5s  queue_length=290   (5.80s)
  t=10s  queue_length=289   (5.78s)
  t=15s  queue_length=288   (5.76s)
  t=20s  queue_length=287   (5.74s)
  t=25s  queue_length=286   (5.72s)
  t=30s  queue_length=285   (5.70s)
  t=35s  queue_length=284   (5.68s)
  t=40s  queue_length=284   (5.68s)
  t=45s  queue_length=283   (5.66s)
  t=50s  queue_length=282   (5.64s)
  t=55s  queue_length=282   (5.64s)
  t=60s  queue_length=281   (5.62s)

  9 frames drained in 60 seconds (0.15 frames/s). At that rate the backlog
  would take ~31 minutes to clear, and the residual drain only happens because
  the test sender ran marginally below 50 fps. The TCP send buffer on the app
  side stayed at 0 bytes throughout, so the backlog is inside Asterisk.

=== PATCHED (catch-up applied) ===

Burst phase - same 100 fps for 6s:
  t=1s   queue_length=6     (0.12s)
  t=2s   queue_length=6
  t=3s   queue_length=6
  t=4s   queue_length=5
  t=5s   queue_length=6
  t=6s   queue_length=6
  -> the backlog never forms; catch-up absorbs the burst in real time

Steady phase - back to 50 fps:
  t= 5s  queue_length=4
  t=10s  queue_length=3
  t=15s  queue_length=1
  t=20s  queue_length=1
  t=25s  queue_length=1
  t=30s  queue_length=1
  t=40s  queue_length=1
  t=45s  queue_length=0
  t=60s  queue_length=0

=== In production (before the patch) ===

Same behaviour observed on live traffic with ~300 concurrent calls: every call
with media in that direction grew its queue at ~1 frame/s regardless of bridge
type (softmix and simple_bridge alike), reaching 897 frames (17.9s) on calls
that had been up for ~15 minutes. The application was sending a steady 50 fps
with an empty TCP send buffer in every sample.
```

### AI disclosure

Per the issue guidelines: this report was prepared with AI assistance — Claude Opus 5 (claude-opus-5, Anthropic), used via Claude Code. It formed the hypothesis and wrote the test tooling. All measurements above come from runs on our own hardware rather than from the model, and the change has been running in our production environment since 2026-08-22.

Verification changed the conclusions: the model's original hypothesis was that CPU contention caused the accumulation, but our load testing contradicted it (a 16-core box at load 82 with 600 concurrent simulated calls kept queues flat), so that hypothesis was dropped. The measured-versus-inferred note in the "Description" section reflects the same verification.

### Asterisk Issue Guidelines

gtjoseph:
Besides the User and Developer notes below, checks have been added to
reject a START_MEDIA_BUFFERING buffering command if it's already
active and to reject a STOP_MEDIA_BUFFERING command if a corresponding
START_MEDIA_BUFFERING command was never issued.

Resolves: #2115

UserNote: A new `unbuffered` chan_websocket mode has been added which
sends frames to the core as they're received from the WebSocket and your
app. It's similar to `passthrough` mode where your app is responsible
for proper framing and timing except that, assuming you're not using an
unbufferable codec (opus, speex, g729), in unbuffered mode, you can
still use START_MEDIA_BUFFERING to send bulk media and have
chan_websocket frame and time it. When you send STOP_MEDIA_BUFFERING,
the channel will go back into unbuffered mode. Unbuffered mode is most
useful when the media your app normally sends to Asterisk is already
properly framed and timed but you may occasionally need to send bulk
media. The mode is set with the `u` dialstring option.

DeveloperNote: Two new fields have been added to the chan_websocket
MEDIA_START event for both the plain text and JSON formats:
    `passthrough: true/false`
    `unbuffered: true/false`
These indicate the mode the channel was created with and correspond
to the `p` and `u` dialstring options.

DeveloperNote: The MARK_MEDIA chan_websocket command can now be used
by an app when in "passthrough" mode.

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -105,6 +105,8 @@ struct websocket_pvt {
 	enum webchan_control_msg_format control_msg_format;
 	int no_auto_answer;
 	int passthrough;
+	int unbuffered;
+	int last_unbuffered;
 	int optimal_frame_size;
 	int bulk_media_in_progress;
 	int report_queue_drained;
@@ -228,7 +230,7 @@ static char *_create_event_MEDIA_START(struct websocket_pvt *instance)
 	char *payload = NULL;
 
 	if (instance->control_msg_format == WEBCHAN_CONTROL_MSG_FORMAT_JSON) {
-		struct ast_json *msg = ast_json_pack("{s:s, s:s, s:s, s:s, s:s, s:i, s:i, s:o }",
+		struct ast_json *msg = ast_json_pack("{s:s, s:s, s:s, s:s, s:s, s:i, s:i, s:b, s:b, s:o }",
 			"event", "MEDIA_START",
 			"connection_id", instance->connection_id,
 			"channel", ast_channel_name(instance->channel),
@@ -236,6 +238,8 @@ static char *_create_event_MEDIA_START(struct websocket_pvt *instance)
 			"format", ast_format_get_name(instance->native_format),
 			"optimal_frame_size", instance->optimal_frame_size,
 			"ptime", instance->native_codec->default_ms,
+			"passthrough", instance->passthrough,
+			"unbuffered", instance->unbuffered,
 			"channel_variables", ast_json_channel_vars(ast_channel_varshead(
 						instance->channel))
 			);
@@ -245,14 +249,16 @@ static char *_create_event_MEDIA_START(struct websocket_pvt *instance)
 		payload = ast_json_dump_string_format(msg, AST_JSON_COMPACT);
 		ast_json_unref(msg);
 	} else {
-		ast_asprintf(&payload, "%s %s:%s %s:%s %s:%s %s:%s %s:%d %s:%d",
+		ast_asprintf(&payload, "%s %s:%s %s:%s %s:%s %s:%s %s:%d %s:%d %s:%s %s:%s",
 			"MEDIA_START",
 			"connection_id", instance->connection_id,
 			"channel", ast_channel_name(instance->channel),
 			"channel_id", ast_channel_uniqueid(instance->channel),
 			"format", ast_format_get_name(instance->native_format),
 			"optimal_frame_size", instance->optimal_frame_size,
-			"ptime", instance->native_codec->default_ms
+			"ptime", instance->native_codec->default_ms,
+			"passthrough", instance->passthrough ? "true" : "false",
+			"unbuffered", instance->unbuffered ? "true" : "false"
 			);
 	}
 
@@ -456,11 +462,13 @@ static __attribute__ ((format (gnu_printf, 2, 3))) char *_create_event_ERROR(
 	(_res); \
 })
 
-/*
- * Reminder...  This function gets called by webchan_read which is
- * triggered by the channel timer firing.  It always gets called
- * every 20ms (or whatever the timer is set to) even if there are
- * no frames in the queue.
+/*!
+ * \internal
+ *
+ * This function gets called by webchan_read which is triggered by the channel
+ * timer firing or by a websocket frame being received while in unbuffered mode.
+ * It always gets called at least every 20ms (or whatever the timer is set to)
+ * even if there are no frames in the queue.
  */
 static struct ast_frame *dequeue_frame(struct websocket_pvt *instance)
 {
@@ -593,8 +601,18 @@ static struct ast_frame *create_frame_from_buffer(struct websocket_pvt *instance
  *
  *   The websocket fd (WS_WEBSOCKET_FDNO) which gets triggered when
  *   there's incoming data to read from the websocket.  In this case,
- *   we read the data and put it ON the queue.  We'll return a null frame.
+ *   we read the data and put it ON the queue.  If in unbuffered mode,
+ *   we'll read it off the queue immediately and return it.  If not in
+ *   unbuffered mode, we'll return a null frame now and the frame read
+ *   will be processed by the timer tick when it comes to the head
+ *   of the queue.
  *
+ *   The reason for queueing and dequeueing immediately in unbuffered
+ *   mode is that read_from_ws_and_queue does a lot of work before adding
+ *   a frame to the frame queue, if it even needs to.  It's much simpler
+ *   to just queue and dequeue, both from a code organization standpoint
+ *   as well as an instruction-path-length standpoint, than it would be
+ *   refactor that code so it can return a frame directly.
  */
 static struct ast_frame *webchan_read(struct ast_channel *ast)
 {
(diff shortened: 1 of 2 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

giandiego:
I built and ran the PR. I found two things.

The unbuffered path itself worked as expected in my runs. Reproducing #2115 against this branch (082d6e415), without `u` the queue peaks at 251 during a 6s burst at 100 fps, settles at 301, and drains 0 frames over the following 55s at a measured 50.0 fps. With `u` on the same build, `queue_length` stays at 0 through every sample, burst included.

**1. The macro checks the wrong flag.** `ERROR_ON_UNBUFFERED_MODE_RTN` evaluates `instance->passthrough` instead of `instance->unbuffered`:

```c
#define ERROR_ON_UNBUFFERED_MODE_RTN(instance, command) \
({ \
    if (instance->passthrough) { \
        send_event(instance, ERROR, "%s not supported in unbuffered mode", command); \
```

With `p()`, or with a small-frame codec, both `passthrough` and `unbuffered` end up at 1, so the macro appears to work. It fails in the case the PR adds: `u` without `p` leaves `unbuffered=1` and `passthrough=0`, so `FLUSH_MEDIA`, `PAUSE_MEDIA` and `CONTINUE_MEDIA` never emit the ERROR and fall through to the queue code.

Built and run against the same build, with `Dial(WebSocket/wsmin/c(alaw)f(json)u)`:

```
<- {"event":"MEDIA_START",...,"optimal_frame_size":160,"ptime":20}
-> PAUSE_MEDIA
-> GET_STATUS
<- {"event":"STATUS",...,"queue_length":0,"media_paused":true}
ERROR events: 0
```

`PAUSE_MEDIA` is accepted and leaves `media_paused=true` on the queue that mode does not use.

**2. Nothing in the protocol distinguishes the mode.** `MEDIA_START` (`chan_websocket.c:231-241`) and `STATUS` (`:360-368`) carry no mode field, and `optimal_frame_size` only goes to zero by codec, in `:1426-1428`, not by the dialstring option. Today the ERROR text is the only indication that separates a channel rejecting those commands from one that does not. This PR adds a new message, `"not supported in unbuffered mode"`, alongside a mode that is not announced anywhere else either.

AI disclosure: I built and ran the PR with AI assistance (Claude). I verified every code reference against the master sources and ran the reproduction myself.

gtjoseph:
Thanks for the feedback.  I've fixed the macro and added fields to the MEDIA_START event to indicate whether passthrough and/or unbuffered modes are set.

mbradeen (approved):

jcolp (approved):

PR #2122, opened 2026-08-26
res_websocket_client: Release lock_obj when retries are exhausted

### LINKED ISSUE #2121: [bug]: res_websocket_client: ast_websocket_client_connect() does not release lock_obj when retries are exhausted
### Severity

Minor

### Versions

20, 22, 23, master

### Components/Modules

`res_websocket_client`

### Operating Environment

Debian 13, x86_64

### Frequency of Occurrence

Constant

### Summary

`ast_websocket_client_connect()` takes `lock_obj` at the top of every pass
through the retry loop (line 415). When the attempts run out it leaves through
`break` (476) and returns NULL without releasing it, so the calling thread ends
up holding it. On top of that, line 480 takes it again instead of releasing it
before sleeping:

```c
		if (lock_obj) {
			ao2_lock(lock_obj);      /* 415: taken at the top of every pass */
		}
		...
			break;                   /* 476: leaves holding it */
		}

		if (lock_obj) {
			ao2_lock(lock_obj);      /* 480: takes it again */
		}
		usleep(wc->reconnect_interval * 1000);
```

The other two exits, the `ast_calloc` failure (428) and the success path (445),
do release it.

The header describes `lock_obj` as "an ao2 object to lock while the connection is
being attempted" (`include/asterisk/websocket_client.h:159-160`). The
`usleep(reconnect_interval)` is the wait between attempts, not an attempt, so
that is where it should be released.

### Effect

`chan_websocket.c:1295` passes the per-call instance as `lock_obj`, allocated at
1386 with `ao2_alloc` and no flags, so it gets the default
`AO2_ALLOC_OPT_LOCK_MUTEX`. When the object is freed, `__ao2_ref()` reaches the
`ast_mutex_destroy()` at `astobj2.c:632` with the mutex still held, and frees the
memory anyway. There is no deadlock; Asterisk mutexes are recursive. In practice
this produces two ERROR lines per failed call.

The other caller, `res/ari/ari_websockets.c:975`, passes `lock_obj = NULL` and is
unaffected.

### Steps to reproduce

`websocket_client.conf`, with nothing listening on the port:

```ini
[wsmin]
type = websocket_client
uri = ws://127.0.0.1:9000/media
connection_type = per_call_config
reconnect_attempts = 5
```

```
CLI> channel originate WebSocket/wsmin/c(alaw)f(json) application Wait 1
```

On master `03b5137` built with `DEBUG_THREADS`, `core show locks` shows the
`lock_obj` count going up by two per pass with no unlock in between, while the
other locks held by the same thread stay at `(1,1)`.

### Proposed fix

It has been there since the function was introduced in `5a3164c0b2`
(2025-05-02), so it is not a regression. The same code is in 20, 22, 23 and
master.

```diff
 		if (lock_obj) {
-			ao2_lock(lock_obj);
+			ao2_unlock(lock_obj);
 		}
 		usleep(wc->reconnect_interval * 1000);
 	}
 
+	if (lock_obj) {
+		ao2_unlock(lock_obj);
+	}
+
 	return NULL;
```

With every pass balanced, the lock is held exactly once when the loop breaks, so
a single unlock before `return NULL` is enough. Tested on master: the failing
scenarios produce no ERROR lines and the success path is unchanged.

### Relevant log output

```shell
WARNING res_websocket_client.c: WebSocket/wsmin/0x...: Websocket connection to ws://127.0.0.1:9000/media failed after 5 tries: Client start error (Transport endpoint is not connected).  Hanging up after exhausting retries.
ERROR lock.c: astobj2.c line 632 (__ao2_ref): Error destroying mutex &obj_mutex->mutex.lock: Device or resource busy
ERROR utils.c: Thread 'pbx_outgoing_exec' still has a lock! - 'lock_obj' (0x...) from 'ast_websocket_client_connect' in res_websocket_client.c:415!
```

### Asterisk Issue Guidelines

giandiego:
ast_websocket_client_connect() takes lock_obj at the top of every pass
through the retry loop.  When the attempts run out it leaves through the
break and returns NULL without releasing it, so the calling thread ends
up holding it.  The other two exits, the ast_calloc failure and the
success path, do release it.

The retry path also takes the lock again instead of releasing it before
sleeping.  The header describes lock_obj as "an ao2 object to lock while
the connection is being attempted", and the usleep is the wait between
attempts rather than an attempt, so that is where it should be released.

chan_websocket.c passes the per-call instance as lock_obj, allocated
with ao2_alloc and no flags, so it gets the default
AO2_ALLOC_OPT_LOCK_MUTEX.  When the object is freed, __ao2_ref() reaches
ast_mutex_destroy() with the mutex still held and frees the memory
anyway.  There is no deadlock, Asterisk mutexes are recursive, but every
call that fails to connect logs two ERROR lines.

With each pass balanced, the lock is held exactly once when the loop
breaks, so a single unlock before the return is enough.

Fixes: #2121

### DIFF
--- a/res/res_websocket_client.c
+++ b/res/res_websocket_client.c
@@ -477,9 +477,13 @@ struct ast_websocket *ast_websocket_client_connect(struct ast_websocket_client *
 		}
 
 		if (lock_obj) {
-			ao2_lock(lock_obj);
+			ao2_unlock(lock_obj);
 		}
 		usleep(wc->reconnect_interval * 1000);
+	}
+
+	if (lock_obj) {
+		ao2_unlock(lock_obj);
 	}
 
 	return NULL;


giandiego:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

giandiego:
AI disclosure: this change was made with AI assistance (Claude). It ran the lab reproduction on a clean master build and drafted the patch and the commit message under my direction. I verified every code reference against the master sources, reran the reproduction and the lock measurements myself, and tested the cherry-picks to 23, 22 and 20. I can debug and own this change.

jcolp (approved):

gtjoseph (approved):

PR #2127, opened 2026-08-28
res_prometheus: add toggle config for metrics

whitetreebug:
In high‑volume call scenarios, Prometheus exports a large number of channels‑related and bridges-related metrics.Not all users require channels_state, channels_duration and bridges_channels.Add conf options to enable or disable channel_details(covering channels_state, channels_duration) and bridge_details(covering bridges_channels).

UserNote: Add new config option 'channels_detail_metrics_enabled' to disable export of 'asterisk_channels_state' and 'asterisk_channels_duration_seconds' metrics,It defaults to 'yes' to preserve existing behavior

Add new config option 'bridges_detail_metrics_enabled' to disable export of 'asterisk_bridges_channels_count' metric, It defaults to 'yes' to preserve existing behavior

Explicitly enable bridges_detail_metrics_enabled for the bridge_to_string unit‑test.

### DIFF
--- a/res/prometheus/bridges.c
+++ b/res/prometheus/bridges.c
@@ -77,6 +77,7 @@ struct bridge_metric_defs {
  */
 static void bridges_scrape_cb(struct ast_str **response)
 {
+	RAII_VAR(struct prometheus_general_config *, config, prometheus_general_config_get(), ao2_cleanup);
 	struct ao2_container *bridge_cache;
 	struct ao2_container *bridges;
 	struct ao2_iterator it_bridges;
@@ -113,6 +114,11 @@ static void bridges_scrape_cb(struct ast_str **response)
 	prometheus_metric_to_string(&bridge_count, response);
 
 	if (num_bridges == 0) {
+		ao2_ref(bridges, -1);
+		return;
+	}
+
+	if (!config || !config->bridges_detail_metrics_enabled) {
 		ao2_ref(bridges, -1);
 		return;
 	}
--- a/res/prometheus/channels.c
+++ b/res/prometheus/channels.c
@@ -129,6 +129,7 @@ static struct prometheus_metric global_channel_metrics[] = {
  */
 static void channels_scrape_cb(struct ast_str **response)
 {
+	RAII_VAR(struct prometheus_general_config *, config, prometheus_general_config_get(), ao2_cleanup);
 	struct ao2_container *channel_cache;
 	struct ao2_container *channels;
 	struct ao2_iterator it_chans;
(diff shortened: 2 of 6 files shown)


whitetreebug:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

whitetreebug:
image
image

res/res_prometheus.c
```diff
@@ -83,6 +83,38 @@
 						</enumlist>
 					</description>
 				</configOption>
+				<configOption name="channels_detail_metrics_enabled" default="yes">
+					<since>
+						<version>20.0.0</version>
```
jcolp: This version is not correct at all. Versions for 20, 22, 23, and 24 need to be present - and you can get the next versions by looking at the current release and incrementing by one.

res/res_prometheus.c
```diff
@@ -83,6 +83,38 @@
...
+						</para>
+						<enumlist>
+							<enum name="no" />
+							<enum name="yes" />
+						</enumlist>
+					</description>
+				</configOption>
+				<configOption name="bridges_detail_metrics_enabled" default="yes">
+					<since>
+						<version>20.0.0</version>
```
jcolp: Same here.

These are important because they become publicly visible so people know when things were introduced.

jcolp (changes requested):

res/res_prometheus.c
whitetreebug: Thanks a lot for the heads‑up. I will fix it.

res/prometheus/bridges.c
```diff
@@ -117,6 +118,11 @@ static void bridges_scrape_cb(struct ast_str **response)
 		return;
 	}
 
+	if (!config || !config->bridges_detail_metrics_enabled) {
```
jcolp, line 121: I don't see why this callback has to do all the work of getting the bridges if it's only just going to not use it.

res/prometheus/channels.c
```diff
@@ -176,6 +177,11 @@ static void channels_scrape_cb(struct ast_str **response)
 		return;
 	}
 
+	if (!config || !config->channels_detail_metrics_enabled) {
```
jcolp, line 180: Same here.

jcolp (changes requested):

whitetreebug:
> image image

@jcolp `asterisk_calls_count` and `asterisk_bridges_count` remain exported in both functions.
In some scenarios, only the total call count is required, not the details.

jcolp (approved):

PR #2136, opened 2026-09-01
codec_ilbc: Build against system libilbc 3.x

### LINKED ISSUE #2135: [improvement]: Build codec_ilbc against system libilbc >= 3.0
### Background

`codec_ilbc` can build against the system libilbc (the WebRTC fixed-point
implementation) only when it is `< 3.0`. Since #84 was resolved by #419,
`configure` requires `libilbc < 3` and falls back to the bundled RFC 3951
reference for any system libilbc >= 3.0. libilbc has been on the 3.x series for
years (e.g. Fedora ships 3.0.4), so on most modern distributions `codec_ilbc`
is effectively always built against the bundled copy.

The only thing 3.x actually broke was the removal of the RFC 3951 compatibility
function names (`initEncode`/`initDecode`/`iLBC_encode`/`iLBC_decode`) that the
`ILBC_WEBRTC` path calls. Those were thin aliases for the underlying
`WebRtcIlbcfix_*` functions, which are unchanged.

### Proposal

1. Map the four removed names to their `WebRtcIlbcfix_*` equivalents in the
   `ILBC_WEBRTC` branch of `codec_ilbc.c`.
2. Relax the `libilbc < 3` pkg-config check to `libilbc`.

This lets distributions that cannot ship bundled copies build `codec_ilbc`
against the system library again; everyone else is unaffected (the bundled copy
is still used when no suitable system libilbc is present).

### API compatibility

Identical arity, parameter order and semantics for all four functions; they
differ only in the buffer element types the `ILBC_WEBRTC` branch already uses
(`uint16_t`/`int16_t`), and trivial return/`mode` widths that `codec_ilbc`
ignores or safely truncates:

| function | bundled (RFC 3951) | system libilbc 3.0.4 (`WebRtcIlbcfix_*`) |
|---|---|---|
| InitEncode | `short (Enc*, int)` | `int (IlbcEncoder*, int16_t)` |
| InitDecode | `short (Dec*, int, int)` | `int (IlbcDecoder*, int16_t, int)` |
| EncodeImpl | `void (uchar*, float*, Enc*)` | `void (uint16_t*, const int16_t*, IlbcEncoder*)` |
| DecodeImpl | `void (float*, uchar*, Dec*, int)` | `int (int16_t*, const uint16_t*, IlbcDecoder*, int16_t)` |

The system header keeps deprecated aliases (`iLBC_Enc_Inst_t` -> `IlbcEncoder`,
etc.), so existing struct declarations resolve.

### Interoperability tested

Rather than rely on "it builds", we wrote a small utility that encodes/decodes
the same PCM with the bundled reference and with system libilbc 3.0.4 — driving
each exactly as `codec_ilbc.c` does — and cross-decoded between them (8 kHz):

| mode | frame | self bundled->bundled | self system->system | interop system->bundled | interop bundled->system |
|---|---|---|---|---|---|
| 20 ms | 38 B | 6.2 dB | 5.7 dB | 5.7 dB | 6.2 dB |
| 30 ms | 50 B | 5.9 dB | 6.0 dB | 6.1 dB | 5.8 dB |

Frame sizes match; encoded streams share framing and differ only in the
expected quantization positions (float reference vs fixed-point); and
cross-decoding lands at the same SNR as self-roundtrip. A wrong mapping, buffer
width or byte order would collapse the interop rows to noise -- they don't. So
frames produced against system libilbc are correct, standards-compliant iLBC
that interoperates with the bundled reference and any other iLBC peer.

### Related
- #84 -- original "fails to build with ilbc 3.0.4"
- #419 -- resolved #84 by disabling system libilbc >= 3

A change is ready (the `codec_ilbc.c` mapping + the `configure` relaxation).

lemenkov:
libilbc 3.x removed the RFC 3951 compatibility function names (initDecode,
initEncode, iLBC_decode, iLBC_encode). Map them to the WebRtcIlbcfix_* symbols
the library exposes, and drop the "libilbc < 3" configure restriction, so
codec_ilbc is built against the system libilbc even when it is version 3.x.

Resolves: #2135

Signed-off-by: Peter Lemenkov <lemenkov@gmail.com>

### DIFF
--- a/codecs/codec_ilbc.c
+++ b/codecs/codec_ilbc.c
@@ -45,6 +45,21 @@
 typedef uint16_t ilbc_bytes;
 typedef int16_t  ilbc_block;
 #define BUF_TYPE i16
+/* libilbc 3.x removed the RFC 3951 compatibility macros that 2.x's <ilbc.h>
+ * provided.  Re-add them only when the header did not, so this is a no-op
+ * against libilbc 2.x and only takes effect on 3.x. */
+#ifndef initDecode
+#define initDecode WebRtcIlbcfix_InitDecode
+#endif
+#ifndef initEncode
+#define initEncode WebRtcIlbcfix_InitEncode
+#endif
+#ifndef iLBC_decode
+#define iLBC_decode WebRtcIlbcfix_DecodeImpl
+#endif
+#ifndef iLBC_encode
+#define iLBC_encode WebRtcIlbcfix_EncodeImpl
+#endif
 #else
 #include "ilbc/iLBC_encode.h"
 #include "ilbc/iLBC_decode.h"
--- a/configure
+++ b/configure
@@ -34380,19 +34380,19 @@ if test "${USE_ILBC}" != "no"; then
       if test "x${PBX_ILBC}" != "x1" -a "${USE_ILBC}" != "no"; then
 
 pkg_failed=no
-{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for libilbc < 3" >&5
-printf %s "checking for libilbc < 3... " >&6; }
+{ printf "%s\n" "$as_me:${as_lineno-$LINENO}: checking for libilbc" >&5
+printf %s "checking for libilbc... " >&6; }
 
 if test -n "$ILBC_CFLAGS"; then
     pkg_cv_ILBC_CFLAGS="$ILBC_CFLAGS"
  elif test -n "$PKG_CONFIG"; then
     if test -n "$PKG_CONFIG" && \
-    { { printf "%s\n" "$as_me:${as_lineno-$LINENO}: \$PKG_CONFIG --exists --print-errors \"libilbc < 3\""; } >&5
-  ($PKG_CONFIG --exists --print-errors "libilbc < 3") 2>&5
+    { { printf "%s\n" "$as_me:${as_lineno-$LINENO}: \$PKG_CONFIG --exists --print-errors \"libilbc\""; } >&5
+  ($PKG_CONFIG --exists --print-errors "libilbc") 2>&5
   ac_status=$?
   printf "%s\n" "$as_me:${as_lineno-$LINENO}: \$? = $ac_status" >&5
   test $ac_status = 0; }; then
-  pkg_cv_ILBC_CFLAGS=`$PKG_CONFIG --cflags "libilbc < 3" 2>/dev/null`
+  pkg_cv_ILBC_CFLAGS=`$PKG_CONFIG --cflags "libilbc" 2>/dev/null`
 		      test "x$?" != "x0" && pkg_failed=yes
 else
   pkg_failed=yes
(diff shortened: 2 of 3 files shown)


lemenkov:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

codecs/codec_ilbc.c
```diff
@@ -45,6 +45,11 @@
 typedef uint16_t ilbc_bytes;
 typedef int16_t  ilbc_block;
 #define BUF_TYPE i16
+/* In the ilbc library were removed RFC 3951 compatibility macros */
+#define initDecode WebRtcIlbcfix_InitDecode
```
jcolp, line 49: Is it possible for ILBC_WEBRTC to be defined in an older non-3 version, and if so does this break compatibility with those?

jcolp (changes requested):
Why is this only going to master? Is there a specific reason?

codecs/codec_ilbc.c
lemenkov: libilbc 2.x's own `<ilbc.h>` already defines these exact compat macros (`#define initEncode WebRtcIlbcfix_InitEncode`, etc.); 3.x removed them, which is what broke the build. So against 2.x re-adding them is a benign, identical macro redefinition — no error or warning, even under `-Werror`. I've now wrapped each in `#ifndef` so it's an explicit no-op when the header still provides them (2.x) and only re-adds them on 3.x.

lemenkov:
No specific reason — you're right that the default is to backport, so I've updated `cherry-pick-to` to 24/23/22/20. The change is safe on those branches: it only takes effect when a system libilbc >= 3 is actually present, otherwise the build is unchanged (bundled, or system libilbc 2.x).

lemenkov:
The previous CI failure here was unrelated to this change: the only failing test was `/main/taskpool/ - selector_skips_busy_executor` (a timing-sensitive taskpool test), and the iLBC codec test was skipped (`no translation path available`), so `codec_ilbc` was not exercised in that job. The same `master-unit` suite passes on other current PRs and on master. I've rebased onto current master and force-pushed to re-run CI.

jcolp (approved):

codecs/codec_ilbc.c
gtjoseph, line 81: Compile failure:

```
codec_ilbc.c:81:9: error: ‘iLBC_Enc_Inst_t’ is deprecated [-Werror=deprecated-declarations]
   81 |         iLBC_Enc_Inst_t enc;
      |         ^~~~~~~~~~~~~~~
In file included from codec_ilbc.c:44:
/usr/include/ilbc.h:333:16: note: declared here
  333 | typedef struct IlbcEncoder_ {
      |                ^~~~~~~~~~~~
codec_ilbc.c:82:9: error: ‘iLBC_Dec_Inst_t’ is deprecated [-Werror=deprecated-declarations]
   82 |         iLBC_Dec_Inst_t dec;
      |         ^~~~~~~~~~~~~~~
/usr/include/ilbc.h:373:16: note: declared here
  373 | typedef struct IlbcDecoder_ {
      |                ^~~~~~~~~~~~
```

Running Fedora 44
```
ilbc.x86_64       3.0.4-19.fc44 fedora
ilbc-devel.x86_64 3.0.4-19.fc44 fedora
```

gtjoseph (changes requested):

codecs/codec_ilbc.c
lemenkov: Good catch, thanks. libilbc 3.x marks the `iLBC_Enc_Inst_t`/`iLBC_Dec_Inst_t` type aliases as deprecated (in favour of `IlbcEncoder`/`IlbcDecoder`), and dev-mode `-Werror=deprecated-declarations` turns that into a build error. I've added a version-guarded mapping to the current type names for libilbc >= 3, alongside the function-name macros:

```c
#if defined(LIBILBC_VERSION_MAJOR) && LIBILBC_VERSION_MAJOR >= 3
#define iLBC_Enc_Inst_t IlbcEncoder
#define iLBC_Dec_Inst_t IlbcDecoder
#endif
```

libilbc 2.x has only the `iLBC_*` names (no `IlbcEncoder`), and `LIBILBC_VERSION_MAJOR` doesn't exist there, so the guard leaves 2.x and the bundled copy untouched. Reproduced the deprecation error against ilbc-devel 3.0.4 and confirmed the fix clears it under `-Werror=deprecated-declarations`. Force-pushed.

jcolp (approved):

codecs/codec_ilbc.c
gtjoseph: Thanks for fixing.  Builds fine on Fedora 44 now.

PR #2137, opened 2026-09-02
app_confbridge: Add announcements option to disable conference announcements

### LINKED ISSUE #2131: [improvement]: Add an option to disable the ConfBridge announcer channel
## Summary

ConfBridge currently creates a `CBAnn` announcer channel for every conference, even when no announcements are configured or used. 

These unnecessary announcer channels introduce additional channel, bridge, and taskprocessor overhead, which can have a measurable negative impact on Asterisk performance, especially in deployments with a high number of concurrent conferences.

It would be useful to have a bridge profile option to disable the announcer channel entirely for conferences that do not use ConfBridge announcements.

For example:

```ini
[my_bridge]
type=bridge
announcements=no
```
The default should remain announcements=yes to preserve the current behavior.

## Reference
The behavior appears to originate from the ConfBridge announcer architecture introduced in the ASTERISK-26289 changes, where the announcer channel became persistent rather than being created only when an announcement needed to be played. https://github.com/asterisk/asterisk/commit/0cdeb2bfb0

f-vicente:
Every conference unconditionally created an announcer channel (CBAnn) and a
taskprocessor thread to play announcements to the conference as a whole, even on
systems that never play any.  That costs a channel, a thread and a bridge member
per conference.

A new bridge profile option, announcements, defaults to yes and preserves the
previous behavior.  When set to no, no announcer channel is created and nothing
is played to the conference as a whole: the join and leave sounds, recorded name
intros, the participant count announced to everyone, sound_begin and
sound_leader_has_left are all suppressed, and users are never prompted to record
their name regardless of the announce_join_leave user profile option.

Prompts played only to the participant that caused them are unaffected, since
they are played on that participant's own channel.  sound_kicked, sound_muted,
sound_unmuted, the join sound played back by hear_own_join_sound and the
participant count requested from the DTMF menu all still play.

The announcer channel exists for the lifetime of the conference, so the value
that takes effect is the one from the bridge profile of the channel that creates
the conference.

Resolves: #2131

UserNote: A new ConfBridge bridge profile option, announcements, can be set to
no to stop announcements from being played to the conference as a whole.  No
announcer channel is then created, which saves a channel and a taskprocessor
thread per conference.  Prompts played only to the participant that caused them
are not affected.  It defaults to yes, which is the previous behavior.

### DIFF
--- a/apps/app_confbridge.c
+++ b/apps/app_confbridge.c
@@ -1127,6 +1127,12 @@ static int announce_user_count(struct confbridge_conference *conference, struct 
 	const char *only_one = conf_get_sound(CONF_SOUND_ONLY_ONE, conference->b_profile.sounds);
 	const char *there_are = conf_get_sound(CONF_SOUND_THERE_ARE, conference->b_profile.sounds);
 
+	if (!user && !conference->playback_queue) {
+		/* Announcements to the entire conference are disabled. Announcements to a
+		 * single caller are played on that caller's channel, so they still work. */
+		return 0;
+	}
+
 	if (conference->activeusers <= 1) {
 		/* Awww we are the only person in the conference bridge OR we only have waitmarked users */
 		return 0;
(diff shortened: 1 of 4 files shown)


f-vicente:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

gtjoseph:
Just FYI...  The 24 branch is now active and should be included in the cherry-picks.  I've taken care of it this time.

jcolp (approved):

PR #2139, opened 2026-09-03
chan_dahdi: Make DAHDI device state cacheable to fix device state sync.

### LINKED ISSUE #2138: [improvement]: chan_dahdi: Device state is uncachable, preventing device state synchronization
### Improvement Description

In chan_dahdi, device state is set as always uncachable, even though it corresponds to real devices. This seems like a regression caused by commit 8fb5bdce9ab9f7f3758545753cbc787653920753

InterLinked1:
Commit 8fb5bdce9ab9f7f3758545753cbc787653920753 disabled caching of
certain device states that weren't associated with real devices in
order to prevent a resource exhaustion attack from guest/unauthenticated
connections.

In most cases, code was converted to remain cacheable; the few exceptions
had mainly to do with certain paths in channel drivers for guest or
unauthenticated access. However, everything in chan_dahdi was made
uncachable for no apparent reason. This is not harmless, since it
prevents device state synchronization from working (e.g. with
res_pjsip_publish_asterisk) since device state updates that are
marked uncachable result in hints on remote systems always showing
"Unavailable" since uncached device state updates are simply discarded.

Since DAHDI devices very much correspond to real devices (perhaps more
so than in any other channel driver), we can safely cache device state
for DAHDI channels, which also ensures that device state
synchronization now works properly.

Resolves: #2138

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -3305,7 +3305,7 @@ static void dahdi_pri_update_span_devstate(struct sig_pri_span *pri)
 	}
 	if (pri->congestion_devstate != new_state) {
 		pri->congestion_devstate = new_state;
-		ast_devstate_changed(AST_DEVICE_UNKNOWN, AST_DEVSTATE_NOT_CACHABLE, "DAHDI/I%d/congestion", pri->span);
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

PR #2141, opened 2026-09-03
ARI, chan_websocket, res_websocket_client: Handle non-blocking and timeout correctly.

### LINKED ISSUE #2140: [bug]: Cascading locks when chan websocket is connected to a server that stops reading packets.
### Severity

Major

### Versions

All current

### Components/Modules

res_http_websocket, res_websocket_client, chan_websocket

### Operating Environment

All

### Frequency of Occurrence

Occasional

### Issue Description

The default mode for websockets is "blocking" and chan_websocket doesn't change that. If the websocket server to which chan_websocket is connected stops consuming TCP packets, its TCP receive buffer will begin to fill and ultimately cause the client's TCP send buffer to start filling.  When the send buffer is completely filled, further writes to the websocket will block.  Given how quickly chan_websocket sends packets, it could take only 10 seconds before the writes start to block.  This can cause a lock cascade because the core is holding the channel lock during this time.  If the calling channel hangs up, it too will block because it needs the peer channel to set hangup cause and RTP stats variables.

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

gtjoseph:
This change was prompted by an issue where if the websocket peer to
which ARI or chan_websocket is connected stops consuming TCP packets,
its TCP receive buffer will begin to fill and ultimately cause the
client's TCP send buffer to start filling. When the send buffer is
completely filled, further writes to the websocket will block. This can
cause a cascading lock situation with chan_websocket because it sends
large amounts of data very quickly but can also affect ARI.

* Added "write_timeout" parameters to websocket_client.conf and
  chan_websocket.conf. ari.conf already had the
  "websocket_write_timeout" parameter. See notes below.

* Updated chan_websocket to use the write_timeout set in
  chan_websocket.conf for incoming/server connections if set and to use
  AST_DEFAULT_WEBSOCKET_WRITE_TIMEOUT (100ms) as the default if not set.

* Updated chan_websocket to use the write_timeout set in
  websocket_client.conf for outgoing/client connections if set and to use
  write_timeout from chan_websocket.conf if not set.  If neither is set,
  AST_DEFAULT_WEBSOCKET_WRITE_TIMEOUT (100ms) is used as the default.

* Set non-blocking mode on the websocket in chan_websocket for both
  client and server connections. ARI already sets it.  This is required
  for the timeouts to operate correctly.

* Updated sample config files and XML documentation for all 3 areas.

Resolves: #2140

UpgradeNote: A new "write_timeout" parameter has been added to
websocket_client.conf that allows setting the maximum amount of time a
write operation can take before returning an error.

UpgradeNote: A new "write_timeout" parameter has been added to
chan_websocket.conf setting the maximum amount of time a write operation
can take before returning an error. Outgoing/client connections can
override this with the new "write_timeout" parameter that was also added
to websocket_client.conf.

DeveloperNote: A new "write_timeout" field has been added to the
ast_websocket_client_options structure that allows setting the maximum
amount of time a write operation can take before returning an error.
The websocket session must be set to nonblocking for this to take effect.

### DIFF
--- a/channels/chan_websocket.c
+++ b/channels/chan_websocket.c
@@ -68,6 +68,7 @@ static const char *msg_format_map[] = {
 struct webchan_conf_global {
 	SORCERY_OBJECT(details);
 	enum webchan_control_msg_format control_msg_format;
+	int write_timeout;
 };
 
 /* This is from the perspective of the app, NOT Asterisk */
@@ -534,10 +535,15 @@ static struct ast_frame *dequeue_frame(struct websocket_pvt *instance)
 			 * We just need to send the data to the websocket.
 			 * The data should already be NULL terminated.
 			 */
-			ast_websocket_write_string(instance->websocket,
+			int res = ast_websocket_write_string(instance->websocket,
 				queued_frame->data.ptr);
-			ast_debug(4, "%s: Sent %s\n",
-				ast_channel_name(instance->channel), (char *)queued_frame->data.ptr);
+			if (res != 0) {
+				ast_log(LOG_ERROR, "%s: Unable to send event %s\n",
+					ast_channel_name(instance->channel), (char *)queued_frame->data.ptr);
+			} else {
+				ast_debug(4, "%s: Sent %s\n",
+					ast_channel_name(instance->channel), (char *)queued_frame->data.ptr);
+			}
 		}
 		/*
 		 * We do NOT send these to the core so we need to free
@@ -1214,6 +1220,14 @@ static int websocket_handoff_to_channel(struct websocket_pvt *instance)
 	}
 
 	/*
+	 * The way write timeouts are handled in iostream requires the socket to be
+	 * in non-blocking mode.  This is fine for reads as well because we already
+	 * set the websocket file descriptor on the channel and let it call
+	 * webchan_read() when data is available.
+	 */
+	ast_websocket_set_nonblock(instance->websocket);
+
+	/*
 	 * Tell res_http_websocket to accumulate incoming WebSocket CONTINUATION frames
 	 * into chunks of 1024 bytes and send us a TEXT or BINARY frame when the threshold
 	 * is reached.
@@ -1266,6 +1280,7 @@ static void _websocket_request_hangup(struct websocket_pvt *instance, int ast_ca
 static int webchan_write(struct ast_channel *ast, struct ast_frame *f)
 {
 	struct websocket_pvt *instance = ast_channel_tech_pvt(ast);
+	int res = 0;
 
 	if (!instance || !instance->websocket) {
 		ast_log(LOG_WARNING, "%s: WebSocket instance or client not found\n",
@@ -1295,8 +1310,13 @@ static int webchan_write(struct ast_channel *ast, struct ast_frame *f)
 		return -1;
 	}
 
-	return ast_websocket_write(instance->websocket, AST_WEBSOCKET_OPCODE_BINARY,
+	res = ast_websocket_write(instance->websocket, AST_WEBSOCKET_OPCODE_BINARY,
 		(char *)f->data.ptr, (uint64_t)f->datalen);
+	if (res != 0) {
+		ast_log(LOG_WARNING, "%s: WebSocket write failure\n", ast_channel_name(ast));
+	}
+
+	return res;
 }
 
 /*!
@@ -1311,6 +1331,10 @@ static int webchan_call(struct ast_channel *ast, const char *dest,
 {
 	struct websocket_pvt *instance = ast_channel_tech_pvt(ast);
 	enum ast_websocket_result result;
+	struct webchan_conf_global *global_cfg = ast_sorcery_retrieve_by_id(sorcery, "global", "global");
+	int global_write_timeout = global_cfg ? global_cfg->write_timeout : AST_DEFAULT_WEBSOCKET_WRITE_TIMEOUT;
+
+	ao2_cleanup(global_cfg);
 
 	if (!instance) {
 		ast_log(LOG_WARNING, "%s: WebSocket instance not found\n",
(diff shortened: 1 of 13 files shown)


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

configs/samples/websocket_client.conf.sample
```diff
@@ -24,6 +28,15 @@
                                ; always retry forever but this setting will control
                                ; how often failure messages are logged.
                                ; Default: 4 for both connection types.
+;write_timeout = 100           ; Write timeout in milliseconds.
```
mbradeen, line 31: Do we want to put any restrictions on this value?  What happens if it is set to 0?

gtjoseph: Ah yeah.  I meant to do that.

gtjoseph:
Converting to DRAFT until at least after the long weekend because I need to make sure ARI is handled correctly since it shares res_websocket_client with chan_websocket.

gtjoseph:
ARI, chan_websocket and res_websocket_client should all be consistent now.

channels/chan_websocket.c
```diff
@@ -68,6 +68,7 @@ static const char *msg_format_map[] = {
 struct webchan_conf_global {
 	SORCERY_OBJECT(details);
 	enum webchan_control_msg_format control_msg_format;
+	int write_timeout;
```
jcolp, line 71: The websocket client itself uses unsigned int, but this uses int. Why the difference?

channels/chan_websocket.c
```diff
@@ -1213,6 +1167,8 @@ static int websocket_handoff_to_channel(struct websocket_pvt *instance)
 		ast_log(LOG_WARNING, "Failed to set TCP_NODELAY on websocket connection: %s\n", strerror(errno));
 	}
 
+	ast_websocket_set_nonblock(instance->websocket);
```
jcolp, line 1228: Add a comment explaining why we use non-blocking.

channels/chan_websocket.c
```diff
@@ -1311,6 +1273,9 @@ static int webchan_call(struct ast_channel *ast, const char *dest,
 {
 	struct websocket_pvt *instance = ast_channel_tech_pvt(ast);
 	enum ast_websocket_result result;
+	RAII_VAR(struct webchan_conf_global *, global_cfg, NULL, ao2_cleanup);
+
+	global_cfg = ast_sorcery_retrieve_by_id(sorcery, "global", "global");
```
jcolp: Is a global section guaranteed to exist? The sample configuration has it entirely commented out.

channels/chan_websocket.c
```diff
@@ -1877,6 +1835,9 @@ static void incoming_ws_established_cb(struct ast_websocket *ast_ws_session,
 	struct ast_variable *v;
 	const char *connection_id = NULL;
 	struct websocket_pvt *instance = NULL;
+	RAII_VAR(struct webchan_conf_global *, global_cfg, NULL, ao2_cleanup);
+
+	global_cfg = ast_sorcery_retrieve_by_id(sorcery, "global", "global");
```
jcolp: Same question applies here

channels/chan_websocket.c
```diff
@@ -2066,6 +2030,11 @@ static int global_apply(const struct ast_sorcery *sorcery, void *obj)
 	ast_debug(1, "control_msg_format: %s\n",
 		control_msg_format_to_str(cfg->control_msg_format));
 
+	if (cfg->write_timeout <= 0) {
```
jcolp, line 2114: If write_timeout is invalid if negative, then can it just match the unsigned int write_timeout on the websocket API?

include/asterisk/websocket_client.h
```diff
@@ -105,6 +106,7 @@ struct ast_websocket_client {
 	int pingpongs;                   /*!< Enable WebSocket PING/PONGs */
 	unsigned int pingpong_interval;  /*!< Send WebSocket PINGs at this interval in seconds */
 	unsigned int pingpong_probes;    /*!< Close connection after this many missed PONG responses */
+	int write_timeout;               /*!< Write timeout (ms) */
```
jcolp, line 109: Same question here about int vs unsigned int

res/ari/ari_websockets.c
```diff
@@ -844,6 +840,9 @@ static void websocket_established_cb(struct ast_websocket *ast_ws_session,
 			remote_addr, session_id);
 	}
 
+	ast_websocket_set_timeout(ast_ws_session,
```
jcolp: Is it possible for general to be NULL?

res/ari/config.c
```diff
@@ -740,7 +745,7 @@ static int ari_conf_init(void)
 	ast_sorcery_register_bool(general, ari_conf_general, enabled, enabled, "yes");
 	ast_sorcery_register_cust(general, pretty, "no");
 	ast_sorcery_register_int(general, ari_conf_general, websocket_write_timeout, write_timeout,
-		AST_DEFAULT_WEBSOCKET_WRITE_TIMEOUT);
+		INT_MAX);
```
jcolp: For the configuration option registrations where INT_MAX is used a comment should be added so that if someone comes along in the future and looks at the documentation and this they don't get immediately confused

jcolp (changes requested):

channels/chan_websocket.c
gtjoseph: The unsigned in ast_websocket_client_options was a holdover from an earlier version.  It should be an int.

channels/chan_websocket.c
gtjoseph: Nope.  Fixed.

channels/chan_websocket.c
gtjoseph: No because this is the only place you can set the write-timeout when we're the websocket server.

res/ari/ari_websockets.c
gtjoseph: Yes.  sigh.

gtjoseph: Heh, actually no because res_ari checks it to determine if ARI is even enabled.

res/ari/config.c
gtjoseph: Here and in chan_websocket I reverted back to using AST_DEFAULT_WEBSOCKET_WRITE_TIMEOUT as the default but I added the comment to res_websocket_client.

configs/samples/chan_websocket.conf.sample
```diff
@@ -8,3 +8,15 @@
                                      ; json: All messages are properly formatted
                                      ; JSON.
                                      ; Default: plain-text
+
+;write_timeout = 100                 ; Write timeout in milliseconds.
+                                     ; The maximum number of milliseconds to wait for a write
+                                     ; to the WebSocket to succeed.  The WebSocket will be closed
+                                     ; and the channel hung up when the timeout is reached.
+                                     ; Outgoing WebSocket connections can overwride this in
```
jcolp: override

res/res_websocket_client.c
```diff
@@ -803,6 +862,11 @@ static int load_module(void)
 	ast_sorcery_register_bool(websocket_client, ast_websocket_client, enable_pingpongs, pingpongs, "no");
 	ast_sorcery_register_uint(websocket_client, ast_websocket_client, pingpong_interval, pingpong_interval, 20);
 	ast_sorcery_register_uint(websocket_client, ast_websocket_client, pingpong_probes, pingpong_probes, 3);
+	/*
+	 * The default for write_timeout needs to be INT_MAX so we can tell that it's not set.
+	 * This allows the modules using this capability to apply their own default value.
+	 */
+	ast_sorcery_register_int(websocket_client, ast_websocket_client, write_timeout, write_timeout, INT_MAX);
```
jcolp, line 869: Will this suffer from the same expression issue that DEFAULT_RECONNECT_ATTEMPTS mentions?

jcolp (dismissed):

res/res_websocket_client.c
gtjoseph: No. That issue is specific to INT_MIN which is a macro `(-INT_MAX - 1)`.
INT_MAX is defined as `__INT_MAX__` which is a scalar provided by the compiler.

jcolp (approved):

Issue #1954, opened 2026-05-26
[bug]: shift count overflow building app_dial.c on ppc64


barf:
### Severity

Minor

### Versions

Asterisk GIT-22-22.0.0-pre1-471-g621784cd6c

### Components/Modules

app_dial.c

### Operating Environment

Debian forky (ppc64)

### Frequency of Occurrence

Constant

### Issue Description

shift count overflow when building app_dial.c on PowerPC (big endian)

the error is included into app_dial.c by `include/asterisk/utils.h`

### Relevant log output

```shell
$ ./configure --enable-dev-mode --enable-coverage --enable-year2038 --without-pjproject-bundled
$ make
...

   [CC] app_dial.c -> app_dial.o
In file included from /home/barf/src/asterisk-22/include/asterisk/alertpipe.h:22,
                 from /home/barf/src/asterisk-22/include/asterisk/channel.h:125,
                 from app_dial.c:43:
app_dial.c: In function ‘do_forward’:
/home/barf/src/asterisk-22/include/asterisk/utils.h:131:44: error: left shift count >= width of type [-Werror=shift-count-overflow]
  131 | #define SWAP64_32(flags) (((uint64_t)flags << 32) | ((uint64_t)flags >> 32))
      |                                            ^~
/home/barf/src/asterisk-22/include/asterisk/utils.h:144:55: note: in expansion of macro ‘SWAP64_32’
  144 |                                         ((p)->flags & SWAP64_32(flag)); \
      |                                                       ^~~~~~~~~
app_dial.c:794:45: note: in expansion of macro ‘ast_test_flag64’
  794 | #define CAN_EARLY_BRIDGE(flags,chan,peer) (!ast_test_flag64(flags, OPT_CALLEE_HANGUP | \
      |                                             ^~~~~~~~~~~~~~~
app_dial.c:1042:54: note: in expansion of macro ‘CAN_EARLY_BRIDGE’
 1042 |                 if (single && !caller_entertained && CAN_EARLY_BRIDGE(peerflags, c, in)) {
      |                                                      ^~~~~~~~~~~~~~~~
/home/barf/src/asterisk-22/include/asterisk/utils.h:131:70: error: right shift count >= width of type [-Werror=shift-count-overflow]
  131 | #define SWAP64_32(flags) (((uint64_t)flags << 32) | ((uint64_t)flags >> 32))
      |                                                                      ^~
[...]

PR #2142, opened 2026-09-04
utils: fix SWAP64_32 macro precedence bug on big-endian

hector-cao:
on my big endian S390x system:

$ asterisk -rx "core waitfullybooted"

gives an interactive prompt instead of executing the command

--- Proposed fix

The SWAP64_32 macro did not parenthesize its argument. When a combined flag mask such as (AST_OPT_FLAG_EXEC | AST_OPT_FLAG_NO_COLOR) was passed, the '<<' shift bound tighter than the '|', so the wrong bits were set. On big-endian systems (e.g. s390x) this caused ast_set_flag64() and a later single-flag ast_test_flag64() to disagree, so 'asterisk -rx <cmd>' dropped into the interactive console instead of executing the command.

Parenthesize the macro argument so the whole expression is evaluated before the word swap. Little-endian was unaffected.

### DIFF
--- a/include/asterisk/utils.h
+++ b/include/asterisk/utils.h
@@ -128,7 +128,7 @@ extern unsigned int __unsigned_int_flags_dummy;
  * \retval The flags with the upper and lower 32 bits swapped if the system is big-endian,
  */
 #if defined(BYTE_ORDER) && (BYTE_ORDER == BIG_ENDIAN)
-#define SWAP64_32(flags) (((uint64_t)flags << 32) | ((uint64_t)flags >> 32))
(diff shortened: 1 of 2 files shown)


seanbright:
I wonder if this fixes #1954 too

hector-cao:
> I wonder if this fixes #2142 too

Hello @seanbright 

#2142 ?

seanbright:
> > I wonder if this fixes #2142 too
> 
> Hello @seanbright
> 
> #2142 ?

Woops! Meant #1954

jcolp:
Please take care of the checklist items:

https://github.com/asterisk/asterisk/pull/2142#pullrequestreview-5114375626

PR #2144, opened 2026-09-05
res_pjsip_session: Add timers_refresher endpoint option

richard611:
res_pjsip_session: Add timers_refresher endpoint option
    
    When an incoming INVITE carries a Session-Expires header without a
    refresher parameter, pjsip defaults to the UAC as the session timer
    refresher. Some peer devices have issues with this default, so there
    is a need to be able to adjust which side refreshes the session for a
    given endpoint.
    
    This change adds the endpoint option timers_refresher with the values
    auto, uas and uac. On an inbound call leg, when the remote sends a
    Session-Expires header without a refresher parameter, the configured
    value is applied before the session timers are negotiated so the
    desired side refreshes the session. auto is the default and keeps the
    normal pjsip negotiation, uas makes this side (the UAS) refresh and uac
    makes the remote side (the UAC) refresh. An explicit refresher parameter
    sent by the remote party is always respected.
    
    UpgradeNote: An Alembic migration script has been added that creates
    the timers_refresher column on the ps_endpoints table for users of the
    realtime database backend.
    
    UserNote: A new endpoint option `timers_refresher` has been added to
    pjsip.conf that lets you control which side refreshes the session when
    the remote sends Session-Expires without a refresher parameter.

### DIFF
--- a/res/res_pjsip/pjsip_config.xml
+++ b/res/res_pjsip/pjsip_config.xml
@@ -895,6 +895,29 @@
 						Maximum session timer expiration period. Time in seconds.
 					</para></description>
 				</configOption>
+				<configOption name="timers_refresher" default="auto">
+					<synopsis>Preferred session timer refresher for inbound call legs</synopsis>
+					<description>
+						<para>Controls which side refreshes the session when the
+						remote party sends a <literal>Session-Expires</literal>
+						header without a <literal>refresher</literal> parameter.
+						</para>
+						<enumlist>
+							<enum name="auto">
+								<para>Do not force a refresher. The session timer
+								refresher is selected by normal negotiation.</para>
+							</enum>
+							<enum name="uas">
+								<para>This side, as the UAS on the inbound call leg,
+								refreshes the session.</para>
+							</enum>
+							<enum name="uac">
+								<para>The remote side, as the UAC, refreshes the
+								session.</para>
+							</enum>
+						</enumlist>
+					</description>
+				</configOption>
 				<configOption name="transport">
 					<since>
 						<version>12.0.0</version>
--- a/res/res_pjsip/pjsip_configuration.c
+++ b/res/res_pjsip/pjsip_configuration.c
@@ -257,6 +257,41 @@ static int timers_to_str(const void *obj, const intptr_t *args, char **buf)
 	return 0;
 }
 
+static const char *timer_refresher_map[] = {
+	[AST_SIP_TIMER_REFRESHER_AUTO] = "auto",
+	[AST_SIP_TIMER_REFRESHER_UAS] = "uas",
+	[AST_SIP_TIMER_REFRESHER_UAC] = "uac",
+};
+
+static int timers_refresher_handler(const struct aco_option *opt, struct ast_variable *var, void *obj)
+{
+	struct ast_sip_endpoint *endpoint = obj;
+
+	if (!strcasecmp(var->value, "auto")) {
+		endpoint->extensions.refresher = AST_SIP_TIMER_REFRESHER_AUTO;
+	} else if (!strcasecmp(var->value, "uas")) {
+		endpoint->extensions.refresher = AST_SIP_TIMER_REFRESHER_UAS;
+	} else if (!strcasecmp(var->value, "uac")) {
+		endpoint->extensions.refresher = AST_SIP_TIMER_REFRESHER_UAC;
+	} else {
+		ast_log(LOG_NOTICE, "Unrecognized option value %s for %s on endpoint %s\n",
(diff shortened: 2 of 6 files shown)


richard611:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seanbright:
> Some peer devices have issues with this default

Without this PR - Is Asterisk behaving correctly according to RFC 4028?

Also, please include a list of the devices that you are referring to.

richard611:
> > Some peer devices have issues with this default
> 
> Without this PR - Is Asterisk behaving correctly according to RFC 4028?
> 
> Also, please include a list of the devices that you are referring to.

Without this PR - Is Asterisk behaving correctly according to RFC 4028: Yes. Under RFC 4028 §9 Table 2 row 4, when the remote UAC supports "timer" and sends Session-Expires without a refresher parameter, the
UAS may choose either side as refresher. Without this PR Asterisk (via pjsip) selects "uac" and designates the peer as refresher, which is fully compliant; if the peer then never refreshes, RFC 4028 §10 requires the session to be torn down, which is what we observe.
```
       UAC supports?  refresher parameter  refresher parameter
                           in request           in response
       -------------------------------------------------------
             N                none                 uas
             N                uac                  NA
             N                uas                  NA
             Y                none             uas or uac
             Y                uac                  uac
             Y                uas                  uas

                        Table 2:  UAS Behavior
```

Devices: ZTE Softswitch. When the ZTE softswitch is under load it fails to process the  UPDATE refresh requests it is responsible for as the designated refresher, so the sessions expire and the calls drop. This option lets us steer the refresher role to our side ("uas") for those inbound legs.

jcolp:
So just a ZTE softswitch not working properly is the only known device?

Is that documented or something?

richard611:
> So just a ZTE softswitch not working properly is the only known device?
> 
> Is that documented or something?

Yes, ZTE Softswitch is the only device we've seen this with so far. It isn't documented anywhere that I'm aware of - it's an operational observation from our deployment, not a published ZTE erratum. What we observe is that only under high concurrency does the ZTE fail to handle the UPDATE-based session refresh correctly, so the session expires and the
call is dropped. Having the UAS side (Asterisk) be the refresher avoids those drops. Per RFC 4028 §9 (Table 2, row 4), when the UAC omits the refresher parameter the UAS is allowed to choose "uac" or "uas", so selecting "uas" here is within spec.

res/res_pjsip_session.c
```diff
@@ -3955,6 +3956,45 @@ static int check_content_disposition(pjsip_rx_data *rdata)
...
+	const struct ast_sip_endpoint *endpoint = session->endpoint;
+	static const pj_str_t str_session_expires = {"Session-Expires", 15};
+	static const pj_str_t str_session_expires_short = {"x", 1};
+	pjsip_sess_expires_hdr *se_hdr;
+
+	if (endpoint->extensions.refresher == AST_SIP_TIMER_REFRESHER_AUTO) {
+		return;
+	}
+
+	se_hdr = (pjsip_sess_expires_hdr *) pjsip_msg_find_hdr_by_names(
```
jcolp, line 3982: I'm not happy with the approach of mutating the received SIP request to alter behavior. An actual API approach should be done with the PJSIP library to allow this to explicitly be specified for the scenario, and then that provided upstream for review/comment.

res/res_pjsip/pjsip_configuration.c
```diff
@@ -257,6 +257,41 @@ static int timers_to_str(const void *obj, const intptr_t *args, char **buf)
...
+	return 0;
+}
+
+static int timers_refresher_to_str(const void *obj, const intptr_t *args, char **buf)
+{
+	const struct ast_sip_endpoint *endpoint = obj;
+
+	if (ARRAY_IN_BOUNDS(endpoint->extensions.refresher, timer_refresher_map)) {
+		*buf = ast_strdup(timer_refresher_map[endpoint->extensions.refresher]);
+	}
```
jcolp, line 291: This can result in *buf being nothing, so something like <unknown> should be set if it is truly unknown.

res/res_pjsip/pjsip_config.xml
```diff
@@ -895,6 +895,29 @@
 						Maximum session timer expiration period. Time in seconds.
 					</para></description>
 				</configOption>
+				<configOption name="timers_refresher" default="auto">
+					<synopsis>Preferred session timer refresher for inbound call legs</synopsis>
```
jcolp, line 899: This will also need the versions specified

jcolp (changes requested):

PR #2146, opened 2026-09-06
res_resolver_unbound: Include libunbound error reason

### LINKED ISSUE #2110: [bug]: res_resolver_unbound: libunbound API calls that fail should include reason
### Severity

Minor

### Versions

22.10.1

### Components/Modules

`res_resolver_unbound.so` loaded, `pjsip` using `type=registration`

### Operating Environment

 Ubuntu 24.04.4 LTS (Noble Numbat), kernel 6.8.0-137-generic, x86_64, Asterisk built from source

### Frequency of Occurrence

Constant

### Issue Description

Summary

`res_resolver_unbound.so` continuously fails to resolve a valid hostname (`sip.brcel.com`), even though:
- the operating system's own DNS resolves the same hostname without any problem (`dig`, `getent`);
- `/etc/asterisk/resolver_unbound.conf` was explicitly configured with `nameserver` entries pointing to public resolvers that work fine outside Asterisk;
- also tested with `resolv = system` (pointing to the same `/etc/resolv.conf` that already works).

The only workaround that actually fixed it was disabling the module entirely (`noload = res_resolver_unbound.so` in `modules.conf`), which makes Asterisk fall back to the default resolver.

This looks like the same behavior reported (and closed without a root cause) in [ASTERISK-28290](https://issues-archive.asterisk.org/ASTERISK-28290) (2019, Asterisk 16.1.1) — same error text, same workaround, still present 7 years later on Asterisk 22.10.1.

## Environment

- Asterisk 22.10.1 (built from source)
- Ubuntu (Linux kernel, VPS server)
- `res_resolver_unbound.so` loaded, `pjsip` using `type=registration` (outbound registration) for a real VoIP provider DID/trunk (brcel.com.br)

## Symptom

Log repeated indefinitely, on every outbound registration attempt (every 60s, `retry_interval=60`):

```
[Aug 20 20:06:53] ERROR[723846] res_resolver_unbound.c: Failed to perform async DNS resolution of 'sip.brcel.com'
[Aug 20 20:06:53] WARNING[723846] res_pjsip_outbound_registration.c: No response received from 'sip:sip.brcel.com:5060' on registration attempt to 'sip:1151978427@sip.brcel.com:5060', retrying in '60'
```

This repeated indefinitely (confirmed for ~15 consecutive minutes, one attempt per minute) until `Maximum retries reached`.

## Evidence that system-level DNS works fine

```
$ cat /etc/resolv.conf
nameserver 127.0.0.53   # systemd-resolved stub
...

$ getent hosts sip.brcel.com
186.209.52.196  sip.brcel.com

$ dig +short sip.brcel.com @8.8.8.8
186.209.52.196

$ resolvectl status | grep -A3 "DNS Servers"
DNS Servers: 153.92.2.6 1.1.1.1 8.8.4.4
```

Resolution works through every path tested outside of Asterisk (local stub resolver, direct public server).

## What was tried and did NOT fix it

**1. `/etc/asterisk/resolver_unbound.conf` with explicit nameservers** (syntax verified against `res_resolver_unbound.c` source of the installed version — option name `nameserver`, filename `resolver_unbound.conf` confirmed correct):

```ini
[general]
nameserver = 1.1.1.1
nameserver = 8.8.4.4
nameserver = 153.92.2.6
```

`module reload res_resolver_unbound.so` reported success, and a full Asterisk process restart was also done (not just a reload) — the error was identical afterward in both cases.

**2. `resolv = system`** (use the same `/etc/resolv.conf` that already works for `dig`/`getent`):

```ini
[general]
resolv = system
```

Full Asterisk restart performed — error was identical afterward.

## Possible lead: DNSSEC

`delv` (a real DNSSEC-validating tool, unlike `dig`/`getent` which don't validate by default) reports that the hostname's zone has an invalid DNSSEC signature:

```
$ delv sip.brcel.com @1.1.1.1
;; validating sip.brcel.com/A: no valid signature found
; unsigned answer
sip.brcel.com.  300  IN  A     186.209.52.196
sip.brcel.com.  300  IN  RRSIG A 13 3 300 ...
```

So this is a domain with a misconfigured DNSSEC setup on the hostname's own DNS provider (RRSIG present but not valid). **We are not certain this is the root cause** — reading the source (`res_resolver_unbound.c`), DNSSEC validation only appears to be enabled if `ta_file` is configured (`ub_ctx_add_ta_file`), and we never set that option in any of our tests. Still, it's possible the bundled unbound context has some default trust anchor compiled in, or that a malformed RRSIG breaks resolution even without strict validation enabled. Leaving this here as a lead for whoever investigates the root cause.

## What fixed it

`/etc/asterisk/modules.conf`:

```ini
[modules]
autoload=yes
noload = res_resolver_unbound.so
```

+ a full Asterisk restart (kill the process + start — `systemctl restart` via the packaged LSB init script did not reliably kill the process in this setup, but that's a quirk of the package/init script, not of the bug itself).

After that, **zero occurrences** of the `Failed to perform async DNS resolution` error in the log, and outbound REGISTER started going out correctly with the hostname resolved (confirmed via packet capture — `tcpdump`):

```
21:00:26 eth0 Out IP <server> > 186.209.52.196.5060: SIP: REGISTER sip:sip.brcel.com:5060 SIP/2.0
```

(the provider's own non-response is a completely separate issue — a source-IP restriction on their end, unrelated to this Asterisk bug.)

## Question for maintainers

- Why doesn't `resolver_unbound.conf` (`nameserver` or `resolv = system`) change the failure behavior at all, even with filename/option syntax verified against the source of the installed version?
- Is it expected that a domain with misconfigured DNSSEC breaks resolution via `res_resolver_unbound` even without `ta_file` configured (i.e., without DNSSEC validation properly enabled)?
- Is there any documented way to make `res_resolver_unbound` actually ignore/skip DNSSEC when no `ta_file` is set, since the observed behavior suggests it isn't simply passing through an unvalidated domain?

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

seifzadeh:
Failures returned by ub_resolve_async() are currently logged without the
reason provided by libunbound, making resolver configuration failures
difficult to diagnose.

Include ub_strerror() in the error message so the underlying libunbound
failure reason is available in the Asterisk log.

Resolves: #2110

### DIFF
--- a/res/res_resolver_unbound.c
+++ b/res/res_resolver_unbound.c
@@ -329,8 +329,8 @@ static int unbound_resolver_resolve(struct ast_dns_query *query)
 		ao2_bump(query), unbound_resolver_callback, &data->id);
 
 	if (res) {
-		ast_log(LOG_ERROR, "Failed to perform async DNS resolution of '%s'\n",
-			ast_dns_query_get_name(query));
+		ast_log(LOG_ERROR, "Failed to perform async DNS resolution of '%s': %s\n",
+			ast_dns_query_get_name(query), ub_strerror(res));
 		ao2_ref(query, -1);
 	}
 


seifzadeh:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seifzadeh:
AI usage: ChatGPT was used to assist with issue analysis, reproduction planning, branch/cherry-pick verification, and review of the proposed change. I reviewed and tested the change myself.

jcolp (approved):

PR #2147, opened 2026-09-07
res_config_pgsql: Quote identifiers when needed

### LINKED ISSUE #1655: [bug]: res_conf_pgsql realtime driver does not properly quote column names
### Severity

Medium

### Versions

23.1.0

### Components/Modules

res_config_pgsql

### Operating Environment

Debian 13

### Frequency of Occurrence

Constant

## Description

The Postgres Realtime driver (`res_config_pgsql.c`) does not properly quote column names in SQL queries, causing Postgres to reject queries when column names start with digits, contain special characters, or are reserved keywords.

## Steps to Reproduce

1. Follow the ARI Push Configuration guide: https://docs.asterisk.org/Configuration/Interfaces/Asterisk-REST-Interface-ARI/The-Asterisk-Resource/ARI-Push-Configuration
2. Replace the `odbc` configuration with `pgsql` configuration
3. Use provided Alembic migrations with Postgres config
4. Attempt to create or update an endpoint via ARI

## Expected Behavior

The INSERT/UPDATE queries should succeed with all valid PJSIP column names.

## Actual Behavior

Postgres rejects the query with a syntax error:

```
ERROR[315]: res_config_pgsql.c:168 _pgsql_exec: PostgreSQL RealTime: Query Failed: INSERT INTO ps_endpoints (id, rpid_immediate, webrtc, ..., 100rel, timers, direct_media, ...) VALUES (...)
ERROR[315]: res_config_pgsql.c:169 _pgsql_exec: PostgreSQL RealTime: Query Failed because: ERROR:  trailing junk after numeric literal at or near "100rel"
LINE 1: ..._setup, connected_line_method, g726_non_standard, 100rel, ti...
                                                             ^
 (PGRES_FATAL_ERROR)
```

## Root Cause

PostgreSQL interprets `100rel` as a numeric literal. In Postgres, identifiers that start with digits must be enclosed in double quotes. It appears that a similar issue has been fixed in Alembic migrations (#274), but never in this driver.

It appears that this might have been attempted to handle, but incorrectly? I can see:
https://github.com/asterisk/asterisk/blob/20e40a9a4866289e73ef157b121d32830bf06cb6/res/res_config_pgsql.c#L1015
Where `ESCAPE_STRING` uses `PQescapeStringConn`, but for column names I think [PQescapeIdentifier](https://www.postgresql.org/docs/18/libpq-exec.html#LIBPQ-PQESCAPEIDENTIFIER) should be used?

## Workaround

Currently, there is no workaround other than using ODBC driver instead of native Postgres driver. I was a bit surprised since such bug would imply pretty much no one uses this driver, as I would imagine almost anyone using it would attempt to insert endpoint with it?

### Relevant log output

```shell
ERROR[315]: res_config_pgsql.c:168 _pgsql_exec: PostgreSQL RealTime: Query Failed: INSERT INTO ps_endpoints (id, rpid_immediate, webrtc, ..., 100rel, timers, direct_media, ...) VALUES (...)
ERROR[315]: res_config_pgsql.c:169 _pgsql_exec: PostgreSQL RealTime: Query Failed because: ERROR:  trailing junk after numeric literal at or near "100rel"
LINE 1: ..._setup, connected_line_method, g726_non_standard, 100rel, ti...
                                                             ^
 (PGRES_FATAL_ERROR)
```

### Asterisk Issue Guidelines

Jayllyz:
Fixes #1655.

Identifiers PostgreSQL can't accept unquoted (e.g. `100rel`) are now
double-quoted in all realtime queries. Valid identifiers are left as-is,
so existing installs keep their case-insensitive behaviour, and
`schema.table` is quoted per part.

New `quote_identifiers` option (default `no`) quotes everything, for
reserved-keyword columns. Names then become case-sensitive.

Adds unit test `/res/res_config_pgsql/test_pgsql_quote_identifier`.

### DIFF
--- a/res/res_config_pgsql.c
+++ b/res/res_config_pgsql.c
@@ -39,6 +39,7 @@
 #include "asterisk/lock.h"
 #include "asterisk/utils.h"
 #include "asterisk/cli.h"
+#include "asterisk/test.h"
 
 AST_MUTEX_DEFINE_STATIC(pgsql_lock);
 AST_THREADSTORAGE(sql_buf);
@@ -46,6 +47,8 @@ AST_THREADSTORAGE(findtable_buf);
 AST_THREADSTORAGE(where_buf);
 AST_THREADSTORAGE(escapebuf_buf);
 AST_THREADSTORAGE(semibuf_buf);
+AST_THREADSTORAGE(identifierbuf_buf);
+AST_THREADSTORAGE(tablebuf_buf);
 
 #define RES_CONFIG_PGSQL_CONF "res_pgsql.conf"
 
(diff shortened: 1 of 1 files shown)


res/res_config_pgsql.c
```diff
@@ -388,12 +391,70 @@ static struct columns *find_column(struct tables *t, const char *colname)
...
+ * \brief Append a double-quoted, properly escaped SQL identifier to buf.
+ *
+ * PostgreSQL identifiers that start with a digit, contain special
+ * characters, or collide with reserved keywords must be double-quoted.
+ * Any double quote within the identifier itself must be doubled.
+ *
+ * \param buf pointer to destination ast_str
+ * \param identifier the raw (unquoted) identifier, e.g. a column or table name
+ */
+static void pgsql_append_quoted_identifier(struct ast_str **buf, const char *identifier)
```
jcolp, line 404: This module supports schema.table format. By quoting the entire thing does that break schema.table support?

This also introduces a behavior change. Unquoted is case-insensitive, quoted is case-sensitive[1]. I could see that breaking existing some installs so I think it should be optional and default to off to preserve existing behavior.

[1] https://www.postgresql.org/docs/current/sql-syntax-lexical.html

res/res_config_pgsql.c
```diff
@@ -388,12 +391,70 @@ static struct columns *find_column(struct tables *t, const char *colname)
...
+{
+	const char *chunk;
+
+	ast_str_append(buf, 0, "%s", "\"");
+	for (chunk = identifier; *chunk; chunk++) {
```
jcolp, line 409: Is it possible for identifier to be NULL?

jcolp (changes requested):

jcolp (changes requested):
The commit message needs to be reformatted and use UserNote[1] so that the new option appears in the release notes so people are aware.

[1] https://docs.asterisk.org/Development/Policies-and-Procedures/Commit-Messages/?h=commit+messages#special-trailers-for-commit-messages

PR #2155, opened 2026-09-10
app_amd: Correct and clarify the AMD documentation.

### LINKED ISSUE #2154: [bug]: app_amd: AMD documentation does not match the implementation
### Severity

Trivial

### Versions

20.21.0, 22.11.0, 23.5.0

### Components/Modules

app_amd

### Operating Environment

n/a

### Frequency of Occurrence

None

### Issue Description

The documentation block in apps/app_amd.c has drifted from the code.

- Two parameter names are misspelled: `miniumWordLength` and  `betweenWordSilence`, instead of `minimumWordLength` and `betweenWordsSilence`.

- The `NOAUDIODATA` value of AMDCAUSE is undocumented, although it is set whenever the analysis window expires without a single voice or CNG frame.

- `MAXWORDLENGTH` is documented as "Word Length - max length of a single word", implying two appended values, but only one is appended: `sprintf(amdCause, "MAXWORDLENGTH-%d", consecutiveVoiceDuration)`

- Nothing states how AMDCAUSE is composed: the measurement that triggered the cause, and for most causes the value of the parameter whose limit it breached, appended and separated by '-'.

- The time-valued parameters do not state that they are in milliseconds.

- It is not documented that AMDSTATUS and AMDCAUSE are both set to the empty string when the detection cannot be performed at all.

- `maximumWordLength` is described as the maximum duration of "a word", but the check measures a run of continuous voice, which may never have been counted as a word.

- Grammatical error: "Those default values get overwritten when the calling AMD with parameters."

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

jlaine:
Fix the miniumWordLength and betweenWordSilence parameter names, which did not match the arguments AMD() accepts.

Document the NOAUDIODATA cause, explain how values are appended to AMDCAUSE, and name the parameter each cause reports. MAXWORDLENGTH appends a single value, so its phantom second field is removed. Also note that both variables are set to the empty string when detection cannot be performed, state the millisecond unit once, and describe maximumWordLength as continuous voice, which is what the check measures.

Resolves: #2154

### DIFF
--- a/apps/app_amd.c
+++ b/apps/app_amd.c
@@ -77,10 +77,10 @@
 				<para>Is the maximum time allowed for the algorithm</para>
 				<para>to decide on whether the audio represents a HUMAN, or a MACHINE</para>
 			</parameter>
-			<parameter name="miniumWordLength" required="false">
+			<parameter name="minimumWordLength" required="false">
 				<para>Is the minimum duration of Voice considered to be a word</para>
 			</parameter>
-			<parameter name="betweenWordSilence" required="false">
+			<parameter name="betweenWordsSilence" required="false">
 				<para>Is the minimum duration of silence after a word to
 				consider the audio that follows to be a new word</para>
 			</parameter>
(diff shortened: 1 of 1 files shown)


jlaine:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

jcolp:
For future reference 24 is now a supported branch. I have adjusted your comment accordingly.

jlaine:
> For future reference 24 is now a supported branch. I have adjusted your comment accordingly.

Noted for future PRs!

jcolp (approved):

PR #2156, opened 2026-09-10
app_externalivr: Process buffered commands

### LINKED ISSUE #2148: [bug]: ExternalIVR can lose a command line buffered by ast_iostream_gets(), leaving the channel stuck in the application
### Severity

Major

### Versions

23.0.0

### Components/Modules

apps/app_externalivr.c

### Operating Environment

`Debian GNU/Linux 12 (bookworm), x86_64, kernel 6.1`

### Frequency of Occurrence

Occasional

### Issue Description

> **Disclaimer on authorship:** this report was drafted by an AI agent — Anthropic's
> **Claude Opus 5** (model ID `claude-opus-5[1m]`), running in Claude Code — during
> analysis of a production incident, and reviewed by a human before submission. The
> diagnosis, code references and timings were derived from a live system and from the
> Asterisk source; the wording is the model's. Please treat the reasoning as you would any
> third-party analysis and verify it against the code. If any part is unclear or looks
> wrong, say so and a human will follow up directly.

## Summary

`ExternalIVR` reads commands from the child process with `ast_iostream_gets()`, which is
**buffered**, but decides *whether to read* by polling the **raw file descriptor** with
`ast_waitfor_nandfds()`. When two or more command lines arrive in a single `read()`, the
surplus is left in the iostream's internal buffer, where the poll cannot see it. If no
further bytes ever arrive on that descriptor, the buffered command is **never processed**.

When the lost command is `E` (exit) or `H` (hangup) — commands a child typically sends last,
and therefore exactly the ones with nothing following them — `ExternalIVR` never terminates.
The channel remains up in the application indefinitely, the caller hears the silence
generator, and the dialplan never advances.

## The mechanism

In `eivr_comm()` (`apps/app_externalivr.c`):

```c
waitfds[0] = ast_iostream_get_fd(eivr_commands);          /* master:645 - the RAW fd */
...
rchan = ast_waitfor_nandfds(&chan, 1, waitfds, ..., &ready_fd, &ms);   /* master:672 */
...
} else if (ready_fd == waitfds[0]) {                       /* master:720 */
        ...
        r = ast_iostream_gets(eivr_commands, input, sizeof(input));    /* master:726 */
```

`ast_iostream_gets()` (`main/iostream.c`) reads up to `sizeof(stream->rbuf)` — **2048
bytes** — into the stream's buffer and returns only the first `\n`-terminated line,
retaining the remainder in `stream->rbufhead` / `stream->rbuflen`:

```c
len = iostream_read(stream, stream->rbuf, sizeof(stream->rbuf));
...
stream->rbuflen += len;
```

So one `read()` can consume several command lines from the pipe while only one is acted on.
The kernel-side pipe is then empty, `ast_waitfor_nandfds()` reports the descriptor
not-ready on every subsequent iteration, and the `else if (ready_fd == waitfds[0])` branch
is never entered again. Any command still sitting in `rbuf` is stranded permanently.

Two things make this easier to hit than it first looks:

1. **The branch is `else if`.** A channel with a frame ready takes priority over a ready
   command descriptor in the same iteration. On an answered channel carrying audio there is
   a frame roughly every 20 ms, so the command descriptor is only ever serviced
   opportunistically.
2. **Nothing ever flushes the buffer.** There is no "drain what is buffered" step, and the
   loop's only other exits are hangup, child death and `ast_waitfor_nandfds()` failure. A
   channel that stays up simply spins on audio frames forever.

The same pattern is used for the error stream at `master:883`.

## Observed failure

A child playing a single prompt received the `F` (file finished) event, then wrote three
command lines back-to-back, roughly 500 µs apart:

```
V,<var1>=<value1>
V,<var2>=<value2>
E,<reason>
```

Result:

* **Both `V` commands took effect** — both channel variables were confirmed set with
  `core show channel`. So Asterisk read and dispatched through line 2.
* **The `E` command was never acted on.** The channel stayed `Up` in `ExternalIVR` for
  **25 minutes 34 seconds**, until it was hung up manually from the CLI.
* The child remained blocked in `read()` on its stdin pipe for that entire period.

That last point is what rules out the obvious alternative — that `E` *was* processed and
the application then hung during cleanup. On `break`, `eivr_comm()` returns and `app_exec()`
closes `stream_stdin` immediately, which would give the child EOF. The child never saw EOF,
so `eivr_comm()` had not returned: the application was still in its `while (1)` loop, having
never seen the `E`.

The most likely sequence, consistent with the ~500 µs write spacing: Asterisk read line 1
alone; lines 2 and 3 then arrived together and were pulled into `rbuf` by a single
`iostream_read()`; `ast_iostream_gets()` returned line 2; the pipe was now empty, so the
descriptor never went ready again and line 3 was stranded.

Hanging the channel up externally released it instantly — `ast_check_hangup()` became true,
the loop took the `H` path, closed the pipes, and the child received EOF and exited within
milliseconds. The pipe and the reader were healthy throughout; only the poll was blind.

## Impact

On one production platform, across roughly five months of call records, **18 calls** were
left in this state long enough to be unmistakable (over five minutes each), on 6 hosts. The
longest stranded a live caller in dead air for **42 minutes 56 seconds**, ending only when
the caller themselves hung up.

Two properties make this worse than the raw numbers suggest:

* **It is invisible.** Every one of those calls was recorded as a normal, successfully
  cleared call. Nothing in Asterisk logs, warns or otherwise marks the condition.
* **The true rate is unmeasurable.** A stranded `E` on a call where the caller hangs up
  quickly is indistinguishable from a normal call — the hangup drives the loop out through
  a different branch and everything cleans up correctly. Only the cases where the caller
  waited are detectable, so 18 is a floor on customer impact, not the defect's frequency.

## Reproduction

Not reliably reproducible on demand, since it depends on write timing, but the conditions
are straightforward:

1. Run `ExternalIVR` on an **answered** channel that is carrying audio (so the channel
   branch competes for loop iterations).
2. Have the child write **two or more command lines back-to-back** with no intervening
   events from Asterisk — for example a `V,...` immediately followed by `E,...`, which is a
   natural thing to do when setting result variables before exiting.
3. Repeat. Occasionally the trailing command is stranded and the application never exits.

Widening the write spacing makes it rarer but does not eliminate it; any two lines that land
in one `read()` can trigger it.

## Suggested fix

After a successful `ast_iostream_gets()`, continue processing while the iostream still holds
a complete buffered line, rather than returning to the poll after exactly one line. That
requires either a small accessor for "bytes currently buffered" or an internal drain loop in
`eivr_comm()`.

Separately — and independently worth considering — making the channel-frame branch not
exclusive of the command branch (`if` rather than `else if`, servicing both when both are
ready) would remove the starvation that makes this reachable in the first place.

Happy to test a patch against the affected production workload if that would help.

---

# Relevant log output

> Sanitised. `<uuid>` replaces channel identifiers, `<var>`/`<value>` replace channel
> variable names and values, and the application path is genericised.

```
Child (IVR application) debug log - the final events of the stuck call:

12:49:13  sendSounds()  -> wrote: S,<prompt-file>
12:49:39  received event: F,<timestamp>,<prompt-file>     (file finished playing)
12:49:39  last prompt in playlist -> wrote, ~500us apart:
              V,<var1>=<value1>
              V,<var2>=<value2>
              E,<reason>
12:49:39  blocked in read() on stdin - no further events received
...
13:14:45  received event: H          (channel hung up manually from the CLI, 25m34s later)
13:14:46  cleanup ran, child exited normally

Asterisk side, during those 25 minutes:

  core show channel <chan>
      State:        Up (6)
      Application:  ExternalIVR
      Data:         /path/to/ivr-app.php(<args>)
      Elapsed Time: 0h20m20s
      Variables:
        <var1>=<value1>          <-- both V commands applied
        <var2>=<value2>          <-- so the reader consumed through line 2

  No "Exiting: <reason>" NOTICE was ever emitted for this channel.
  (Note: on this system NOTICE is not routed to a log file, so this is
   consistent with the diagnosis but is not by itself evidence.)

Child process state while stuck (pid of the ExternalIVR child):

  /proc/<pid>/stack
      [<0>] pipe_read+0x3df/0x470
      [<0>] vfs_read+0x2db/0x310
      [<0>] ksys_read+0xb7/0xf0
      [<0>] do_syscall_64+0x55/0xb0

  /proc/<pid>/fd
      0 -> pipe:[...]      (stdin from Asterisk - blocked reading this)
      1 -> pipe:[...]
      2 -> pipe:[...]

  Blocked here for the full 25 minutes with the pipe still open, i.e. Asterisk
  had not closed stream_stdin, i.e. eivr_comm() had not returned.
```

### Relevant log output

```shell

```

### Asterisk Issue Guidelines

seifzadeh:
Process complete lines already buffered by ast_iostream_gets() before
waiting on the file descriptor again. This prevents trailing commands
such as E from being left unprocessed.

Resolves: #2148

DeveloperNote: Add ast_iostream_has_buffered_line().

### DIFF
--- a/apps/app_externalivr.c
+++ b/apps/app_externalivr.c
@@ -643,6 +643,8 @@ static int eivr_comm(struct ast_channel *chan, struct ivr_localuser *u,
 	struct ast_channel *rchan;
 	int res = -1;
 	int hangup_info_sent = 0;
+	int command_pending = 0;
+	int error_pending = 0;
 
 	waitfds[0] = ast_iostream_get_fd(eivr_commands);
 	waitfds[1] = eivr_errors ? ast_iostream_get_fd(eivr_errors) : -1;
@@ -669,7 +671,16 @@ static int eivr_comm(struct ast_channel *chan, struct ivr_localuser *u,
 		errno = 0;
 		exception = 0;
 
-		rchan = ast_waitfor_nandfds(&chan, 1, waitfds, (eivr_errors) ? 2 : 1, &exception, &ready_fd, &ms);
+		if (command_pending) {
+			rchan = NULL;
+			ready_fd = waitfds[0];
+		} else if (error_pending) {
+			rchan = NULL;
+			ready_fd = waitfds[1];
+		} else {
+			rchan = ast_waitfor_nandfds(&chan, 1, waitfds, (eivr_errors) ? 2 : 1,
+				&exception, &ready_fd, &ms);
+		}
 
 		if (ast_channel_state(chan) == AST_STATE_UP && !AST_LIST_EMPTY(&u->finishlist)) {
 			AST_LIST_LOCK(&u->finishlist);
@@ -724,6 +735,7 @@ static int eivr_comm(struct ast_channel *chan, struct ivr_localuser *u,
 			}
 
 			r = ast_iostream_gets(eivr_commands, input, sizeof(input));
+			command_pending = ast_iostream_has_buffered_line(eivr_commands);
 			if (r <= 0) {
 				if (r == 0) {
 					ast_chan_log(LOG_ERROR, chan, "Child process went away\n");
@@ -881,6 +893,7 @@ static int eivr_comm(struct ast_channel *chan, struct ivr_localuser *u,
 			}
 
 			r = ast_iostream_gets(eivr_errors, input, sizeof(input));
+			error_pending = ast_iostream_has_buffered_line(eivr_errors);
 			if (r > 0) {
 				ast_chan_log(LOG_NOTICE, chan, "stderr: %s\n", ast_strip(input));
 			} else if (r == 0) {
--- a/include/asterisk/iostream.h
+++ b/include/asterisk/iostream.h
@@ -142,6 +142,16 @@ int ast_iostream_get_fd(struct ast_iostream *stream);
 int ast_iostream_wait_for_input(struct ast_iostream *stream, int timeout);
 
 /*!
+ * \brief Check whether the iostream read buffer contains a complete line.
+ *
+ * \param stream A pointer to an iostream
+ *
+ * \retval 0 if no complete line is buffered
+ * \retval 1 if a complete line is buffered
+ */
+int ast_iostream_has_buffered_line(struct ast_iostream *stream);
+
+/*!
  * \brief Make an iostream non-blocking.
  *
  * \param stream A pointer to an iostream
--- a/main/iostream.c
+++ b/main/iostream.c
@@ -87,6 +87,11 @@ int ast_iostream_get_fd(struct ast_iostream *stream)
 	return stream->fd;
 }
 
+int ast_iostream_has_buffered_line(struct ast_iostream *stream)
+{
+	return stream->rbuflen && memchr(stream->rbufhead, '\n', stream->rbuflen) != NULL;
+}
+
 int ast_iostream_wait_for_input(struct ast_iostream *stream, int timeout)
 {
 #if defined(DO_SSL)


seifzadeh:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

seifzadeh:
AI disclosure: I used ChatGPT to assist with analysis, patch review, and drafting the reproduction/regression tests. I reviewed and built the changes locally and verified the fix against the original reproducer, partial-line buffering, and stderr buffering.

jcolp:
The original reporter should confirm that it resolves the issue for them.

seifzadeh:
Thanks Joshua.

@l7s, when you have a chance, could you please test this patch against the original affected environment/workload and confirm whether it resolves the issue for you?

l7s:
@seifzadeh we will test and report back within next 24 hours

PR #2160, opened 2026-09-11
chan_dahdi: Avoid transitioning device state from INUSE to RINGING.

### LINKED ISSUE #2159: [improvement]: chan_dahdi: Avoid INUSE -> RINGING device state transitions for analog lines
### Improvement Description

When an analog station receives a call, its device state is momentarily "INUSE" and then almost immediately transitions to "RINGING". This can cause problems for other CPE that have busy lamp fields (BLFs) which tend to consider this kind of transition to be illegal, and thus ignore it.

To avoid creating issues, we should avoid this transition and ensure that the first device state is RINGING rather than INUSE.

InterLinked1:
By default, DAHDI relies on the Asterisk code to provide most device
state, especially for analog lines, in which case it relies entirely
on the core. When an incoming call arrives to a DAHDI line, it
immediately asks the core to update device state for the channel.

This can sometimes lead to unoptimal outcomes; for example, when
ringing an FXS line (station), the device state is momentarily
"INUSE" and then almost immediately transitions to "RINGING". However,
this is incompatible with the state machines of some CPE. For example,
Polycom IP phones' BLFs will ignore any "RINGING" received after an
"INUSE". This makes sense because a line can't ring after being in use
without first being made idle. (The reverse case, RINGING followed by
INUSE, is allowed, since one can answer a ringing phone.)

Thus, to ensure that BLFs accurately display the status of DAHDI
channels, we need to avoid the momentary "INUSE" device state
for ringing channels.

Resolves: #2159

### DIFF
--- a/channels/chan_dahdi.c
+++ b/channels/chan_dahdi.c
@@ -9826,12 +9826,22 @@ static struct ast_channel *dahdi_new(struct dahdi_pvt *i, int state, int startpb
 	ast_jb_configure(tmp, &global_jbconf);
 
 	/* Set initial device state */
-	ast_copy_string(device_name, ast_channel_name(tmp), sizeof(device_name));
-	dashptr = strrchr(device_name, '-');
-	if (dashptr) {
-		*dashptr = '\0';
+	if (IS_FXO_SIG(i) && idx == SUB_REAL && state == AST_STATE_RESERVED && !startpbx && requestor) {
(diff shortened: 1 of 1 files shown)


InterLinked1:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (approved):

PR #2168, opened 2026-09-17
res_geolocation: Fix alembic scripts for MySQL.

gtjoseph:
Depending on the backend storage type, MySQL rows could have a maximum
length of 65535 bytes including storage overhead. The number and size of
the fixed-size string columns in the geoloc_profile table cause this to
be exceeded. Unfortunately, this isn't detected until you try to
actually create the table. To address this, the string columns types
have been changed to TEXT, which both MySQL and PostgreSQL handle
without counting to maximum row size. For long, variable-length strings,
TEXT is also more storage efficient.

MySQL also had an issue with the way the Enums were created.  These
were also fixed.


gtjoseph:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

pr-checklist-exception: alembic-upgrade

jcolp (approved):

PR #2169, opened 2026-09-19
file.c: Reduce number of allocations when searching for sound files

seanbright:
Depending on the requested format, the number of format modules loaded, and the order those modules were loaded, it's possible for dozens of small allocations to occur for each file playback on a channel.

So instead we now allocate a single buffer big enough for any registered file format extension and we use that for all lookups.


seanbright:
cherry-pick-to: 20
cherry-pick-to: 22
cherry-pick-to: 23
cherry-pick-to: 24

main/file.c
```diff
@@ -580,13 +645,11 @@ static int filehelper(const char *filename, const void *arg2, const char *fmt, c
 		stringp = storage;
 		while ( (ext = strsep(&stringp, "|")) ) {
 			struct stat st;
-			char *fn = build_filename(filename, ext);
 
-			if (fn == NULL)
-				continue;
+			/* Append the extension to our template */
+			strcpy(extension_start, translate_extension(ext));
```
jcolp, line 664: Since the result of translate_extension has to be stored within the same buffer size as the original extension it can't be larger, document that in the translate_extension function just in case

main/file.c
```diff
@@ -351,28 +365,69 @@ static int copy(const char *infile, const char *outfile)
...
+ * \internal
+ * \brief Allocate a filename template with room for all registered extensions
+ *
+ * \note You must hold the formats list lock when calling this function
+ *
+ * \param[in] filename The name of the file, which may be an absolute path
+ * \param[out] extension_start On return, points to the memory offset within the returned
+ * string where candidate extensions should be stored
+ *
+ * \retval An allocated string large enough to accomodate all registered filename
```
jcolp: accommodate

main/file.c
```diff
@@ -122,6 +122,9 @@ static int publish_format_update(const struct ast_format_def *f, struct stasis_m
 	return 0;
 }
 
+/* Protected by the formats list lock */
+static unsigned maximum_extension_length;
```
jcolp, line 134: This is only calculated on registration so if one is unloaded it won't shrink, I'd document it in case someone goes looking

jcolp (changes requested):

jcolp (approved):

PR #2171, opened 2026-09-20
audiohook.c: Fix frame leaks and timer starvation in whisper framehook.

jxmx:
Note: I've been trying to track down a bug in app_rpt (https://github.com/AllStarLink/app_rpt/pull/1260, https://github.com/AllStarLink/app_rpt/issues/1240) and after some extended sessions with Claude Code Opus this appears to be the root cause. I read the AI policy for Asterisk. Please let me know if I need to change anything in this PR. I've not done much PRs of this nature.

------

The whisper framehook added in "Audiohooks: Allow whisper audiohooks to work without an underlying stream" leaks two frames per timer driven read and can leave its timer fd permanently readable.

ast_write() does not take ownership of the frame it is handed, so the silent frame the framehook injects was only freed when the write failed. The callback also returned &ast_null_frame while discarding the frame it was given; framehook_list_push_event() does not free the original frame, that is the callback's responsibility. Both frames are AST_FRAME_VOICE, so each leak also holds a reference on the cached slin format, and a long lived channel with a whisper audiohook eventually trips the ao2 excessive refcount assertion:

```
  ERROR frame.c: Excessive refcount 100000 reached on ao2 object 0x...
  ERROR frame.c: FRACK!, Failed assertion Excessive refcount 100000 ...
  # 1: /usr/sbin/asterisk(__ast_frdup+0x22c)
```

The discarded frame is also a real read frame that the whisper audiohooks have already been mixed into, so audio was being dropped.

Separately, the timer was only acknowledged when the read event happened to see ast_channel_fdno() equal to the framehook's own fd slot. ast_waitfor_nandfds() records one fd per channel per wakeup and lets later channels override earlier ones, so a caller waiting on several channels routinely reads the hooked channel with fdno pointing elsewhere. The timer fd is level triggered, so a skipped ack leaves it readable forever, every later ast_waitfor() on that channel returns immediately, and the caller spins at 100% CPU. The DAHDI and pthread timing modules make this permanent because they clear only a single expiration per ack, while timerfd drains all of them, which is why this only reproduces on some systems.

Acknowledge the timer on any read event and drain all pending expirations, guarded by a zero timeout poll: the timing modules open their fds blocking (timerfd_create() without TFD_NONBLOCK), so acknowledging a timer that has not fired would block until it does. Injection stays gated on the framehook's own fd so behaviour is otherwise unchanged.


jxmx:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
Did Claude write the commit message, cause there's a Resolves that's seemingly expecting you to have put an issue number in it which you didn't. We're also likely to adopt the Assisted-by header like Linux does[1] instead of Co-authored-by so please use that. I would also ask how much of this was written by you.

[1] https://docs.kernel.org/process/coding-assistants.html#attribution

jcolp:
And is there a specific test case in normal Asterisk for the starvation issue that reproduces it?

main/audiohook.c
```diff
@@ -141,10 +180,21 @@ static struct ast_frame *whisper_framehook_event_cb(struct ast_channel *chan,
 
 	switch (event) {
 	case AST_FRAMEHOOK_EVENT_READ:
-		/* We only care about timer driven READ events */
-		if (ast_channel_fdno(chan) != hook_data->timer_fd_slot
-			|| !hook_data->timer
-			|| ast_timer_ack(hook_data->timer, 1) < 0) {
+		/*
+		 * Acknowledge the timer on any read event, not only on a read this channel
```
jcolp, line 184: And doesn't this defeat the entire purpose of the original change?

If every read acks the timer completely, then ast_channel_fdno may never get set to timer_fd_slot, and so whispering will never occur.

mkmer: I think Claude is having some difficulties understanding what an Unreal Channel is/does and how it is connected.

We do appear to have an issue related to an `who = ast_waitfor_n(cs, n, &ms);` not waiting when we attach an audiohook to :1 of an unreal channel.
 
Changing the order of the selection (putting the unreal channel last in the list of 2 channels) appears to make this problem go away, but I don't think we fully understand the underlying problem at this point.

Our internal architecture: Bridge <-> local channel:2 <-> local channel:1 <-> thread which processes both local channel:1 and IAX channel using `who = ast_waitfor_n(cs, n, &ms);` <-> IAX channel -> internet.

mkmer: "Maybe" this https://github.com/asterisk/asterisk/pull/1967 is having problems with unreal channels.
We will be researching and follow up.

jxmx:
I got rabbit-holed on fixing things further and further upstream rather than looking at making a refactor in our own code. Abandoning this PR.

PR #2174, opened 2026-09-21
res_manager_devicestate: Add AMI actions to set device/mailbox state.

### LINKED ISSUE #2173: [new-feature]: res_manager_devicestate: Add AMI actions to set device/mailbox states
### Feature Description

Currently, there is no way for an external process to set device/mailbox states in the way flexible that res_pjsip_publish_asterisk can - func_devstate only allows setting Custom device states and res_mwi_external_ami has dependencies and conflicts. Add generic AMI actions that are more flexible.

InterLinked1:
Add AMI actions that can be used to set device states and mailbox
states. This allows these states to be manually set in the same
way that res_pjsip_publish_asterisk can be used to manually
synchronize these states. This overcomes some existing limitations:

* func_devstate only allows setting Custom device states
* res_mwi_external_ami has a conflict with app_voicemail

Resolves: #2173

UserNote: The new DeviceStateChange and MailboxStateChange manager
actions allow setting device and mailbox states via AMI (the latter
of which, unlike the existing MWIUpdate action, doesn't require
res_mwi_external or conflict with app_voicemail).


InterLinked1:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp (changes requested):
This is not equivalent to the res_pjsip_publish_asterisk or the multi-Asterisk support and doesn't exactly "set" the state.

In this module you are essentially overwriting the local calculated or persisted state of things, which for the most part derive/get their state from other modules. This means you have two sources of truth feeding the same thing in a non-aggregated way which can overwrite each other. For example if you set a custom device state this way, then it won't be persisted.

This might be fine for your usage but this is far too shoot yourself in the foot easily to be something I would considering merging.

InterLinked1:
> This is not equivalent to the res_pjsip_publish_asterisk or the multi-Asterisk support and doesn't exactly "set" the state.

The module itself isn't equivalent, I'm referring to the ability of the module to set state for anything.

> In this module you are essentially overwriting the local calculated or persisted state of things, which for the most part derive/get their state from other modules. This means you have two sources of truth feeding the same thing in a non-aggregated way which can overwrite each other. For example if you set a custom device state this way, then it won't be persisted.

I could exempt Custom devices, but how is this any different from what res_pjsip_publish_asterisk can do? (And while not enforced, one would probably use this instead of res_pjsip_publish_asterisk, not in addition to it.)
The only difference here is allowing something besides PJSIP to set the states, for cases where that isn't an option.

> This might be fine for your usage but this is far too shoot yourself in the foot easily to be something I would considering merging.

What approach would you prefer then to allow people to manually synchronize/set device states? Right now, it's not possible at all - not from the dialplan, not from the CLI, not from AMI.

jcolp:
There's two ways it works:

1. Setting local state - which is fundamentally against the way things work and I won't accept, it is far too shoot yourself in the foot, you could keep the patch for yourself but wouldn't be accepted - that's what this does right now
2. Providing remote state which is done using an entity ID to identify each server, and then an aggregate is done which is published internally and used - that would work, but would need to protect against setting state using the local entity ID, so that only a remote entity ID can be provided

InterLinked1:
> There's two ways it works:
> 
> 1. Setting local state - which is fundamentally against the way things work and I won't accept, it is far too shoot yourself in the foot, you could keep the patch for yourself but wouldn't be accepted - that's what this does right now
> 2. Providing remote state which is done using an entity ID to identify each server, and then an aggregate is done which is published internally and used - that would work, but would need to protect against setting state using the local entity ID, so that only a remote entity ID can be provided

I see, that makes sense. Yes, 2 is the use case, not 1. So you would be okay if the actions accepted an EntityID as well and enforced that it wasn't the same as the local one?

jcolp:
Yes.

res/res_manager_devicestate.c
```diff
@@ -118,10 +186,90 @@ static int action_devicestatelist(struct mansession *s, const struct message *m)
...
+		astman_send_error(s, m, "Missing or invalid entity ID");
+		return 0;
+	}
+	if (!strchr(device, '/') && !strchr(device, ':')) {
+		astman_send_error(s, m, "Invalid device name");
+		return 0;
+	}
+
+	state_val = ast_devstate_val(state);
+	if (state_val == AST_DEVICE_UNKNOWN) {
```
jcolp, line 212: UNKNOWN is documented as a valid value, but will always fail because this will always say it is invalid. Either it can't be allowed, or an explicit string comparison has to be done here to allow it.

res/res_manager_devicestate.c
```diff
@@ -145,6 +298,13 @@ static int load_module(void)
 		return AST_MODULE_LOAD_DECLINE;
 	}
 
+	if (ast_manager_register_xml("DeviceStateChange", EVENT_FLAG_CALL | EVENT_FLAG_REPORTING, action_devicestatechange)) {
+		return AST_MODULE_LOAD_DECLINE;
```
jcolp, line 302: These two added register calls fail to clean up prior registered/created things.

res/res_manager_devicestate.c
```diff
@@ -118,10 +186,90 @@ static int action_devicestatelist(struct mansession *s, const struct message *m)
...
+	if (ast_strlen_zero(mbox) || ast_strlen_zero(newmsgs) || ast_strlen_zero(oldmsgs)) {
+		astman_send_error(s, m, "Missing required parameters");
+		return 0;
+	}
+	if (ast_strlen_zero(entity_id) || ast_str_to_eid(&eid, entity_id)) {
+		astman_send_error(s, m, "Missing or invalid entity ID");
+		return 0;
+	}
+
+	new_msgs = atoi(newmsgs);
```
jcolp, line 247: These should use sscanf as other things do or ast_str_to_int so that they actually validate

res/res_manager_devicestate.c
```diff
@@ -118,10 +186,90 @@ static int action_devicestatelist(struct mansession *s, const struct message *m)
 	return 0;
 }
 
+static int action_devicestatechange(struct mansession *s, const struct message *m)
+{
+	struct ast_eid eid;
+	enum ast_device_state state_val;
+	const char *device = astman_get_header(m, "Device");
+	const char *state = astman_get_header(m, "State");
+	const char *cachable = astman_get_header(m, "Cachable");
```
jcolp, line 195: This isn't documented in the XML

res/res_manager_devicestate.c
```diff
@@ -65,6 +65,73 @@
...
+				<para>The name of the mailbox (with optional context).</para>
+			</parameter>
+			<parameter name="NewMessages" required="true">
+				<para>The number of new messages.</para>
+			</parameter>
+			<parameter name="OldMessages" required="true">
+				<para>The number of old messages.</para>
+			</parameter>
+			<parameter name="EntityID" required="true">
+				<para>The Entity ID of the remote Asterisk system that originated this device state updated.</para>
```
jcolp, line 119: mailbox state update.

res/res_manager_devicestate.c
```diff
@@ -65,6 +65,73 @@
...
+				<para>Device name. Does not need to be Custom; however, set device states for non-Custom devices with caution.</para>
+			</parameter>
+			<parameter name="State" required="true">
+				<para>The new device state value.</para>
+				<para>Should be one of the following:</para>
+				<para>The possible values for both uses of this function are:</para>
+				<para>UNKNOWN | NOT_INUSE | INUSE | BUSY | INVALID | UNAVAILABLE | RINGING | RINGINUSE | ONHOLD</para>
+			</parameter>
+			<parameter name="EntityID" required="true">
+				<para>The Entity ID of the remote Asterisk system that originated this device state updated.</para>
```
jcolp, line 87: device state update.

res/res_manager_devicestate.c
```diff
@@ -65,6 +65,73 @@
...
+		</synopsis>
+		<syntax>
+			<xi:include xpointer="xpointer(/docs/manager[@name='Login']/syntax/parameter[@name='ActionID'])" />
+			<parameter name="Device" required="true">
+				<para>Device name. Does not need to be Custom; however, set device states for non-Custom devices with caution.</para>
+			</parameter>
+			<parameter name="State" required="true">
+				<para>The new device state value.</para>
+				<para>Should be one of the following:</para>
+				<para>The possible values for both uses of this function are:</para>
```
jcolp, line 83: This only has a single use

res/res_manager_devicestate.c
```diff
@@ -118,10 +186,90 @@ static int action_devicestatelist(struct mansession *s, const struct message *m)
...
+		astman_send_error(s, m, "Invalid device state value");
+		return 0;
+	}
+
+	if (!ast_eid_cmp(&ast_eid_default, &eid)) {
+		astman_send_error(s, m, "Entity ID is ourself (must be from a different Asterisk system)");
+		return 0;
+	}
+
+	ast_publish_device_state_full(device, ast_devstate_val(state), ast_false(cachable) ? AST_DEVSTATE_NOT_CACHABLE : AST_DEVSTATE_CACHABLE, &eid);
```
jcolp, line 222: There's no need to call ast_devstate_val again as you already have it in state_val

res/res_manager_devicestate.c
```diff
@@ -65,6 +65,73 @@
 			</managerEvent>
 		</responses>
 	</manager>
+	<manager name="DeviceStateChange" language="en_US">
+		<since>
+			<version>24.0.0</version>
```
jcolp, line 70: 24.0.0 has already been put out as release candidates, so the earliest this would be is 24.1.0 plus these need the other versions too

jcolp (changes requested):

PR #2175, opened 2026-09-22
res_pjsip: Fix crash on NULL endpoint snapshot in CLI output

### LINKED ISSUE #2172: [bug]: Segfault in cli_endpoint_print_body() running 'pjsip show endpoints' under endpoint churn
**Severity:** Minor

**Versions:** 20.16.0 (crash observed). Unguarded dereference also present on 18, 20, 21, 22, certified/20.7 and master.

**Components/Modules:** res_pjsip

**Operating Environment:** Debian 12 (bookworm), x86_64, Linux 6.1. Asterisk 20.16.0 built from source. PJSIP endpoints created and destroyed dynamically under sustained call volume (~600 calls/min/node). Crashes coincided with `pjsip show endpoints` polled once per minute.

**Frequency of Occurrence:** Occasional

---

## Issue Description

`cli_endpoint_print_body()` dereferences the result of
`ast_sip_get_endpoint_snapshot()` without checking it for NULL:

```c
/* res/res_pjsip/pjsip_configuration.c — line numbers from master */

2141:  RAII_VAR(struct ast_endpoint_snapshot *, endpoint_snapshot,
2141:           ast_sip_get_endpoint_snapshot(endpoint), ao2_cleanup);
...
2166:          endpoint_snapshot->num_channels,     /* <-- no NULL check */
```

`ast_sip_get_endpoint_snapshot()` is a thin wrapper around
`ast_endpoint_latest_snapshot()`, which returns NULL when the endpoint has no
entry in the Stasis cache:

```c
/* main/stasis_endpoints.c */
msg = stasis_cache_get(ast_endpoint_cache(), ast_endpoint_snapshot_type(), id);
ast_free(id);
if (!msg) {
        return NULL;
}
```

That state is reachable for an endpoint that is still in the sorcery container
but no longer has a cache entry. The clearest path is teardown:
`ast_endpoint_shutdown()` publishes a `stasis_cache_clear` message that removes
the snapshot, while sorcery dispatches its deleted-observer notification
asynchronously through a taskprocessor (`main/sorcery.c`). The CLI iterates
sorcery, not the cache, so it can still be handed an endpoint whose snapshot
has already gone. Systems that create and destroy PJSIP endpoints dynamically
hit that window far more often than a static configuration does.

Marking this as inference rather than fact: I have traced these paths in the
source but have not proven the specific window from a dump. The create side
looks narrower than teardown, since `endpoint_publish_snapshot()` dispatches
synchronously to the caching subscription, so I would not claim it with equal
confidence.

The same file already handles this correctly in a sibling function on 20, 21,
22 and master:

```c
1843:          endpoint_snapshot ? endpoint_snapshot->num_channels : 0);   /* guarded */
2166:          endpoint_snapshot->num_channels,                            /* unguarded */
```

On 18 and certified/20.7 that sibling — `active_channels_to_str()`, at lines
1706 and 1701 respectively — reads instead:

```c
if (endpoint_snapshot) {
        return;
}

ast_sip_for_each_channel_snapshot(endpoint_snapshot,
                                  active_channels_to_str_cb, str);
```

which returns when the snapshot is present and falls through to use it when it
is NULL. That looks like an inverted test rather than a guard, and may warrant
separate attention; I have not observed a crash on that path.

### Reachability

`cli_endpoint_print_body()` is the formatter for:

- `pjsip show endpoints`
- `pjsip list endpoints`
- `pjsip show endpoint <id>`

So any operator or monitoring script running a read-only CLI command can crash
the entire process. In our case a health-metrics cron job called
`pjsip show endpoints` once a minute; two separate machines crashed within a few
hours of each other. One of those had crashed with an identical signature three
months earlier, before that cron job existed — I have not been able to establish
what invoked the formatter on that occasion, but the earlier crash suggests the
polling raised the frequency of a race that was already there rather than
introducing it.

### Evidence that this is the faulting instruction

Four crashes, all on `res_pjsip.so`, all reported by the kernel as
`segfault at 50`:

```
node A  Jun 08 14:25:03  segfault at 50  ip ...604e78  res_pjsip.so[...5e4000+3a000]
node A  Sep 21 12:44:16  segfault at 50  ip ...35de78  res_pjsip.so[...33d000+3a000]
node A  Sep 21 16:12:01  segfault at 50  ip ...c78e78  res_pjsip.so[...c58000+3a000]
node B  Sep 21 17:57:01  segfault at 50  ip ...a49731  res_pjsip.so[...a29000+3a000]
```

The base the kernel prints is the start of the VMA containing the fault — the
executable `PT_LOAD` segment, mapped at `p_vaddr 0x12000` — not the start of the
file. `addr2line` wants a file-relative offset, so the correction is
`(ip - base) + 0x12000`. The `p_memsz` of that segment rounds to exactly the
`0x3a000` the kernel reports, which is how the segment was identified.

Without the correction the offset is `0x20e78`, which resolves to
`ast_sip_initialize_sorcery_domain_alias` — an init-time function that cannot be
running under live traffic. That inconsistency is what prompted the check.
With it:

```
$ addr2line -f -C -e res_pjsip.so 0x32e78
cli_endpoint_print_body
/usr/src/asterisk-20.16.0/res/res_pjsip/pjsip_configuration.c:2040
```

The three crashes on node A span separate ASLR loads and all give
`ip - base = 0x20e78` exactly; node B, a separately compiled binary, gives
`0x20731`. Both resolve to the same function and the same line (2040 is the
start of the `ast_str_append()` call whose argument list contains the
dereference).

The faulting instruction and the struct layout agree with the kernel's
`at 50`:

```
$ objdump -d res_pjsip.so
  32e78:  45 8b 75 50    mov 0x50(%r13),%r14d        # %r13 == NULL

$ gdb -batch -ex 'ptype /o struct ast_endpoint_snapshot' res_pjsip.so
  /*     72      |       4 */    enum ast_endpoint_state state;
  /*     76      |       4 */    int max_channels;
  /*     80      |       4 */    int num_channels;      <-- 80 == 0x50
```

So the read is `NULL->num_channels`.

### Why there is no core dump attached

In place of a dump, the faulting address was resolved to a single source line
on two independently compiled binaries, and the struct offset behind the
kernel's `at 50` was confirmed against the type layout. The code path can also
be read directly: the dereference at 2166 is unconditional and the pointer it
uses is documented as NULL-able.

That said, the guidelines ask for a backtrace from a core dump, and I do not
have one. The reason is worth stating plainly.

All four crashes produced no dump. `asterisk.conf` had `dumpcore = yes` and the
unit ran with `LimitCORE=infinity`, so that side was correct, but
`kernel.core_pattern` was the stock relative name `core`. A relative pattern
makes the kernel write into the crashing process's working directory, which for
Asterisk is `/`, and the `asterisk` user cannot write there. Every dump was
discarded silently — no error anywhere, which is why it went unnoticed until I
went looking for the dumps and found none.

I fixed that across the fleet once the crashes were understood, roughly an hour
after the last one:

```
16:12  crash            (core_pattern still relative, dump lost)
17:04  core_pattern fixed on the machine that had crashed
17:57  crash on the second machine  (fix not yet applied there)
18:15  core_pattern fixed there too
18:34  monitoring stopped calling 'pjsip show endpoints'
```

The last line is why no dump has appeared since. Removing the once-a-minute
call was necessary to keep the service up, but it also removed what was making
the race frequent. With the polling in place the same machine crashed twice in
about three and a half hours; without it, the interval between the only two
crashes on that machine was three and a half months. So a dump may take a long
time to arrive naturally, and I did not want to hold the report for it.

Core dumps are now enabled and verified working on every node, so if this
recurs I will attach a real backtrace to this issue. I am also happy to run a
controlled reproduction on an idle node if that would be more useful than
waiting.

### Suggested fix

Two call sites need guarding, not one. The first matches the sibling at 1843:

```diff
-		endpoint_snapshot->num_channels,
+		endpoint_snapshot ? endpoint_snapshot->num_channels : 0,
```

The second is `ast_sip_for_each_channel_snapshot()` at line 1804, which
dereferences the same pointer:

```diff
-	int num, num_channels = endpoint_snapshot->num_channels;
+	int num, num_channels;
+
+	if (!endpoint_snapshot) {
+		return 0;
+	}
+	num_channels = endpoint_snapshot->num_channels;
```

Both are reachable from the command that crashed: `pjsip show endpoints` sets
`context.recurse = 1` (`res_pjsip/pjsip_cli.c`), which reaches the second site
via `cli_endpoint_print_child_body("channel", ...)` → `cli_channel_iterate()`
→ `ast_sip_for_each_channel()`. Guarding only the first would move the crash
rather than remove it. The observed faulting address is the first site; I have
not seen a crash at the second, but it takes the same NULL from the same
command.

I am happy to submit this as a PR if the approach looks right.

### Workaround

For automated collection, read `/ari/endpoints` instead of parsing
`pjsip show endpoints`. ARI enumerates the Stasis cache directly
(`stasis_cache_dump()` in `res/ari/resource_endpoints.c`), so it never reaches
this formatter.

The two are not equivalent, and the difference is the bug itself: the CLI walks
the sorcery container (`ast_sorcery_retrieve_by_regex()`), while ARI dumps the
cache. An endpoint that is in sorcery but has no cached snapshot — exactly the
one that crashes the CLI — is therefore visible to the CLI and absent from ARI.
Sampled simultaneously on a busy node both happened to return the same 23
endpoints, and ARI's `state` field lines up with the CLI's status column
(`offline` for `Unavailable`), which was close enough for our health checks.

There is no workaround for an operator typing `pjsip show endpoints` at the
CLI, which is the part that makes this worth fixing rather than just avoiding.

## Relevant log output

```shell
# Kernel log, node A (Asterisk 20.16.0, Debian 12)
Jun 08 14:25:03 nodeA kernel: asterisk[2216708]: segfault at 50 ip 00007fec0d604e78 sp 00007febd57f40c0 error 4 in res_pjsip.so[7fec0d5e4000+3a000] likely on CPU 1 (core 1, socket 0)
Sep 21 12:44:16 nodeA kernel: asterisk[3894464]: segfault at 50 ip 00007f041635de78 sp 00007f0394e2e0c0 error 4 in res_pjsip.so[7f041633d000+3a000] likely on CPU 3 (core 3, socket 0)
Sep 21 16:12:01 nodeA kernel: asterisk[3923824]: segfault at 50 ip 00007f6b81c78e78 sp 00007f6b13b1c0c0 error 4 in res_pjsip.so[7f6b81c58000+3a000] likely on CPU 0 (core 0, socket 0)

# Kernel log, node B (Asterisk 20.16.0, Debian 12, separately compiled binary)
Sep 21 17:57:01 nodeB kernel: asterisk[459519]: segfault at 50 ip 00007f3c3da49731 sp 00007f3c0d0f10e0 error 4 in res_pjsip.so[7f3c3da29000+3a000] likely on CPU 2 (core 1, socket 0)

# Resolving the faulting address on each node's own binary
nodeA $ addr2line -f -C -e /usr/lib/asterisk/modules/res_pjsip.so 0x32e78
cli_endpoint_print_body
/usr/src/asterisk-20.16.0/res/res_pjsip/pjsip_configuration.c:2040

nodeB $ addr2line -f -C -e /usr/lib/asterisk/modules/res_pjsip.so 0x32731
cli_endpoint_print_body
/usr/src/asterisk-20.16.0/res/res_pjsip/pjsip_configuration.c:2040
```

enesgur:
ast_sip_get_endpoint_snapshot() returns NULL when an endpoint has no
entry in the Stasis cache, which can happen for an endpoint that is
still in the sorcery container but whose snapshot has already been
removed. cli_endpoint_print_body() dereferenced the result without
checking, crashing the process on 'pjsip show endpoints',
'pjsip list endpoints' and 'pjsip show endpoint <id>'.

ast_sip_for_each_channel_snapshot() takes the same pointer and has the
same unchecked dereference. 'pjsip show endpoints' reaches it through
the recursive channel formatter, using a separately fetched snapshot,
so guarding only the CLI formatter would move the crash rather than
remove it. Guard both.

active_channels_to_str() in the same file already handles a NULL
snapshot this way.

Assisted-by: Claude Opus 5

Resolves: #2172


enesgur:
cherry-pick-to: 24
cherry-pick-to: 23
cherry-pick-to: 22
cherry-pick-to: 20

jcolp:
The CLA needs to be signed, and the AI disclosure should go in via the Assisted-by footer.

jcolp (approved):

